PR #1652, opened 2014-01-11
Fixed installing styles containing templates during WCFSetup

MeinAccount:
The WCF pseudo-application wasn't loaded (even after its creation) during the WCFSetup. That caused an error in `Template::__construct()` [when trying to get the `packageID` of the application](https://github.com/WoltLab/WCF/blob/master/wcfsetup/install/files/lib/data/template/Template.class.php?source=cc#L57). This is called [by `StyleEditor::import()`](https://github.com/WoltLab/WCF/blob/master/wcfsetup/install/files/lib/data/style/StyleEditor.class.php#L438) when creating a template for a style.


MeinAccount:
The WCF 2.1 seems to have resolved the issue.

PR #1664, opened 2014-01-28
add some stuff to implement new 3rd party login over plugins

Stricted:


joshuaruesweg:
I think it would be great if this milestone will be merged.

PR #1859, opened 2014-12-25
Removing exit in exception handler

Morik:
Php is automatically exit the current Script after an Exception is thrown.
Because of this, it is not necessary to manually exit the Script inside the exception handler.
While this won't change anything with normal php, it will break hhvm to use the default exception handler


colin-kiegel:
Tip:
flush(); exit;

This will mimic the default behaviour of php in HHVM inside error and exception handlers.

PR #1865, opened 2015-01-18
add "support-likes" attribute

joshuaruesweg:
atm it is not possible to have two commentLists on one page, without that both supports likes


TimWolla:
Doing this vice versa would be better for backwards compatibility. We are in Beta already.

joshuaruesweg:
And if this attribute isn't set, the default value is then ``0``?

TimWolla:
> And if this attribute isn't set, the default value is then 0?

Yes. Otherwise every developer would have to add the attribute.

joshuaruesweg:
I think every developer include the ``commentList.tpl``-Template. 

But I can rewrite this.

TimWolla:
> I think every developer include the commentList.tpl-Template.

Looks like I missed that this was a general use template. I'll let @dtdesign decide.

joshuaruesweg:
@TimWolla maybe this is a PR for `next` if @dtdesign won't decide ^^

PR #1868, opened 2015-01-23
Update global.less

jastend:
Overflow: hidden causes incompatibility with latest Chrome for Android version (v40.0.2214.89).

Details: https://community.woltlab.com/thread/234872-4-x-scrollen-in-chrome-f%C3%BCr-android-nicht-mehr-m%C3%B6glich/


dtdesign:
This is already a modified version which uses JavaScript to detect WebKit on iOS (Safari) which requires this fix. The previous version was just targeting WebKit in general which led to this change.

jastend:
Sorry for that mistake, should have been obvious.

PR #1870, opened 2015-01-24
Update year in copyright notice

KingDome24:
Since nearly one month it's 2015, therefore it makes sense to update the copyright notices.


TimWolla:
Three commits for that small change?

PR #1873, opened 2015-01-29
add userID to authors UserProfile

joshuaruesweg:


dtdesign:
Guests have no user id, or to be more precise, the user id is defined as 0/null?!

joshuaruesweg:
See here: https://community.woltlab.com/thread/235057-userfollowing-event/?postID=1412776#post1412776

The commit haven't anything to do whether the author is guest or not.

PR #1880, opened 2015-02-10
remove PostgreSQL relict

joshuaruesweg:


joshuaruesweg:
I thought the PostgreSQL class had been removed, is apparently not the case, maybe you should be performed if it is not supported anyway.

PR #1883, opened 2015-02-18
de-informal added

andreasd:
added translation de-informal for wcf


andreasd:
Please consider adding de-informal to WCF and thus encouraging application/plugin developers to do the same.

CobbleCity:
Please change the languagename attribute to get a difference between de and de-informal :)

andreasd:
thanks, overlooked that one. added "(Du)" to the language name

My1:
I think when this (honestly great thing) is closed nobody will attempt to merge it because every thinks "okay not needed anymore" or whatever

andreasd:
this was closed because woltlab did not reply at all to the pull request. neither here, nor on their forum. they did not give the slightest hint, that they would plan to support this. as the pull request was outdated, it was closed. 

over a year has passed since then, there are multiple new posts in the forum thread regarding this topic (https://community.woltlab.com/thread/235745-deutsch-informal/?pageNo=1) every few weeks and it is the second highest rating request in the forum.

still, as there is no response from woltlab at all, I honestly don't care anymore...

My1:
well that sadly makes sense. well they always say they dont ignore stuff that doesnt get commented but well...

honestly I think it's annoying that a german based community software actually removed it. it's not like removing a regional variant of German which is hard to read anyway but just plain informal German which is imo a very essential part of a community.

andreasd:
you can create your own de-informal translation by downloading the de xml files for your installation (containing all the variables of the installed plugins) and then translating this XML file. there is a list of common search&replace phrases in the thread linked above that makes this actually quite fast. i translated our whole installation in about 1 hour with the help of the search&replace list and then searching for remaining phrases that needed further refinement.

My1:
well I think this should get back into default, honestly I mean phpBB isnt even german based but they for example have it (and they are free)

PR #1893, opened 2015-03-18
Update LanguageEditor.class.php

schlimpf:
Also escape backslash characters (as first escape operation!). Otherwise this will lead to invalid PHP code in the language files, e.g. for the language item value "test\'test" or "test\".


TimWolla:
@dtdesign `master` or `2.0`?

schlimpf:
both

PR #1898, opened 2015-03-22
Add all POST variables to the GET request

joshuaruesweg:


joshuaruesweg:
@dtdesign

dtdesign:
I haven't forgot these, especially since there are still some bug reports open. I'm always building the update once it is ready to be deployed on woltlab.com, so we can run it in a live environment while I work on the issues that are irrelevant there.

joshuaruesweg:
:+1:

PR #1904, opened 2015-05-01
Apply multiple changes to PasswordUtil

SoftCreatR:
- Update encryption type to `2y` (since `2a` is considered to be
insecure). This is supported since PHP 5.3.7
- Update `isBlowfish()` to detect `2y` as valid blowfish encryption
type
- Improve `isDifferentBlowfish()` to detect encryption type changes as
well (next to cost factor changes)

Oh and i am not sure, what `$2f$` is, since it's not a valid type, but it will be detected as a valid mode.


dtdesign:
`2y` is not supported in PHP 5.3.2 which is the minimum required version for WCF 2.1.x.

SoftCreatR:
Damn, i thought the minimum requirement is 5.3.20 :(

PR #1905, opened 2015-05-01
AbstractCacheBuilder change $arrayIndex default value to null

ramiusGitHub:
Direct access to array index '' is not possible, e.g. in the ACPMenuCacheBuilder (Top Level Menu Items are in cached in array index '')


TimWolla:
Create self contained pull requests, please. Also changing core APIs is not fun.

PR #1910, opened 2015-05-27
add application support

joshuaruesweg:
That would be very helpful, then I would not again reproduce countless files.


TimWolla:
Merged into `next`.

joshuaruesweg:
So I must reflect the whole class? It's a feature which should be added in master, too, because this isn't a change which affects current plugins in any way :(

dtdesign:
This is not going to happen in `master`, `next` is sufficient.

PR #1912, opened 2015-05-29
Add RedisCacheSource

max-m:
This adds a new `RedisCacheSource` to the core.

Please have another look at the language variable `wcf.acp.option.category.general.cache.redis.description`, it is nearly the same as the `memcached` one as both `Redis`and `memcached` are quite similar in function.


max-m:
Just to mention it, because it can be overlooked quickly:
The Travis build failed because of misordered use statements in `wcf\data\AbstractDatabaseObjectAction`

PR #1921, opened 2015-06-14
Option for rebuilding all data at once

cytl:
It's very helpful when there is a option for rebuilding **all** data at once.


TimWolla:
Did you choose the wrong branch by accident?

cytl:
Yes I did. Sorry.

PR #1922, opened 2015-06-14
Option for rebuilding all data at once

cytl:
It's very helpful when there is a option for rebuilding **all** data at once.

### DIFF
--- a/wcfsetup/install/files/acp/templates/rebuildData.tpl
+++ b/wcfsetup/install/files/acp/templates/rebuildData.tpl
@@ -35,6 +35,7 @@
 		
 		{foreach from=$objectTypes item=objectType}
 			<dl class="wide">
+				<dd>
 			                <a class="button small" id="rebuildAllData">{lang}wcf.acp.rebuildAllData{/lang}</a>
 					<small>{lang}wcf.acp.rebuildAllData.description{/lang}</small>
 					


wcfsetup/install/files/acp/templates/rebuildData.tpl
```diff
@@ -35,7 +35,9 @@
 		
 		{foreach from=$objectTypes item=objectType}
 			<dl class="wide">
-				<dd>
+			                <a class="button small" id="rebuildAllData">{lang}wcf.acp.rebuildAllData{/lang}</a>
```
cytl: Added button

TimWolla:
Is this pull request complete? I wonder how this should work.

wcfsetup/install/files/acp/templates/rebuildData.tpl
```diff
@@ -35,7 +35,9 @@
 		
 		{foreach from=$objectTypes item=objectType}
 			<dl class="wide">
-				<dd>
```
max-m: This line should not be removed.

cytl:
Sure, not at all. I will deliver the full code within the next few days.

max-m:
So it is incomplete …

dtdesign:
Feel free to submit a new pull request once everything is in place. Additionally please try to submit as few commits as possible.

PR #1930, opened 2015-07-21
Renaming of "Benutzer online" to "Mitglieder online"

LKDevelopment:
Renaming of "Benutzer online" to "Mitglieder online" as is everywhere else spoken from "Mitgliedern" and not form "Benutzern"


dtdesign:
Guests are also considered and they're anything but members.

PR #1931, opened 2015-07-26
Add missing parent::execute();

SoftCreatR:
Cannot attach event listeners to `GetUpdateInfoCronjob` and `RefreshSearchRobotsCronjob` because of missing `parent::execute($cronjob);`


TimWolla:
They don't have a parent class.

PR #1932, opened 2015-07-28
Improve JSON::encode()

SoftCreatR:
`json_encode` originally allows to add optional parameters, while `JSON::encode` does not (see http://php.net/manual/de/function.json-encode.php)


SoftCreatR:
Just noticed, that this change is not compatible with 5.3.2

PR #1933, opened 2015-07-30
Add EDITABILITY_OWNER_ONLY_REGISTRATION options

joshuaruesweg:
The changes I have of course tested and they work.

### DIFF
--- a/wcfsetup/install/lang/de.xml
+++ b/wcfsetup/install/lang/de.xml
@@ -1645,6 +1645,7 @@ Wenn Sie unter <em>System -&gt; Optionen -&gt; Allgemein -&gt; E-Mails</em> alle
 		<item name="wcf.acp.user.option.editable.1"><![CDATA[Eigentümer]]></item>
 		<item name="wcf.acp.user.option.editable.2"><![CDATA[Administrator]]></item>
 		<item name="wcf.acp.user.option.editable.3"><![CDATA[Eigentümer und Administrator]]></item>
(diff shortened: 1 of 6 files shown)


wcfsetup/install/lang/de.xml
```diff
@@ -1645,6 +1645,7 @@ Wenn Sie unter <em>System -&gt; Optionen -&gt; Allgemein -&gt; E-Mails</em> alle
 		<item name="wcf.acp.user.option.editable.1"><![CDATA[Eigentümer]]></item>
 		<item name="wcf.acp.user.option.editable.2"><![CDATA[Administrator]]></item>
 		<item name="wcf.acp.user.option.editable.3"><![CDATA[Eigentümer und Administrator]]></item>
+		<item name="wcf.acp.user.option.editable.8"><![CDATA[Eigentümer bei Erstanmeldung und Administrator]]></item>
```
TimWolla: This does not really make sense. Each bit is for a specific permission. You would use `4` for Eigentümer bei Erstanmeldung and `6` would be both, as `4` + `2` is `6`.

joshuaruesweg: The problem is, that `EDITABILITY_ALL` is `3` and `6 & 3` is `true`, but `8 & 3` not

TimWolla: And where exactly is `& EDITABILITY_ALL` used?

PR #1944, opened 2015-08-30
Escape delimiter in validationPattern

SoftCreatR:
Using the tilde sign in validation patterns may cause PHP errors, because it's used as delimiter but won't be escaped.


TimWolla:
`preg_quote` quotes everything.

PR #1949, opened 2015-09-19
remove superfluous interface treatment

ShatteredMotion:
Since PHP 5.3.9, `is_subclass_of` supports checks against interfaces as
well. Therefore, the special interface treatment with `ReflectionClass`
is obsolete.

### DIFF
--- a/wcfsetup/install/files/lib/acp/action/WorkerProxyAction.class.php
+++ b/wcfsetup/install/files/lib/acp/action/WorkerProxyAction.class.php
@@ -4,7 +4,6 @@ use wcf\action\AbstractSecureAction;
 use wcf\action\AJAXInvokeAction;
 use wcf\system\exception\SystemException;
 use wcf\system\WCF;
-use wcf\util\ClassUtil;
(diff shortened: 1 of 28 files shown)


TimWolla:
It looks like this makes the method obsolete. Can you mark it as deprecated and remove the uses of it, as well?

/cc @dtdesign

ShatteredMotion:
`wcf\util\ClassUtil::isInstanceOf` throws an exception if one of both classes does not exist while `is_subclass_of` would simply return `false`.
Not sure if other parts of the framework rely on this behaviour. In that case, depreciating this method might not be the best thing to do.

TimWolla:
> is_subclass_of would simply return false.

The usual PHP insanity. I'll let @dtdesign decide.

dtdesign:
Checking for class/interface existence is a side-effect of this method. It's not even a clever one because it is anything but obvious considering its boolean return value.

`is_subclass_of` is fine and the method should be marked deprecated.

TimWolla:
Thanks.

PR #1950, opened 2015-09-27
Parse report message with 'SimpleMessageParser'

Fabii547:
This change makes it possible to click on links that are submitted with the report message.

This is useful if the user reports a double thread and send the link to the other thread within the report message.


TimWolla:
If Smileys are disabled it LGTM.

TimWolla:
Thanks!

PR #1954, opened 2015-10-05
Use 'XMLWriter' to export languages

Fabii547:
Improves the readability of the code.

### DIFF
--- a/wcfsetup/install/files/lib/data/language/LanguageEditor.class.php
+++ b/wcfsetup/install/files/lib/data/language/LanguageEditor.class.php
@@ -140,31 +140,24 @@ class LanguageEditor extends DatabaseObjectEditor implements IEditableCachedObje
 		// sort categories
 		ksort($items);
 
-		// write language file
-		$filename = FileUtil::getTemporaryFilename($this->languageCode, '.xml');
-
 		$xml = new XMLWriter();
-		$xml->beginDocument('language', 'http://www.woltlab.com', 'http://www.woltlab.com/XSD/maelstrom/language.xsd', array('languagecode' => $this->languageCode, 'languagename' => $this->languageName, 'countrycode' => $this->countryCode));
+		$xml->beginDocument('language', 'http://www.woltlab.com', 'http://www.woltlab.com/XSD/maelstrom/language.xsd', ['languagecode' => $this->languageCode, 'languagename' => $this->languageName, 'countrycode' => $this->countryCode]);
 
 		foreach ($items as $category => $categoryItems) {
 			// sort items
 			ksort($categoryItems);
 
-			$xml->startElement('category', array('name' => $category));
+			$xml->startElement('category', ['name' => $category]);
 			
 			// items
 			foreach ($categoryItems as $item => $value) {
-				$xml->writeElement('item', StringUtil::escapeCDATA($value), array('name' => $item));
+				$xml->writeElement('item', $value, ['name' => $item]);
 			}
 			
 			$xml->endElement();
 		}
 		
-		$xml->endDocument($filename);
-
-		// send language file to client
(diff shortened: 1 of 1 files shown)


wcfsetup/install/files/lib/data/language/LanguageEditor.class.php
```diff
@@ -141,25 +139,32 @@ public function export($packageIDArray = array(), $exportCustomValues = false) {
...
+			$xml->endElement();
 		}
 		
-		// footer
-		echo "</language>";
+		$xml->endDocument($filename);
+
+		// send language file to client
+		readfile($filename);
+		@unlink($filename);
```
ShatteredMotion: Why do you write the output into a temporary file instead of outputting it directly to the client?

Fabii547: I simply not realized that this is possible if I call `$xml->endDocument()` without parameter.

wcfsetup/install/files/lib/data/language/LanguageEditor.class.php
```diff
@@ -141,25 +139,32 @@ public function export($packageIDArray = array(), $exportCustomValues = false) {
...
-			
-			// category header
-			echo "\t<category name=\"".$category."\">\n";
+
+			$xml->startElement('category', array('name' => $category));
 			
 			// items
 			foreach ($categoryItems as $item => $value) {
-				echo "\t\t<item name=\"".$item."\"><![CDATA[".StringUtil::escapeCDATA($value)."]]></item>\n";
+				$xml->writeElement('item', StringUtil::escapeCDATA($value), array('name' => $item));
```
TimWolla: Please `writeCData`, this makes the output way more readable: https://secure.php.net/manual/en/function.xmlwriter-write-cdata.php

Check whether you still need `escapeCDATA` afterwards.

wcfsetup/install/files/lib/data/language/LanguageEditor.class.php
```diff
@@ -141,25 +139,32 @@ public function export($packageIDArray = array(), $exportCustomValues = false) {
...
+		$xml->beginDocument('language', 'http://www.woltlab.com', 'http://www.woltlab.com/XSD/maelstrom/language.xsd', array('languagecode' => $this->languageCode, 'languagename' => $this->languageName, 'countrycode' => $this->countryCode));
+
 		foreach ($items as $category => $categoryItems) {
 			// sort items
 			ksort($categoryItems);
-			
-			// category header
-			echo "\t<category name=\"".$category."\">\n";
+
+			$xml->startElement('category', array('name' => $category));
```
TimWolla: Short array notation for `next`, please.

dtdesign:
This PR is missing a crucial aspect: Why?

wcfsetup/install/files/lib/data/language/LanguageEditor.class.php
Fabii547: `writeCData` is already used inside [`writeElement()`](https://github.com/WoltLab/WCF/blob/master/wcfsetup/install/files/lib/util/XMLWriter.class.php#L133) but you're right `escapeCDATA` is unnecessary.

Fabii547:
It improves the readability of the code and the consistence because on other places the `XMLWriter` is used too.

wcfsetup/install/files/lib/data/language/LanguageEditor.class.php
Fabii547: How are your guidelines for the short array notation?

`[]` or `[ ]`
 `['foo']` or `[  'foo' ]` 

 Because actually there is no unique notation visible.

PR #1960, opened 2015-10-30
Support for phpFox 3.x Password Hash Method

SocialOfficer:
Added support for phpFox 3.x password hash implementation.
-
Database sample:
```
phpfox3:phpfox_user.password:phpfox_user.password_salt
```


ginomessmer:
Here's an article I wrote for those who are interested importing their phpFox 3 users to WBB: http://ginomessmer.me/2016/07/import-phpfox3-users-to-wbb

PR #1965, opened 2015-12-05
Fix mixed content warnings for Google Fonts API

max-m:
Use protocol relative URLs for Google Fonts API.


dtdesign:
It shouldn't be hardcoded at all though.

PR #1968, opened 2015-12-23
Improve display of Facebook shares

SoftCreatR:
As of April '15, `shares` includes both, likes and shares. By using an FQL query, you can get the exact number of shares without likes or other factors, which are counted in.

### DIFF
--- a/wcfsetup/install/files/js/WCF.Message.js
+++ b/wcfsetup/install/files/js/WCF.Message.js
@@ -3421,9 +3421,9 @@ WCF.Message.Share.Page = Class.extend({
 	 * Fetches number of Facebook likes.
 	 */
 	_fetchFacebook: function() {
-		this._fetchCount('https://graph.facebook.com/?id={pageURL}', $.proxy(function(data) {
-			if (data.shares) {
-				this._provider.facebook.link.children('span.badge').show().text(data.shares);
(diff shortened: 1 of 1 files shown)


TimWolla:
Superseded by #1973

SoftCreatR:
#1973 is for WCF 2.0, where the code structure is slightly different.

TimWolla:
Yeah, saw this myself (after seeing your second pull request for Twitter). I'll take a closer look after the holidays.

TimWolla:
Are you sure this is correct? `data.data` is an `Array' literal for me.

Whoknowsit:
I did not test it, yet. However, the first `data` is the function param, the 2nd is from the object that is being returned by Facebook. See https://graph.facebook.com/fql?q=SELECT%20share_count%20FROM%20link_stat%20WHERE%20url=%27https://facebook.com%27 for example.

SoftCreatR:
I did not test it, yet. However, the first `data` is the function param, the 2nd is from the object that is being returned by Facebook. See https://graph.facebook.com/fql?q=SELECT%20share_count%20FROM%20link_stat%20WHERE%20url=%27https://facebook.com%27 for example.

TimWolla:
> However, the first data is the function param, the 2nd is from the object that is being returned by Facebook.

I am aware of that. It returns:

```json
{
   "data": [
      {
         "share_count": 24964670
      }
   ]
}
```

for me, `data` is an array literal here. This would require `data.data[0].share_count`, instead of `data.data.share_count`.

SoftCreatR:
You are right. Just fixed it.

PR #1969, opened 2015-12-23
Remove Twitter share count

SoftCreatR:
Twitter has abandoned the Share Count API, so calling their API is obsolete and should be prevented.


TimWolla:
Thanks.

PR #1970, opened 2015-12-25
Fixed grammar, added comma

jacboy:

PR #1971, opened 2015-12-25
Fix phrase

SoftCreatR:
In `wcf.global.error.exception`, a colon is missing. Furthermore (despite the bad grammar) i would recommend not to abbreviate „Administrationsbereich“ because not everyone knows, what „ACP“ means.

This PR replaces #1970


TimWolla:
Applied manually with small changes. Thanks!

PR #1983, opened 2016-01-17
use ObjectTypeAction instead of ObjectTypeEditor

joshuaruesweg:


joshuaruesweg:
The branch should be `next` not `master` -.-'

PR #1988, opened 2016-02-08
Update MySQLDatabase.class.php

ternes3:
This change allow to supply socket as hostname.


TimWolla:
nah, this check is completely bogus. What if my company is using the host `mysql.sockfactory.example.com` for the MySQL database? And no, I won't merge it if you check whether the `.sock` is at the end, either.

dtdesign:
It is already possible to define a socket, just provide something like this:  `localhost;unix_socket=/var/run/mysqld/mysqld.sock`

PR #1992, opened 2016-03-05
prevent undefined offset with invalid tokens

joshuaruesweg:
Here is a topic for this bug: https://community.woltlab.com/thread/247132-aufrufen-von-index-php-boardfeed-at-error/?postID=1510524#post1510524


TimWolla:
Thanks

PR #1995, opened 2016-03-11
Missing index on wcf_user.authData

NBZ4live:
wcf1_user.authData has no index but is used in SELECTS by 3rd party auth.
A full table scan can lead to massive performance issues on systems with a big number of users.

This PR adds an index to this column.

Further it adds a User::getUserByAuthData() method to improve code reuse in the 3rd party auth.

PS: I think that the authData index should be unique, but added a normal one because this seems to be the default behavior in WCF

### DIFF
--- a/com.woltlab.wcf/package.xml
+++ b/com.woltlab.wcf/package.xml
@@ -42,6 +42,10 @@
 		<instruction type="aclOption">aclOption.xml</instruction>
 		<instruction type="script">acp/post_install.php</instruction>
 	</instructions>
+
+	<instructions type="update" fromversion="2.1.10">
+		<instruction type="sql" run="standalone">update_user6_authData.sql</instruction>
+	</instructions>
 	
 	<instructions type="update" fromversion="2.0.13">
 		<!-- pre-update check to prevent update from WCF < 2.0.10 -->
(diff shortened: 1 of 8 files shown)


com.woltlab.wcf/update_user6_authData.sql
```diff
@@ -0,0 +1 @@
+CREATE INDEX authData ON wcf1_user (authData);
```
TimWolla: `ALTER TABLE … ADD INDEX (…)` please.

com.woltlab.wcf/package.xml
```diff
@@ -42,6 +42,10 @@
 		<instruction type="aclOption">aclOption.xml</instruction>
 		<instruction type="script">acp/post_install.php</instruction>
 	</instructions>
+
+	<instructions type="update" fromversion="2.1.10">
```
TimWolla: Please leave this out, this will be added when bumping the version.

com.woltlab.wcf/update_user6_authData.sql
TimWolla: In fact: Leave this out as well.

TimWolla:
Thanks, applied the INDEX to `master` and the PHP changes to `next`.

I'd appreciate a short comment once you change your pull request in the future, otherwise I might not notice.

NBZ4live:
Ok. thx.

PR #1997, opened 2016-03-16
Fix noscript version of reCaptcha

SoftCreatR:
See https://community.woltlab.com/thread/247376-captcha-fallback-zu-klein

Current:

image

Fixed / Modified:

image

### DIFF
--- a/com.woltlab.wcf/templates/recaptcha.tpl
+++ b/com.woltlab.wcf/templates/recaptcha.tpl
@@ -105,13 +105,13 @@
 			<dd>
 				<div id="recaptchaBucket{$recaptchaBucketID}"></div>
 				<noscript>
-					<div style="width: 302px; height: 352px;">
-						<div style="width: 302px; height: 352px; position: relative;">
-							<div style="width: 302px; height: 352px; position: absolute;">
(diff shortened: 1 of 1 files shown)


TimWolla:
Where did you take the changes from? The HTML differs from the one at: https://developers.google.com/recaptcha/docs/faq#does-recaptcha-support-users-that-dont-have-javascript-enabled

SoftCreatR:
I've made them by my own, a while ago, because you cannot adopt it 1:1 in WCF due to some CSS rule incompabilities.

TimWolla:
Thanks.

My1:
this change looks great. (b4 anyone asks: yes I'm the same My1 who posted the thread).

My1:
by the way, will it kill the next update if I just patch this up myself (by using this new file directly)?

next patchday does take quite some time.

TimWolla:
> by the way, will it kill the next update if I just patch this up myself (by using this new file directly)?

Unless you break the file system permissions you are able to replace the template without breaking the update.

My1:
well it might have been like "oh this file is not what it was, let's not update to not break anything"

SoftCreatR:
WCF is mercyless in this case.

My1:
okay, well I know that upgrading phpbb can be a pain because of this (maybe not anymore, but at least it iirc was teh cae once) but well when plugins are made manually it's legit.

TimWolla:
> well it might have been like "oh this file is not what it was, let's not update to not break anything"

You are not supposed to change anything in the file system at all, so there is no reason to keep track of checksums or anything.

PR #1999, opened 2016-03-20
Assign System Variables after Clearing

jens1o:
Hi,

I've added a boolean option, which is useful for short code.


TimWolla:
> Clears assignment of all template variables. This should not be called during runtime as it could leed to an unexpected behaviour.

In fact this method should rather be removed. What do you need it for?

jens1o:
Hi,

I use the function to make my custom logger, which also logging the output.

TimWolla:
> I use the function to make my custom logger, which also logging the output.

Sorry, I don't understand how this change is relevant to a logger logging "output". Please provide some real world example code or try rephrasing your use case.

PR #2002, opened 2016-03-23
don't show packageserver password on edit form

mutec:
at the moment the value of the password-field is set to the clear password string of the stored data for the edited packageserver
this is risky in case you have more than one user accessing the packageserver list

please also patch this for WCF 2.2


My1:
very helpful fix although for the future (WCF2.2 or even farther) it might be helpful to use an oauth/token style login for package servers, so that the forum literally doesnt even know any passwords.

wcfsetup/install/files/lib/acp/form/PackageUpdateServerEditForm.class.php
```diff
@@ -49,11 +49,18 @@ public function save() {
...
+		$serverData = array('data' => array_merge($this->additionalFields, array(
 			'serverURL' => $this->serverURL,
-			'loginUsername' => $this->loginUsername,
-			'loginPassword' => $this->loginPassword
-		))));
+			'loginUsername' => $this->loginUsername
+		)));
+		// renew password if not empty
+		if (!empty($this->loginUsername) && !empty($this->loginPassword))
+			$serverData['data']['loginPassword'] = $this->loginPassword;
```
joshuaruesweg: Codestyle - In the whole WCF if-conditions are in one line.

joshuaruesweg:
With this change you can't delete the password. So you must enter a password, if you edit the server and had a password before.

My1:
then let's prefll it for example with just 
*
that's almost impossuble as a password

mutec:
@joshuaruesweg The password get's deleted, when the username and password are both empty. I cant' image any case you'd want to store the username, but not the password. If the server requires data, you'll have to use both (in general).
See also lines 60+61

My1:
@mutec that also works fine.

joshuaruesweg:
@mutec Sorry that I have overlooked. But there are still codestyle issues :)

My1:
if you know WCF code style so good, make it better.
not everyone codes on the WCF.

joshuaruesweg:
I think, if there are codestyle issues, WoltLab will not merge the change. But that's just my opinion, because the WCF should be consistent. That make it easier for developers. Let's wait for feedback from WoltLab. I'm just a external developer.

My1:
yeah but I have literally no Idea. I just proposed the change (in some private messages, so bad admins dont get bad ideas.

TimWolla:
I was first contacted in private about this “issue”, here is the translation of my reply: 

> If the license owner does not trust the other administrators he is able to revoke the `admin.system.package.canEditServer` permission. Keep in mind: By allowing them to use the package system they are able to circumvent any restrictions anyway, as plugins are able to execute arbitrary scripts during installation and update.

This pull request does not improve security, as security through obscurity does not work in the real world. Apart from the obvious issue mentioned in the quote there are other problems that are not fixed by this pull request:

1. A rogue administrator still is able to change the package server URL to one of his choice, thus receiving the credentials in plain text.
2. The user experience is diminished (it may leave the user wondering why the password was not saved) for security improvement whatsoever.

Regarding:
> very helpful fix although for the future (WCF2.2 or even farther) it might be helpful to use an oauth/token style login for package servers, so that the forum literally doesnt even know any passwords.

This is entirely up to choice of the provider of the package server. But even if the provider would give out opaque tokens for package server authentication one possibly still violates the license agreement (IANAL), as one gives non-authorized users access to the packages (they are able to obtain the credentials).

For this reasons this pull request is not going to be merged.

My1:
bout the token stuff:
well what do you like more.
letting someone get a package, which can be probably found on other locations anyway.
or letting somebone being able to take your whole account?
you could even lock the token to a single server by IP or DNS so it could not be used outside.

about the password saving.
well okay changing the package server URL is an issue (which could be overridden by killing the pw on url change)
and if you are not going to merge this at least inform the user that the passowrd is stored in plaintext and can be viewed by any other admin who can change package servers. because usually a checkbox under a login is not exactly saving the password in clear text and letting it be retrievable easy as hell.

TimWolla:
> well okay changing the package server URL is an issue (which could be overridden by killing the pw on url change)

This still does not solve the general issue that one should not willy-nilly give out access to the package system. If the license owner does not trust the other team members to not screw him over he should not give them access to sensitive tools or information.

> you could even lock the token to a single server by IP or DNS so it could not be used outside.

The credentials have to be stored in the installation in some form. If the credentials have to be stored they can be extracted. Locking the token to the ip address would bring no security improvement (as plugins are able to download the package and send it somewhere else), while introducing issues such as intermittently broken updates due to changing ip addresses (especially when doing load balancing).

> also even if the package system gives you full rights, who is actually creative enough to make a package that grabs the password.

Security does not care about “creativeness”. If something is exploitable it will be exploited. The solution is to not give access to exploitable tools.

> this change is mostly to get the less sophicated attacks away.

And thus brings no security improvement. We already give administrators the tools to properly secure their community, if they opt to not use them it is their own choice.

My1:
but at the very least the user should be warned more clearly that the password will be stored in plaintext and any package or admin with packageserver permissions can read it.

TimWolla:
> but at the very least the user should be warned more clearly that the password will be stored in plaintext and any package or admin with packageserver permissions can read it.

I don't think so. We would need to warn users that the database credentials are saved in plain text as well. And plugins might save sensitive information as well. If we go that route: Browser vendors should  prominently warn users that saved credentials are stored in plain text (Firefox allows one to set a master password, this has to be done manually, though). To the best of my knowledge: They don't do it, because it should be fairly obvious that stored information can be retrieved by some means. The only “safe” place is a place that cannot be accessed by a networked computer.

My1:
well that FF stores it retrievably is farily obvious but as I once said: a checkmark with remember usually works more like a cookie than saving the PW directly, and the fact that the password is pre-filled is not obvious at first sight. that a database using software has to save the pw retrievably is always that way, obvious enough.

TimWolla:
> a checkmark with remember usually works more like a cookie than saving the PW directly

No, it does not. Your mail program does not store a cookie. Your favorite FTP program does not store a cookie. Your web browser does not store a cookie. 

In fact the stored cookie is equivalent to a password, as it allows you to access sensitive information usually protected by the password. The common “duplicate login” or “sudo mode” as GitHub calls it does not count, as it does not allow you to save the credentials (for obvious reasons).

Please refrain from reiterating the very same message in different words, I understood it the first time I read it and it just wastes time.

My1:
well FTP and Mail are different because the concept of cookies doent even exist in those protocols but the browser DOES save a cookie. the password storage is usually outside of the website.

TimWolla:
> well FTP and Mail are different because the concept of cookies doent even exist in those protocols but the browser DOES save a cookie. the password storage is usually outside of the website.

Yeah, but this changes nothing. I already explained why the cookie is a password equivalent in my last message.

> also even though a cookie does allow you a lot but your cookie alone doesnt allow you to get into woltlabs customer area. it also asks for the password again

Yeah, I mentioned that:

> The common “duplicate login” or “sudo mode” as GitHub calls it does not count, as it does not allow you to save the credentials (for obvious reasons).

As you keep reiterating the very same message I am going to lock the conversation at this place. This leads to nothing.

PR #2019, opened 2016-04-25
Fix CSS ID selector for 3rdParty

NBZ4live:
CSS selectors will not work if the ID starts with a number.

This pull request fixes the 3rdParty HTML tag ID.


dtdesign:
All though you're correct, JavaScript is able to select the element by id. Changing the id could potentially break JavaScript code relying on that element, thus we're not going to change it at this point.

We can earliest address this with 2.2 (which has already been done to some extent).

PR #2020, opened 2016-05-01
Force lowercase for cookie domain and domain name

SoftCreatR:
Entering a non-lowercase domain name may result in an infinite loop and a non-lowercase cookie domain might result in some unexpected behavior.


dtdesign:
Would you be so kindly to explain what exactly `… might result in some unexpected behavior.` means?

SoftCreatR:
Cookies won't be recognizer on xyz.tld if you create them for XYZ.tld

PR #2023, opened 2016-05-05
(2.2) Use CDN for Font Awesome

SoftCreatR:
Me and many other people prefer always to use the latest version of Font
Awesome. The best way to achieve this (in my opinion) is to use the
official CDN. No more manual updates, loading performance increases and
less memory consumption while compiling the stylesheets are just a few
benefits.

Dunno, if you like the idea of using it in the ACP, too.
Alternatively, we could use a fixed FA version here and use SRI to keep
it safe or ship the pre-compiled font-awesome.min.css as additional
stylesheet exclusively in the acp.


SoftCreatR:
The failed CI checks aren't my fault :P

PR #2028, opened 2016-05-12
Add class to dropdown trigger element

Padarom:
The trigger element of `WCF.Dropdown.Interactive.Instance` will be given/taken the 'dropdownOpen' class when the interface opens/closes.

This allows to style dropdown triggers as active in custom themes without having to manually extend the Dropdown.Interface.Instance

As a comparison, this is what I'm currently doing in my theme:
```js
	WCF.Dropdown.Interactive.Instance = WCF.Dropdown.Interactive.Instance.extend({
		open: function() {
			this._super();
			this._triggerElement.addClass('open');
		},
		
		close: function() {
			this._super();
			this._triggerElement.removeClass('open');
		}
	});
```

### DIFF
--- a/wcfsetup/install/files/js/WCF.js
+++ b/wcfsetup/install/files/js/WCF.js
@@ -1550,6 +1550,7 @@ WCF.Dropdown.Interactive.Instance = Class.extend({
 		WCF.Dropdown._closeAll();
 		
 		this._container.addClass('open');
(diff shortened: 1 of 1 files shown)


dtdesign:
This is already the case for WCF 2.2, but for compatibility reasons I'll not perform this change for WCF 2.1; I can simply not foresee potential conflicts by adding a CSS class to these elements.

PR #2033, opened 2016-05-25
Enable installation in HTTPS-only enviroment

agp8x:
Added HTTPS dectection


TimWolla:
Can you check whether you are able to use `RouteHandler::getProtocol()` inside WCFSetup: https://github.com/WoltLab/WCF/blob/c6aed51a95eb1787c5489e49e85cbeb30d4345d0/wcfsetup/install/files/lib/system/request/RouteHandler.class.php#L222-L233?

agp8x:
Thanks for the hint, that looks indeed nicer, and works straight-away

TimWolla:
Thanks!

PR #2038, opened 2016-05-28
Fix MariaDB / MySQL-Version detection

SoftCreatR:
Some MariaDB installations report the MySQL version instead of the MariaDB version (e.g. 5.6.20 instead of 10.1.14) which blocks the installation process.


dtdesign:
This does not even remotely fix the issue, it will even fail for the 10.1.* version that report a lower 5.5 version (but are effectively above that). This can be solved by using `SHOW VERSION()` which will return the true (= non-fake) version for both MySQL and MariaDB. We'll take care of that ourselves.

SoftCreatR:
Aww... Good point.

dtdesign:
For clarification: The MySQL version returned by MariaDB does not represents the version it equals to, but is instead a dirty work-around to prevent MySQL clusters from rejecting MariaDB because they consider `10.*` as `MySQL 1.*`.

PR #2040, opened 2016-06-06
Optimize Controller Detection

Morik:
The current Implementation does not properly detect Controllers beginning with multiple uppercase characters like BBCodeList, AJAXProxy etc.
This PR should optimize the detection to fit all of the current controllers without manually adding exceptions for them.


dtdesign:
The hack for AJAXProxy is on purpose to ensure everything works during an update, there is something else in place for all other controller violating the name schema: https://github.com/Morik/WCF/blob/8ff2cd798c6b2fbaf035fb222cd6271f23452389/wcfsetup/install/files/lib/system/request/ControllerMap.class.php#L47

It somehow doesn't seem to work, but this is something I can work on :)

dtdesign:
https://github.com/WoltLab/WCF/commit/578513198ff9daf726135aeda85eb8c94342aaa3

PR #2052, opened 2016-06-30
Update de.xml

MDMAN:
Sprachvariable "wcf.acp.rebuildData.com.woltlab.wcf.article" und "wcf.acp.rebuildData.com.woltlab.wcf.article.description" hinzugefügt.


TimWolla:
This pull request is bad for several reasons:

1. It does not update en.xml.
2. The commit message is written in German.
3. It brings little value at the current point. Lots of language items are missing right now (esp. language items related to the CMS), if we would have distinct commits for every single one of these it would completely clutter the commit history.

Therefore I'll close this pull request. I suggest waiting until everything has settled down, and then creating issues instead of pull requests, as it improves consistency in the language if it is not written by several different persons which are not *that* familiar with the overall style of writing.

PR #2055, opened 2016-07-08
Fix typos

Fabii547:


dtdesign:
@Fabii547 Much appreciated, that must have taken a while to track these down!

wcfsetup/install/files/lib/data/comment/CommentAction.class.php
```diff
@@ -84,7 +84,7 @@ class CommentAction extends AbstractDatabaseObjectAction {
 	public $createdResponse = null;
 	
 	/**
-	 * errors occuring durch the validation of addComment or addResponse
```
TimWolla: I had to laugh :joy:

PR #2060, opened 2016-07-12
add required to required optiontypes

mutec:
see also: https://community.woltlab.com/thread/245055/?postID=1533283#post1533283

Required optiontypes don't get marked as required at the moment which makes it impossible to highlight them via JS/CSS and allows the user sending the form with empty fields which produces unnecessary loss of time and traffic because validate will return an exception if empty.
I might have forgotten to change some optiontypes like radios or checkboxes.


dtdesign:
This change is welcome, but not for the stable branch, please submit this PR for the `next` branch.

PR #2088, opened 2016-08-21
Remove deprecated methods in WoltLabImage.js

joshuaruesweg:


dtdesign:
The current state of the plugin is just fine.

The deprecated method calls are currently the only clean way to access the functionality without using real `require()` calls (not those where you have to pray that they return in time!). On top of that the built-in  language system of Redactor is something I try to avoid whenever possible.

I appreciate your work and effort, but this pull request isn't really an improvement.

PR #2090, opened 2016-08-22
Use 'CryptoUtil::secureCompare()' instead of 'PasswordUtil::secureCompare()'

Fabii547:
`PasswordUtil::secureCompare()` is deprecated in favor of `CryptoUtil::secureCompare()`.


TimWolla:
Thanks.

PR #2103, opened 2016-09-02
Fix length for indexed varchar fields

0xLeon:
Trying to install with those fields length set to 255 results in a blank page without any error message on a All-Inkl webspace running PHP 5.6 and MySQL 5.6. Using this fix, installation went just fine without any problems.

For documentation, the error message in the thrown PDOException wrapped by a DatabaseQueryExecutionException was: 1071 - Specified key was too long; max key length is 767 bytes.


0xLeon:
Also, this is probably a fix for: [https://community.woltlab.com/thread/250155-can-t-install-woltlab-suite-3-0-0-alpha-1/]

TimWolla:
Can you please change the `maxvalue` of `register_username_max_length` and `register_username_min_length` as well? Something “sane” like `100` or so.

And please add a check to [`Package::isValidPackageName`](https://github.com/WoltLab/WCF/blob/acdb7e64dcd359ba62df2de46940298d5ef41c46/wcfsetup/install/files/lib/data/package/Package.class.php#L252) for good measure as well.

Thanks.

0xLeon:
Added those checks. Maybe you should consider using exceptions indicating the specific error in `UserUtil::isValidEmail` because of the needed length limitation which is not complying with the maximum length of 256 characters for a mail address as stated in RFC 821. Right now an invalid mail address just results in a generic error message.

0xLeon:
Failed travis build wasn't my fault, see 0fff082bfb784073ad101eb9860cc359946d221d 😃

Gravatronics:
@0xLeon I triggered a new build.

TimWolla:
> maximum length of 256 characters for a mail address as stated in RFC 821

Technically the limit is 320 bytes, but anything above 254 is undeliverable, because SMTP allows only 256 bytes (including the angle brackets not part of the mail address). Also I am not sure, whether the regular expression allows *any* valid email adress anyway. 

I don't think any user expects to seriously use an email address with more than 191 characters ...

TimWolla:
Thanks.

PR #2108, opened 2016-09-21
easy way to disable WCF's branding via AbstractPage

mutec:
implemented the boolean `\wcf\page\AbstractPage::$showWoltLabBranding`.

This change is needed to provide an easy way (for 3rdparty-developers)
to hide WoltLab's copyright notice within the footer on specific pages
by setting it to `false`.
This gives every developer the possibility to show a specific copyright
without showing the general one. So far so good we could also do this by
setting `WOLTLAB_BRANDING` to `false`. But setting this option would
disable WoltLab's copyright at all which is not the wanted effect in
most cases and would cause much trouble in case WoltLab's products are
installed and no copyright-free-license has been purchased.
Maybe it would even more senseful to remove `WOLTLAB_BRANDING` at all
and use `AbstractPage::$showWoltLabBranding` only.

In a nutshell: Without such a solution there is no easy way to disably
only WoltLab's copyright on only a few pages.

### DIFF
--- a/com.woltlab.wcf/templates/pageFooterCopyright.tpl
+++ b/com.woltlab.wcf/templates/pageFooterCopyright.tpl
@@ -1,3 +1,3 @@
 {event name='copyright'}
 
-{if !'WOLTLAB_BRANDING'|defined || WOLTLAB_BRANDING}<div class="copyright">{lang}wcf.page.copyright{/lang}</div>{/if}
+{if (!'WOLTLAB_BRANDING'|defined || WOLTLAB_BRANDING) && (!$showWoltLabBranding|isset || $showWoltLabBranding)}<div class="copyright">{lang}wcf.page.copyright{/lang}</div>{/if}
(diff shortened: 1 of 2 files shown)


com.woltlab.wcf/templates/pageFooterCopyright.tpl
```diff
@@ -1,3 +1,3 @@
 {event name='copyright'}
 
-{if !'WOLTLAB_BRANDING'|defined || WOLTLAB_BRANDING}<div class="copyright">{lang}wcf.page.copyright{/lang}</div>{/if}
+{if (!'WOLTLAB_BRANDING'|defined || WOLTLAB_BRANDING) && $showWoltLabBranding}<div class="copyright">{lang}wcf.page.copyright{/lang}</div>{/if}
```
TimWolla: Check whether `showWoltLabBranding` is defined (and assume `$showWoltLabBranding = true` if it is not). I'm fairly sure that this will crash when a `NamedUserException` is thrown.

mutec: I changed that. Sorry for the mess of commits. I had already 3 instances of my IDE opened. :(

Will there be a package for easy copyright-removal which sets `WOLTLAB_BRANDING` false (and `true` on deinstallation again), or what was the initial thought about implementing such an option?

dtdesign:
The branding is always in place because regardless of any own software being used, it is still using the WoltLab Suite, at the very least WoltLab Suite Core.

PR #2109, opened 2016-09-27
removed double class attribute from section tag on update overview

mutec:
The `section` got two `class="x"` attributes which makes it impossible to run any update from the PackageUpdatePage because the second gets ignored and `WCF.ACP.Package.Update.Manager` won't find any `.jsPackageUpdate`.


TimWolla:
Thanks.

PR #2112, opened 2016-10-03
add event WCF::destruct()

joshuaruesweg:


joshuaruesweg:
I create a plugin, which checks some conditions (with the build-in condition system) in an event listener. This listener must always be performed every request. If I use `WCF::initialized()`, it is to early for page checks. If I use the destructor, everything is fine (because the page has been initialized.

TimWolla:
I don't quite understand the reasoning yet (you are being pretty vague), but I can give two advice:

- Check whether a new event at the end of `RequestHandler::handle()` is feasible and then create a PR for that event
- If it is not: Register your own shutdown function using `register_shutdown_function` using an appropriate event.

PR #2113, opened 2016-10-04
Speed up parseSmilies in MessageParser

ramiusGitHub:
If the WCF holds many smilies (thousands, e.g. due to the emoji plugin by SoftCreatR), the loop in the parseSmilies function needs a significant amount of time. Checking beforehand with strpos and only calling preg_replace, if the smiley was found at least one time, leads to an improved performance.

### DIFF
--- a/wcfsetup/install/files/lib/system/bbcode/MessageParser.class.php
+++ b/wcfsetup/install/files/lib/system/bbcode/MessageParser.class.php
@@ -133,8 +133,10 @@ class MessageParser extends BBCodeParser {
 	 */
 	protected function parseSmilies($text, $enableHtml = false) {
 		foreach ($this->smilies as $code => $html) {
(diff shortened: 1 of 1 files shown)


ramiusGitHub:
My WBB has around 3000 smilies (about 100 standard smilies, 1000 for SoftCreatR's emoji plugin and 2000 icons from the game Dota 2). A thread page with 20 posts each around two to three lines of text needed 0.8s before and 0.5s after the above optimization.
I know having 3000 smilies is not the point of the smiley system, however, this change does not have a negative effect on any other types of using the WBB.

dtdesign:
Feel free to make such a modification to your own system, I seriously doubt we'll make any changes to that specific file in the 2.1 branch.

This is far beyond what I'm comfortable merging into the stable branch anyway and there has never been a real need for such a change, making it a risky move without a significant reason for us.

ramiusGitHub:
I guess with the Emoji Plugin by SoftCreatR there are many WBBs that suffer at least 0.1 seconds per thread view.

I can apply that change locally, but won't it be overriden with every WCF update that uses the file PIP?

ShatteredMotion:
> I can apply that change locally, but won't it be overriden with every WCF update that uses the file PIP?

Only by updates with changes to the `MessageParser`. WoltLab usually provides file archives which only contain the files changed since the last version. Therefore, only those files are overwritten with the version from the archive and other files remain untouched.

PR #2117, opened 2016-10-18
[master] fixed smiley bottleneck, major performance improvement

Slind14:
see https://github.com/WoltLab/WCF/pull/2116


TimWolla:
Only one PR is necessary. We can merge upward and we can change the base branch, if we want. Thus I'm closing this one.

PR #2119, opened 2016-10-21
Update composer dependencies

TimWolla:
ad Purifier:

> The primary backwards-incompatible change is that HTML Purifier will now add rel="noreferrer" to all links with target attributes (you can disable this with %HTML.TargetNoReferrer.)

see also: https://github.com/WoltLab/WCF/pull/2119/files#diff-fbd783b9f31efdbd3ec5e07e7e23e98f


TimWolla:
The emogrifier changes have been reviewed by me and are good to go.

jens1o:
Is Parsedown obsolete, is not it?

TimWolla:
> Is Parsedown obsolete, is not it?

It is not.

jens1o:
Where is it used? Did I looked into the wrong "Core" files? Is it used by Forum?

TimWolla:
> Where is it used? Did I looked into the wrong "Core" files? Is it used by Forum?

https://github.com/WoltLab/com.woltlab.wcf.exporter/blob/5107e490c2c58d9641357dd279e172ae7b4016cc/files/lib/system/exporter/NodeBB0xRedisExporter.class.php#L542

jens1o:
Ah, there it is. Thanks!

PR #2124, opened 2016-10-27
Add sidebarRightBottom to the sidebar

Griborim:


dtdesign:
There is already a suitable event in place, just a few lines down. What is the motivation behind this request?

Griborim:
You can only Add Boxes on the Top from the sidebar with {capture assign='sidebarRight'}.

With {capture assign='sidebarRightBottom'} can you add Boxes after the loaded boxes.

dtdesign:
The top placeholder is exclusively for static sidebar content, such as filter options on the thread list in the forums and similar. For all other purposes use a proper box that can then be freely arranged by the administrator.

Griborim:
And what is with static Boxes after dynamic boxes?

dtdesign:
Static boxes are those that can only be used on that specific page and they are designed to always appear at the top. I've already given you an example on what kind of boxes this covers, but everything beyond that should be a dynamic box.

PR #2131, opened 2016-11-05
fix balloonTooltip for stats

joshuaruesweg:


jens1o:
Ah thanks. I thought it was removed 😄

PR #2142, opened 2016-11-27
Fix "DailyMotion" media bbcode

woouki:
With the current code, the video is not converted correctly. Only a blank post is displayed.

See https://developer.dailymotion.com/player#player-parameters


wcfsetup/setup/db/install.sql
```diff
@@ -2150,7 +2150,7 @@ INSERT INTO wcf1_template_group (parentTemplateGroupID, templateGroupName, templ
...
 	-- DailyMotion
-	INSERT INTO wcf1_bbcode_media_provider (title, regex, html) VALUES ('DailyMotion', 'https?://(?:www\\.)?dailymotion\\.com/video/(?P<ID>[a-zA-Z0-9]+)', '<iframe width="480" height="208" src="http://www.dailymotion.com/embed/video/{$ID}"></iframe>');
+	INSERT INTO wcf1_bbcode_media_provider (title, regex, html) VALUES ('DailyMotion', 'https?://(?:www\.)?dailymotion\.com/video/(?P<ID>[a-zA-Z0-9_-]+)', '<iframe width="480" height="270" src="//www.dailymotion.com/embed/video/{$ID}"></iframe>');
```
BurntimeX: Why did you remove the double-backlashes here?

woouki: With double-backlashes, the videos are not recognized, only a link or a blank post appears after saving a new thread/post. After the double-backlashes are removed, the video is converted and displayed correctly. Testet with Beta 5 and this video:
http://www.dailymotion.com/video/x53ishb_gedruckte-pfoten-diese-hunde-laufen-zum-ersten-mal_animals

PR #2148, opened 2016-12-15
Load language neutral content

Griborim:
Load language neutral content if content does not exist in the language

### DIFF
--- a/wcfsetup/install/files/lib/data/box/Box.class.php
+++ b/wcfsetup/install/files/lib/data/box/Box.class.php
@@ -189,6 +189,8 @@ class Box extends DatabaseObject {
 		if ($this->isMultilingual || $this->boxType == 'system') {
 			if (isset($this->boxContents[WCF::getLanguage()->languageID])) {
 				return $this->boxContents[WCF::getLanguage()->languageID]->title;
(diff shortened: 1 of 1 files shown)


wcfsetup/install/files/lib/data/box/Box.class.php
```diff
@@ -189,6 +189,8 @@ public function getBoxContentTitle() {
 		if ($this->isMultilingual || $this->boxType == 'system') {
 			if (isset($this->boxContents[WCF::getLanguage()->languageID])) {
 				return $this->boxContents[WCF::getLanguage()->languageID]->title;
+			} else if (isset($this->boxContents[0])) {
```
joshuaruesweg: code style

joshuaruesweg:
Also I recommend to merge this with the first if condition. 

So:
```php
if (($this->isMultilingual || $this->boxType == 'system') && isset($this->boxContents[0])) {
  	if (isset($this->boxContents[WCF::getLanguage()->languageID])) {
  		return $this->boxContents[WCF::getLanguage()->languageID]->title;
 	}
}
else if (isset($this->boxContents[0])) {
```

AND merge this behavior with the other functions (e.g. `getImage()`) otherwise it is not the same behavior.

PR #2150, opened 2016-12-16
Fix cronjob month calculation

Morik:
The current implementation ignores the month parameter at all

### DIFF
--- a/wcfsetup/install/files/lib/util/CronjobUtil.class.php
+++ b/wcfsetup/install/files/lib/util/CronjobUtil.class.php
@@ -147,7 +147,7 @@ final class CronjobUtil {
 		self::$result['month'] = $values['month'][$index];
 		
 		// swap to the next year if the next execution month is before the current month
-		if(self::$result['month'] < gmdate('n', self::$timeBase)) {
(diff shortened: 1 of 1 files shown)


dtdesign:
These changes do not take into account that the combination of the calculated day and month are invalid, which would require recalculating the day based on the next possible month, that is the month dictated by the instructions and matching the day.

Furthermore, this change is targeting WCF 2.1 and following our strict stability policy it must solve a show-stopper type issue to be valid for merging. This issue can be annoying, but it isn't critical and can be easily worked around by manually checking the month when setting the cronjob to once per month.

Corrupting the cronjob calculation in the (old-)stable release should and will not be taken lightly. As a result this change does not qualify for merging into WCF 2.1 at all.

jens1o:
But, instead, could you remove the month field in the cronjob editor, since it's obsolete?

Morik:
I actually have already a fix for the month / year problem but cant push it as im currently on holidays in Sweden ;)
I'll detected it while trying all currently used cronjobs to detect any problems with the new implementation.
I will publish it together with my test results to show that there is no side effect by the changes and push it to the wsc3 tree.

PR #2151, opened 2016-12-17
Fix anti-aliasing issue with Safari/WebKit

azerusDE:
See: https://community.woltlab.com/thread/252285-anti-aliasing-effekt-im-safari-korrigieren/


dtdesign:
Not sure if it is a good idea to target all browsers, also the thread implies that there might be more than one place where Safari runs into issues.

azerusDE:
Does not work properly with the code. Dropdown can not be opened anymore.

dtdesign:
@azerusDE Thanks for your efforts, but I'll run my own tests anyway, especially since I want to limit the changes to Safari as best as possible.

PR #2152, opened 2016-12-21
Add afterBoxes event

jens1o:
This PR adds the possibility to add a Box into the current user online box, for example when adding new statistics/categories etc.

### DIFF
--- a/com.woltlab.wcf/templates/usersOnlineInfoBox.tpl
+++ b/com.woltlab.wcf/templates/usersOnlineInfoBox.tpl
@@ -34,4 +34,5 @@
 			</div>
 		{/if}
 	</section>
+	{event name='afterBoxes'}
 {/if}


dtdesign:
I can't think of any reason why just adding another box doesn't cut it? Additionally, this isn't a generic statistics box where you can throw in random stats.

Please include a real world example why such an event would be required and why adding an own box won't have the same result.

jens1o:
Because you can ensure that the box added by this Template Listener is always right behind the box(which is not possible when you're creating a plugin). For example, when you want to celebrate a user, you could place a box right behind it instead of having it around anywhere else. The disadvantages are that the box added isn't available in the Box Menu.

Okey, I think when the box added isn't in the Box Menu, it will frustrating the users. Closing... Sorry!

PR #2153, opened 2016-12-23
Add events to UserProfile::isAccessible, UserProfile::isProtected and UsersOnlineList::isVisible

Feuerfuchs:
Use case for me is a friend system plugin that, among other features, allows users to make their profile or status only visible to friends. To implement this, I need to be able to change the return value of the three methods listed in the title.
I had actually found a "workaround" by replacing the decorator class of the UserProfileList with a subclass of UserProfile that introduced such events, but it was a pretty ugly hack and didn't work everywhere. That's why I was advised by a staff member to create this patch.

### DIFF
--- a/wcfsetup/install/files/lib/data/user/UserProfile.class.php
+++ b/wcfsetup/install/files/lib/data/user/UserProfile.class.php
@@ -12,6 +12,7 @@ use wcf\data\DatabaseObjectDecorator;
 use wcf\data\ITitledLinkObject;
 use wcf\system\cache\builder\UserGroupPermissionCacheBuilder;
 use wcf\system\cache\runtime\UserProfileRuntimeCache;
+use wcf\system\event\EventHandler;
 use wcf\system\user\signature\SignatureCache;
 use wcf\system\user\storage\UserStorageHandler;
 use wcf\system\WCF;
@@ -20,12 +21,12 @@ use wcf\util\StringUtil;
 
 /**
  * Decorates the user object and provides functions to retrieve data for user profiles.
- * 
+ *
  * @author	Marcel Werk
  * @copyright	2001-2016 WoltLab GmbH
  * @license	GNU Lesser General Public License <http://opensource.org/licenses/lgpl-license.php>
  * @package	WoltLabSuite\Core\Data\User
- * 
+ *
  * @method	User	getDecoratedObject()
  * @mixin	User
  */
@@ -34,89 +35,89 @@ class UserProfile extends DatabaseObjectDecorator implements ITitledLinkObject {
 	 * @inheritDoc
 	 */
 	protected static $baseClass = User::class;
-	
+
 	/**
 	 * cached list of user profiles
 	 * @var	UserProfile[]
 	 */
 	protected static $userProfiles = [];
-	
+
 	/**
 	 * list of ignored user ids
 	 * @var	integer[]
 	 */
 	protected $ignoredUserIDs = null;
-	
+
 	/**
 	 * list of follower user ids
 	 * @var	integer[]
 	 */
 	protected $followerUserIDs = null;
-	
+
 	/**
 	 * list of following user ids
 	 * @var	integer[]
 	 */
 	protected $followingUserIDs = null;
-	
+
 	/**
 	 * user avatar
 	 * @var	IUserAvatar
 	 */
 	protected $avatar = null;
-	
+
 	/**
 	 * user rank object
 	 * @var	UserRank
 	 */
 	protected $rank = null;
-	
+
 	/**
 	 * age of this user
 	 * @var	integer
 	 */
 	protected $__age = null;
-	
+
 	/**
 	 * group data and permissions
 	 * @var	mixed[][]
 	 */
 	protected $groupData = null;
-	
+
 	/**
 	 * current location of this user.
 	 * @var	string
 	 */
 	protected $currentLocation = null;
-	
+
 	const GENDER_MALE = 1;
 	const GENDER_FEMALE = 2;
-	
+
 	const ACCESS_EVERYONE = 0;
 	const ACCESS_REGISTERED = 1;
 	const ACCESS_FOLLOWING = 2;
 	const ACCESS_NOBODY = 3;
-	
+
 	/**
 	 * @inheritDoc
 	 */
 	public function __toString() {
 		return $this->getDecoratedObject()->__toString();
 	}
-	
+
 	/**
 	 * Returns a list of all user ids being followed by current user.
-	 * 
+	 *
 	 * @return	integer[]
 	 */
 	public function getFollowingUsers() {
 		if ($this->followingUserIDs === null) {
 			$this->followingUserIDs = [];
-			
+
 			if ($this->userID) {
 				// get ids
 				$data = UserStorageHandler::getInstance()->getField('followingUserIDs', $this->userID);
-				
+
 				// cache does not exist or is outdated
 				if ($data === null) {
 					$sql = "SELECT	followUserID
@@ -125,7 +126,7 @@ class UserProfile extends DatabaseObjectDecorator implements ITitledLinkObject {
 					$statement = WCF::getDB()->prepareStatement($sql);
 					$statement->execute([$this->userID]);
 					$this->followingUserIDs = $statement->fetchAll(\PDO::FETCH_COLUMN);
-					
+
 					// update storage data
 					UserStorageHandler::getInstance()->update($this->userID, 'followingUserIDs', serialize($this->followingUserIDs));
 				}
@@ -134,23 +135,23 @@ class UserProfile extends DatabaseObjectDecorator implements ITitledLinkObject {
 				}
 			}
 		}
-		
+
 		return $this->followingUserIDs;
 	}
-	
+
 	/**
 	 * Returns a list of user ids following current user.
-	 * 
+	 *
 	 * @return	integer[]
 	 */
 	public function getFollowers() {
 		if ($this->followerUserIDs === null) {
 			$this->followerUserIDs = [];
-			
+
 			if ($this->userID) {
 				// get ids
 				$data = UserStorageHandler::getInstance()->getField('followerUserIDs', $this->userID);
-				
+
 				// cache does not exist or is outdated
 				if ($data === null) {
 					$sql = "SELECT	userID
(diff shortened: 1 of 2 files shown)


wcfsetup/install/files/lib/data/user/UserProfile.class.php
```diff
@@ -438,23 +439,28 @@ public static function getUserProfilesByUsername(array $usernames) {
 	 */
 	public function isAccessible($name) {
 		/** @noinspection PhpVariableVariableInspection */
+		$data = ['result' => true, 'args' => [$name]];
```
TimWolla: Use `$data = ['result' => true, 'name' => $name]`.

wcfsetup/install/files/lib/data/user/UserProfile.class.php
```diff
@@ -463,7 +469,9 @@ public function isAccessible($name) {
 	 * @return	boolean
 	 */
 	public function isProtected() {
-		return (!WCF::getSession()->getPermission('admin.general.canViewPrivateUserOptions') && !$this->isAccessible('canViewProfile') && $this->userID != WCF::getUser()->userID);
+		$data = ['result' => (!WCF::getSession()->getPermission('admin.general.canViewPrivateUserOptions') && !$this->isAccessible('canViewProfile') && $this->userID != WCF::getUser()->userID)];
```
TimWolla: Insert a linebreak after the opening bracket and before the closing bracket for readability.

wcfsetup/install/files/lib/data/user/online/UsersOnlineList.class.php
```diff
@@ -156,20 +157,24 @@ public function checkRecord() {
 	 * @return	boolean
 	 */
 	public static function isVisible($userID, $canViewOnlineStatus) {
-		if (WCF::getSession()->getPermission('admin.user.canViewInvisible') || $userID == WCF::getUser()->userID) return true;
+		$data = ['result' => false, 'args' => [$userID, $canViewOnlineStatus]];
```
TimWolla: Likewise: Please refrain from the args array and opt for explicit array indicies.

wcfsetup/install/files/lib/data/user/UserProfile.class.php
```diff
@@ -810,3 +818,4 @@ public static function getGuestUserProfile($username) {
 		return new UserProfile(new User(null, ['username' => $username]));
 	}
 }
+
```
TimWolla: Duplicate empty line.

wcfsetup/install/files/lib/data/user/online/UsersOnlineList.class.php
```diff
@@ -156,20 +157,24 @@ public function checkRecord() {
...
+				if (WCF::getUserProfileHandler()->isFollower($userID)) $data['result'] = true;
 				break;
 		}
 		
-		return false;
+		EventHandler::getInstance()->fireAction(get_called_class(), 'isVisible', $data);
+		return $data['result'];
 	}
 }
+
```
TimWolla: Duplicate empty line.

TimWolla:
LGMT

/cc @dtdesign Do you have any objections?

dtdesign:
I'm not really convinced by this at all, mostly because it's redundant and basically reinvents the wheel. When we designed the follow system back in WCF 2.0, it was quite important for us to mimic the friendship system.

Mutually following each other has the same effect as the friendship system and was explicitly build this way.

Feuerfuchs:
@dtdesign There's actually a large demand from the community I'm activate at for a friend system despite there being the follower system. The main reasons are that 1. friends aren't necessarily interested in what their friends do, and 2. mutually following each other is not necessarily a signal that both persons are actually friends.

As for 1: In our forum we have a "forum games" section where threads consist of ones such as "say something about the previous poster's avatar", "…signature", and so on. It's basically a pretty spammy off topic area, and as you might have guessed, it can clutter the dashboard very badly. So what did many members do when we had switched form WBB 3 to 4.1? They removed some of their friends from their follower list because they wanted to have a clean dashboard with posts they are actually interested in. But not only were some friends removed from follower lists, now our members were able to add random people without their consent to the follower list, which is perfect to keep track of what interesting people write. So in the end, the follower list for many of them consist of some (but not all) friends and basically strangers who post interesting things. Of course many friends are upset about that — both for when they are removed from the follower list, or if they follow someone they consider as friend and expect them to follow them back.

Which brings me to point 2: So as I said above, friends were removed from the follower list, and that already means mutually following each other is not a good way to convey friendship. But also, it can happen that two members find each other interesting and thus follow each other, despite them being strangers and not having contact at all, thus not actually being friends.

Now I want to add a third point: The changes I made may have more use cases than the one I listed. What our community also often requested is an extension of the ignore function by denying blocked users access to your profile. This would be made possible.
Another feature request by our community was a whitelist/blacklist for who has access to the profile. Now that was a feature I won't implement because the friend system is already a whitelist basically, but perhaps other administrators would need a plugin that implements this.

I don't think the follower system is bad. As I said, it's great to keep track of interesting people, which friends don't necessarily are, as harsh as that sounds. That's why a friend system is necessary for our community, but it wouldn't invalidate the necessity of the follower system. Actually, it would improve it because people could remove friends from their follower list without having to worry about them getting upset.

dtdesign:
I can see that there are edge cases that could be better handled with a friend system, but this doesn't change the fact, that a follower and friendship system are the same at the basic level. It will also create another level of complexity when it comes to permissions and make it transparent when a permission will actually be effective.

Widely overlapping components will make it almost impossible to allow them to run side-by-side, without making things more awkward than they already are. When we build the follower system, we actually evaluated the possibility to have a friendship system at the same time, but the further we dug, the more obvious were the flaws of such an approach.

Have you ever considered going the opposite route of extending the ignore feature to control object visibility and accessibility? This does solve quite a lot of your aforementioned concerns without introducing a competing system.

Feuerfuchs:
> I can see that there are edge cases that could be better handled with a friend system, but this doesn't change the fact, that a follower and friendship system are the same at the basic level.

Because they are a list of users? Yes. Apart from that, they offer different advantages that are geared towards followers/friends. Your new article system and the WoltLab Suite Blog are also the same thing at a basic level (people can publish text), and yet you ship both because they offer very different advantages.

> It will also create another level of complexity when it comes to permissions and make it transparent when a permission will actually be effective.

That's true.

> When we build the follower system, we actually evaluated the possibility to have a friendship system at the same time, but the further we dug, the more obvious were the flaws of such an approach.

Is there a public discussion available where these flaws are listed, or alternatively, are you willing to elaborate? I'm genuinely curious as it might help me find a better solution.

> Have you ever considered going the opposite route of extending the ignore feature to control object visibility and accessibility? This does solve quite a lot of your aforementioned concerns without introducing a competing system.

What do you mean?

Also, one last thing: The friend system shouldn't be the subject of the debate here. Regardless of whether you implement this patch or not, I plan to add my plugin to the plugin store. Without the patch, it would simply lack the feature to make your profile/online status only visible to friends.
I just like to also hear the opinion of the community of whether a friend system is needed or not, that's all.

wcfsetup/install/files/lib/data/user/UserProfile.class.php
```diff
@@ -463,7 +469,11 @@ public function isAccessible($name) {
 	 * @return	boolean
 	 */
 	public function isProtected() {
-		return (!WCF::getSession()->getPermission('admin.general.canViewPrivateUserOptions') && !$this->isAccessible('canViewProfile') && $this->userID != WCF::getUser()->userID);
```
dtdesign: This change is not necessary, `isAccessible()` already offers the ability for event listeners to alter the result. Revert the changes to this particular method please.

wcfsetup/install/files/lib/data/user/online/UsersOnlineList.class.php
```diff
@@ -156,20 +157,23 @@ public function checkRecord() {
 	 * @return	boolean
 	 */
 	public static function isVisible($userID, $canViewOnlineStatus) {
-		if (WCF::getSession()->getPermission('admin.user.canViewInvisible') || $userID == WCF::getUser()->userID) return true;
```
dtdesign: These two conditions should always cause the method to return true, regardless of any program logic or restrictions. Please restore the original line and have them return early just as before. The code below should be altered to no not check for these two conditions again.

wcfsetup/install/files/lib/data/user/UserProfile.class.php
```diff
@@ -438,23 +439,28 @@ public static function getUserProfilesByUsername(array $usernames) {
...
 			break;
 			
 			case self::ACCESS_NOBODY:
-				return false;
+				$data['result'] = false;
 			break;
 		}
+		
+		EventHandler::getInstance()->fireAction($this, 'isAccessible', $data);
+		return $data['result'];
```
dtdesign: Please add a blank line between the `EventHandler` call and the return statement, this is true for the other, similar occurrences below.

wcfsetup/install/files/lib/data/user/online/UsersOnlineList.class.php
```diff
@@ -157,19 +158,23 @@ public function checkRecord() {
 	 */
 	public static function isVisible($userID, $canViewOnlineStatus) {
 		if (WCF::getSession()->getPermission('admin.user.canViewInvisible') || $userID == WCF::getUser()->userID) return true;
+
+		$data = ['result' => false, 'userID' => $userID, 'canViewOnlineStatus' => $canViewOnlineStatus];
 		
 		switch ($canViewOnlineStatus) {
 			case 0: // everyone
-				return true;
+				$data['result'] = true;
```
dtdesign: One last thing: This is now a fall-through which serves no purposes and also has no side-effect. Nevertheless, this looks really odd and I was scratching my head while reading your code.

I failed to mark this in my previous review request, but I want to merge this whole change in a single squash commit without having a loosely coupled commit from me to fix this. Please do me a favor and add a `statement` and this is ready for merge.

Feuerfuchs: Dear god, how did I not notice that… I'll fix it right away.

dtdesign: Thanks.

Feuerfuchs:
Thank you so much for reconsidering and ultimately merging this patch. I appreciate it a lot! :)

PR #2155, opened 2016-12-29
Added beforeProcess event to HtmlOutputNodeProcessor

earlyhost:
Added possibility for plugins to change html output on every request (like it is done with keywords highlighting)

This event is needed for the lexicon plugin to support dynamic detection of lexicon entries. This event is a replacement the old wcf\system\bbcode\MessageParser :: beforeParsing event in WCF 2.1.


wcfsetup/install/files/lib/system/html/output/node/HtmlOutputNodeProcessor.class.php
```diff
@@ -59,6 +60,9 @@ public function setOutputType($outputType) {
 	 * @inheritDoc
 	 */
 	public function process() {
+		// fire event action
+		EventHandler::getInstance()->fireAction($this, 'preProcess');
```
Gravatronics: Please use `beforeProcess` as the event name as in `HtmlInputNodeProcessor`.

PR #2161, opened 2017-01-07
Fix cronjob month calculation

Morik:
The current calculation ignores given months, this is fixed with this pull request.
To detect any issues with the new implementation i'll run several tests without any differences with the cronjobs currently used by the wsc.
https://gist.github.com/Morik/27804aaa57916150dd08268cd8e224d8
Running those tests took quite a while but there have been no differences for all calculations without a month selection so this fix should have no problems with them.
It also allows some exotic cronjobs like the 29. February etc.

### DIFF
--- a/wcfsetup/install/files/lib/util/CronjobUtil.class.php
+++ b/wcfsetup/install/files/lib/util/CronjobUtil.class.php
@@ -138,12 +138,28 @@ final class CronjobUtil {
 	 * @param	array		$values
 	 */
 	protected static function calculateTime(array &$values) {
-		// calculation starts with month, thus start with
-		// month of $time (if within values)
-		$currentMonth = gmdate('n', self::$timeBase);
-		self::findKey($currentMonth, $values['month']);
-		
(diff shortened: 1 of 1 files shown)


jens1o:
any news on this?

dtdesign:
It is currently targeted for 3.1 and will be taken into consideration during the development phase. I'm curious, what did you expect?

jens1o:
Hmm, I wanted to set a cronjob(in a plugin I develop) every two months to reduce the load of some servers(api restrictions). But it isn't possible right now when I am right.

dtdesign:
Run once every month and check if `date('n', TIME_NOW) % 2 == 1` which will cause your script to run in jan/mar/may/jul/sep/nov, compare against `0` to execute on feb/apr/jun/aug/oct/dec.

Or just use a real cronjob.

jens1o:
Okey, thanks.

jens1o:
Thanks!

PR #2162, opened 2017-01-13
Support jQuery plugins utilizing AMD

TimWolla:


wcfsetup/install/files/js/require.config.js
```diff
@@ -44,3 +44,12 @@ requirejs.config({
...
 	}
 });
+
+/* Define jQuery shim. We cannot use the shim object in the configuration above,
+   because it tries to load the file, even if the exported global already exists.
+   This shim is needed for jquery plugins supporting an AMD loaded jquery, because
+   we break the AMD support of jquery for BC reasons.
+*/
+define('jquery', function() {
+	return window.$;
```
dtdesign: For consistency have it check for `window.jQuery` to avoid any kind of timing issues if there is some `noConflict()` stuff going on.

PR #2163, opened 2017-01-14
Add support for VieCode Customer Area 5.1.x passwords in PasswordUtil

Fighter456:
This would be useful for our KittMedia Shop.
Would this PR merged into the 2.0 branch if I create an extra PR for that, too?

### DIFF
--- a/wcfsetup/install/files/lib/util/PasswordUtil.class.php
+++ b/wcfsetup/install/files/lib/util/PasswordUtil.class.php
@@ -7,7 +7,7 @@ use wcf\system\Regex;
  * Provides functions to compute password hashes.
  * 
  * @author	Alexander Ebert
- * @copyright	2001-2015 WoltLab GmbH
+ * @copyright	2001-2017 WoltLab GmbH
  * @license	GNU Lesser General Public License <http://opensource.org/licenses/lgpl-license.php>
  * @package	com.woltlab.wcf
  * @subpackage	util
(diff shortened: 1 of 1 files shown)


dtdesign:
There is an almost infinite number of different hashing algorithms out there and we focus on supporting the most common ones. The proposed one does not qualify to be added to the core.

Fighter456:
I am honestly:
This stance against other developers is very annoying. So it's really no fun to develop products based on your software.

A hash algorithm, which is used in a product for your software (still WCF 1.0), is not a qualified hash algorithm for you? Yes, this hash algorithm was created by an external developer but the developer of this product is not a small person in the WoltLab scene.

Then you should urgently consider how other developers will be able to support more hash algorithms with your software. The way (converting passwords after import) created by you seems _not to be intended for external developers_. Please think about it.

dtdesign:
There is always that one additional piece that we could add to the software, that would benefit X and is quite useful with Y. Nobody is going to argue against this and I can clearly see how much you would love to see this addition being added.

At the end of the day, it is us who are in charge of keeping everything maintained, consistent and seamlessly working together. We need to draw a line somewhere and restrict ourselves to not cross it, after all, it is there for a good reason.

Then again, there can be exceptions made if we think that it would benefit a significant amount of users, but in this particular case, it would at best affect a handful of them. I'm not saying that your changes are void by all means, they're simply not sufficiently in demand for them to be merged into the core.

PS: Discarding passwords on import by flagging them as `invalid` is an intended functionality, and was added to provide an escape route for unrecognized hashing algorithms.

PR #2167, opened 2017-01-29
fix theoretical XSS leak in notices

joshuaruesweg:


wcfsetup/install/files/lib/data/notice/Notice.class.php
```diff
@@ -41,8 +42,8 @@ public function __toString() {
 		// replace `{$username}` with the active user's name and `{$email}`
 		// with the active user's email address
 		$text = strtr(WCF::getLanguage()->get($this->notice), [
-			'{$username}' => WCF::getUser()->username,
-			'{$email}' => WCF::getUser()->email
+			'{$username}' => $this->noticeUseHtml ? StringUtil::encodeHTML(WCF::getUser()->username) : WCF::getUser()->username,
+			'{$email}' => $this->noticeUseHtml ? StringUtil::encodeHTML(WCF::getUser()->email) : WCF::getUser()->email
 		]);
 		
 		if (!$this->noticeUseHtml) {
```
TimWolla: Please change the line below to use `StringUtil::encodeHTML` as well.

joshuaruesweg: Done with an additional commit

PR #2174, opened 2017-02-03
js templateplugin: don't load minified JS within ACP

mutec:
After this commit `{js acp='true' file='WCF.ACP'}` to `wcf/acp/js/WCF.ACP.js?v=...` instead of `wcf/acp/js/WCF.ACP.min.js?v=...`
even if `ENABLE_DEBUG_MODE=0` since it's not recommended to load minified JS within acp-templates at all.


dtdesign:
> since it's not recommended to load minified JS within acp-templates at all.

I disagree with that.

PR #2179, opened 2017-02-10
hide condition field, if the user can't select any group

joshuaruesweg:


wcfsetup/install/files/lib/system/condition/UserGroupCondition.class.php
```diff
@@ -111,7 +111,8 @@ public function getData() {
 	 * @inheritDoc
 	 */
 	public function getHTML() {
-		return <<<HTML
+		if (count($this->getUserGroups())) {
```
Gravatronics: Please use `!empty` instead of `count`.

joshuaruesweg:
Done @Gravatronics, sorry for the late response :/

PR #2281, opened 2017-05-24
Fix importing from tt with html code

Cyperghost:


TimWolla:
Can you give an example of a post that previously failed and now doesn't?

Cyperghost:
Yes.

By this code 
`folgende Zeile musst du dann ändern
[tt]<div class="slideshowContainer newsSlideshowContainer">[/tt]
zu
[tt]<div class="slideshowContainer newsSlideshowContainer marginTop">[/tt]`

After upgrading from wcf2.1 i run the rebuild data worker and got an exception `Trying to get property of non-object`

dtdesign:
Can't reproduce this issue with the code provided above, it works perfectly fine. Regardless of that, your change doesn't really fix anything and instead just bails out when there is something odd happening, in particular you're not fixing the bug, your change simply shoves it under the carpet.

TimWolla:
@Cyperghost As an unrelated hint: Multiline code blocks should be formatted using three backticks. This would make it more readable.

PR #2285, opened 2017-06-01
Update RecaptchaHandler.class.php

markusmarkusz:
I doubt that this PR will be accepted, but it is worth trying it. ^^

The effect of this change is probably self-explanatory.


TimWolla:
> I doubt that this PR will be accepted, but it is worth trying it. ^^

Apart from the commit message it's looking good. I'm going to fix the message myself and apply the patch to the 2.0 branch, as this is a bugfix.

Thanks.

TimWolla:
Closed with a1f6a843e157b3d057e93da4d1e158405f0d7860.

PR #2305, opened 2017-06-11
Improve error message for missing instruction file

Fabii547:
At the moment if you use the short instruction syntax (`<instruction type="file" />`) you get the following not helpful error message if the file cannot be found. 

"The install-instructions specify the file “” for the Package Installation Plugin “file”, but it cannot be found in the specified location."

This PR solves this by using the default filename for the error message if the instruction looks like `<instruction type="file" />`.

### DIFF
--- a/wcfsetup/install/files/lib/system/package/validation/PackageValidationArchive.class.php
+++ b/wcfsetup/install/files/lib/system/package/validation/PackageValidationArchive.class.php
@@ -247,10 +247,12 @@ class PackageValidationArchive implements \RecursiveIterator {
 		for ($i = 0, $length = count($instructions); $i < $length; $i++) {
 			$instruction = $instructions[$i];
 			if (!PackageValidationManager::getInstance()->validatePackageInstallationPluginInstruction($this->archive, $instruction['pip'], $instruction['value'])) {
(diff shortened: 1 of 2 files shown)


wcfsetup/install/files/lib/system/package/validation/PackageValidationArchive.class.php
```diff
@@ -247,10 +247,12 @@ protected function validatePackageInstallationPlugins($type, array $instructions
 		for ($i = 0, $length = count($instructions); $i < $length; $i++) {
 			$instruction = $instructions[$i];
 			if (!PackageValidationManager::getInstance()->validatePackageInstallationPluginInstruction($this->archive, $instruction['pip'], $instruction['value'])) {
+				$defaultFilename = PackageValidationManager::getInstance()->getDefaultFilenameForPackageInstallationPlugin($instruction['pip']);
+				
 				throw new PackageValidationException(PackageValidationException::MISSING_INSTRUCTION_FILE, [
 					'pip' => $instruction['pip'],
 					'type' => $type,
-					'value' => $instruction['value']
+					'value' => ($instruction['value']) ? $instruction['value'] : $defaultFilename
```
dtdesign: `$instruction['value'] ?: $defaultFilename`

Fabii547: Fixed.

PR #2309, opened 2017-06-17
Consistently use 'IObjectCondition' instead of 'IUserCondition'

Fabii547:
New conditions  implement `IObjectCondition` instead of `IUserCondition` which provides a more generalized `checkObject()` method. Old ones still stick with `IUserCondition` but implement at the same time `IObjectListCondition` which leads to a mixture of `*User` and `*Object` methods.

For consistency "old" conditions should also implement `IObjectCondition` instead of `IUserCondition`. This can be achieved by using the (already existing!) `TObjectUserCondition` trait which redirects `checkUser()` calls to `checkObject()` if necessary.

### DIFF
--- a/wcfsetup/install/files/lib/system/condition/UserAvatarCondition.class.php
+++ b/wcfsetup/install/files/lib/system/condition/UserAvatarCondition.class.php
@@ -3,6 +3,8 @@ namespace wcf\system\condition;
 use wcf\data\condition\Condition;
 use wcf\data\user\User;
 use wcf\data\user\UserList;
+use wcf\data\DatabaseObject;
+use wcf\data\DatabaseObjectDecorator;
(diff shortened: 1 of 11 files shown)


Gravatronics:
This breaks existing WSC code and breaks backwards compatibility.

Fabii547:
Hmm, okay. 

Can you explain why? I mean the trait would simply redirect the calls?!

Gravatronics:
There are checks for the specific interface, not only for certain methods to be implemented.

Fabii547:
Okay, that's an argument. But even that could be circumvented by leaving `IUserCondition` implemented. But in this case it's possibly better to leave it as it is actually.

Thanks for your explanation ;)

PR #2330, opened 2017-07-08
inherited language system

mutec:
This feature enables inheritance of languages.
E.g.: You translated German into informal German. If some items of informal German are missing, they will be fetched from it's parent language "German". Same with e.g. en-us and en-gb or the Chinese language variants. Event if you translate English into some different languages you could have a fallback to the English language instead of ugly language item strings.
By this change the user will also be able to decide whether she/he prefers formal or informal German.

Basically the request creates a new column at `wcf1_language` and then calls the parent language's method within `Language::get()` and `Language::getDynamicVariable()`. This feature won't cause big performance-issues because the fallback is used only! if the language item does not exist and that shouldn't be often (but useful if some plugins don't support the specific language and you just don't realize that).

The request also adds a fallback for article contents and page contents.

### DIFF
--- a/wcfsetup/install/lang/de-informal.xml
+++ b/wcfsetup/install/lang/de-informal.xml
@@ -1,0 +1,3927 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<language xmlns="http://www.woltlab.com" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.woltlab.com http://www.woltlab.com/XSD/vortex/language.xsd" languagecode="de-informal" languagename="Deutsch informell" countrycode="de">
+	<category name="wcf.acl">
+		<item name="wcf.acl.access"><![CDATA[Zugangsbeschränkung]]></item>
+		<item name="wcf.acl.access.grant"><![CDATA[Zugang gewähren]]></item>
+		<item name="wcf.acl.access.granted"><![CDATA[Erlaubte Benutzer und Benutzergruppen]]></item>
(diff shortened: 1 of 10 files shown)


wcfsetup/setup/db/install.sql
```diff
@@ -568,6 +568,7 @@ CREATE TABLE wcf1_language (
 	isDefault TINYINT(1) NOT NULL DEFAULT 0,
 	hasContent TINYINT(1) NOT NULL DEFAULT 0,
 	isDisabled TINYINT(1) NOT NULL DEFAULT 0,
+	parentID INT(10),
```
TimWolla: Use an explicit `NULL` for consistency.

wcfsetup/install/lang/de-informal.xml
```diff
@@ -0,0 +1,3927 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
TimWolla: This does not even leverage the inheritance feature.

dtdesign:
This is a bit too much of a change to be considered for 3.1, especially given it significant impact on the language API which in return is one of the core components. We're currently in the process of wrapping things up and getting the last planned features in before we start focusing on the stabilization period.

That said, I'm not against such a change at all, but with a future release, earliest with 3.2. I've added this PR to the internal feature list, allowing us to review it at it at a later stage.

jens1o:
So, what about now?

mutec:
*knock* *knock* the door. Is this still "too much of change" how you called it two years ago?

PR #2332, opened 2017-07-12
http => https in social networks links

jens1o:
quite a tiny change, but it will look better in the statusbar when hovering about the link and it saves (close to no tbh) time.


TimWolla:
We should go the full circle here. There's a few instances of external links in `page.xml` and the language files as well. Would you consider updating them as well?

jens1o:
I'd love to. I wasn't unsure whether you wanted to do that yourself, but this should actually a heads-up thing.

dtdesign:
@jens1o Can you please update your PR to be in sync with the current master?

jens1o:
@dtdesign done

PR #2339, opened 2017-07-16
Added template event to content meta data

fragsalat:
I added a template event to the article and thread meta data for those who want to render additional data like aggregated reviews, share data, etc

I only could find template events which allows the developer to add new navigation buttons.

Since the contribution guideline doesn't mention the need of an issue, I omitted this additional effort.

PS: Would be nice if you could add this to the Forum and maybe Blog as well :)


fragsalat:
Thx for the fast merge. Are there any plans on when you want to release a new version?

dtdesign:
There is no ETA.

PR #2351, opened 2017-07-23
get rid of custom http protocol versions and fix an E_NOTICE error

jens1o:
The E_NOTICE error has been fixed by me because I want to get rid of a full-of-spam error log I had due to some persons who want to check out what that endpoint is doing. You can see what will happen here:
https://woltlab.com/font/getFont.php

Some webbrowsers(Internet Explorer) and especially some server-software(Apache) are confused when they're forced to use an alternative http version for a single endpoint, and in some places it can ruin the use of the advantages of HTTP 2.

### DIFF
--- a/com.woltlab.wcf/templates/user.tpl
+++ b/com.woltlab.wcf/templates/user.tpl
@@ -183,12 +183,19 @@
 					{event name='userDataRow1'}
 				</ul>
 				
-				{if $user->canViewOnlineStatus() && $user->getLastActivityTime()}
+				{hascontent}
 					<ul class="inlineList commaSeparated">
-						<li>{lang}wcf.user.usersOnline.lastActivity{/lang}: {@$user->getLastActivityTime()|time}</li>
-						{if $user->getCurrentLocation()}<li>{@$user->getCurrentLocation()}</li>{/if}
+						{content}
+							{if $user->canViewOnlineStatus() && $user->getLastActivityTime()}
+								<li>{lang}wcf.user.usersOnline.lastActivity{/lang}: {@$user->getLastActivityTime()|time}</li>
+								{if $user->getCurrentLocation()}<li>{@$user->getCurrentLocation()}</li>{/if}
+							{/if}
+							{if $__wcf->session->getPermission('admin.user.canViewIpAddress') && $user->registrationIpAddress}
+								<li>{lang}wcf.user.registrationIpAddress{/lang}: <span class="userRegistrationIpAddress">{$user->getRegistrationIpAddress()}</span></li>
+							{/if}
+						{/content}
 					</ul>
-				{/if}
+				{/hascontent}
 				
 				<dl class="plain inlineDataList">
 					{include file='userInformationStatistics'}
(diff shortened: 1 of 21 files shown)


TimWolla:
> Some webbrowsers(Internet Explorer) and especially some server-software(Apache) are confused when they're forced to use an alternative http version for a single endpoint, and in some places it can ruin the use of the advantages of HTTP 2.

Can you expand on that, please? This seems hard to believe and I'd like to know how I could check myself that the changes are working as advertised.

TimWolla:
> The E_NOTICE error has been fixed by me because I want to get rid of a full-of-spam error log I had due to some persons who want to check out what that endpoint is doing.

I just applied a slightly different bugfix for this issue, while maintaining you as the author of the commit, as it basically was your work.

This pull request now conflicts, but there is no need to rush fixing it, I'd like to discuss the changes first (and can fix the merge conflict myself if necessary) :smile:

jens1o:
Having this file:
```php
<?php

$version = '1.' . (random_int(0, 1) >= 0.5 ? '1' : '0');

header("HTTP/{$version} 200 OK");
?>
<p>Hello World</p>
```

IE will drop the cache for this route when you give it a refresh and the version changes, which is annoying when you have a form, where the webserver will normally respond `1.1`, but you overwrite it back to `1.0` and thus IE drops it cache to be somehow minimalistic.

Sometimes, you see a tiny blink, which means the cache is dropped, and sometimes you don't, when the http protocol number remains the same.

Some webservers([my reference how I got onto that issue](https://secure.php.net/manual/en/function.header.php#92305)), like apache are still thinking of being in their own mode, and (that's my explaination for this behaviour) are using compress technologies, but the webbrowsers don't understand them because they were forced to use a lower version where there wasn't such a specification.

TimWolla:
> Some webservers(my reference how I got onto that issue), like apache are still thinking of being in their own mode, and (that's my explanation for this behavior) are using compress technologies, but the webbrowsers don't understand them because they were forced to use a lower version where there wasn't such a specification.

This is about Apache 2.0, which is like … horribly outdated. But thanks for the comment, we'll check how the situation is in 2017.

jens1o:
You're still supporting PHP 5.4, which is horribly outdated, too. ¯\_(ツ)_/¯ Either way, I like to behave friendly and be a good friend with the webserver. c:

dtdesign:
We're supporting PHP 5.5, because Ubuntu 14.04 LTS comes with PHP 5.5 only. To put it in perspective, it alsocomes with Apache 2.4.

*mumble mumble* Yet people thought that Debian back in the old days was bad.

jens1o:
Other way around: When will the LTS version die?

dtdesign:
April 2019 ;)

TimWolla:
@jens1o My recherche has shown the following:

1. Anything using FastCGI does not care (the HTTP version is entirely ignored). This are probably the most likely setups today, because mod_php is a security nightmare.
2. mod_php has a special casing for `HTTP/1.0` inside the PHP source code, forcing a compatibility flag in Apache. The HTTP version is completely ignored otherwise as well.

Thus: Please set `HTTP/1.1` statically everywhere. There is no need to read `SERVER_PROTOCOL`.

jens1o:
Okay, I thought you rather wanted to close this inquiry, but okay.

TimWolla:
@jens1o Thanks. Applied manually to squash the commits and fix a few places you missed.

PR #2355, opened 2017-07-27
fix wcf1_package.packageDir silently

mutec:
This change tries to fix the `is missing or does not implement 'wcf\system\application\IApplication'` if the folder of an has been moved, but the admin forgot to update the `wcf1_package`-table. The column `packageDir` will be fixed and the config file will be rewritten automatically. This works only of the app is placed in a sub-folder of the WSC itself.

### DIFF
--- a/wcfsetup/install/files/lib/system/WCF.class.php
+++ b/wcfsetup/install/files/lib/system/WCF.class.php
@@ -1,8 +1,10 @@
 <?php
 namespace wcf\system;
 use wcf\data\application\Application;
+use wcf\data\application\ApplicationEditor;
 use wcf\data\option\OptionEditor;
 use wcf\data\package\Package;
+use wcf\data\package\PackageAction;
 use wcf\data\package\PackageCache;
 use wcf\data\package\PackageEditor;
 use wcf\data\page\Page;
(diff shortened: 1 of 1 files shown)


dtdesign:
I do like the general idea, but I don't like the way these changes have been implemented. It would make much more sense to check for the class first and on mismatch try to guess the path. This avoids duplicating the code that is used to load and register an application.

Oh, and line 622 in your version violates our code style (`else` in the same line as `}`).

TimWolla:
> This avoids duplicating the code that is used to load and register an application.

Another suggestion: Once you attempted to fix the values perform a “self redirect”.

mutec:
I thought about that too. The reason why I tried it without a reload is, that I want to keep the POST-data (if it happened while someone is writing a post for example). If I redirect, the data would get lost - or am I wrong?

TimWolla:
HTTP 307 / 308 is specified to preserve the protocol. But I consider it unlikely that this issue is encountered during a POST request.

dtdesign:
This change is only relevant when moving things around on the filesystem, something you shouldn't do while the site is still live anyway, therefore this is something we should not account for.

dtdesign:
@mutec I'm currently wrapping up the updates for the WSC 3.0 series, afterwards I will check back on the development version: Are you going to adjust your pull request according to the above feedback, or should I take care of it?

mutec:
I'll have a look the next days or today. Not sure when I'll have enough time.

dtdesign:
@mutec Do you had some spare time to look into this? If not, do you want me to take care of implementing the changes on my own?

PR #2373, opened 2017-08-11
Remove quotation marks around links in trophy recent activity event

Fabii547:
Other recent activity events doesn't use quotation marks.


joshuaruesweg:
Thanks.

PR #2382, opened 2017-08-16
Updating recipientIDs after event

Cyperghost:


dtdesign:
Changing the list of participants is not supported, additionally this targets 2.1 instead of 3.1 (next).

PR #2383, opened 2017-08-17
Fix {foreach} for non-`\Countable` `\Traversable`s

TimWolla:


TimWolla:
Don't merge yet.

wcfsetup/install/files/lib/system/template/TemplateScriptingCompiler.class.php
```diff
@@ -689,19 +689,27 @@ protected function compileForeachTag($foreachTag) {
 			$foreachProp = "\$this->v['tpl']['foreach'][".$args['name']."]";
 		}
 		
+		$hash = StringUtil::getRandomID();
```
dtdesign: Is there a particular reason behind using `$hash` and `"\$_foreach_".$hash` everywhere instead of using something like this:
```
$foreachHash = '$_foreach_' . StringUtil::getRandomID();
```
After all, it's always the same variabl ename, and the constant string concats make the code a bit harder to read than it should be.

wcfsetup/install/files/lib/system/template/TemplateScriptingCompiler.class.php
```diff
@@ -689,19 +689,27 @@ protected function compileForeachTag($foreachTag) {
...
+		
 		$phpCode = "<?php\n";
+		$phpCode .= "\$_foreach_".$hash." = ".$args['from'].";\n";
+		$phpCode .= "\$_foreach_".$hash."_cnt = 1;\n";
```
dtdesign: This will effectively evaluate to `1` for `NULL` values, but `count()` explicitly evaluates to `0` for this special value.

TimWolla:
Good catches, both fixed.

PR #2385, opened 2017-08-21
hide <small> when possible in contact form

jens1o:


dtdesign:
This behavior is consistent with all other types of generated option forms, also your change did not account for phrases without an actual match.

jens1o:
> This behavior is consistent with all other types of generated option forms
So, a blown-up DOM should be the standard? This slows down(when spamming it) querying for dom elements... You're confusing me, at least, it was worth it a try.

PR #2393, opened 2017-08-23
optimize composer autoloader

jens1o:
It makes the installation filesize a little bit larger, but this will improve the autoloader performance more than 30% according to http://mouf-php.com/optimizing-composer-autoloader-performance in a time-critical scope.


Morik:
I'm not sure, if this will have any performance benefits at all.
Since there aren't packages with the same namespace, there is no need to do multiple `file_exists` calls.
This said you have therefore the additional overhead to parse a larger array at runtime.

dtdesign:
The proprosed change has a negative impact on the overall performance, composer's autoloader is barely every used on runtime. HTMLPurifier is pretty much the only case in which the autoloader is used regulary, but in that case saving a few ms doesn't make HTMLPurifier any faster, that request will be slower than normal anyway.

Long story short: The current autoloaders work just fine and have been tested for performance already, both by time measuring and by examining calls via xdebug dumps.

PR #2398, opened 2017-08-24
pull gravatars via https

jens1o:
https://gravatar.com/avatar/8b464a8b587c833f435541619d605f40?s=200&r=g works quite fine, too.

### DIFF
--- a/wcfsetup/install/files/lib/data/user/avatar/Gravatar.class.php
+++ b/wcfsetup/install/files/lib/data/user/avatar/Gravatar.class.php
@@ -19,7 +19,7 @@ class Gravatar extends DefaultAvatar {
 	 * gravatar base url
 	 * @var	string
 	 */
-	const GRAVATAR_BASE = 'http://gravatar.com/avatar/%s?s=%d&r=g&d=%s';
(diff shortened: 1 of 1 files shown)


TimWolla:
Not every web hoster supports outgoing TLS connections.

jens1o:
How are you going to do that with GitHub Auth Actions, Twitter, Google, and Facebook?

For example: https://github.com/WoltLab/WCF/blob/master/wcfsetup/install/files/lib/action/FacebookAuthAction.class.php#L48

TimWolla:
> How are you going to do that with GitHub Auth Actions, Twitter, Google, and Facebook?

These require TLS from their side, this is documented as well and provides a real value add, because secrets are transmitted.

Gravatars OTOH transmit public information. While I agree that encrypting everything is a good thing in general in this case there are more downsides.

- Servers don't usually get MITMd anyway.
- The end user receives the Gravatar properly encrypted by WoltLab Suite.
- PHP may not support outgoing connections.
- It is not possible to re-use negotiated TLS sessions for multiple gravatar requests, because of PHP's shared nothing request model.

jens1o:
Okay, I get your point.

PR #2436, opened 2017-10-08
fix php 7.2 compatibility problem

jens1o:
happens when there were no tags entered but the module is activated. (`count(null)` is a bad idea):
https://github.com/WoltLab/WCF/blob/93f181958dcd0e88e54d0686712aecc50979cde4/wcfsetup/install/files/lib/page/AbstractArticlePage.class.php#L131-L160

### DIFF
--- a/com.woltlab.wcf/templates/article.tpl
+++ b/com.woltlab.wcf/templates/article.tpl
@@ -263,7 +263,7 @@
 	</div>
 {/if}
 
-{if $relatedArticles|count}
+{if !$relatedArticles|empty}
 	<section class="section relatedArticles">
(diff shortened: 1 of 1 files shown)


dtdesign:
This doesn't work out, because `empty($var)` will return true for objects, but does not check against `\Countable::count()` to see if it is considered to be empty. In that case, an empty list of related articles will pass the `empty` check, despite it holding zero records.

It would make more sense to check if `$relatedArticles` is not null, and use `count` too to check for records. The other choice would be to use `{hascontent}`, but that's much slower than just a plain `count()`-invokation. Also it makes sense to leave a comment before that line to make absolutely clear, why there isn't "just" an check for `empty`.

PR #2445, opened 2017-10-15
Update favico.js

SoftCreatR:
Please update favico.js to 0.3.10 or even 0.3.7, which supports favicons to be loaded via CORS (if the source server supports it).


TimWolla:
Thanks.

PR #2454, opened 2017-10-27
Improved pasting from clipboard into the editor

dtdesign:
* Uses the native clipboard API to process text and HTML in all major browsers.
* Better cross-browser detection of pasted images for both URLs (using ``) and plain image data.
* Uses the legacy implementation for Internet Explorer 11 (ancient API) and iOS Safari (incomplete and broken API).
* Works-around the ~1s second paste delay in Chrome for Android.
* Almost entirely avoids page movement on paste in iOS Safari.

### DIFF
--- a/wcfsetup/install/files/js/3rdParty/redactor2/plugins/WoltLabPaste.js
+++ b/wcfsetup/install/files/js/3rdParty/redactor2/plugins/WoltLabPaste.js
@@ -153,16 +153,6 @@ $.Redactor.prototype.WoltLabPaste = function() {
 					// prevent page scrolling
 					this.tmpScrollTop = undefined;
 					
-					// DEBUG CODE TO VERIFY PASTE HANDLING -- REMOVE BEFORE RELEASE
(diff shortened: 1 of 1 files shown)


wcfsetup/install/files/js/3rdParty/redactor2/plugins/WoltLabPaste.js
```diff
@@ -33,34 +70,102 @@ $.Redactor.prototype.WoltLabPaste = function() {
...
+				// do nothing
 			}).bind(this);
 			
 			var mpGetPasteBoxCode = this.paste.getPasteBoxCode;
 			this.paste.getPasteBoxCode = (function (pre) {
+				if (pastedHtml !== null || pastedPlainText !== null) {
+					// prevent page scrolling
+					this.tmpScrollTop = undefined;
+					
+					// DEBUG CODE TO VERIFY PASTE HANDLING -- REMOVE BEFORE RELEASE
```
TimWolla: Remove before merge.

PR #2463, opened 2017-10-31
array with clipboard-actions which require a page-reload after execution

mutec:
`wcf\system\clipboard\action\AbstractClipboardAction::$refreshPageActions`
this makes it more easy to provide simple clipboard-actions without the need of special JavaScript which handles the marking after delete/enable/disable etc.

### DIFF
--- a/wcfsetup/install/files/js/WoltLabSuite/Core/Controller/Clipboard.js
+++ b/wcfsetup/install/files/js/WoltLabSuite/Core/Controller/Clipboard.js
@@ -365,6 +365,11 @@ define(
 			if (fireEvent) {
 				triggerEvent();
 			}
+
+			//noinspection JSUnresolvedVariable
+			if (data.parameters.refreshPageAfterExecution === 'true') {
+				window.location.reload();
+			}
 		},
 		
 		/**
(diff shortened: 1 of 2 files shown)


wcfsetup/install/files/js/WoltLabSuite/Core/Controller/Clipboard.js
```diff
@@ -410,6 +415,11 @@ define(
 				
 				this._loadMarkedItems();
 			}).bind(this));
+
+			//noinspection JSUnresolvedVariable
```
dtdesign: Bad location, this code is executed before the response from the server was received (see the code a few lines earlier).

mutec: Which position would you recommend?

dtdesign:
Played around with the basic idea behind this change and came across a few issues related to the asynchronous behavior of some actions, making the code path quite unpredictable. Additionally, forced page reloads should generally go with some sort of notifications to make it more clear.

I'm not going to pretend that _all_ built-in actions behave according to the requirements above, but the changes suggested in this change neither fix them, nor does it reduce the amount of misbehaving code parts.

PR #2464, opened 2017-10-31
support the license-tag within the package.xml

mutec:
multilingual license-tag within the packageinformation-tag, containing the attribute `url` with an URL.
example:
```xml
	<packageinformation>
		<packagename>Test-Paket</packagename>
		<version>1.0.0 Alpha 1</version>
		<date>2017-10-05</date>
		<license url="https://www.mysterycode.de/licenses/kostenlose-plugins/">Kostenlose Plugins</license>
	</packageinformation>
```
or multilingual:
```xml
	<packageinformation>
		<packagename>Test-Paket</packagename>
		<version>1.0.0 Alpha 1</version>
		<date>2017-10-05</date>
		<license language="de" url="https://www.mysterycode.de/licenses/kostenlose-plugins/">Kostenlose Plugins</license>
	</packageinformation>
```

The license will be shown on the detail-page of a package and on the confirm-form before the installation.

### DIFF
--- a/XSD/package.xsd
+++ b/XSD/package.xsd
@@ -97,6 +97,16 @@
 					</xs:simpleContent>
 				</xs:complexType>
 			</xs:element>
+			<xs:element name="license" minOccurs="0" maxOccurs="unbounded">
+				<xs:complexType>
+					<xs:simpleContent>
+						<xs:extension base="xs:string">
+							<xs:attribute name="language" type="xs:string" />
+							<xs:attribute name="url" type="xs:string" />
+						</xs:extension>
+					</xs:simpleContent>
+				</xs:complexType>
+			</xs:element>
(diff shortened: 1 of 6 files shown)


dtdesign:
This doesn't cover all aspects of the license confirmation and in particular does not handle installations via the package servers or changes to licenses applied during updates. Also, the user is not explicitly prompt to accept the license terms and therefore is the same as not displaying it at all.

We do consider supporting license prompts in future versions, but none of the current solutions (including this PR) satisfy both our expectations and the legal requirements to have them actually work.

mutec:
Maybe it's time to support a license prompt with WSC 5.2?
However having a license-Tag would make it easier to find the license of a package - and every package contains this file in contrast to PHP/JavaScript-Files.
Your updateserver could also read the license from the package automatically - oh wait you still don't serve the license information. :)

PR #2465, opened 2017-10-31
i18n database object list class

mutec:
this makes it possible to provide a list of objects in a real alphabetical order
for performance-reasons it should be used when needed only (e.g. when sortField is title and title might contain language items)


wcfsetup/install/files/lib/data/I18nDatabaseObjectList.class.php
```diff
@@ -0,0 +1,54 @@
...
+	 * Array of fields filled with language items
+	 * this slows down the system, use this only when sorting
+	 * based on an i18n-value.
+	 * The value represents the original field name.
+	 * The key represents the new field name and is optional.
+	 * If no key is provided, the new field will be 'i18n'.$value
+	 * 
+	 * use table.field schema!
+	 * 
+	 * @example   [ 'title' => 'titleSortField' ]
```
TimWolla: This example contradicts the explanation above. Should it read `[ 'titleSortField' => 'title' ]`?

wcfsetup/install/files/lib/data/I18nDatabaseObjectList.class.php
```diff
@@ -0,0 +1,54 @@
...
+abstract class I18nDatabaseObjectList extends DatabaseObjectList {
+	/**
+	 * Array of fields filled with language items
+	 * this slows down the system, use this only when sorting
+	 * based on an i18n-value.
+	 * The value represents the original field name.
+	 * The key represents the new field name and is optional.
+	 * If no key is provided, the new field will be 'i18n'.$value
+	 * 
+	 * use table.field schema!
```
TimWolla: What does this mean?

wcfsetup/install/files/lib/data/I18nDatabaseObjectList.class.php
```diff
@@ -0,0 +1,54 @@
...
+ * @author	Florian Gail
+ * @copyright	2001-2017 WoltLab GmbH
+ * @license	GNU Lesser General Public License <http://opensource.org/licenses/lgpl-license.php>
+ * @package	WoltLabSuite\Core\Data
+ * @since	3.1
+ */
+abstract class I18nDatabaseObjectList extends DatabaseObjectList {
+	/**
+	 * Array of fields filled with language items
+	 * this slows down the system, use this only when sorting
```
TimWolla: Use proper grammar with complete sentences, please.

wcfsetup/install/files/lib/data/I18nDatabaseObjectList.class.php
```diff
@@ -0,0 +1,54 @@
...
+				
+				if (!empty($this->sqlSelects)) $this->sqlSelects .= ", ";
+				
+				if (is_numeric($key)) {
+					$tmp = explode('.', $value);
+					$key = 'i18n' . $tmp[count($tmp) - 1];
+				}
+				
+				$this->sqlSelects .= "IF (" . $matchTable . ".languageItemValue IS NULL, " . $value . ", " . $matchTable . ".languageItemValue ) AS " . $key;
+				$this->sqlJoins .= " LEFT JOIN wcf" . WCF_N . "_language_item " . $matchTable . " ON " . $matchTable . ".languageItem = " . $value . " AND " . $matchTable . ".languageID = " . WCF::getLanguage()->languageID;
```
TimWolla: Allow explicit specification of a language while defaulting to the global language.

wcfsetup/install/files/lib/data/I18nDatabaseObjectList.class.php
```diff
@@ -0,0 +1,54 @@
...
+	 * @inheritDoc
+	 */
+	public function readObjects() {
+		if (!empty($this->i18nFields)) {
+			foreach ($this->i18nFields as $key => $value) {
+				$matchTable =  'i18n_' . StringUtil::getHash($value);
+				
+				if (!empty($this->sqlSelects)) $this->sqlSelects .= ", ";
+				
+				if (is_numeric($key)) {
```
TimWolla: I'm not sure about this. I feel that an explicit name should be given in all cases.

wcfsetup/install/files/lib/data/I18nDatabaseObjectList.class.php
```diff
@@ -0,0 +1,54 @@
...
+					$tmp = explode('.', $value);
+					$key = 'i18n' . $tmp[count($tmp) - 1];
+				}
+				
+				$this->sqlSelects .= "IF (" . $matchTable . ".languageItemValue IS NULL, " . $value . ", " . $matchTable . ".languageItemValue ) AS " . $key;
```
TimWolla: This is: `COALESCE($matchTable.languageItemValue, $value)`.

wcfsetup/install/files/lib/data/I18nDatabaseObjectList.class.php
TimWolla: Break this string into multiple lines for readability.

wcfsetup/install/files/lib/data/I18nDatabaseObjectList.class.php
mutec: Yes. But I'll change the behavior since key = original column and value = new column's name should be more logical.

TimWolla: @mutec That would allow defining the same field multiple times. Not sure whether that is better. Add an explicit validation then.

wcfsetup/install/files/lib/data/I18nDatabaseObjectList.class.php
```diff
@@ -0,0 +1,65 @@
...
+		parent::__construct();
+		
+		if ($this->languageID === null) {
+			$this->languageID = WCF::getUser()->languageID;
+		}
+		
+		if (!empty($this->i18nFields)) {
+			foreach ($this->i18nFields as $key => $value) {
+				if (is_numeric($key)) {
+					throw new SystemException("'".$this->className."::\$i18nFields' should contain key-value pairs. '".$value."' has a no or a numeric key.");
```
Gravatronics: Please do not use `SystemException` but one of PHP's own exception classes, probably `\LogicException`. Also, `$this->className` contains the DBO class name, not the list name.

wcfsetup/install/files/lib/data/I18nDatabaseObjectList.class.php
mutec: Better now? I'll clean up the commits if everything is correct.

wcfsetup/install/files/lib/data/I18nDatabaseObjectList.class.php
```diff
@@ -0,0 +1,72 @@
...
+		if (!empty($this->i18nFields)) {
+			if (count($this->i18nFields) != count(array_flip($this->i18nFields))) {
+				throw new SystemException("'".$this->className."::\$i18nFields' should unique values.");
+			}
+			
+			foreach ($this->i18nFields as $key => $value) {
+				if (is_numeric($key)) {
+					throw new SystemException("'" . $this->className . "::\$i18nFields' should contain alphanumeric key-value pairs. '" . $value . "' has an empty or numeric key.");
+				}
+				if (!preg_match('/^([0-9a-zA-Z]+)$/', $value)) {
```
TimWolla: I suggest a single if with this regex: `/^[a-z][a-zA-Z0-9]*$/` and the following message:

> Array keys of MyDBO::$i18nFields must start with a small letter and consist of letters and number only.

wcfsetup/install/files/lib/data/I18nDatabaseObjectList.class.php
```diff
@@ -0,0 +1,72 @@
...
+	public function __construct() {
+		parent::__construct();
+		
+		if ($this->languageID === null) {
+			$this->languageID = WCF::getUser()->languageID;
+		}
+		
+		if (!empty($this->i18nFields)) {
+			if (count($this->i18nFields) != count(array_flip($this->i18nFields))) {
+				throw new SystemException("'".$this->className."::\$i18nFields' should unique values.");
```
TimWolla: You already 👍 the comment by @Gravatronics, but it was marked outdated for some reason: I agree with his review, but suggest the more specific `DomainException` (which is a subclass of `LogicException`), instead.

wcfsetup/install/files/lib/data/I18nDatabaseObjectList.class.php
```diff
@@ -0,0 +1,72 @@
...
+	 */
+	public $i18nFields = [];
+	
+	/**
+	 * ID of the language that should be used
+	 * the value defaults to the current user's language
+	 *
+	 * @var integer
+	 */
+	public $languageID;
```
TimWolla: This does not allow you to change the languageID, because it already is embedded into the string once the constructor finished. I suggest adding a `$languageID` parameter to the constructor (and possibly a `setSortingLanguage` – I'm open for a better method name – method that allows you to change the language after construction, obviously use that method in the constructor instead of repeating code).

mutec: How do you want to change the ID after construction? It's an integer within a string that could contain anything. That would be possible if you handle the custom joins/selects and those for i18n in different strings/arrays and rebuild the query right before every execution like this:
```php
$tmpSelects = $this->sqlSelects; // the same for joins…
$this->sqlSelects .= (!empty($this->sqlSelects) ? ', ' : '') . $this->i18nSelects;
// query
$this->sqlSelects = $tmpSelects;
```

wcfsetup/install/files/lib/data/I18nDatabaseObjectList.class.php
mutec: I read the commit after pushing my commits and wanted to wait for your feedback before I mess up everything with commits. :)

TimWolla: Yes, that's what I guessed. But GitHub should not have collapsed his comment as outdated, because it is newer than all the commits *and* refers to a line existing in the current diff.

mutec:
see https://github.com/WoltLab/WCF/pull/2490
ma fault, sorry…

PR #2468, opened 2017-11-11
use `PreparedStatement#fetchSingleColumn` where possible

jens1o:
As far as I know, it frees up the connection for that statement, and makes the connection itself more stable.


TimWolla:
It does not make a difference if the statement object is immediately disposed by the garbage collector afterwards, but it is arguably cleaner, because it makes the intent clear.

PR #2473, opened 2017-11-19
Add `chars` suffix option

jens1o:
This is useful for codes that needs to be generated... Perhaps this could be used at the max signature length?


wcfsetup/install/lang/en.xml
```diff
@@ -1347,6 +1347,7 @@ When prompted for the notification URL for the instant payment notifications, pl
 		<item name="wcf.acp.option.category.general.payment.paidSubscription"><![CDATA[Paid Subscriptions]]></item>
 		<item name="wcf.acp.option.google_maps_api_key"><![CDATA[Browser API Key]]></item>
 		<item name="wcf.acp.option.google_maps_api_key.description"><![CDATA[Google provides a detailed description on how to get an API key <a href="{@$__wcf->getPath()}acp/dereferrer.php?url={'https://developers.google.com/maps/documentation/javascript/get-api-key'|rawurlencode}" class="externalURL">here</a>.]]></item>
+		<item name="wcf.acp.option.suffix.chars"><![CDATA[Char(s)]]></item>
```
dtdesign: It should read `Chars`.

jens1o:
Thanks :)

PR #2475, opened 2017-11-23
Jodel as media provider

mutec:


jens1o:
why? :(

dtdesign:
The list of pre-defined media providers are intentionally short, as they represent only major sites that have also been around for a while.

Users can always add additional providers at their own discrection.

PR #2483, opened 2017-11-26
Improve plain/text formatting of anchor tags and smilies

max-m:
Truncating URLs in `plain/text` mode leads to obvious problems, this behavior has been switched off.
Also smilies are converted to the appropriate smiley code in `plain/text` now, previously they would have been completely omitted.


max-m:
Converting images (not smilies) in `plain/text` into the following format could be a nice addition too:
```
<img src="https://example.com/example.png"> → [Image https://example.com/example.png]
<img src="https://example.com/example.png" alt="Example Image"> → Example Image [Image https://example.com/example.png]
```

PR #2490, opened 2017-11-29
i18n database object list class

mutec:
this makes it possible to provide a list of objects in a real alphabetical order
for performance-reasons it should be used when needed only (e.g. when sortField is title and title might contain language items)

### DIFF
--- a/wcfsetup/install/files/lib/data/I18nDatabaseObjectList.class.php
+++ b/wcfsetup/install/files/lib/data/I18nDatabaseObjectList.class.php
@@ -1,0 +1,60 @@
+<?php
+namespace wcf\data;
+use wcf\system\WCF;
+use wcf\util\StringUtil;
+
+/**
+ * Abstract class for a list of database objects with better sorting of i18n-based columns.
+ *
+ * @author	Florian Gail
+ * @copyright	2001-2017 WoltLab GmbH
+ * @license	GNU Lesser General Public License <http://opensource.org/licenses/lgpl-license.php>
+ * @package	WoltLabSuite\Core\Data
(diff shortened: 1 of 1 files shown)


wcfsetup/install/files/lib/data/I18nDatabaseObjectList.class.php
```diff
@@ -0,0 +1,60 @@
...
+	public function __construct($languageID = null) {
+		parent::__construct();
+		
+		if ($languageID === null) {
+			$languageID = WCF::getUser()->languageID;
+		}
+		
+		if (!empty($this->i18nFields)) {
+			if (count($this->i18nFields) !== count(array_flip($this->i18nFields))) {
+				throw new SystemException("Array values of '".$this->className."::\$i18nFields' must be unique.");
```
Gravatronics: Please also use an SPL exception here (`\DomainException`).

wcfsetup/install/files/lib/data/I18nDatabaseObjectList.class.php
```diff
@@ -0,0 +1,60 @@
...
+	 *
+	 * @example   [ 'title' => 'titleSortField' ]
+	 *
+	 * @var string[]
+	 */
+	public $i18nFields = [];
+	
+	/**
+	 * @inheritDoc
+	 * @param integer $languageID id of the language that should be used
```
Gravatronics: Please add a `@throws` tag for the thrown exceptions.

mutec: done :)

PR #2494, opened 2017-12-05
Add possibility to change sidebar open/close labels per page

ChristopherWalz:
A sidebar for example could contain a contact form. Instead of "Show sidebar" you want to show an explicit action label like "Contact us". This pull request makes that possible:

```
{capture assign='sidebarRightShow'}Contact us{/capture}
{capture assign='sidebarRightHide'}Close contact form{/capture}
```

### DIFF
--- a/com.woltlab.wcf/templates/footer.tpl
+++ b/com.woltlab.wcf/templates/footer.tpl
@@ -43,7 +43,13 @@
 			{/capture}
 				
 			{if $__sidebarRightContent|trim}
-				<aside class="sidebar boxesSidebarRight" data-show-sidebar="{lang}wcf.global.button.showSidebar{/lang}" data-hide-sidebar="{lang}wcf.global.button.hideSidebar{/lang}">
+				{if !$sidebarRightShow|isset}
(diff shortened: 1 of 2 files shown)


ChristopherWalz:
Can I expect that to be merged before the final release? Otherwise I need to change the templates for my project on my own.

com.woltlab.wcf/templates/footer.tpl
```diff
@@ -43,7 +43,13 @@
 			{/capture}
 				
 			{if $__sidebarRightContent|trim}
-				<aside class="sidebar boxesSidebarRight" data-show-sidebar="{lang}wcf.global.button.showSidebar{/lang}" data-hide-sidebar="{lang}wcf.global.button.hideSidebar{/lang}">
+				{if !$sidebarRightShow|isset}
```
dtdesign: Rename this variable to `$__sidebarRightShow` in order to avoid conflicts with existing variables. This applies to `$sidebarRightHide` (new: `$__sidebarRightHide`) as well as the variables in the `header.tpl` too.

Additionally, turn these `if`-statements into one-liners, splitting them across three lines doesn't really improve readability.

PR #2500, opened 2017-12-18
update instructions from WSC 3.0.11 pl 1 to WSC 3.1

mutec:


dtdesign:
Thank you for your effort, but unfortunately this isn't helpful at this time. I'll have to run extensive tests to get the upgrade path to work, in particular when running it against real data, like I already did back then. Your commit contains untested changes (`ROP TABLE`, I'm looking at you!), so I cannot rely on them to be working.

I will take care of the upgrade path once everything is ready and no more changes are being done. Adding the instructions at this time offers no value, because they may or may not require changes in the future, also it would enable upgrade support way earlier than intended.

mutec:
It's tested, seems like the WSC doesn't care about that illegal MySQL-Statement.

PR #2503, opened 2017-12-19
use UserProfileRuntimeCache in UserBirthdayAction

jens1o:
I took a look using the developer tools and saw that in one request multiple user profiles has been requests multiple times. Let's try to avoid that.


dtdesign:
That looks like a micro-optimization to me, including an increase in code complexity. Do you have any numbers?

jens1o:
Well, only if you use one of my plugins, which does additional checks before outputting sensitive data and thus I'm also in need of the user profiles, and thus I saw that those were fetched twice.

PR #2512, opened 2018-01-07
use no cookie version of youtube in YouTubeBBCodeMediaProvider

jens1o:
see https://community.woltlab.com/thread/264689-was-wird-woltlab-konkret-an-tools-f%C3%BCr-die-dsgvo-anbieten/?postID=1649529#post1649529


dtdesign:
Have you verified, that it does not break on mobile devices again? It was removed quite some time ago in c38e96cfb7ee5386131bb932f34c5eebb377483e, because it behaves differently than the default youtube player.

jens1o:
At least for Firefox Focus and Firefox Mobile on Android 7 it's working fine. :)

PR #2513, opened 2018-01-08
Fix displayed day for timestamps

neon-dev:
If the server time compared to the users system time is off by some seconds, the displayed day for timestamps could be incorrect.
Only affected timestamps at/near midnight.
See https://community.woltlab.com/thread/264879-fehlerhafte-sprechende-zeitangaben/

### DIFF
--- a/wcfsetup/install/files/js/WoltLabSuite/Core/Date/Time/Relative.js
+++ b/wcfsetup/install/files/js/WoltLabSuite/Core/Date/Time/Relative.js
@@ -55,7 +55,11 @@ define(['Dom/ChangeListener', 'Language', 'WoltLabSuite/Core/Date/Util', 'WoltLa
 			
 			var date = new Date();
 			var timestamp = (date.getTime() - date.getMilliseconds()) / 1000;
-			if (_offset === null) _offset = timestamp - window.TIME_NOW;
+			// initialize server / client time offset in minute granularity
+			if (_offset === null) {
+				_offset = timestamp - window.TIME_NOW;
+				_offset = (_offset >= 0 ? Math.floor(_offset / 60) : Math.ceil(_offset / 60)) * 60;
+			}
 			
 			for (var i = 0, length = _elements.length; i < length; i++) {
 				var element = _elements[i];


jens1o:
@dtdesign Is there a reason for this?

dtdesign:
Although it is still left unclear what the reason of this date mismatch is, using arbitrary number rounding isn't going to solve it. The suggested change only works for offsets <= 60 seconds, but actual differences can span up to 10 minutes in total due to incorrect system times.

That said, this PR does not fix anything, but merely hides the mismatch for small offsets.

neon-dev:
Ok after putting more thought into it, I see two problems:
1. The offset calculation has a rather large margin of error, since the server time is created before rendering the page and the client date is instantiated pretty late in the rendering process due to the nature of JS in browsers.
2. The ```midnight``` date is created on client side, without offset adjustment, but is subtracted with an adjusted timestamp ([see](https://github.com/WoltLab/WCF/blob/068cf48a9e7da544253953e2d3adfff048b8d3dd/wcfsetup/install/files/js/WoltLabSuite/Core/Date/Time/Relative.js#L90))

I guess after both issues (especially no. 2) are adressed, the problem should disappear. Not sure about what you could do to fix no. 1 since we can not figure out how much time passed between creating the server's and the client's date. Polling the servertime from the script would be an option, but maybe an overkill and still not completely exact, although almost.

dtdesign:
Actually, issue no. 1 is a non-issue because this is already taken care of by the offset, it contains the difference between server and client timestamp. This is quite important, because it accounts for both different timezones and any differences between the server's "now" and the client's "now" on page load. That said, the offset shifts the calculation towards the server time and syncs up this why (at least with a <1 second precision).

The bottom line is, the calculation for midnight appears to be indeed off and needs to be adjusted to include the offset. I'll take care of it at some point, unless you would like to address this on your own, in which case a PR against WCF 2.1 is sufficient, I'll merge the changes to WSC 3.0 and 3.1 anyway.

neon-dev:
As I understand, the ```_offset``` is only to mitigate differences between the server's system clock and the clients and should not consider page load delays in the client time offset.
An offset between two timestamps is only accurate, if both are created at the exact same point in time.
So If the server's system clock says it's 08:04:50 UTC when client's says it's 08:04:58 UTC, we have an offset of 8s. But if the client millis are captured 2 seconds later than the servers, we now have an offset of 10s, which moves everything 2s too far back when adjusting dates.
As for the timezones, you maybe got confused like I did two weeks ago in the forums. They don't affect the timestamps at all and are no factor in this topic :D

If I find the time to properly fix and test no. 2 in the next days, I'll create a new PR in WCF 2.1 as you said. Unfortunately I cannot promise anything, since I'm pretty busy these days.
Thanks

PR #2514, opened 2018-01-08
Fix displayed day for timestamps (WCF 2.1)

neon-dev:
If the server time compared to the users system time is off by some seconds, the displayed day for timestamps could be incorrect.
Only affected timestamps at/near midnight.
See https://community.woltlab.com/thread/264879-fehlerhafte-sprechende-zeitangaben/


neon-dev:
This fix can also be applied to WCF 2.0: https://github.com/WoltLab/WCF/blob/2.0/wcfsetup/install/files/js/WCF.js#L3129

PS: Seems like Travis is misconfigured or something. The build fail message is a false error.

dtdesign:
PR suffers from the same issue as #2513

PR #2536, opened 2018-02-11
Typo

SoftCreatR:


TimWolla:
Redid the patch manually, as the target branch was incorrect and the patch did not apply cleanly in `3.0`. Thanks.

PR #2540, opened 2018-02-16
Experimental fix for table name and alias resolution

dtdesign:
Relying on a local static variable is dangerous, because it is bound to the implementing class, but once set, it becomes fixed for all derived classes too.

This is similar to the `wcf\system\SingletonFactory` work-around that uses a lookup table rather than LSB fields for the same reasons.

### DIFF
--- a/wcfsetup/install/files/lib/data/DatabaseObject.class.php
+++ b/wcfsetup/install/files/lib/data/DatabaseObject.class.php
@@ -42,24 +42,6 @@ abstract class DatabaseObject implements IStorableObject {
 	protected static $sortOrder = null;
 	
 	/**
-	 * The list of derived database table names based on the class name.
-	 * 
-	 * WARNING: This is strictly an internal lookup table. DO NOT USE IT.
-	 * 
-	 * @var string[] 
-	 */
-	protected static $_derivedDatabaseTableName = [];
-	
-	/**
-	 * The list of derived database table aliases based on the class name.
-	 * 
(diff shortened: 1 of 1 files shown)


wcfsetup/install/files/lib/data/DatabaseObject.class.php
```diff
@@ -41,6 +41,24 @@
 	 */
 	protected static $sortOrder = null;
 	
+	/**
+	 * The list of derived database table names based on the class name.
+	 * 
+	 * WARNING: This is strictly an internal lookup table. DO NOT USE IT.
```
TimWolla: As you are using an array now anyway: You *could* leave this as a static variable inside the method.

dtdesign: Could be worth it, bonus points for hiding the internals, but I haven't tried it. Did you check if it behaves properly with derived classes, especially given a random initialization order?

TimWolla: Looking good: https://3v4l.org/gefrF.

See also this top comment: http://php.net/manual/en/language.variables.scope.php#59676

dtdesign:
Changed it into a local static variable and everything appears to be working just fine, including derived classes.

@TimWolla Please check these changes again, preferably in a test installation, and merge them at your own discrection. I'm really not feeling comfortable making changes to a core class that close to a release :D

PR #2548, opened 2018-03-12
Add new "SameSite"-cookie-attribute

Scrayos:
Hey,

this adds the "[SameSite](https://www.owasp.org/index.php/SameSite)"-cookie-attribute. This attribute was already accepted (and implemented) by Chrome, Firefox and Opera (at least). It was added to effectively tackle a few CSRF vulnerabilities by not sending cookies on sub-requests or only sending them under specific circumstances.

I added this to the HeaderUtil-setCookie utility, so it's easy to maintain and I also included a config option for the different operation modes, aswell as the "off", to retain backwards-compatibility.

This is my first PR to this project, so I expect I gotta correct a few things for this getting merged, I am, however, willing to make those adjustments, as I'm really looking forward to having this within my favourite framework! :)

### DIFF
--- a/com.woltlab.wcf/option.xml
+++ b/com.woltlab.wcf/option.xml
@@ -618,8 +618,7 @@ private:wcf.acp.option.exception_privacy.private</selectoptions>
 				<optiontype>select</optiontype>
 				<defaultvalue>none</defaultvalue>
 				<selectoptions>none:wcf.acp.option.cookie_same_site_attribute.none
-					lax:wcf.acp.option.cookie_same_site_attribute.lax
-					strict:wcf.acp.option.cookie_same_site_attribute.strict</selectoptions>
+lax:wcf.acp.option.cookie_same_site_attribute.lax</selectoptions>
 			</option>
 			<!-- /general.system.cookie-->
 			
--- a/com.woltlab.wcf/option.xml
+++ b/com.woltlab.wcf/option.xml
@@ -613,6 +613,14 @@ private:wcf.acp.option.exception_privacy.private</selectoptions>
 				<defaultvalue>wsc31_</defaultvalue>
 				<validationpattern>.+</validationpattern>
 			</option>
+			<option name="cookie_same_site_attribute">
+				<categoryname>general.system.cookie</categoryname>
+				<optiontype>select</optiontype>
+				<defaultvalue>none</defaultvalue>
+				<selectoptions>none:wcf.acp.option.cookie_same_site_attribute.none
+					lax:wcf.acp.option.cookie_same_site_attribute.lax
+					strict:wcf.acp.option.cookie_same_site_attribute.strict</selectoptions>
+			</option>
 			<!-- /general.system.cookie-->
 			
 			<!-- general.system.http -->
--- a/wcfsetup/install/files/lib/util/HeaderUtil.class.php
+++ b/wcfsetup/install/files/lib/util/HeaderUtil.class.php
@@ -49,8 +49,7 @@ final class HeaderUtil {
 		}
 		
 		$sameSitePolicy = '';
-		$sameSiteOption = COOKIE_SAME_SITE_ATTRIBUTE;
-		if ($sameSiteOption != 'none') {
+		if (COOKIE_SAME_SITE_ATTRIBUTE != 'none') {
 			$sameSitePolicy = '; SameSite='.COOKIE_SAME_SITE_ATTRIBUTE;
 		}
 		
--- a/wcfsetup/install/files/lib/util/HeaderUtil.class.php
+++ b/wcfsetup/install/files/lib/util/HeaderUtil.class.php
@@ -8,7 +8,7 @@ use wcf\system\WCF;
 
 /**
  * Contains header-related functions.
- * 
+ *
  * @author	Marcel Werk
  * @copyright	2001-2018 WoltLab GmbH
  * @license	GNU Lesser General Public License <http://opensource.org/licenses/lgpl-license.php>
@@ -35,7 +35,7 @@ final class HeaderUtil {
 	
 	/**
 	 * Alias to php setcookie() function.
-	 * 
+	 *
 	 * @param	string		$name
 	 * @param	string		$value
 	 * @param	integer		$expire
@@ -47,8 +47,15 @@ final class HeaderUtil {
 		if ($addDomain && strpos($cookieDomain, ':') !== false) {
 			$cookieDomain = explode(':', $cookieDomain, 2)[0];
 		}
+	    
+	    	$sameSitePolicy = "";
+	    	$sameSiteOption = COOKIE_SAME_SITE_ATTRIBUTE;
+	    	if ($sameSiteOption != "none") {
+	    		$sameSitePolicy = "; SameSite=".COOKIE_SAME_SITE_ATTRIBUTE;
+		}
 		
-		@header('Set-Cookie: '.rawurlencode(COOKIE_PREFIX.$name).'='.rawurlencode($value).($expire ? '; expires='.gmdate('D, d-M-Y H:i:s', $expire).' GMT; max-age='.($expire - TIME_NOW) : '').'; path=/'.($addDomain ? '; domain='.$cookieDomain : '').(RouteHandler::secureConnection() ? '; secure' : '').'; HttpOnly', false);
+		
+		@header('Set-Cookie: '.rawurlencode(COOKIE_PREFIX.$name).'='.rawurlencode($value).($expire ? '; expires='.gmdate('D, d-M-Y H:i:s', $expire).' GMT; max-age='.($expire - TIME_NOW) : '').'; path=/'.($addDomain ? '; domain='.$cookieDomain : '').(RouteHandler::secureConnection() ? '; secure' : '').'; HttpOnly'.$sameSitePolicy, false);
 	}
 	
 	/**
@@ -107,7 +114,7 @@ final class HeaderUtil {
 	
 	/**
 	 * Parses the rendered output.
-	 * 
+	 *
 	 * @param	string		$output
 	 * @return	string
 	 */
@@ -156,10 +163,10 @@ final class HeaderUtil {
 	
 	/**
 	 * Redirects the user agent to given location.
-	 * 
+	 *
 	 * @param	string		$location
 	 * @param	boolean		$sendStatusCode
-	 * @param	boolean		$temporaryRedirect 
+	 * @param	boolean		$temporaryRedirect
 	 */
 	public static function redirect($location, $sendStatusCode = false, $temporaryRedirect = true) {
 		if ($sendStatusCode) {
@@ -172,7 +179,7 @@ final class HeaderUtil {
 	
 	/**
 	 * Does a delayed redirect.
-	 * 
+	 *
 	 * @param	string		$location
 	 * @param	string		$message
 	 * @param	integer		$delay
--- a/wcfsetup/install/files/lib/util/HeaderUtil.class.php
+++ b/wcfsetup/install/files/lib/util/HeaderUtil.class.php
@@ -47,13 +47,12 @@ final class HeaderUtil {
 		if ($addDomain && strpos($cookieDomain, ':') !== false) {
 			$cookieDomain = explode(':', $cookieDomain, 2)[0];
 		}
-	    
-	    	$sameSitePolicy = "";
-	    	$sameSiteOption = COOKIE_SAME_SITE_ATTRIBUTE;
-	    	if ($sameSiteOption != "none") {
-	    		$sameSitePolicy = "; SameSite=".COOKIE_SAME_SITE_ATTRIBUTE;
-		}
 		
+		$sameSitePolicy = "";
+		$sameSiteOption = COOKIE_SAME_SITE_ATTRIBUTE;
+		if ($sameSiteOption != "none") {
+			$sameSitePolicy = "; SameSite=".COOKIE_SAME_SITE_ATTRIBUTE;
+		}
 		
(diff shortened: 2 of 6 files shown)


com.woltlab.wcf/option.xml
```diff
@@ -613,6 +613,14 @@ private:wcf.acp.option.exception_privacy.private</selectoptions>
 				<defaultvalue>wsc31_</defaultvalue>
 				<validationpattern>.+</validationpattern>
 			</option>
+			<option name="cookie_same_site_attribute">
+				<categoryname>general.system.cookie</categoryname>
+				<optiontype>select</optiontype>
+				<defaultvalue>none</defaultvalue>
+				<selectoptions>none:wcf.acp.option.cookie_same_site_attribute.none
+					lax:wcf.acp.option.cookie_same_site_attribute.lax
```
TimWolla: Do not indent these.

wcfsetup/install/files/lib/util/HeaderUtil.class.php
```diff
@@ -8,7 +8,7 @@
 
 /**
  * Contains header-related functions.
- * 
+ *
```
TimWolla: Do not remove these.

wcfsetup/install/files/lib/util/HeaderUtil.class.php
```diff
@@ -47,8 +47,15 @@ public static function setCookie($name, $value = '', $expire = 0) {
 		if ($addDomain && strpos($cookieDomain, ':') !== false) {
 			$cookieDomain = explode(':', $cookieDomain, 2)[0];
 		}
+	    
+	    	$sameSitePolicy = "";
```
TimWolla: This value is immediately reassigned.

com.woltlab.wcf/option.xml
```diff
@@ -613,6 +613,14 @@ private:wcf.acp.option.exception_privacy.private</selectoptions>
...
+					strict:wcf.acp.option.cookie_same_site_attribute.strict</selectoptions>
```
TimWolla: Remove the `strict` option, it provides horrible UX.

wcfsetup/install/files/lib/util/HeaderUtil.class.php
```diff
@@ -47,8 +47,15 @@ public static function setCookie($name, $value = '', $expire = 0) {
...
+	    	$sameSiteOption = COOKIE_SAME_SITE_ATTRIBUTE;
+	    	if ($sameSiteOption != "none") {
```
TimWolla: I wonder if we forcibly should set some cookies to `SameSite` (notably the `cookieHash` cookie).

TimWolla: Use single quotes for regular strings.

Scrayos: Would be pleased to add it, if there's consent about this.

wcfsetup/install/files/lib/util/HeaderUtil.class.php
Scrayos: Well, it's only reassigned if $sameSiteOption is not 'none'. Otherwise it never gets assigned, and I believe it is a good practice to always declare variables that are used.

Scrayos:
@TimWolla I implemented the requested changes/attached comments to two of your requested changes.

wcfsetup/install/lang/de.xml
```diff
@@ -1051,6 +1051,10 @@
 		<item name="wcf.acp.option.cookie_path"><![CDATA[Cookiepfad]]></item>
 		<item name="wcf.acp.option.cookie_path.description"><![CDATA[Der Cookiepfad wird absolut zum Document Root angegeben - also z.B. "/forum" für https://www.woltlab.de/forum.]]></item>
 		<item name="wcf.acp.option.cookie_prefix"><![CDATA[Präfix für Cookienamen]]></item>
+		<item name="wcf.acp.option.cookie_same_site_attribute"><![CDATA[Cookies vor externen Zugriffen schützen]]></item>
+		<item name="wcf.acp.option.cookie_same_site_attribute.description"><![CDATA[Fügt beim Setzen von Cookies zusätzlich das <a href="{@$__wcf->getPath()}acp/dereferrer.php?url={'https://en.wikipedia.org/wiki/HTTP_cookie#SameSite_cookie|rawurlencode}" class="externalURL">„SameSite“</a> Attribut hinzu, um das Nutzen der Cookies bei externer Referenzierung dieser Seite zu verhindern.]]></item>
+		<item name="wcf.acp.option.cookie_same_site_attribute.none"><![CDATA[Deaktiviert]]></item>
+		<item name="wcf.acp.option.cookie_same_site_attribute.lax"><![CDATA[Locker]]></item>
```
TimWolla: Add the literal names of the values. Something like „Lasch (Lax)“ should do.

wcfsetup/install/files/lib/util/HeaderUtil.class.php
```diff
@@ -48,7 +48,13 @@ public static function setCookie($name, $value = '', $expire = 0) {
 			$cookieDomain = explode(':', $cookieDomain, 2)[0];
 		}
 		
-		@header('Set-Cookie: '.rawurlencode(COOKIE_PREFIX.$name).'='.rawurlencode($value).($expire ? '; expires='.gmdate('D, d-M-Y H:i:s', $expire).' GMT; max-age='.($expire - TIME_NOW) : '').'; path=/'.($addDomain ? '; domain='.$cookieDomain : '').(RouteHandler::secureConnection() ? '; secure' : '').'; HttpOnly', false);
+		$sameSitePolicy = '';
+		$sameSiteOption = COOKIE_SAME_SITE_ATTRIBUTE;
```
TimWolla: continuing my comment in the fresh diff, because it is being collapsed otherwise:

> Well, it's only reassigned if $sameSiteOption is not 'none'. Otherwise it never gets assigned, and I believe it is a good practice to always declare variables that are used.

I misread one of the variable names as both being the same. Do not use a temporary variable just for the `if` expression then (i.e. remove this line).

Scrayos: Ah, yeah. No I realize that it's not needed anymore. It was used previously, because - in my first attempt - I checked with empty(). And that requires its parameter to be a variable. Gonna remove the variable of course.

Scrayos:
@TimWolla I resolved everything you requested, once again.

dtdesign:
I'm leaning towards not implementing it at all:
1. All POST requests already use a CSRF token. Regular forms have to include it (this is a requirement of the `AbstractForm` class) and the AJAX-API strictly requires the token too, even for simple "GET"-style requests.
2. It will only offer some sort of protection if it is implemented by the browser, so we cannot drop the CSRF tokens anyway. But with proper CSRF tokens (and we implement/force them at very low levels!) these lax-style cookies do not add any additional security. It is very easy to overestimate the choice that users have with their browsers, just think of the average workspace environment.
3. It is a rather obscure mechanic that is also not well-known among most people, and I'm pretty sure this applies to the average developer too. For example, we're setting `X-Frame-Options: SAMEORIGIN` for a very long time already and there have been a lot of error reports that eventually resolve to this setting.

The biggest issue is the combination of 2. and 3., where it will cause issues for users (and thus leading to a bad experience) without adding any significant benefit. This is especially true with setups where our software is only a part of the website and have to work together with other software pieces, most of the time they're sharing cookies. Breaking their setup with a feature that does not offer anything of value (again: our software _is_ protected against CSRF _by design_) is simply not worth it.

Scrayos:
> All POST requests already use a CSRF token. Regular forms have to include it (this is a requirement of the AbstractForm class) and the AJAX-API strictly requires the token too, even for simple "GET"-style requests.

That is good to know, and I absolutely understand, that this already assures CSRF isn't possible anymore. I don't know too much about how one may (accidentally) break the requirements of an AbstractForm, but I guess it's some kind of "design by contract", so one COULD in theory work around it, but it is neither supported, nor advised. Anyways, someone could also break the security in much more direct ways, so this shouldn't be a problem.


> It will only offer some sort of protection if it is implemented by the browser, so we cannot drop the CSRF tokens anyway. But with proper CSRF tokens (and we implement/force them at very low levels!) these lax-style cookies do not add any additional security. It is very easy to overestimate the choice that users have with their browsers, just think of the average workspace environment.

As I've written, it was already widely adopted by the browser, but of course we cannot rely on this attribute alone, because - as you've written - by no means, everyone always runs the latest version of a modern browser. I think however, that it is a good measure to let the browser know, that we intend to block these kind of requests. I deem the SameSite cookie attribute smoother, in terms of message flow. While a cookie with this attribute won't even be sent to the site, the CSRF-Token-approach feels a lot more like an error, instead of a cleanly handled, intended behavior, at least in my perspective.

I'm having troubles explaining what I mean, but my basic point is, that this cookie-attribute actively tells the browser what kind of behavior we want to achieve, and the browser helps us to achieve it, while the CSRF-Token is more like "Ooops, that shouldn't have happened in the first place!". While this - of course - does not add much (if any) to the security, I think it improves the integration into sophisticated security techniques.


> It is a rather obscure mechanic that is also not well-known among most people, and I'm pretty sure this applies to the average developer too. For example, we're setting X-Frame-Options: SAMEORIGIN for a very long time already and there have been a lot of error reports that eventually resolve to this setting.

That may be right, but I think with the default setting of "none", we're already reducing the error-rate by a lot, and as you stated, "lax" is so lax, that a person wouldn't run into many problems using it, I think. It should also integrate well with other software-pieces on the same domain, as it permits _everything_ within that domain. And their setup will never be broken, if they do not break it theirselves, because - as already stated - it's disabled by default.

My strongest point however, is about compliance and transparency. While one has to know the internals about the CSRF-token, to research more about how it is implemented, this is a technique, that is already adopted and taken into consideration by browser-influencers like Mozilla. They take SameSite cookies already into their [scoring for observatory](https://github.com/mozilla/http-observatory/blob/master/httpobs/docs/scoring.md) and actively encourage developers to adopt this attribute. The transparency is about everyone (including scan-tools, that are often used) being able to directly see, that this cookie is protected against CSRF-usage.

(Sorry for such a long answer)

dtdesign:
> I think however, that it is a good measure to let the browser know, that we intend to block these kind of requests.
It's nice, but at the same time doesn't increase security at all. Our cookies are set to `HttpOnly` anyway, so even JavaScript cannot take advantage of additional cookies that aren't supposed to be present.

It's tempting to adopt it for the sake of consistency, after all that's what CSRF tokens were invented for? Actually, it is even bad, because it will cause requests to silently fail without giving any hints whatsoever on why things do not work as expected.

For instance, have you ever had to deal with both secure and "non-secure" cookies for the same domain and path? You would assume that when visiting the site using TLS that the browser sends only the secure cookies? No, it also includes the non-secure cookies. However, things get nasty when the same cookie exists for both modes, eventually causing the browser to arbitrarily pick one of them. Even worse, you cannot unset a non-secure cookie using a secure-cookie, you'll have to explicitly invalidate it using a non-secure cookie. Oh, and the browser doesn't tell you with what kind of cookie you're dealing.

Silently failing browser requests are a hard-to-debug nightmare.

> That may be right, but I think with the default setting of "none"
I'm sorry, but this equals "does not exist". To put it into perspective, if you have 100 people, less than 10 people will actually enable it and half of them don't even understand the full implications of it (of which half of them will encounter issues because of it).

We're left with two choices: Either implement it and set it to "lax" by default, or not implement it at all. There is no in-between.

Scrayos:
> it is even bad, because it will cause requests to silently fail without giving any hints whatsoever on why things do not work as expected.
[...]
Silently failing browser requests are a hard-to-debug nightmare.

I don't think this is any harder to debug than failed CSRF-token-challenges. Both, CSRF-token and SameSite-cookies work with a check for the existence of cookies (or the absence of those). While I agree, that a failed token-challenge makes it more clear, that this is an unintended behavior, I am of the opinion, that dealing with those users, as if they're unauthorized is more reliable, than throwing an error, if something fails.

That may sound counter-intuitive, but clinging to "If it's not this, that or this kind of request, no cookies are sent" is easier than "Cookies are sent under all circumstances, but if this value isn't sent aswell, we'll just answer with an error.", at least in my opinion.


> For instance, have you ever had to deal with both secure and "non-secure" cookies for the same domain and path? You would assume that when visiting the site using TLS that the browser sends only the secure cookies? No, it also includes the non-secure cookies. However, things get nasty when the same cookie exists for both modes, eventually causing the browser to arbitrarily pick one of them. Even worse, you cannot unset a non-secure cookie using a secure-cookie, you'll have to explicitly invalidate it using a non-secure cookie. Oh, and the browser doesn't tell you with what kind of cookie you're dealing.

No, fortunately I only work on HSTS-preloaded domains, so I never ran into these kind of problems. But I don't think this problem would get any worse/more unreliable with this SameSite cookie attribute. This attribute is completely independent from HttpOnly and secure.


> I'm sorry, but this equals "does not exist". To put it into perspective, if you have 100 people, less than 10 people will actually enable it and half of them don't even understand the full implications of it (of which half of them will encounter issues because of it).
We're left with two choices: Either implement it and set it to "lax" by default, or not implement it at all. There is no in-between.

In that case I'd - obviously - still vote for "set it to „lax“ by default". I don't see any potential for errors, caused by that, as the policy is (like the name implies) so lax. It should cover all use cases, that I can think of at the moment. [link, prepender and form get](https://www.sjoerdlangkemper.nl/2016/04/14/preventing-csrf-with-samesite-cookie-attribute/#same-site-cookie-attribute) are still allowed, and I think the `X-Frame-Options: SAMEORIGIN` you already mentioned earlier have way more potential to mess ones setup up, and they still made it into the WCF.

dtdesign:
Scheduled for re-evaluation for 3.2, there will be no such change for the 3.1 series.

dtdesign:
Implemented in #3593

StevenLeonKara:
anyone know a quick fix

PR #2549, opened 2018-03-29
[Optimization] Replace strstr with strpos because it’s faster

LKDevelopment:
As mentioned from @dtdesign in the commit fef5bd1f21876de170d181e7c81b92ff893efa6f (see comments) this little pr replaces all `strstr` calls with `strpos` because the docs for the strstr function (https://secure.php.net/manual/en/function.strstr.php) says that it is better to use strpos if you only want to determine if a string is in another string. I think doesn't brings so much performance, but actually on all other checks of this type strpos is already used, so why don't "optimize" the last few findings?

Signed-off-by: Lukas Kämmerling <kontakt@lukas-kaemmerling.de>

### DIFF
--- a/wcfsetup/install/files/lib/action/PaypalCallbackAction.class.php
+++ b/wcfsetup/install/files/lib/action/PaypalCallbackAction.class.php
@@ -43,7 +43,7 @@ class PaypalCallbackAction extends AbstractAction {
 				throw new SystemException('connection to paypal.com failed: ' . $e->getMessage());
 			}
 			
-			if (strstr($content, "VERIFIED") === false) {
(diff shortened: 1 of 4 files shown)


LKDevelopment:
Looks like the tests are failing because of an issue with packagist:
````
 [Composer\Downloader\TransportException]                                                                                                                                     
  The "http://packagist.org/p/provider-2018-01%241b64fe7375236daced9d549cbf9ae687f9ec01bb353d84e461c5af3136163641.json" file could not be downloaded (HTTP/1.1 404 Not Found)
````

LKDevelopment:
@dtdesign @timwollna

How often do you Review PRs :)?

PR #2555, opened 2018-04-11
Add width/height properties for default avatars

SoftCreatR:
Under certain circumstances*, it is possible, that the style tag will be stripped from from SVG images. This leads to a display of very tiny avatar images as shown below:

image

Since `width` and `height` are valid properties and respected by every browser, it's not the worst idea to include them.

* = An example scenario is the Mirage option from Cloudflare.

### DIFF
--- a/wcfsetup/install/files/lib/data/user/avatar/DefaultAvatar.class.php
+++ b/wcfsetup/install/files/lib/data/user/avatar/DefaultAvatar.class.php
@@ -79,7 +79,7 @@ SVG;
 	public function getImageTag($size = null) {
 		if ($size === null) $size = $this->size;
 		
-		return '<img src="'.StringUtil::encodeHTML($this->getURL($size)).'" style="width: '.$size.'px; height: '.$size.'px" alt="" class="userAvatarImage">';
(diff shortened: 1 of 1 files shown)


wcfsetup/install/files/lib/data/user/avatar/DefaultAvatar.class.php
```diff
@@ -79,7 +79,7 @@ public function getURL($size = null) {
 	public function getImageTag($size = null) {
 		if ($size === null) $size = $this->size;
 		
-		return '<img src="'.StringUtil::encodeHTML($this->getURL($size)).'" style="width: '.$size.'px; height: '.$size.'px" alt="" class="userAvatarImage">';
+		return '<img src="'.StringUtil::encodeHTML($this->getURL($size)).'" style="width: '.$size.'px; height: '.$size.'px" width="'.$size.'" height="'.$size.'" alt="" class="userAvatarImage">';
```
TimWolla: Remove the `style` tag then (it's redundant).

dtdesign:
Wouldn't it make much more sense to get in touch with Cloudflare and tell them to fix Mirage instead? I mean, it's labeled as "BETA" for a reason and it is designed to reduce the images without breaking the layout - which at this point it fails to accomplish.

SoftCreatR:
Already did that and from what i know, it's a known issue OR works as designed.

PR #2558, opened 2018-04-25
Remove deprecated/removed permissions

SoftCreatR:
As of April 4, 2018 the permissions `user_about_me` and `user_website` have been deprecated (see https://developers.facebook.com/docs/graph-api/changelog/breaking-changes/#4-4-2018).

App developers will see a notice, that they should update their app to no longer request these permissions. 

Since there isn't any alternative (or i havent found any), it can be completely removed.


SoftCreatR:
Heads up: This issue prevents users from logging in using Facebook due to a bug at Facebooks end which leads to an endless loop.

TimWolla:
Applied manually because of the incorrect target branch.

Thanks.

PR #2561, opened 2018-04-28
Change the DefaultAvatar’s size to 128x128 pixels

max-m:
Change the DefaultAvatar’s size to a more reasonable size of 128x128 pixels.
Throughout the Suite Core avatars are usually embedded via the `getImageTag` function
but when using the direct avatar URL a larger image is usually wanted.

Firefox for example does not upscale images used in `window.Notification`
thus the avatar would be rendered as a tiny 16x16 image in this case.
This change allows to downscale accordingly in this case.

### DIFF
--- a/wcfsetup/install/files/lib/data/user/avatar/DefaultAvatar.class.php
+++ b/wcfsetup/install/files/lib/data/user/avatar/DefaultAvatar.class.php
@@ -56,7 +56,7 @@ class DefaultAvatar implements IUserAvatar {
 			
 			// the <path> is basically a shorter version of a <rect>
 			$svg = <<<SVG
-<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 16 16" width="16" height="16"><path fill="#{$backgroundColor}" d="M0 0h16v16H0z"/><text x="8" y="8" fill="#{$textColor}" text-anchor="middle" dy=".3em" font-family="Arial" font-size="7">{$text}</text></svg>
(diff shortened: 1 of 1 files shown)


wcfsetup/install/files/lib/data/user/avatar/DefaultAvatar.class.php
```diff
@@ -56,7 +56,7 @@ public function __construct($username = '') {
 			
 			// the <path> is basically a shorter version of a <rect>
 			$svg = <<<SVG
-<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 16 16" width="16" height="16"><path fill="#{$backgroundColor}" d="M0 0h16v16H0z"/><text x="8" y="8" fill="#{$textColor}" text-anchor="middle" dy=".3em" font-family="Arial" font-size="7">{$text}</text></svg>
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 16 16" width="128" height="128"><path fill="#{$backgroundColor}" d="M0 0h16v16H0z"/><text x="8" y="8" fill="#{$textColor}" text-anchor="middle" dy=".3em" font-family="Arial" font-size="7">{$text}</text></svg>
```
dtdesign: Have you checked this change in IE11? I remember it being very picky on different dimensions for the svg and the `viewBox`?

SweetieChry: Looking good in Internet Explorer 11

unknown

PR #2697, opened 2018-06-12
Add guard against invalid dates in Date/Picker.init()

max-m:
See https://community.woltlab.com/thread/270307-fehlerhafte-f%C3%BCllung-eines-profilfeldes/


dtdesign:
Thanks

PR #2701, opened 2018-07-05
Make it possible for plugins to return different box content

ChristopherWalz:
Make it possible for plugins to return different box content

### DIFF
--- a/wcfsetup/install/files/lib/data/box/Box.class.php
+++ b/wcfsetup/install/files/lib/data/box/Box.class.php
@@ -261,6 +261,7 @@ class Box extends DatabaseObject {
 	 * @return	string
 	 */
 	public function render() {
+		$data = [];
 		EventHandler::getInstance()->fireAction($this, 'beforeRender', $data);
 		
 		// an event decided to return something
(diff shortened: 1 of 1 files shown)


dtdesign:
Would you be so kindly to explain the motivation behind this change, in particular what are you trying to achieve with it? Off-loading the entire process to a plugin isn't transparent and can be hard to debug.

PS: Please consider testing the changes before submitting a PR, `$data` is never defined.

ChristopherWalz:
I want to disable the initial rendering (and data fetching) of specific boxes and load the content via AJAX after the page has been fully loaded.

dtdesign:
That sounds a bit like the original box author would want to modify their code instead to have the data loaded asynchronously? After all, it isn't the purpose of the API to fix other author's code when they could just modify it themselves. Additionally, this approach is rather dangerous as it makes assumptions on the internal state of the box and its environment, such as when the box changes it behavior depending on the request object.

ChristopherWalz:
How else could I solve what I'm trying to achieve?

NetzwergX:
Extend the original box, add your desired behavior. That way you have proper control over whats happening and the EL does not need tor rely on assumptions.

ChristopherWalz:
Well I wanted to let the admin decide, which boxes should be loaded asynchronously. Guess it's not possible to provide this functionality with the current code/ELs.

NetzwergX:
Can't you replace the loaded box ad-hoc in the handler? I haven't checked whether thats possible, but I'd simply replace the implementation live. Or you take some event that is before he box loading and prepend an auto-loader that loads your own implementation. There are ways to get around this.

PR #2702, opened 2018-07-08
support app installation via cli

mutec:
this implements support for installation of apps via php-cli.
The user will be prompted to provide additional information like installation path, domain and domain path.
Example:
```mysterycode@demo: / $ php cli.php
                                                                           WoltLab Suite (tm) 3.1.4
   Benutzername> Administrator
   Kennwort> **********
   >package install https://update.mysterycode.de/package-file/460/
   Installationsverzeichnis> /guestbook/
   Domain> 499bf03.demo.mysterycode.de:8081
   Pfad> /guestbook/
   100% [###############################################################################################################] Installation abgeschlossen
   >
```


dtdesign:
The package system in the CLI API is in a poor state, caused by the immutability of loaded classes on runtime, which is circumvented in the browser by issuing new requests. We need to rework how the API works, possibly by spawning new processes to mimic the behavior in the browser.

I agree with @TimWolla that this PR does improve the package CLI API in general.

PR #2721, opened 2018-07-20
Be more accurate with benchmark

jens1o:
Measure once the request has been received, and not when it has been already handled.

Especially noticeable on development environments with slower hard disk drives.

### DIFF
--- a/wcfsetup/install/files/lib/system/benchmark/Benchmark.class.php
+++ b/wcfsetup/install/files/lib/system/benchmark/Benchmark.class.php
@@ -20,7 +20,7 @@ class Benchmark extends SingletonFactory {
 	 * general benchmark start time
 	 * @var	float
 	 */
-	protected $startTime = $_SERVER['REQUEST_TIME_FLOAT'];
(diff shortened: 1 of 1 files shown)


Morik:
The Benchmark mode is designed to detect Bottlenecks in your php implementation, not on your system.
It could be interesting to add this as an additional information, but should not replace the default implementation...

dtdesign:
I agree with what @Morik said, the primary intend of this feature is to detect issues with the PHP code.

However, I also agree that the benchmark could be extended to include this information as a dedicated piece in the details overlay, measuring the offset between the request start and the benchmark start. The benchmark is also missing the "hot" feature for queries which was intended to highlight queries that had a significant impact.

If you're interested to work on this feature, I suggest that you open a new issue so that we can discuss which modifications should be made to the benchmark.

jens1o:
> If you're interested to work on this feature, I suggest that you open a new issue so that we can discuss which modifications should be made to the benchmark.

I am, thank you! I'm going to create the issue.

Issue #2699, opened 2018-06-21
Cronjobs are not multi-process safe


lgandras:
Hi,

In version 3.1.3 cronjobs can be run multiple times in parallel. The problem is as follows. Cronjobs that are for potential execution must meet:
```
(afterNextExec <= TIME_NOW || nextExec <= TIME_NOW) && isDisabled = 0 && failCount < MAX_FAIL_COUNT
```
Afterwards, a cronjob is *not* put into PENDING *if it doesn't meet any* of the following conditions:
```
afterNextExec <= TIME_NOW && state = EXECUTING && failCount + 1 == MAX_FAIL_COUNT
afterNextExec > TIME_NOW && nextExec <= TIME_NOW && state != READY
```
The previous can be written as follows:
```
afterNextExec <= TIME_NOW && state = EXECUTING && failCount + 1 == MAX_FAIL_COUNT || afterNextExec > TIME_NOW && nextExec <= TIME_NOW && state != ready
```
The opposite of that condition (represents cronjobs that *are* going to be executed under this thread) is:
```
!(afterNextExec <= TIME_NOW && state = EXECUTING && failCount + 1 == MAX_FAIL_COUNT || afterNextExec > TIME_NOW && nextExec <= TIME_NOW && state != ready)
```
==
```
!(afterNextExec <= TIME_NOW && state = EXECUTING && failCount + 1 == MAX_FAIL_COUNT) && !(afterNextExec > TIME_NOW && nextExec <= TIME_NOW && state != ready))
```
==
```
(afterNextExec > TIME_NOW || state != EXECUTING || failCount + 1 != MAX_FAIL_COUNT) && (afterNextExec <= TIME_NOW || nextExec > TIME_NOW || state = ready)
```
Which in conjunction with the mysql read condition (represents cronjobs that were read AND *are* going to be executed under this thread) is:
```
(afterNextExec <= TIME_NOW || nextExec <= TIME_NOW) && isDisabled = 0 && failCount < MAX_FAIL_COUNT && (afterNextExec > TIME_NOW || state != EXECUTING || failCount + 1 != MAX_FAIL_COUNT) && (afterNextExec <= TIME_NOW || nextExec > TIME_NOW || state = ready)
```
Now, in order to label a cronjob as "it is going to be run under this thread", the following change is done:
```
state = PENDING
```
So, the following represents cronjobs that: 1. were read in the query 2. are going to be executed in this thread 3. are going to be executed in some other thread (already state=PENDING):
[...]

PR #2726, opened 2018-07-30
Reset previously failed cronjobs before executing them

dtdesign:
Fix for #2699


wcfsetup/install/files/lib/system/cronjob/CronjobScheduler.class.php
```diff
@@ -105,6 +105,56 @@ public function getNextExec() {
 		return $this->cache['nextExec'];
 	}
 	
+	/**
+	 * Resets any cronjobs that have previously failed to execute. Cronjobs that have failed too often will
+	 * be disabled automatically.
+	 */
+	protected function resetFailedCronjobs() {
```
TimWolla: This method is never called.

PR #2729, opened 2018-08-01
Append REQUEST_URI to database queries

TimWolla:
This is a first proof of concept commit to discuss whether we want something like this. I opted to directly access `$_SERVER['REQUEST_URI']` to keep the enhancement lightweight – it is a debugging feature for production after all.

----
This might come in handy to debug long running queries in
production.

Idea stolen from: https://chris-lamb.co.uk/projects/django-append-url-to-sql

image

### DIFF
--- a/wcfsetup/install/files/lib/system/database/Database.class.php
+++ b/wcfsetup/install/files/lib/system/database/Database.class.php
@@ -240,7 +240,22 @@ abstract class Database {
 		$statement = $this->handleLimitParameter($statement, $limit, $offset);
 		
 		try {
-			$pdoStatement = $this->pdo->prepare($statement);
+			// Append routing information of the current request as a comment.
+			// This allows the system administrator to find offending requests
+			// in MySQL's slow query log and / or MySQL's process list.
+			// Note: This is meant to be run unconditionally in production to be
+			//       useful. Thus the code to retrieve the request information
+			//       must be absolutely lightweight.
+			$requestInformation = '';
+			if (isset($_SERVER['REQUEST_URI'])) {
+				$requestInformation = substr($_SERVER['REQUEST_URI'], 0, 90);
+				if (isset($_REQUEST['className']) && isset($_REQUEST['actionName'])) {
+					$requestInformation .= ' ('.$_REQUEST['className'].':'.$_REQUEST['actionName'].')';
+				}
+			}
+			$requestInformation = substr($requestInformation, 0, 180);
+			
+			$pdoStatement = $this->pdo->prepare($statement." -- ".$this->pdo->quote($requestInformation));
 			
 			return new $this->preparedStatementClassName($this, $pdoStatement, $statement);
 		}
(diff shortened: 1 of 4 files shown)


dtdesign:
I like the general idea, although using just the request uri isn't enough:
1. Include the `className` and `actionName` for requests to the ajax API endpoints (including uploads).
2. Add support for CLI commands, it needs to include at least the script name and if possible meta data, such as the executed action for workers. However, this is more on the "nice to have" side, because it should be fairly easy to tell apart CLI connections from regular requests.

Morik:
How about adding additional debug infos ?
If the Debug Mode is activated you could use debug_backtrace to trace the file from which it was called, as well as the line and function.
This is of course not exactly cheap in terms of performance, but would make things much more comfortable.

TimWolla:
> If the Debug Mode is activated you could use debug_backtrace to trace the file from which it was called, as well as the line and function.

When developing you are the only person to request pages and can just use the Benchmark to show you the offending methods and functions.

This feature is explicitly targeted to be unconditionally run in high concurrency production environments to be able to recognize offending pages in MySQL's slow query log / MySQL's process list. Once you know the offending page you can reproduce the issue within a clone of your installation with the Benchmark enabled.

That being said I agree with @dtdesign that the className and actionName for AJAX requests needs to be shown to be useful. CLI actions are less of a concern, because you usually know that you are executing some CLI worker.

TimWolla:
Update this PR to:
- Show actionName + className. Whenever they are given instead of for AJAX requests only to not pull the RequestHandler into the Database class. The system administrator is smart enough to figure it out.
- Add escaping using PDO::quote to turn the data into a string, not containing newlines.

bildschirmfoto vom 2018-08-09 15-24-08

wcfsetup/install/files/lib/system/database/Database.class.php
```diff
@@ -240,7 +240,22 @@ public function prepareStatement($statement, $limit = 0, $offset = 0) {
...
+			//       useful. Thus the code to retrieve the request information
+			//       must be absolutely lightweight.
+			$requestInformation = '';
+			if (isset($_SERVER['REQUEST_URI'])) {
+				$requestInformation = substr($_SERVER['REQUEST_URI'], 0, 90);
+				if (isset($_REQUEST['className']) && isset($_REQUEST['actionName'])) {
+					$requestInformation .= ' ('.$_REQUEST['className'].':'.$_REQUEST['actionName'].')';
+				}
+			}
+			$requestInformation = substr($requestInformation, 0, 180);
```
jens1o: You might only want to do this in debug mode?

wcfsetup/install/files/lib/system/database/Database.class.php
```diff
@@ -240,7 +240,22 @@ public function prepareStatement($statement, $limit = 0, $offset = 0) {
...
+			//       useful. Thus the code to retrieve the request information
+			//       must be absolutely lightweight.
+			$requestInformation = '';
+			if (ENABLE_PRODUCTION_DEBUG_MODE && isset($_SERVER['REQUEST_URI'])) {
+				$requestInformation = substr($_SERVER['REQUEST_URI'], 0, 90);
+				if (isset($_REQUEST['className']) && isset($_REQUEST['actionName'])) {
+					$requestInformation .= ' ('.$_REQUEST['className'].':'.$_REQUEST['actionName'].')';
+				}
+			}
+			$requestInformation = substr($requestInformation, 0, 180);
```
dtdesign: Move this line into the `if` above.

TimWolla: Done.

PR #2736, opened 2018-08-12
 Update composer dependencies 

TimWolla:
Care should be taken to review the changes to htmlpurifier (message processing) and emogrifier (emails).


jens1o:
What's up with the SCSS Compiler?

TimWolla:
An upgrade was attempted in 2016 and then reverted: https://github.com/WoltLab/WCF/commits/master/wcfsetup/install/files/lib/system/style/scssphp/scss.inc.php

The performance of newer version was horrible, if I remember correctly.

Morik:
And what about scssphp?
Do you plan to stick on that outdated version?
The current version is 0.7.7...

PR #2749, opened 2018-09-11
Support pluginStoreFileID tag from third party servers

mutec:
Remove the restriction to WoltLab's servers, since there is no good reason for a restriction.
In case anyone has stupid thoughts (simulating it's a checked/other package from the official plugin store or whatever) in his mind, he'd get what he want's anyway. And most or all _known_ third party developers are reliable.


dtdesign:
The ID is meant for internal purposes to make it easier for us to identify installed packages that are also available in the Plugin-Store.

The official servers already provide the IDs for all available packages in the Plugin-Store, so whatever information is provided by a third party server, it is (a) potentially unreliable/outdated and (b) redundant. There is literally zero value for us to open this up for any third party servers.

Issue #2614, opened 2018-05-25
Event access for the syntax highlighter guessing


dtdesign:
n/t

ref https://community.woltlab.com/thread/263279-plugin-event-wenn-sourcecode-highlighter-gesucht-wird/

PR #2752, opened 2018-09-27
WIP: JavaScript based syntax highlighting

TimWolla:
bildschirmfoto vom 2018-09-27 17-46-04

This is a work in progress pull request to solicit some early feedback on the [Prism](https://prismjs.com/) based syntax highlighter for the code metacode.

- The current `<ol>` based markup has been replaced with a proper `<pre><code>` markup to better work with Prism. The line numbers are added using CSS `content: attr(data-number)`. This removed the need for the “Copy Code” dialog box, as the code can properly be selected.
- [ ] Question: Should the copy code box be reintroduced?
- Code Highlighters can be loaded on demand using require.js, but probably not all highlighters should be shown in the redactor UI.
- [ ] Question: Should an option be added that selects the highlighters shown in the redactor UI?

ToDo:
- [ ] Clean up markup in codeMetaCode.tpl
- [ ] Clean up CSS
- [ ] Re-add syntax highlighter autodetection (ref #2614)
- [ ] Check whether stable line anchors can be created using the objectType + objectID
- [ ] Check whether anything else needs to be changed on the server side


Morik:
I like the idea a lot, have you considered to optimize the language Detection?
I like the idea how highlight.js does it:
They have a list of Keywords (like function names etc) for each language and sum then every occurrence of a keyword to get the Language with the highest score.
https://github.com/highlightjs/highlight.js/blob/master/src/languages/php.js#L34-L41
Would be nice to have something similar added to the wsc...

jens1o:
What's the advantage of this solution?

dtdesign:
1. It appears to fail to load a highlighter when none was provided in the first place. Tried it with valid PHP code that had a leading `<?php`, it then tries to load `…/prism-.js?t=…` and the title is set to `wcf.bbcode.code..title`.
2. The color scheme for PHP is awful, who came up with that orange color for keywords that are usually blue or green (PhpStorm: violet)? People have different preferences, but I'm pretty certain that orange will displease the majority. Besides, it's difficult to read using the default style due to the low contrast ratio (2.19:1).
3. Wrapped lines start right at the left boundary, but it should start at the virtual ruler that separates the line numbers from the content:
image

dtdesign:
> Question: Should the copy code box be reintroduced?
No, but I would love to see a button `Copy to Clipboard` using the HTML5 Clipboard API.

> Question: Should an option be added that selects the highlighters shown in the redactor UI?
This has been requested more than once, so I would say yes. We can use sane defaults for this option, e. g. remove brainfuck.

max-m:
Some issues I encountered:
- When using the “copy to clipboard” function newlines are doubled (FF63, Chromium 70, IE11)
- IE11 has some kind of issue with the `DocumentFragment` in `WoltLabSuite/Core/Prism.js`:
```
Possible Unhandled Promise Rejection: TypeError: Das Objekt unterstützt diese Aktion nicht.
   "Possible Unhandled Promise Rejection:"
   {
      [functions]: ,
      __proto__: {
         [functions]: ,
         __proto__: {
            [functions]: ,
            __proto__: { },
            message: "",
            name: "Error"
         },
         message: "",
         name: "TypeError"
      },
      description: "Das Objekt unterstützt diese Aktion nicht.",
      message: "Das Objekt unterstützt diese Aktion nicht.",
      name: "TypeError",
      number: -2146827843,
      stack: "TypeError: Das Objekt unterstützt diese Aktion nicht.
   at Prism.wscSplitIntoLines (https://**host**/wcf/2018-11-20/js/WoltLabSuite/Core/Prism.js?t=1542757866:15:3)
   at Anonymous function (https://**host**/wcf/2018-11-20/js/WoltLabSuite/Core/Bbcode/Code.js?t=1542757866:104:5)
   at body (https://**host**/wcf/2018-11-20/js/WoltLabSuite/Core/Bbcode/Code.js?t=1542757866:27:7)"
   }```

File: wcfsetup/install/files/js/WoltLabSuite/Core/Prism.js [hunk 1]
```diff
@@ -0,0 +1,59 @@
...
+ * @license	GNU Lesser General Public License <http://opensource.org/licenses/lgpl-license.php>
+ * @module	WoltLabSuite/Core/Prism
+ */
+
+window.Prism = window.Prism || {}
+window.Prism.manual = true
+
+define(['prism/prism'], function () {
+	Prism.wscSplitIntoLines = function (container) {
+		var frag = new DocumentFragment();
```
max-m: The `DocumentFragment` constructor is still “experimental” and unsupported by IE11.
```suggestion
		var frag = document.createDocumentFragment();
```

wcfsetup/install/files/js/WoltLabSuite/Core/Prism.js
```diff
@@ -0,0 +1,59 @@
...
+		
+		function newLine() {
+			var line = elCreate('span');
+			elData(line, 'number', lineNo++);
+			frag.appendChild(line);
+			
+			return line;
+		}
+		
+		it = document.createNodeIterator(container, NodeFilter.SHOW_TEXT);
```
max-m:
```suggestion
		// IE11 expects a fourth, non-standard, parameter (entityReferenceExpansion) and a valid function as third
		it = document.createNodeIterator(container, NodeFilter.SHOW_TEXT, function () {
			return NodeFilter.FILTER_ACCEPT
		}, false);
```

wcfsetup/install/files/js/WoltLabSuite/Core/Clipboard.js
```diff
@@ -0,0 +1,34 @@
...
+				promise = Promise.reject('navigator.clipboard is not supported');
+			}
+			
+			return promise.catch(function () {
+				if (!window.getSelection) throw new Error('window.getSelection is not supported');
+				
+				var range = document.createRange();
+				range.selectNode(element);
+				window.getSelection().addRange(range);
+				if (!document.execCommand('copy')) {
```
max-m: `document.execCommand('copy')` fails in Edge, probably because it is being called in a Promise and not synchronously in the event handler.

TimWolla: Included in 8d1094cc519f0056c80f24613ac1596c56921e22, but I was not able to verify the fix due to the lack of Edge. Internet Explorer 11 works fine.

wcfsetup/install/files/js/WoltLabSuite/Core/Clipboard.js
```diff
@@ -0,0 +1,41 @@
...
+				return navigator.clipboard.writeText(text);
+			}
+			else if (window.getSelection) {
+				var textarea = elCreate('textarea');
+				textarea.style.cssText = 'position: absolute; left: -9999px; width: 0;';
+				document.body.appendChild(textarea);
+				try {
+					textarea.value = text;
+					textarea.select();
+					if (!document.execCommand('copy')) {
```
joshuaruesweg: You can't use `document.execCommand('copy')` on iOS, it fails.

joshuaruesweg: See https://stackoverflow.com/a/34046084

PR #2753, opened 2018-10-05
fixed logged IP-address of authentication failure while logging is disabled

mutec:
maybe we should add a notice that in case the IPs are not logged this security feature get's useless


dtdesign:
This isn't really a bug, because disabling the limited ip storing seriously cripples the self-defense mechanism against large scale login attempts. While there is a limit per user account, the ip address is used to enforce a limit on the total number of attempts from one source, regardless of the account that is given.

Removing the ip address effectively opens up a serious vulnerability against bots that use large lists of known username to password combinations, e. g. from yet another of these large hacks, and run it against the entire userbase.

The logged ip addresses are automatically discarded after some time and no other kind of processing is performed on this data, it doesn't even track successful logins (!). This means that we protect the privacy of the visitor without sacrificing parts of the system security against automated attacks.

Issue #2754, opened 2018-10-11
Transparently forward `Mail` usage to `Email` and extend `SimpleEmail`


dtdesign:
The legacy `wcf\system\mail\Mail` object is deprecated since WSC 3.0 in favor of the new `wcf\system\email\Email` that features both HTML and queue support. Despite the new API being available for quite some time, there are still a lot of implementation that use the old `Mail` class for historic reasons.

However. the lack of the support for the background queue is a significant downside. Preserving backwards compatibility is important, therefore we cannot drop support for this legacy class at all, instead we should investigate options to transform the class into a wrapper that forwards all calls to the `Email` class instead.

The `SimpleEmail` class is a great utility class that offers a much simplified interface while still making use of the powerful `Email` API in the background. However, it lacks the ability to handle attachments which is a significant limitation when it comes to porting legacy code to `SimpleEmail`.

mutec:
`SimpleEmail` expects `UserMailbox` which makes it unattractive for forwarding `wcf\system\mail\Mail` to it.

PR #2756, opened 2018-10-23
Map \wcf\system\mail\Mail to \wcf\system\email\Email

TimWolla:
see #2754


mutec:
Removing some methods isn't a good idea. You should remove the class completely or keep the API useable. I'm going to push another solution later.

dtdesign:
Stalled until #2761 has been handled.

PR #2761, opened 2018-10-28
fully pipe wcf\system\mail\Mail to wcf\system\email\Email

mutec:


wcfsetup/install/files/lib/system/mail/Mail.class.php
```diff
@@ -247,7 +132,15 @@ public static function buildAddress($name, $email, $encodeName = true) {
 	 * Sends this mail.
 	 */
 	public function send() {
-		MailSender::getInstance()->sendMail($this);
+		if (!empty($this->attachmments)) {
```
TimWolla:
```suggestion
		if (!empty($this->attachments)) {
```

wcfsetup/install/files/lib/system/mail/Mail.class.php
```diff
@@ -21,84 +26,32 @@ class Mail {
...
-	/**
-	 * mail message
-	 * @var	string
-	 */
-	protected $message = '';
-	
-	/**
-	 * mail sender
-	 * @var	string
+	 * mail object this class get's forwarded to
```
TimWolla:
```suggestion
	 * mail object this class gets forwarded to
```

wcfsetup/install/files/lib/system/mail/Mail.class.php
```diff
@@ -21,84 +26,32 @@ class Mail {
...
+	 * @var Email
+	 * @since WSC 3.2.0
```
TimWolla:
```suggestion
	 * @since 3.2.0
```

wcfsetup/install/files/lib/system/mail/Mail.class.php
```diff
@@ -308,7 +201,8 @@ public function getSubject() {
 	 * @param	string		$message
 	 */
 	public function setMessage($message) {
-		$this->message = $message;
+		if ($this->contentType == 'text/html') $this->email->setBody(new HtmlTextMimePart($message));
+		else $this->email->setBody(new PlainTextMimePart($message));
```
TimWolla: This change is bogus: You are mapping anything to `text/plain`.

wcfsetup/install/files/lib/system/mail/Mail.class.php
```diff
@@ -327,10 +221,26 @@ public function getMessage() {
 	 */
 	public function setFrom($from) {
 		if (is_array($from)) {
-			$this->from = self::buildAddress(key($from), current($from));
+			$this->email->setSender(new Mailbox(current($from), key($from)));
+		}
+		else {
+			$this->email->setSender(new Mailbox($from));
```
TimWolla: This change is bogus. `$from` might be an address specification in `name-addr` format.

wcfsetup/install/files/lib/system/mail/Mail.class.php
```diff
@@ -327,10 +221,26 @@ public function getMessage() {
...
+	 * @param	string		$type
+	 * @since 3.2.0
+	 */
+	protected function addRecipient($recipient, $type = 'to') {
+		if (is_array($recipient)) {
+			$this->email->addRecipient(new Mailbox(current($recipient), key($recipient), $this->language), $type);
 		}
 		else {
-			$this->from = $from;
+			$this->email->addRecipient(new Mailbox($recipient, null, $this->language), $type);
```
TimWolla: This change is bogus. `$recipient` might be an address specification in `name-addr` format.

wcfsetup/install/files/lib/system/mail/Mail.class.php
```diff
@@ -445,7 +355,7 @@ public function addAttachment($path, $name = '') {
 	 * @param	integer		$priority
 	 */
 	public function setPriority($priority) {
-		$this->priority = $priority;
+		$this->email->addHeader('X-Priority', $priority);
```
TimWolla: This will add a new header with every call.

wcfsetup/install/files/lib/system/mail/Mail.class.php
```diff
@@ -497,8 +396,19 @@ public function setContentType($contentType) {
 	 * @param	string		$header
 	 */
 	public function setHeader($header) {
```
TimWolla: The new implementation is bogus. `setHeader` expected to receive a string that possibly contains multiple headers. You would need to parse this string according to RFC 5322 to extract the headers.

wcfsetup/install/files/lib/system/mail/Mail.class.php
```diff
@@ -272,7 +158,14 @@ public function addTo($to) {
 	 * @return	mixed
 	 */
 	public function getTo() {
```
TimWolla: The change violates the contract that:
```
$mail = new Mail();
$mail->addTo($x);
// $mail->getTo()[0] === $x; for all $x given !is_array($x)
```

Specifically the `display-name` of the address specification is being stripped. In fact this is one of the reasons I did not implement the getters in #2756: It is impossible to map the methods without a behavioural change. Thus I opted for a highly visible exception instead of a silent change in behaviour. If people absolutely have to use the getters (which I doubt) they would need to rewrite their code anyway and thus can simply use the new email API (which exists since 2 years). If they don't, then #2756 is fine for them and they get to reap the benefits of the new email API without doing anything in the short term.

mutec: The display name has never been there, as you can see:
https://github.com/WoltLab/WCF/blob/3.0/wcfsetup/install/files/lib/system/mail/Mail.class.php#L283-L285
https://github.com/WoltLab/WCF/blob/3.0/wcfsetup/install/files/lib/system/mail/Mail.class.php#L261
https://github.com/WoltLab/WCF/blob/3.0/wcfsetup/install/files/lib/system/mail/Mail.class.php#L242-L244

TimWolla: If a raw string (i.e. no array) is given to `addTo` it is passed through straight. I've seen examples in the wild where a string in `name-addr` format is explicitely given instead of letting `Mail` construct it. See: https://github.com/WoltLab/WCF/commit/b694faf3bfa8a7be7eafcc21568728785ec24c4e#diff-dcefc405b8d94fc948f8ae36d00d2e26L178

PR #2763, opened 2018-10-28
add TDatabaseObjectToggle trait

### LINKED ISSUE #2728: Add `TToggleAction` trait
Integrating `IToggleAction` requires duplicating the same code over and over again which could be avoided by introducing a `TToggleAction` trait which would provide a default implementation of the two methods.

```php
trait TToggleAction {
	
	public function validateToggle() {
		$this->validateUpdate();
	}
	
	public function toggle() {
		foreach ($this->getObjects() as $object) {
			$object->update([
				'isDisabled' => $object->isDisabled ? 0 : 1
			]);
		}
	}
}
```

If you like I would create a pull request for that.

mutec:
see #2728 
Closes #2728

### DIFF
--- a/wcfsetup/install/files/lib/data/TDatabaseObjectToggle.class.php
+++ b/wcfsetup/install/files/lib/data/TDatabaseObjectToggle.class.php
@@ -1,0 +1,33 @@
+<?php
+namespace wcf\data;
+
+/**
+ * Default implementation of the `IToggleAction` interface.
+ * 
+ * @author	Florian Gail
+ * @copyright	2001-2018 WoltLab GmbH
+ * @license	GNU Lesser General Public License <http://opensource.org/licenses/lgpl-license.php>
+ * @package	WoltLabSuite\Core\Data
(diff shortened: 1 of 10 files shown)


wcfsetup/install/files/lib/data/TDatabaseObjectToggle.class.php
```diff
@@ -0,0 +1,33 @@
+<?php
+namespace wcf\data;
+
+/**
+ * Provides a method for toggling the "isDisabled" status of database objects.
```
Gravatronics:
```suggestion
 * Default implementation of the `IToggleAction` interface.
```

ghost (changes requested):

Gravatronics:
Would you mind changing the existing classes in Core implementing `IToggleAction` this way to use this trait instead?

ghost (approved):

dtdesign:
Thanks!

PR #2764, opened 2018-10-28
rewrite rule generator

### LINKED ISSUE #2704: Built-in tool to generate rewrites for an Apache .htaccess file
n/t

mutec:
see #2704
Closes #2704


wcfsetup/install/files/acp/templates/__optionRewriteRules.tpl
```diff
@@ -0,0 +1,14 @@
+{if $category->categoryName === 'general'}
+	<div id="dialogRewriteRules" style="display: none" />
```
joshuaruesweg: The JavaScript can generate the element itself (see https://docs.woltlab.com/javascript_new-api_dialogs.html)

wcfsetup/install/files/lib/data/option/OptionAction.class.php
```diff
@@ -41,7 +45,7 @@ class OptionAction extends AbstractDatabaseObjectAction {
 	/**
 	 * @inheritDoc
 	 */
-	protected $requireACP = ['create', 'delete', 'emailSmtpTest', 'import', 'update', 'updateAll'];
+	protected $requireACP = ['create', 'delete', 'emailSmtpTest', 'import', 'update', 'updateAll', 'parseRewriteRules'];
```
dtdesign:
```suggestion
	protected $requireACP = ['create', 'delete', 'emailSmtpTest', 'import', 'update', 'updateAll', 'generateRewriteRules'];
```

wcfsetup/install/files/lib/data/option/OptionAction.class.php
```diff
@@ -111,4 +115,84 @@ public function emailSmtpTest() {
...
+				}
+			}
+		}
+		
+		foreach ($dirs as $dir => $domainPaths) {
+			krsort($domainPaths);
+			
+			foreach ($domainPaths as $domainPath => $value) {
+				$path = FileUtil::removeTrailingSlash(substr($value, strlen($dir)));
+				$snippet = 'RewriteCond %{SCRIPT_FILENAME} !-d
```
dtdesign: Consider using Heredoc instead, it makes the code much easier to read and puts an emphasis on the exact look of it:
```php
$snippet = <<<SNIPPET
[…]
SNIPPET;
```

wcfsetup/install/lang/de.xml
```diff
@@ -1932,6 +1932,12 @@ Als Benachrichtigungs-URL in der Konfiguration der sofortigen Zahlungsbestätigu
 		<item name="wcf.acp.rescueMode.description"><![CDATA[Die eingestellte Domains und/oder Pfade stimmen nicht mit der für den Aufruf verwendeten Adresse überein; Dies kann in Folge eines Umzugs auf eine andere Domain oder Veränderung der Verzeichnisstruktur auftreten. Mit Hilfe des unten stehenden Formulars {if LANGUAGE_USE_INFORMAL_VARIANT}kannst du{else}können Sie{/if} die Angaben schnell und einfach anpassen.]]></item>
 		<item name="wcf.acp.rescueMode.username.notAuthorized"><![CDATA[Diese Benutzer ist nicht berechtigt die Konfiguration zu verändern.]]></item>
 	</category>
+
+	<category name="wcf.acp.rewrite">
+		<item name="wcf.acp.rewrite"><![CDATA[Rewrite-Regeln]]></item>
```
dtdesign: We should put an emphasis on the fact, that this Apache only; The description below is a bit vague.

dtdesign:
I've taken the time to look through it and would like to see a few changes:
1. The `validateGenerateRewriteRules()` method should check for the permission `admin.configuration.canEditOption`.
2. The `codeMetaCode` template does not exist in the ACP, I recommend that you use a simple `<pre>` instead.
3. The Heredoc block is faulty, it doesn't use apostrophes to exit the block, instead you need to use variable interpolation, e. g.  `Hello {$world}!`.

image

wcfsetup/install/files/lib/data/option/OptionAction.class.php
```diff
@@ -111,4 +115,89 @@ public function emailSmtpTest() {
 		
 		return ['validationResult' => $smtp->testConnection()];
 	}
+	
+	/**
+	 * Validates the "generateRewriteRules" action
+	 * @throws \wcf\system\exception\AJAXException
+	 */
+	public function validateGenerateRewriteRules() {
+		if (!function_exists('apache_get_version')) {
```
jens1o: `FileUtil::isApacheModule()`

mutec:
> 2\. The `codeMetaCode` template does not exist in the ACP, I recommend that you use a simple `<pre>` instead.
That's why I used `$tpl`. In case we use `<pre>`, we'd need a new template for that in case we don't want an ugly appearance. Sometimes it's required to have multiple files, so we can't just remove the file's location.

> 3\. The Heredoc block is faulty, it doesn't use apostrophes to exit the block, instead you need to use variable interpolation, e. g.  `Hello {$world}!`.
Didn't recognize I used variables in there when I changed that. 🤨

wcfsetup/install/files/lib/data/option/OptionAction.class.php
```diff
@@ -111,4 +114,89 @@ public function emailSmtpTest() {
...
+	protected function fetchRewriteRules() {
+		$dirs = $rules = [];
+		foreach (ApplicationHandler::getInstance()->getApplications() as $app) {
+			$test = $app->getPackage()->getAbsolutePackageDir();
+			$insert = true;
+			
+			foreach ($dirs as $dir => $apps) {
+				if (strpos($dir, $test) !== false) {
+					unset($dirs[$dir]);
+					$insert = false;
```
mutec: this has to be true I think. let me check that

PR #2766, opened 2018-11-01
clipboard action for triggering a new activation email

mutec:
might be useful being able to send an activation email again after a job got stuck and removed from the queue


wcfsetup/install/files/lib/data/user/UserAction.class.php
```diff
@@ -828,4 +828,38 @@ public function validateSaveSocialNetworkPrivacySettings() {
...
+	
+	/**
+	 * Triggers a new activation email.
+	 * @since 3.2
+	 */
+	public function resendActivationMail() {
+		foreach ($this->objects as $object) {
+			$email = new Email();
+			$email->addRecipient(new UserMailbox($object->getDecoratedObject()));
+			$email->setSubject(WCF::getLanguage()->getDynamicVariable('wcf.user.register.needActivation.mail.subject'));
```
TimWolla: You need to use the user’s `Language`.

wcfsetup/install/files/lib/data/user/UserAction.class.php
```diff
@@ -828,4 +828,38 @@ public function validateSaveSocialNetworkPrivacySettings() {
...
+			}
+		}
+	}
+	
+	/**
+	 * Triggers a new activation email.
+	 * @since 3.2
+	 */
+	public function resendActivationMail() {
+		foreach ($this->objects as $object) {
```
TimWolla: A new `activationCode` should be generated for each user (similar to `RegisterNewActivationCodeForm`).

wcfsetup/install/lang/en.xml
```diff
@@ -2997,6 +2997,7 @@ Errors are:
 		<item name="wcf.clipboard.item.com.woltlab.wcf.user.enable"><![CDATA[Approve]]></item>
 		<item name="wcf.clipboard.item.com.woltlab.wcf.user.exportMailAddress"><![CDATA[Export Email Addresses]]></item>
 		<item name="wcf.clipboard.item.com.woltlab.wcf.user.merge"><![CDATA[Merge]]></item>
+		<item name="wcf.clipboard.item.com.woltlab.wcf.user.resendActivationMail"><![CDATA[Send activation Email again]]></item>
```
```suggestion
		<item name="wcf.clipboard.item.com.woltlab.wcf.user.resendActivationMail"><![CDATA[Send New Activation Code]]></item>
```

wcfsetup/install/lang/de.xml
```diff
@@ -2652,6 +2652,7 @@ Fehler sind beispielsweise:
 		<item name="wcf.clipboard.item.com.woltlab.wcf.user.enable"><![CDATA[Aktivieren]]></item>
 		<item name="wcf.clipboard.item.com.woltlab.wcf.user.exportMailAddress"><![CDATA[E-Mail-Adressen exportieren]]></item>
 		<item name="wcf.clipboard.item.com.woltlab.wcf.user.merge"><![CDATA[Zusammenfügen]]></item>
+		<item name="wcf.clipboard.item.com.woltlab.wcf.user.resendActivationMail"><![CDATA[Aktivierungs-E-Mail erneut senden]]></item>
```
TimWolla:
```suggestion
		<item name="wcf.clipboard.item.com.woltlab.wcf.user.resendActivationMail"><![CDATA[Neuen Aktivierungscode versenden]]></item>
```

wcfsetup/install/files/lib/data/user/UserAction.class.php
```diff
@@ -828,4 +828,47 @@ public function validateSaveSocialNetworkPrivacySettings() {
...
+			$action = new UserAction($this->objects, 'update', [
+				'data' => [
+					'activationCode' => $code
+				]
+			]);
+			$action->executeAction();
+			
+			$email = new Email();
+			// get current/updated user object for activation code
+			$email->addRecipient(new UserMailbox(new User($object->userID)));
```
TimWolla: Create a `UserList` with all the users you need updated (saves queries). This is most easily done using a two-step process:

1. Update the activationCode for every user.
2. Select all the users using a `UserList`.
3. Send email as usual.

joshuaruesweg:
There's one more thing I'd like to say: Not everyone has set that users can activate themselves via email. You should only offer this option if the users are not always activated anyway or have to be activated by the administrator. 

In addition, I question creating a new activation code each time. This will invalidate old emails. Is it a problem to simply continue using the old code?

PR #2767, opened 2018-11-01
Don't truncate cronjob descriptions

mutec:
Truncating after 50 chars is very ugly for high resolution displays like there is 4cm text and 15cm white space.
No matter which resolution, the text will be cut off after three lines, which makes it possible to display the full description on large screens but prevent large text blocks on mobile/small screens.

Before:
image

After:
image

### DIFF
--- a/wcfsetup/install/files/acp/templates/cronjobList.tpl
+++ b/wcfsetup/install/files/acp/templates/cronjobList.tpl
@@ -86,9 +86,9 @@
 							<td class="columnDate columnStartDow">{$cronjob->startDow|truncate:30}</td>
 							<td class="columnText columnDescription" title="{$cronjob->description|language}">
 								{if $cronjob->isEditable()}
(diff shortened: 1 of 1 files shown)


Morik:
The line-clamp property isn't currently standardized and not supportet by all browsers.
It seams to me a little bit messed up a specially as it depends on the old webkit-box and doesn't work with regular flexboxes.

mutec:
I know. Maybe we should remove this truncation completely. I don't know any package that would break the layout and a few lines (on a mobile, which isn't the best device for administration anyway) aren't critical.

Morik:
Why not add a regular `overflow: hidden;` with `white-space: nowrap;` for mobile devices only?
This would add the full text on desktop browsers while not breaking the mobile design at all.

mutec:
We'd need a fixed width then - otherwise it would become a snake/spaghetti.

dtdesign:
I don't see any reason to have the content truncated at all, iirc this dates back to WCF 1.x.

dtdesign:
Thanks!

Issue #2686, opened 2018-05-29
The ip addresses listed in the users online list should be linked to a search engine


dtdesign:
* [ ] Defaults to querying `google.com`.
* [ ] Include an event-based override to support other sites.

ref https://community.woltlab.com/thread/266425-add-google-ip-search-query-link-to-ip-s-listed-on-online-list/

jens1o:
>  Defaults to querying google.com.

Please. Make. This. Configurable.

Issue #2691, opened 2018-05-29
Option to anonymize ip addresses


dtdesign:
Apply anomization similar to what Google Analytics does (see https://support.google.com/analytics/answer/2763052?hl=en). It makes sense to do this late in the development process to evaluate if there have been any changes due to the GDPR.

ref https://community.woltlab.com/thread/267773-ip-adressen-anonymisieren/

TimWolla:
see also: #2656

PR #2768, opened 2018-11-01
ipSearch-template plugin for linking ip addresses to a search engine

mutec:
see #2686
Solves #2686


wcfsetup/install/files/lib/system/template/plugin/IpSearchModifierTemplatePlugin.class.php
```diff
@@ -0,0 +1,42 @@
...
+	 */
+	const SEARCH_ENGINE_URL_DEFAULT = 'https://www.google.com/search?q=%ip%';
+	
+	/**
+	 * @inheritDoc
+	 */
+	public function execute($tagArgs, TemplateEngine $tplObj) {
+		$domain = Url::parse( USER_IP_ADDRESS_SEARCHENGINE ?: self::SEARCH_ENGINE_URL_DEFAULT)['host'];
+		$ipAddress = StringUtil::trim(MessageUtil::stripCrap($tagArgs[0]));
+		$url = str_replace('%ip%', $ipAddress, USER_IP_ADDRESS_SEARCHENGINE ?: self::SEARCH_ENGINE_URL_DEFAULT);
```
TimWolla: Use `sprintf`.

wcfsetup/install/files/lib/system/template/plugin/IpSearchModifierTemplatePlugin.class.php
```diff
@@ -0,0 +1,42 @@
...
+	 * default search engine schema
+	 * @var string
+	 */
+	const SEARCH_ENGINE_URL_DEFAULT = 'https://www.google.com/search?q=%ip%';
+	
+	/**
+	 * @inheritDoc
+	 */
+	public function execute($tagArgs, TemplateEngine $tplObj) {
+		$domain = Url::parse( USER_IP_ADDRESS_SEARCHENGINE ?: self::SEARCH_ENGINE_URL_DEFAULT)['host'];
```
TimWolla: Erroneous space:
```suggestion
		$domain = Url::parse(USER_IP_ADDRESS_SEARCHENGINE ?: self::SEARCH_ENGINE_URL_DEFAULT)['host'];
```

wcfsetup/install/lang/en.xml
```diff
@@ -4086,6 +4088,7 @@ Email: {@$emailAddress} {* this line ends with a space *}
 		<item name="wcf.user.email.error.notUnique"><![CDATA[The email is already in use.]]></item>
 		<item name="wcf.user.email.error.invalid"><![CDATA[The email is invalid.]]></item>
 		<item name="wcf.user.ipAddress"><![CDATA[IP Address]]></item>
+		<item name="wcf.user.ipAddress.searchEngine"><![CDATA[Search for {if !$ipAddress|empty}{$ipAddress}{else}this IP-Address{/if} {if !$host|empty}on {$host}{/if}]]></item>
```
```suggestion
		<item name="wcf.user.ipAddress.searchEngine"><![CDATA[Search for {if !$ipAddress|empty}{$ipAddress}{else}this IP address{/if} {if !$host|empty}on {$host}{/if}]]></item>
```

wcfsetup/install/lang/en.xml
```diff
@@ -1152,6 +1152,8 @@
 		<item name="wcf.acp.option.session_enable_virtualization.description"><![CDATA[Users can login from multiple devices (desktop, tablet, smartphone, etc.) concurrently; all devices will share the same session. Disabling this will restrict users to only one device (session) at a time, causing all the other device sessions to terminate. It is not recommended to disable this option.]]></item>
 		<item name="wcf.acp.option.timezone"><![CDATA[Timezone]]></item>
 		<item name="wcf.acp.option.timezone.description"><![CDATA[The default timezone of your page.]]></item>
+		<item name="wcf.acp.option.user_ip_address_searchengine"><![CDATA[Search engine for IP Addresses]]></item>
+		<item name="wcf.acp.option.user_ip_address_searchengine.description"><![CDATA[IP Addresses will be linked to this search engine; by default IP Addresses are linked to www.google.com. <span class="inlineCode">%ip%</span> will be replaced by the IP Address.]]></item>
```
```suggestion
		<item name="wcf.acp.option.user_ip_address_searchengine.description"><![CDATA[IP addresses will be linked to this search engine; by default IP addresses are linked to www.google.com. <span class="inlineCode">%ip%</span> will be replaced by the IP address.]]></item>
```

wcfsetup/install/lang/en.xml
```diff
@@ -1152,6 +1152,8 @@
 		<item name="wcf.acp.option.session_enable_virtualization.description"><![CDATA[Users can login from multiple devices (desktop, tablet, smartphone, etc.) concurrently; all devices will share the same session. Disabling this will restrict users to only one device (session) at a time, causing all the other device sessions to terminate. It is not recommended to disable this option.]]></item>
 		<item name="wcf.acp.option.timezone"><![CDATA[Timezone]]></item>
 		<item name="wcf.acp.option.timezone.description"><![CDATA[The default timezone of your page.]]></item>
+		<item name="wcf.acp.option.user_ip_address_searchengine"><![CDATA[Search engine for IP Addresses]]></item>
```
```suggestion
		<item name="wcf.acp.option.user_ip_address_searchengine"><![CDATA[Search Engine for IP Addresses]]></item>
```

wcfsetup/install/files/lib/system/template/plugin/IpSearchModifierTemplatePlugin.class.php
```diff
@@ -0,0 +1,42 @@
...
+		$ipAddress = StringUtil::trim(MessageUtil::stripCrap($tagArgs[0]));
+		$url = str_replace('%ip%', $ipAddress, USER_IP_ADDRESS_SEARCHENGINE ?: self::SEARCH_ENGINE_URL_DEFAULT);
+		
+		return '<a href="' . $url . '"' . (EXTERNAL_LINK_REL_NOFOLLOW ? ' rel="nofollow"' : '') .(EXTERNAL_LINK_TARGET_BLANK ? ' target="_blank"' : '') . ' title="' . WCF::getLanguage()->getDynamicVariable('wcf.user.ipAddress.searchEngine', ['host' => $domain, 'ipAddress' => $ipAddress]) . '">' . (!empty($tagArgs[1]) && is_integer($tagArgs[1]) ? StringUtil::truncate($ipAddress, $tagArgs[1]) : $ipAddress) . '</a>';
```
TimWolla: Construct the title as a separate variable for readability.

wcfsetup/install/files/lib/system/template/plugin/IpSearchModifierTemplatePlugin.class.php
```diff
@@ -0,0 +1,42 @@
...
+use wcf\system\template\TemplateEngine;
+use wcf\system\WCF;
+use wcf\util\MessageUtil;
+use wcf\util\StringUtil;
+use wcf\util\Url;
+
+/**
+ * IP address search modifier plugin which links the IP address to a search engine.
+ *
+ * In case the second parameter is given, the shown IP address gets truncated based on the given amount
```
TimWolla: Truncation based on a number of characters does not make sense: IPv6 addresses are way longer than IPv4 adresses and even the latter might differ greatly in length. Either they are cut too short or are cut too long.

You should null trailing bits of the IP address based on CIDR, adding a method in a `*Util` might make sense.

TimWolla: This Util will also come in handy for #2691.

TimWolla: > Didn't forget this one, but I'll push the code since I'm going to be away for a few days again.

Looking at this once again: You did not truncate the IP address for the search engine link, but for `textContent` of the `<a>` only. I thought this was for privacy, but apparently it is not. Why did you the second parameter?

mutec: I just kept the existing behavior:
https://github.com/WoltLab/WCF/blob/af3d18c74b9a1abc5ec9658b48d81308d0c0e419/com.woltlab.wcf/templates/usersOnlineList.tpl#L97
I think it's there since WCF 2 where the list was splitted. Same with the user agent which gets truncated.

wcfsetup/install/files/lib/system/template/plugin/IpSearchModifierTemplatePlugin.class.php
```diff
@@ -0,0 +1,43 @@
...
+ * @license	GNU Lesser General Public License <http://opensource.org/licenses/lgpl-license.php>
+ * @package	WoltLabSuite\Core\System\Template\Plugin
+ * @since	3.2
+ */
+class IpSearchModifierTemplatePlugin implements IModifierTemplatePlugin {
+	/**
+	 * default search engine schema
+	 * @var string
+	 */
+	const SEARCH_ENGINE_URL_DEFAULT = 'https://www.google.com/search?q=%s';
```
jens1o: please make this configurable.

wcfsetup/install/files/lib/system/template/plugin/IpSearchModifierTemplatePlugin.class.php
TimWolla: I see. I'd say remove this: A full length IPv6 that is going to be seen in the wild only exceeds this by 9 characters (technically it's 45 characters total, but we convert IPv4 mapped addresses to real IPv4) and the `…` that is appended visually takes up the space of at least two characters. So there is hardly any gain. /cc @dtdesign 

I believe that the CIDR truncation is not required then either. It should rather be implemented as part of #2691.

dtdesign: I don't see a point in truncating the IP either. In this particular situation, the usefulness of the ip address is strictly boolean, either I don't care about it at all or I need the exact ip address for whatever reasons. Truncating it contributes to neither of those goals, while also actively working against the user when the full ip address does matter.

mutec: Then I'll remove the truncation.

dtdesign:
Thanks!

PR #2770, opened 2018-11-07
Automatic client-side attachment scaling

### LINKED ISSUE #2714: Pre-resize images before uploading
see https://hacks.mozilla.org/2011/01/how-to-develop-a-html5-image-uploader/

max-m:
This implementation handles drag-n-drop, normal uploads and copy-pasted files in attachments.

TODO / Points for discussion:
- [x] Basic functionality
  - [x] Scale if file size is exceeded
  - [x] Scale if image dimensions are exceeded
- [x] Prevent upscaling images of small dimensions that exceed the file size limit
- [ ] ACP options
  - [x] Add ACP options
  - [ ] Check the default values
- [x] Scale inside a WebWorker if possible to not block the UI thread
- [x] Evaluate if a custom downsampling algorithm like Hermite downsampling should be implemented

Closes #2714


wcfsetup/install/lang/de.xml
```diff
@@ -1202,6 +1202,16 @@
 		<item name="wcf.acp.option.attachment_thumbnail_width"><![CDATA[Breite der Vorschaugrafiken]]></item>
 		<item name="wcf.acp.option.attachment_storage"><![CDATA[Speicherort der Dateianhänge]]></item>
 		<item name="wcf.acp.option.attachment_storage.description"><![CDATA[{if LANGUAGE_USE_INFORMAL_VARIANT}Du kannst{else}Sie können{/if} optional einen vom Standard abweichenden Speicherort für die Dateianhänge definieren. Bereits existierende Dateianhänge müssen bei einer Änderung des Speicherortes manuell in den neuen Ort übertragen werden.]]></item>
+		<item name="wcf.acp.option.category.message.attachment.autoscale"><![CDATA[Automatische Bildskalierung]]></item>
+		<item name="wcf.acp.option.attachment_image_autoscale"><![CDATA[Browserseitige Bildskalierung aktivieren]]></item>
+		<item name="wcf.acp.option.attachment_image_autoscale.description"><![CDATA[Wenn diese Option aktiviert ist, werden angehangene Bilder klientseitig skaliert, falls diese die maximalen Dimensionen oder das Dateigrößenlimit überschreiten.]]></item>
```
joshuaruesweg:
```suggestion
		<item name="wcf.acp.option.attachment_image_autoscale.description"><![CDATA[Wenn diese Option aktiviert ist, werden angehangene Bilder clientseitig skaliert, falls diese die maximalen Dimensionen oder das Dateigrößenlimit überschreiten.]]></item>
```

wcfsetup/install/files/js/WCF.js
```diff
@@ -6067,24 +6084,40 @@ if (COMPILER_TARGET_DEFAULT) {
...
+					return $uploadID;
 				}
 				
-				return $uploadID;
+				return null;
+				
 			}
 			
-			return null;
+			if (files instanceof Promise) {
```
TimWolla: See above.

wcfsetup/install/files/js/WoltLabSuite/Core/ImageResizer.js
```diff
@@ -0,0 +1,145 @@
...
+				
+				case 'image/gif':
+					ext = '.gif';
+					break;
+				
+				default:
+					ext = blob.type.split('/')[1];
+			}
+			
+			return new File([blob], name.split(/(.+)(\..+?)$/)[1] + "_autoscaled" + ext);
```
TimWolla: Using `match` probably is more intuitive:
```suggestion
			return new File([blob], name.match(/(.+)\..+?$/)[1] + "_autoscaled" + ext);
```

wcfsetup/install/files/js/WoltLabSuite/Core/ImageResizer.js
```diff
@@ -0,0 +1,145 @@
...
+		},
+		
+		getFile: function (fileName) {
+			var self = this;
```
TimWolla: Prefer `.bind` over this.

wcfsetup/install/files/js/WoltLabSuite/Core/ImageResizer.js
```diff
@@ -0,0 +1,145 @@
...
+			
+			return this.getBlob().then(function (blob) {
+				return self.blobToFile(blob, fileName);
+			});
+		},
+		
+		getBlob: function () {
+			var self = this;
```
TimWolla: Ditto

wcfsetup/install/files/js/WoltLabSuite/Core/ImageResizer.js
```diff
@@ -0,0 +1,145 @@
...
+			
+			return new Promise(function (resolve, reject) {
+				// Create a new image blob
+				try {
+					if (HTMLCanvasElement.prototype.toBlob) {
```
TimWolla: Check specifically for `self.canvas.toBlob`.

wcfsetup/install/files/js/WoltLabSuite/Core/ImageResizer.js
```diff
@@ -0,0 +1,145 @@
...
+						
+						for (var i = 0; i < length; i++) {
+							data[i] = binary.charCodeAt(i);
+						}
+						
+						resolve(new Blob([data], {type: self.fileType}));
+					}
+				}
+				catch (error) {
+					reject(error);
```
TimWolla: This happens automatically when an Error is thrown.

wcfsetup/install/lang/de.xml
TimWolla: klientseitig should be made consistent with whatever is used above.

wcfsetup/install/lang/de.xml
```diff
@@ -1202,6 +1202,16 @@
 		<item name="wcf.acp.option.attachment_thumbnail_width"><![CDATA[Breite der Vorschaugrafiken]]></item>
 		<item name="wcf.acp.option.attachment_storage"><![CDATA[Speicherort der Dateianhänge]]></item>
 		<item name="wcf.acp.option.attachment_storage.description"><![CDATA[{if LANGUAGE_USE_INFORMAL_VARIANT}Du kannst{else}Sie können{/if} optional einen vom Standard abweichenden Speicherort für die Dateianhänge definieren. Bereits existierende Dateianhänge müssen bei einer Änderung des Speicherortes manuell in den neuen Ort übertragen werden.]]></item>
+		<item name="wcf.acp.option.category.message.attachment.autoscale"><![CDATA[Automatische Bildskalierung]]></item>
+		<item name="wcf.acp.option.attachment_image_autoscale"><![CDATA[Browserseitige Bildskalierung aktivieren]]></item>
```
TimWolla:
```suggestion
		<item name="wcf.acp.option.attachment_image_autoscale"><![CDATA[Bildskalierung durch den Webbrowser aktivieren]]></item>
```

wcfsetup/install/lang/en.xml
```diff
@@ -1190,6 +1190,16 @@
 		<item name="wcf.acp.option.attachment_thumbnail_width"><![CDATA[Thumbnail Width]]></item>
 		<item name="wcf.acp.option.attachment_storage"><![CDATA[Storage Location]]></item>
 		<item name="wcf.acp.option.attachment_storage.description"><![CDATA[You can specify a different storage location for attachments. Attention! You are responsible to move any existing attachments to the new storage location.]]></item>
+		<item name="wcf.acp.option.category.message.attachment.autoscale"><![CDATA[Automatic Image Scaling]]></item>
+		<item name="wcf.acp.option.attachment_image_autoscale"><![CDATA[Enables client side image scaling]]></item>
```
```suggestion
		<item name="wcf.acp.option.attachment_image_autoscale"><![CDATA[Enables image scaling in the web browser]]></item>
```

wcfsetup/install/files/js/WoltLabSuite/Core/ImageResizer.js
```diff
@@ -0,0 +1,145 @@
...
+				else {
+					self.canvas.width = self.maxHeight * (image.width / image.height);
+					self.canvas.height = self.maxHeight;
+				}
+				
+				// Scale the image down
+				// TODO: Evaluate if a custom downsampling algorithm like Hermite downsampling should be implemented
+				self.ctx.drawImage(image, 0, 0, self.canvas.width, self.canvas.height);
+				
+				resolve()
```
TimWolla: I'm not sure whether this empty “return value” is good API design. You should return the `Blob` at the very least (or create a new canvas for every resize operation and return that canvas). Consider this:

```
const resizer = new ImageResizer();
const myFiles = [ …, …, … ];
const resized = Promise.all(
  myFiles.map(function (file) {
    return resizer.resize(file).then(resizer.getFile.bind(resizer, file.src))
  })
)
```

If I am not mistaken you would have the same resized file for every item, because the `getFile` asynchronously runs in the next iteration of the event loop, after `resize` has been called for every file.

wcfsetup/install/files/js/WCF.Attachment.js
```diff
@@ -308,6 +315,85 @@ if (COMPILER_TARGET_DEFAULT) {
...
+											if (!needsResize) {
+												resolve(file);
+											}
+											else {
+												resizer.resize(image)
+													.then(function () {
+														return resizer.getFile(file.name)
+													})
+													.then(resolve)
+													.catch(function (error) {
```
TimWolla: This catch should be moved before `.then(resolve)` and simply `return file`. Otherwise you are going to catch errors in the invocation of `resolve`, leading calling `resolve` twice which violates the Promise's contract.

wcfsetup/install/files/js/WCF.Attachment.js
```diff
@@ -308,6 +315,85 @@ if (COMPILER_TARGET_DEFAULT) {
...
+						// Create an ImageResizer instance
+						var resizer = new ImageResizer()
+							.setMaxWidth(self._options.autoScaleMaxWidth)
+							.setMaxHeight(self._options.autoScaleMaxHeight)
+							.setFileType(self._options.autoScaleFileType)
+							.setQuality(self._options.autoScaleQuality);
+						
+						for (var i = 0; i < files.length; i++) {
+							if (files[i].type.match(/^image\//i)) {
+								(function (i) {
```
TimWolla: This IIFE is redundant, because there is a second closure (the parameter of the `Promise` constructor).

wcfsetup/install/files/js/WoltLabSuite/Core/ImageResizer.js
TimWolla: What I intuitively would expect from the API is something like this (passing along parameters, instead of storing state in the object):

```
resizer.resize(file).then(function (canvas) {
 return resizer.getFile(filename, canvas);
});
// equivalent to:
resizer.resize(file).then(resizer.getFile.bind(resizer, filename));
```

wcfsetup/install/files/js/WCF.Attachment.js
```diff
@@ -308,6 +315,85 @@ if (COMPILER_TARGET_DEFAULT) {
...
+									}));
+								})(i);
+							}
+							else {
+								promises.push(Promise.resolve(files[i]));
+							}
+						}
+						
+						resolve(Promise.all(promises));
+					});
```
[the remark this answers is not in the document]
max-m: Instead of rejecting the promise I would resolve with the untouched list of files, thus I don’t have to worry about rejected promises in WCF.Upload.
I think that failure to load the required modules is the uncommon case, which makes trying to upload the untouched files sound like a sane failure mode :smile:

wcfsetup/install/files/js/WCF.js
```diff
@@ -6005,56 +6009,69 @@ if (COMPILER_TARGET_DEFAULT) {
...
+						}
+					});
+				};
 				
-				$fd.append('actionName', this._options.action);
-				$fd.append('className', this._className);
-				var $additionalParameters = this._getParameters();
-				for (var $name in $additionalParameters) {
-					$fd.append('parameters[' + $name + ']', $additionalParameters[$name]);
+				if ($uploadID instanceof Promise) {
```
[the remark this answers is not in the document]
max-m: Now that you mention it: In the case that `$uploadID` is not a promise, everything should work as expected, but if it is a promise `_upload` will return it which needs handling in all methods that call `_upload` and use its return value (if I didn’t miss something that should be just `Attachment._editorUpload`).

wcfsetup/install/files/js/WCF.js
max-m: I wanted to keep the old code path as untouched as possible but makes sense to just return a promise as we have to handle them anyway.

wcfsetup/install/files/js/WCF.Attachment.js
```diff
@@ -308,6 +317,104 @@ if (COMPILER_TARGET_DEFAULT) {
...
+													if (self._options.autoScale.fileType === 'keep') {
+														fileType = file.type;
+													}
+													
+													return resizer.getFile(canvas, file.name, fileType);
+												})
+												.catch(function (error) {
+													// In case of an error return the original file
+													console.debug('[WCF.Attachment] Failed to resize', error);
+													resolve(file);
```
TimWolla:
```suggestion
													return file;
```

wcfsetup/install/files/js/WCF.Attachment.js
```diff
@@ -308,6 +317,104 @@ if (COMPILER_TARGET_DEFAULT) {
...
+						
+						return Promise.all(promises);
+					})
+					.catch(function (error) {
+						console.debug('[WCF.Attachment] Error while loading ImageResizer: '+error);
+						
+						// In case of a failure to load the required modules, return the untouched list of files
+						return files;
+					})
+					.then(function (promiseAllOrFiles) {
```
TimWolla: This whole `.then` is a no-op.

wcfsetup/install/files/js/WoltLabSuite/Core/ImageResizer.js
```diff
@@ -0,0 +1,147 @@
...
+					var length = binary.length;
+					var data = new Uint8Array(length);
+					
+					for (var i = 0; i < length; i++) {
+						data[i] = binary.charCodeAt(i);
+					}
+					
+					resolve(new Blob([data], {type: fileType}));
+				}
+			}).bind(this));
```
TimWolla: Unneeded `.bind`?

com.woltlab.wcf/option.xml
```diff
@@ -1029,7 +1032,40 @@ redis:cache_source_redis_host,!cache_source_memcached_host</enableoptions>
...
+keep:wcf.acp.option.attachment_image_autoscale_file_type.keep</selectoptions>
+			</option>
+			<option name="attachment_image_autoscale_quality">
+				<categoryname>message.attachment.autoscale</categoryname>
+				<optiontype>integer</optiontype>
+				<defaultvalue>80</defaultvalue>
+				<minvalue>1</minvalue>
+				<maxvalue>100</maxvalue>
+				<suffix>percent</suffix>
+			</option>
```
Gravatronics: Remember to add the new options to `constants.php`.

wcfsetup/install/files/js/WoltLabSuite/Core/ExifUtil.js
```diff
@@ -0,0 +1,190 @@
...
+		 * Extracts the EXIF / XMP sections of a JPEG blob.
+		 *
+		 * @param       blob    {Blob}                                  JPEG blob
+		 * @param       exif    {Uint8Array}                            APP1 sections
+		 * @returns             {Promise<Uint8Array | TypeError>}       Promise resolving with the altered JPEG blob
+		 */
+		getExifBytesFromJpeg: function (blob) {
+			return new Promise(function (resolve, reject) {
+				if (!(blob instanceof Blob) && !(blob instanceof File)) {
+					return reject(TypeError('The argument must be a Blob or a File'));
```
TimWolla:
```suggestion
					return reject(new TypeError('The argument must be a Blob or a File'));
```

wcfsetup/install/files/js/WoltLabSuite/Core/ExifUtil.js
```diff
@@ -0,0 +1,190 @@
...
+		/**
+		 * Removes all EXIF and XMP sections of a JPEG blob.
+		 *
+		 * @param       blob    {Blob}                          JPEG blob
+		 * @returns             {Promise<Blob | TypeError>}     Promise resolving with the altered JPEG blob
+		 */
+		removeExifData: function (blob) {
+			return new Promise(function (resolve, reject) {
+				if (!(blob instanceof Blob) && !(blob instanceof File)) {
+					return reject(TypeError('The argument must be a Blob or a File'));
```
TimWolla:
```suggestion
					return reject(new TypeError('The argument must be a Blob or a File'));
```

wcfsetup/install/files/js/WoltLabSuite/Core/ExifUtil.js
```diff
@@ -0,0 +1,190 @@
...
+	var _signatureEXIF = 'Exif';
+	var _signatureXMP  = 'http://ns.adobe.com/xap/1.0/';
+	
+	return {
+		/**
+		 * Extracts the EXIF / XMP sections of a JPEG blob.
+		 *
+		 * @param       blob    {Blob}                                  JPEG blob
+		 * @param       exif    {Uint8Array}                            APP1 sections
+		 * @returns             {Promise<Uint8Array | TypeError>}       Promise resolving with the altered JPEG blob
```
TimWolla: The JPEG blob is not being altered, no?

wcfsetup/install/files/js/WoltLabSuite/Core/ExifUtil.js
```diff
@@ -0,0 +1,190 @@
...
+					for (var i = 2; i < bytes.length;) {
+						// each sequence starts with 0xFF
+						if (bytes[i] !== 0xFF) break;
+						
+						var length = 2 + ((bytes[i + 2] << 8) | bytes[i + 3]);
+						
+						// Check if the next byte indicates an EXIF sequence
+						if (bytes[i + 1] === _tagNames.APP1) {
+							var signature = '';
+							for (let j = i + 4; bytes[j] !== 0 && j < bytes.length; j++) {
```
TimWolla:
```suggestion
							for (var j = i + 4; bytes[j] !== 0 && j < bytes.length; j++) {
```

wcfsetup/install/files/js/WoltLabSuite/Core/ExifUtil.js
```diff
@@ -0,0 +1,190 @@
...
```
TimWolla:
```suggestion
							for (var j = i + 4; bytes[j] !== 0 && j < bytes.length; j++) {
```

wcfsetup/install/files/js/WoltLabSuite/Core/ExifUtil.js
```diff
@@ -0,0 +1,190 @@
...
+		 * @param       blob    {Blob}                  JPEG blob
+		 * @param       exif    {Uint8Array}            APP1 sections
+		 * @returns             {Promise<Blob | never>} Promise resolving with the altered JPEG blob
+		 */
+		setExifData: function (blob, exif) {
+			return this.removeExifData(blob).then(function (blob) {
+				return new Promise(function (resolve) {
+					var reader = new FileReader();
+					
+					reader.addEventListener("loadend", function () {
```
TimWolla: Is it possible that the load fails and the `Promise` needs to be `reject`ed?

wcfsetup/install/files/js/WoltLabSuite/Core/ExifUtil.js
```diff
@@ -0,0 +1,190 @@
...
+		 */
+		removeExifData: function (blob) {
+			return new Promise(function (resolve, reject) {
+				if (!(blob instanceof Blob) && !(blob instanceof File)) {
+					return reject(TypeError('The argument must be a Blob or a File'));
+				}
+				
+				var reader = new FileReader();
+				
+				reader.addEventListener("loadend", function () {
```
TimWolla: Is it possible that the load fails and the `Promise` needs to be `reject`ed?

wcfsetup/install/files/js/WoltLabSuite/Core/ExifUtil.js
```diff
@@ -0,0 +1,190 @@
...
+		 */
+		getExifBytesFromJpeg: function (blob) {
+			return new Promise(function (resolve, reject) {
+				if (!(blob instanceof Blob) && !(blob instanceof File)) {
+					return reject(TypeError('The argument must be a Blob or a File'));
+				}
+				
+				var reader = new FileReader();
+				
+				reader.addEventListener("loadend", function() {
```
TimWolla: Is it possible that the load fails and the `Promise` needs to be `reject`ed?

wcfsetup/install/files/js/WoltLabSuite/Core/ImageResizer.js
```diff
@@ -0,0 +1,145 @@
...
+ * @copyright	2001-2018 WoltLab GmbH
+ * @license	GNU Lesser General Public License <http://opensource.org/licenses/lgpl-license.php>
+ * @module	WoltLabSuite/Core/ImageResizer
+ */
+define([
+	'WoltLabSuite/Core/FileUtil',
+	'WoltLabSuite/Core/ExifUtil',
+	'Pica'
+], function(FileUtil, ExifUtil, Pica) {
+	var pica = Pica({ features: [ 'js', 'wasm', 'ww' ] });
```
TimWolla:
```suggestion
	var pica = new Pica({ features: [ 'js', 'wasm', 'ww' ] });
```

wcfsetup/install/files/js/WoltLabSuite/Core/ImageResizer.js
```diff
@@ -0,0 +1,145 @@
...
+				.then(function (blob) {
+					if (fileType === 'image/jpeg' && typeof exif !== 'undefined') {
+						blob = ExifUtil.setExifData(blob, exif);
+					}
+					
+					var matches = fileName.match(/(.+)(\..+?)$/);
+					
+					// If we inserted EXIF data blob has been converted to a Promise
+					return Promise.resolve(blob)
+						.then(function (blob) {
```
TimWolla: - Append this `.then` to the outer Promise
- Move `var matches` into the `.then`.
- You now can `return ExifUtil.setExifData(blob, exif)` in the `if` above and `return blob` otherwise.

wcfsetup/install/files/js/WoltLabSuite/Core/ImageResizer.js
```diff
@@ -0,0 +1,145 @@
...
+		
+		setFileType: function (value) {
+			if (value == null) value = ImageResizer.prototype.fileType;
+			
+			this.fileType = value;
+			return this;
+		},
+		
+		getFile: function (result, fileName, fileType, quality) {
+			var image = result.image;
```
TimWolla: No need for these local variables, they are only read once, each.

wcfsetup/install/files/js/require.config.js
```diff
@@ -3,12 +3,13 @@ requirejs.config({
...
 		favico: '3rdParty/favico',
-		'perfect-scrollbar': '3rdParty/perfect-scrollbar'
+		'perfect-scrollbar': '3rdParty/perfect-scrollbar',
+		'Pica': '3rdParty/pica'
 	},
 	shim: {
 		enquire: { exports: 'enquire' },
 		favico: { exports: 'Favico' },
-		'perfect-scrollbar': { exports: 'PerfectScrollbar' }
+		'perfect-scrollbar': { exports: 'PerfectScrollbar' },
```
TimWolla:
```suggestion
		'perfect-scrollbar': { exports: 'PerfectScrollbar' }
```

wcfsetup/install/files/js/require.config.js
```diff
@@ -44,7 +45,7 @@ requirejs.config({
 			'Ui/TabMenu': 'WoltLabSuite/Core/Ui/TabMenu',
 			'Upload': 'WoltLabSuite/Core/Upload',
 			'User': 'WoltLabSuite/Core/User'
-		}
+		},
```
TimWolla:
```suggestion
		}
```

wcfsetup/install/files/js/WoltLabSuite/Core/ImageResizer.js
```diff
@@ -0,0 +1,145 @@
...
+			reader.readAsDataURL(file);
+			
+			return exif.then(function (exif) {
+				return new Promise(function (resolve, reject) {
+					image.onerror = reject;
+					
+					image.onload = function () {
+						var canvas = document.createElement('canvas');
+						
+						// Prevent upscalingq
```
TimWolla:
```suggestion
						// Prevent upscaling
```

wcfsetup/install/files/js/WoltLabSuite/Core/ImageResizer.js
```diff
@@ -0,0 +1,145 @@
...
+						else if (quality >= 0.4) {
+							resizeQuality = 2;
+						}
+						
+						var options = {
+							quality: resizeQuality,
+							cancelToken: cancelPromise
+						};
+						
+						resolve(pica.resize(image, canvas, options).then(function (result) {
```
TimWolla: This should read:

```
pica.resize(image, canvas, options).then(function (result) {
 resolve({ image: result, exif: exif });
}, reject)
```

You forgot to handle the “error case”.

wcfsetup/install/files/js/WoltLabSuite/Core/ImageResizer.js
```diff
@@ -0,0 +1,197 @@
...
+		 */
+		getFile: function (result, fileName, fileType, quality) {
+			fileType = fileType || this.fileType;
+			quality = quality || this.quality;
+			
+			var basename = fileName.match(/(.+)(\..+?)$/);
+			
+			return pica.toBlob(result.image, fileType, quality)
+				.then(function (blob) {
+					if (fileType === 'image/jpeg' && typeof exif !== 'undefined') {
```
joshuaruesweg: `exif` should be `result.exif`

wcfsetup/install/files/js/WCF.Attachment.js
```diff
@@ -285,20 +292,112 @@ if (COMPILER_TARGET_DEFAULT) {
...
-				this._removeButton();
-				this._createButton();
-			}
-			
-			return $uploadID;
+			require([
+				'WoltLabSuite/Core/FileUtil',
+				'WoltLabSuite/Core/ImageResizer',
+				'WoltLabSuite/Core/Ajax/Status'
+			]).then((function (modules) {
```
dtdesign: Please use the regular syntax that we use pretty much everywhere: `require([…], function(…) {});`.

wcfsetup/install/files/js/WCF.Attachment.js
```diff
@@ -285,20 +292,112 @@ if (COMPILER_TARGET_DEFAULT) {
...
+					}
+					
+					// Resize the images in series.
+					// As our resizer is based on Pica it will use multiple workers per image if possible.
+					$promise = Array.prototype.reduce.call($files, (function (acc, file) {
+						return acc.then((function (arr) {
+							var $timeout = new Promise(function (resolve, reject) {
+								setTimeout(function () {
+									resolve(file);
+								}, 10e3); // Timeout per image
```
dtdesign: `10000` is much more legible than `10e3`, you could even write it as `10 * 1000` if you're concerned about this. Also, the comment should highlight that the timeout is asynchronous for each image and thus runs in parallel.

wcfsetup/install/files/js/WCF.Attachment.js
```diff
@@ -285,20 +292,112 @@ if (COMPILER_TARGET_DEFAULT) {
...
+							});
+							
+							var $promise = $resizer.resize(file, $maxWidth, $maxHeight, $quality, file.size > $maxSize, $timeout)
+							.then((function (result) {
```
dtdesign: Inconsistent indentation, it should be one level deeper than the `var $promise` statement above.

wcfsetup/install/files/js/WCF.Attachment.js
```diff
@@ -285,20 +292,112 @@ if (COMPILER_TARGET_DEFAULT) {
...
+					var $maxHeight = this._options.autoScale.maxHeight;
+					var $quality = this._options.autoScale.quality;
+					
+					if (this._options.autoScale.fileType !== 'keep') {
+						$resizer.setFileType(this._options.autoScale.fileType);
+					}
+					
+					// Resize the images in series.
+					// As our resizer is based on Pica it will use multiple workers per image if possible.
+					$promise = Array.prototype.reduce.call($files, (function (acc, file) {
```
dtdesign: Is there any reason for not calling `$files.reduce()`? It is always an array because of line 308.

wcfsetup/install/files/js/WCF.Attachment.js
```diff
@@ -285,20 +292,112 @@ if (COMPILER_TARGET_DEFAULT) {
...
+					
+					if (this._fileUpload) {
+						// remove and re-create the upload button since the 'files' property
+						// of the input field is readonly thus it can't be reset
+						this._removeButton();
+						this._createButton();
+					}
+					
+					return $uploadID;
+				}).bind(this)).finally(AjaxStatus.hide);
```
dtdesign: Inconsistent use of newlines, `finally()` should go into a new line.

wcfsetup/install/files/js/WCF.Attachment.js
```diff
@@ -285,20 +292,112 @@ if (COMPILER_TARGET_DEFAULT) {
...
+			}).bind(this)).catch(function (error) {
```
dtdesign: Same as above, `catch()` should be on its own line.

wcfsetup/install/files/js/WCF.js
```diff
@@ -5966,39 +5966,46 @@ if (COMPILER_TARGET_DEFAULT) {
...
 			else {
-				$files = this._fileUpload.prop('files');
+				if (file) {
+					$files.push(file);
+				}
+				else if (blob) {
+					var $ext = '';
+					switch (blob.type) {
+						case 'image/png':
+							$ext = '.png';
```
dtdesign: That's a bit counter-intuitive, we usually only store the actual extension in variables, not the dot itself.

wcfsetup/install/files/js/WoltLabSuite/Core/ExifUtil.js
```diff
@@ -0,0 +1,204 @@
...
+								bytes.set(start, 0);
+								bytes.set(end, start.length);
+							}
+						}
+						else {
+							i += length;
+						}
+					}
+					
+					resolve(new Blob([ bytes ], { type: blob.type }));
```
dtdesign: Superfluous spaces, it should read `[bytes]`.

wcfsetup/install/files/js/WoltLabSuite/Core/ExifUtil.js
```diff
@@ -0,0 +1,204 @@
...
+						
+						var start = bytes.slice(0, offset);
+						var end = bytes.slice(offset);
+						
+						bytes = new Uint8Array(start.length + exif.length + end.length);
+						bytes.set(start);
+						bytes.set(exif, offset);
+						bytes.set(end, offset + exif.length);
+						
+						resolve(new Blob([ bytes ], { type: blob.type }));
```
dtdesign: `[ bytes ]` -> `[bytes]`

wcfsetup/install/files/js/WCF.Attachment.js
```diff
@@ -285,20 +292,112 @@ if (COMPILER_TARGET_DEFAULT) {
...
+					$files.push(file);
+				}
+				else if (blob) {
+					$files.push(FileUtil.blobToFile(blob, 'pasted-from-clipboard'));
+				}
+				else {
+					$files = this._fileUpload.prop('files');
+				}
+				
+				// Default action is to resolve with the unaltered list of files
```
dtdesign: In general: Comments should be a whole sentence ending with a period and they should describe _why_ something is done, not _how_.

wcfsetup/install/files/js/WCF.Attachment.js
```diff
@@ -285,20 +292,112 @@ if (COMPILER_TARGET_DEFAULT) {
...
+				var $promise = Promise.resolve($files);
+				
+				if (this._options.autoScale && this._options.autoScale.enable) {
+					var $maxSize = this._buttonSelector.data('maxSize');
+					
+					// Create an ImageResizer instance
```
dtdesign: Remove this comment please, it's useless.

wcfsetup/install/files/js/3rdParty/pica.js
```diff
@@ -0,0 +1,2223 @@
+/* (The MIT License)
```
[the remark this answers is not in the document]
TimWolla: pica is automatically pulled into `WoltLabSuite.Core.min.js` by r.js / when running _buildCore.js, because it is loaded as a require module. The same happens to favico, perfect-scrollbar and enquire.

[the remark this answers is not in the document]
dtdesign: ✔️

com.woltlab.wcf/option.xml
```diff
@@ -970,6 +973,39 @@ redis:wcf.acp.option.cache_source_type.redis</selectoptions>
 				<minvalue>96</minvalue>
 				<suffix>pixel</suffix>
 			</option>
+			<option name="attachment_image_autoscale">
+				<categoryname>message.attachment.autoscale</categoryname>
+				<optiontype>boolean</optiontype>
+				<defaultvalue>1</defaultvalue>
```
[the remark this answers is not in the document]
dtdesign: How about 1920×1200? That covers the largest typical desktop resolution and everything above gets scaled down.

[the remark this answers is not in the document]
max-m: My though for defaulting to square dimensions was to accommodate smartphone screenshots, which usually come in portrait orientation but scaling them to 675×1200 in case of Full HD smartphone panels should still ok, I guess.

com.woltlab.wcf/option.xml
```diff
@@ -970,6 +973,39 @@ redis:wcf.acp.option.cache_source_type.redis</selectoptions>
...
+				<optiontype>integer</optiontype>
+				<defaultvalue>1024</defaultvalue>
+				<minvalue>1</minvalue>
+			</option>
+			<option name="attachment_image_autoscale_file_type">
+				<categoryname>message.attachment.autoscale</categoryname>
+				<optiontype>radioButton</optiontype>
+				<defaultvalue>keep</defaultvalue>
+				<selectoptions>keep:wcf.acp.option.attachment_image_autoscale_file_type.keep
+image/jpeg:wcf.acp.option.attachment_image_autoscale_file_type.jpeg</selectoptions>
```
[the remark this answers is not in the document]
dtdesign: I would suggest to test the performance of these pixel checks, especially on smartphones, and see how it works. Unless there are some significant issues, we should try to preserve the transparency if possible.

wcfsetup/install/files/js/WCF.Attachment.js
```diff
@@ -285,20 +292,112 @@ if (COMPILER_TARGET_DEFAULT) {
...
+				'WoltLabSuite/Core/ImageResizer',
+				'WoltLabSuite/Core/Ajax/Status'
+			]).then((function (modules) {
+				var FileUtil = modules[0];
+				var ImageResizer = modules[1];
+				var AjaxStatus = modules[2];
+				
+				AjaxStatus.show();
+				
+				var $files = [];
```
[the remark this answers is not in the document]
max-m: Oh, well, I was told to keep the style consistent and use the `var $foo` scheme in the old module and `var foo` in new require-modules. :)

wcfsetup/install/files/js/WCF.Attachment.js
max-m: Using the `Promise` returned by require allows nice error handling, consider the following

```
require(['WoltLabSuite/Core/Event/Handler'], function (x) { return Promise.reject(new Error('Callback')) }, function (err) { console.error('Error caught!', err) })

Promise {<resolved>: Arguments(1)}
VM217:1 Uncaught (in promise) Error: Callback
    at <anonymous>:1:83
    at WoltLabSuite.Core.min.js?v=1539810053:16

require(['WoltLabSuite/Core/Event/Handler']).then(function (x) { return Promise.reject(new Error('Callback')) }).catch(function (err) { console.error('Error caught!', err) })

VM223:1 Error caught! Error: Callback
    at <anonymous>:1:88
Promise {<resolved>: undefined}
```

com.woltlab.wcf/option.xml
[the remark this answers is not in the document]
dtdesign: I thought about this too, but smartphone differ greatly in size and there really isn't a common ground. I have a few ideas about auto-scaling images from 4K screens, but in the end they don't end up well.

Recently, I dug through the various screen resolutions of handheld devices and I cam up with this range of screen dimensions. The values denote the range of sizes for the *sm*allest and the *w*i*d*est side regardless of the orientation. They're not perfect, but they do give an idea of the ranges to expect (notably, there are actually no overlaps!).

*iPhone*
sm: 320-414
wd: 568-896

*iPad*
sm: 768
wd: 1024

*Android Phone*
sm: 360-480
wd: 640-853

*Android Tablet*
sm: 600-850
wd: 960-1024

wcfsetup/install/files/js/WCF.Attachment.js
dtdesign: That's a bogus example, you're abusing the error handler for `require()` itself which also triggers if loading the module fails for whatever reasons. Remember, you're rejecting the promise of `require` and not of whatever you're doing inside.

```
require([…], function(…) {
  (new Promise()).catch(…);
});
```

wcfsetup/install/files/js/WCF.Attachment.js
max-m: Line 316, `$files` can be a `FileList` which is iterable but does not inherit the Array prototype.

wcfsetup/install/files/js/WCF.js
max-m: That code has been left untouched :)

TimWolla:
Please check on windows (IE, Edge), Mac (Safari) and iOS (Safari) whether uploading still works. Scaling does not need to happen, but it would be great if it did.

dtdesign:
It works in Internet Explorer 11 and Microsoft Edge 44, both on Windows 10 x64.

However, the uploaded image was not resized twice in Edge, the resulting image was larger than the maximum allowed resolution, in fact, it wasn't even resized at all. The console printed this error:
```
[WCF.Attachment] Failed to resize image "OCIFOG0_1300.jpg": SecurityError
```
`WCF.Attachment.js (383,10)`

I tried it 7 times with the same exact image and it failed on attempt 1 and 3, all uploads were performed in the thread reply window, not in the thread add form. It's somewhat reproducible, because I am able to trigger this error from time by time, by doing the exact same thing. It doesn't yield a stack trace or anything else.

dtdesign:
I'm unable to reproduce this issue in Edge anymore, tried it over 30 times in different combinations, but it doesn't happen again. It doesn't make much sense, considering that it sometimes did not work while creating successive posts without reloading the page.

Could've been just Edge having some hiccups. We should keep an eye on this during the beta phase, but for now I see no point in pursing this any further.

LGTM

PR #2786, opened 2018-12-03
Update Codemirror to version 5.42.0

TimWolla:


TimWolla:
see: https://community.woltlab.com/thread/273598-css-editor-multipliziert-textebausteine-mobil-nach-belieben/?postID=1718496#post1718496

TimWolla:
Does not fix the issue it should fix.

PR #2795, opened 2018-12-08
Use dataset to get the data attributes

Cyperghost:
- using HTML5 dataset
- Optimize some JS Doc´s


dtdesign:
`Element.dataset` transform the name of data properties to camelCase, e. g. `data-object-id` is accessible via `dataset['objectId']`, but not `dataset['object-id']`.

This changes breaks the entire software without offering any advantages.

Issue #2492, opened 2017-11-30
Future Changes


dtdesign:
Disclaimer: This is a rough list of things that may or may not be considered for the next major release. Things will get added, removed or postponed till eternity - do not take anything here for granted.

Things listed here are **not guaranteed** to be implemented **EVER**.

* **Simple Comment Integration for Page Classes**
  Integrating comments on a page requires a bit too much repetitive PHP code, it should be off-loaded into a trait to reduce copy&paste. – @BurntimeX 
* **Efficient DOM-Element Creation**
  Creating even simple DOM elements is rather cumbersome because of the rather verbose element API, with `elCreate()` being merely a wrapper around `document.createElement()`. We could integrate [hyperscript](https://github.com/hyperhype/hyperscript) that has proven to be both efficient and simple (see [preact](https://github.com/developit/preact)), in order to reduce the boilerplate-noise created by the native element API. – @dtdesign 
* **PIP GUI: Required Language Items**
  Certain PIPs require additional language items, for example for objects types based on the object type identifier. It would be useful if such language items could be edited directly as part of editing the entry. – Gravatronics 
* **Reinstall package via devtools**
  Devtools should support uninstallation and installation of the same package with one click of a button. – Gravatronics 
* **Simplify Notification System**
  Implementing new notifications is too complex. Possibilities to reduce the complexity:
  * Work the DBO classes directly as most already implement all of the relevant methods and instead of `IUserNotificationObjectType::getObjectsByIDs()` use runtime caches or object providers.
  * Simplify language items: Only require one language item for titles (which internally handles stacked vs. non-stacked notifications via the `{plural}` template plugin) and one language item for messages which is also used in the user interface and for emails.
  * An event class for event will probably still be required but it should be simplified by providing default implementations for most methods as current implementations for title and message getters almost only differ in the names of the language items (which could be determined via the object type of the event object and the event name) and the variables made available in the language items.
  -- Gravatronics 
* **Mark deprecated methods and fields**
  The use of PhpDoc comments to flag deprecated methods is quite inconsistent and cumbersome, including the lack of a reporting ability. #2798 was an earlier attempt before PHP8's attributes were introduced that appear to be exactly what we are looking for. We should monitor https://wiki.php.net/rfc/deprecated_attribute for a possible built-in attribute in the future. -- TimWolla

## Archive

* **`SameSite` Attribute for Cookies** (#3593)
[...]

PR #2798, opened 2018-12-09
Add option to make deprecated functionality inaccessible

### LINKED ISSUE #2789: Add option to make deprecated functionality inaccessible
To ease migrating from deprecated functionality to the “new version” it should be possible to prevent using this deprecated functionality and make the old calls fail hard.

- [ ] Option `ENABLE_DEPRECATION_WARNINGS`
- [ ] Function `\wcf\functions\deprecatedFeature($since, $linkToDocs)` (*) which is a no-op if the option is disabled and `throw`s otherwise.
- [ ] Similar function for JavaScript
- [ ] Update deprecated methods to first call this function
- [ ] Update deprecated class files to call this function before defining the class (i.e. when the file is being autoloaded)

(*) `$linkToDocs` points to an article (or an anchor within one) on `docs.woltlab.com` that explains what changes have to be made in order to migrate to the new code.

TimWolla:
Closes #2789 

image


dtdesign:
The example perfectly outlines why simple text annotations won't cut it. Throwing a deprecation error that leads deeper into the rabbit doesn't help a developer, they need to know *how* they can migrate their API calls. This is the whole point of why I wanted the function to include a URL that points to some kind of docs.

joshuaruesweg:
Is the subordinate clause "while 'ENABLE_DEPRECATION_WARNINGS' is enabled:" really needed? That sounds more like overhead to me.

com.woltlab.wcf/option.xml
```diff
@@ -1615,6 +1615,11 @@ DESC:wcf.global.sortOrder.descending</selectoptions>
 				<optiontype>boolean</optiontype>
 				<defaultvalue>0</defaultvalue>
 			</option>
+			<option name="enable_deprecation_warnings">
```
joshuaruesweg: You should probably set a dependency on the dev tool option here.

Gravatronics: I have done the same for `log_missing_language_items`.

wcfsetup/install/files/lib/core.functions.php
```diff
@@ -90,6 +90,97 @@ function wcfDebug() {
 	}
 }
 
+namespace wcf\functions\deprecated {
```
Gravatronics: Change the namespace to `wcf\functions` (see comments on function names below).

wcfsetup/install/files/lib/core.functions.php
```diff
@@ -90,6 +90,97 @@ function wcfDebug() {
...
+	use \wcf\system\exception\IExtraInformationException;
```
Gravatronics: We do not use leading backslashes for imports.

wcfsetup/install/files/lib/core.functions.php
```diff
@@ -90,6 +90,97 @@ function wcfDebug() {
...
+				public function getExtraInformation() {
+					return $this->extraInformation;
+				}
+			}
+		}
+		
+		return new DeprecationError($message, $extraInformation);
+	}
+	
+	function feature($message, $extraInformation) {
```
Gravatronics: Change to `deprecatedFeature`.
Also, when using the function (except for the two functions below), wouldn't you want to pass a string as the second parameter?

wcfsetup/install/files/lib/core.functions.php
```diff
@@ -90,6 +90,97 @@ function wcfDebug() {
...
+		if (!ENABLE_DEPRECATION_WARNINGS) return;
+		
+		throw DeprecationError("A deprecated feature is used, while 'ENABLE_DEPRECATION_WARNINGS' is enabled: ".$message, $extraInformation);
```
Gravatronics: May-be change the name of the `DeprecationError()` function to `getDeprecationError()`? At first glance, this looks like a `new` is missing.

wcfsetup/install/files/lib/core.functions.php
```diff
@@ -90,6 +90,97 @@ function wcfDebug() {
...
+	}
+	
+	/**
+	 * Sends a deprecation warning for the given class when the returned
+	 * object is destructed.
+	 * 
+	 * Usage:
+	 * 
+	 * $_ = \wcf\functions\deprecated\_class(Foo::class);
```
Gravatronics: Is there a reason for the assignment?

wcfsetup/install/files/lib/core.functions.php
```diff
@@ -90,6 +90,97 @@ function wcfDebug() {
...
+	 * class Foo { ... }
+	 * 
+	 * @param	string	$className
+	 * @return	object
+	 */
+	function _class($className) {
```
Gravatronics: Change to `deprecatedClass`.

wcfsetup/install/files/lib/core.functions.php
```diff
@@ -90,6 +90,97 @@ function wcfDebug() {
...
+		if (!ENABLE_DEPRECATION_WARNINGS) return;
+
+		return new class($className) {
```
Gravatronics: Please add a comment explaining why this indirect way of throwing the exception is necessary.

wcfsetup/install/files/lib/core.functions.php
```diff
@@ -90,6 +90,97 @@ function wcfDebug() {
...
+		};
+	}
+	
+	/**
+	 * Sends a deprecation warning for the given method.
+	 * 
+	 * Usage:
+	 * 
+	 * public function foo() {
+	 *  \wcf\functions\deprecated\_method(__CLASS__, __FUNCTION__);
```
Gravatronics: `__METHOD__` instead of `__FUNCTION__`?

wcfsetup/install/files/lib/core.functions.php
```diff
@@ -90,6 +90,97 @@ function wcfDebug() {
...
+	 *  ...
+	 * }
+	 * 
+	 * @param	string	$className
+	 * @param	string	$method
+	 */
+	function _method($className, $method) {
```
Gravatronics: Change to `deprecatedMethod`.

wcfsetup/install/lang/de.xml
```diff
@@ -1505,6 +1505,8 @@ Als Benachrichtigungs-URL in der Konfiguration der sofortigen Zahlungsbestätigu
 		<item name="wcf.acp.option.message_sidebar_enable_articles"><![CDATA[Artikelanzahl der Autoren anzeigen]]></item>
 		<item name="wcf.acp.option.log_missing_language_items"><![CDATA[Fehlende Sprachvariablen protokollieren]]></item>
 		<item name="wcf.acp.option.log_missing_language_items.description"><![CDATA[Die fehlenden Sprachvariablen werden in der Datei <code>{'WCF_DIR'|constant}log/missingLanguageItems.txt</code> protokolliert.]]></item>
+		<item name="wcf.acp.option.enable_deprecation_warnings"><![CDATA[Fehler bei Verwendung von @deprecated code]]></item>
```
Gravatronics: Put `@deprecated` in a `kbd` element? (Same question for the English version and the description.)

wcfsetup/install/lang/de.xml
```diff
@@ -1505,6 +1505,8 @@ Als Benachrichtigungs-URL in der Konfiguration der sofortigen Zahlungsbestätigu
...
+		<item name="wcf.acp.option.enable_deprecation_warnings.description"><![CDATA[Verhindert die Verwendung von @deprecated Klassen und Methoden, indem bei Verwendung ein <kbd>\Error</kbd> geworfen wird.]]></item>
```
Gravatronics: `Klassen und Methoden` might be too restrictive as it can also be relevant for parts of methods which would then also not rely on `@deprecated` (thus, simply replace with `Code`?). On the other hand, may-be this objection it is too specific.

ghost (changes requested):
Should this feature be enabled by default in developer mode? (cc @dtdesign)

wcfsetup/install/files/lib/core.functions.php
TimWolla: `__METHOD__` is `__CLASS__.'::'.__FUNCTION` which needs to be split manually to pass to reflection.

wcfsetup/install/files/lib/core.functions.php
TimWolla: Added explanation to the comment.

wcfsetup/install/lang/de.xml
TimWolla: I think that the specific examples are helpful to understand what exactly happens here. “Preventing using code” is somewhat vague.

wcfsetup/install/files/lib/core.functions.php
TimWolla: >  Also, when using the function (except for the two functions below), wouldn't you want to pass a string as the second parameter?

You would use `$message` for a quick message (“`The $baz parameter of Foo::bar() is deprecated`”) and `$extraInformation` for additional details (such as the web link to the docs).

wcfsetup/install/lang/de.xml
Gravatronics: It would `Verhindert die Verwendung von @deprecated Code`.

TimWolla:
I just finished up the PHP part and added some calls for the “easy to migrate” functions / classes. Does this look good so far?

wcfsetup/install/files/js/wcf.globalHelper.js
```diff
@@ -289,6 +289,24 @@
 		return obj.hasOwnProperty(property);
 	};
 	
+	window.deprecatedFeature = function(message) {
+		console.log("A deprecated feature is used, while 'ENABLE_DEPRECATION_WARNINGS' is enabled: " + message);
```
dtdesign: This creates a significant nuisance during normal operations, these should be suppressed unless `ENABLE_PRODUCTION_DEBUG_MODE` is enabled.

Also, we should consider including `new Error().stack` as an additional parameter, otherwise this message will be a nightmare to debug.

dtdesign (changes requested):
Regarding the JS API only:
The functions for deprecating code in the JS API should be used for functions that are _really_ deprecated, that is, the entire function was already replaced by a sole log message or an error.

Anything that is a soft deprecation, using a plain comment or printing a message but continuing, should be excempted for now. The JS API is in a bit of a weird state, where deprecated features are actively used by us to this day (e. g. `WCF.Language`) and there won't be any changes to this in 5.2. It should be postponed until a future release which cleans up this mess.

dtdesign:
@TimWolla Please look into this issue and decide if anything needs to be changed or if it should be postponed to 5.3.

dtdesign:
This PR is quite old and had a few issues, along with a non trivial detection of deprecated API components and difficulties with proper reporting. PHP 8 introduces attributes which would be a perfect fit for this, providing a native and familiar experience, including but not limited to access to deprecated properties.

For this reason, I suggest to burry this PR for good and revisit the concept again once we've moved to PHP 8.

TimWolla:
PHP 8 would make this way more elegant indeed, but I guess it's quite some time out until we are able to use PHP 8 sanely, because of all those slow upgrading webhosters out there.



> This PR is quite old and had a few issues, along with a non trivial detection of deprecated API components and difficulties with proper reporting.

As you might've seen I've updated this PR recently. On the PHP side it's working quite well from my tests. The major issue right now is that I'm not always able to fill in an actionable description within the `@deprecated` tag, because the class / method is not my area of expertise and the commit is either hard to find or useless with regard to that. For old stuff this will now change with PHP 8 attributes and with new stuff this PR just as easy to use as PHP 8 attributes.

TimWolla:
This PR is more than 2 years old without getting anywhere. PHP 8's attribute syntax is now backwards compatible by being interpreted as a comment in PHP 7 and PHP 8.1 might get [a `Deprecated` attribute](https://wiki.php.net/rfc/deprecated_attribute) in the standard library.

Implementing this should be revisited at a later point, while taking the research that went into this PR into account.

see also:
#3880 
#3881 
#2492

PR #2813, opened 2019-01-02
Add missing data-identifier to mobile menu

Morik:
All menus items contain their identifier as data attribute.
This allows admins to easily add css specific to those links (like icons etc).
see: https://github.com/WoltLab/WCF/blob/master/com.woltlab.wcf/templates/__menu.tpl#L6
This is missing on the mobile menu version and added with this pull request.

### DIFF
--- a/com.woltlab.wcf/templates/pageMenuMobile.tpl
+++ b/com.woltlab.wcf/templates/pageMenuMobile.tpl
@@ -3,7 +3,7 @@
 	<ol class="menuOverlayItemList" data-title="{lang}wcf.menu.page{/lang}">
 		<li class="menuOverlayTitle">{lang}wcf.menu.page.navigation{/lang}</li>
 		{foreach from=$__wcf->getBoxHandler()->getBoxByIdentifier('com.woltlab.wcf.MainMenu')->getMenu()->getMenuItemNodeList() item=menuItemNode}
-			<li class="menuOverlayItem">
+			<li class="menuOverlayItem" data-identifier="{@$menuItemNode->identifier}">
 				{assign var=__outstandingItems value=$menuItemNode->getOutstandingItems()}
 				<a href="{$menuItemNode->getURL()}" class="menuOverlayItemLink{if $__outstandingItems} menuOverlayItemBadge{/if}{if $menuItemNode->isActiveNode()} active{/if}"{if $menuItemNode->isExternalLink() && EXTERNAL_LINK_TARGET_BLANK} target="_blank"{/if}>
 					<span class="menuOverlayItemTitle">{lang}{$menuItemNode->title}{/lang}</span>


dtdesign:
I'm a bit hesitant, I've seen attributes carelessly being used in CSS selectors without adding any sort of context, such as a class name. Introducing this into the master is somewhat dangerous, essentially leaving us with only two options:

1. Postpone this change and apply it in 5.2 only.
2. Add it to `3.0`, but slightly alter the identifier to prevent collisions, e. g. `{...->identifier}-mobile`.

That said, I'm not really big into option 2 for obvious reasons, so unless you can give me a good reason, I'm leaning towards option 1.

Morik:
The mobile menu has already an other class structure so that a change should not impact other css code.
And i've targeted the 3.1 branch as this identifier wasn't added in the 3.0 branch.
see: https://github.com/WoltLab/WCF/blob/3.0/com.woltlab.wcf/templates/__menu.tpl#L4

Morik:
Ok, i get it :P
But this could also happen in the 3.2 tree, it would only delay potential collisions.
If this is a serious problem we should prefer the altered version anyway...

dtdesign:
The difference is that it's likely that the site owner will review their site after an upgrade to check for any issues. This is a bit different with regular updates that are meant to be applied as soon as possible in a live environment, especially given that we strive to avoid making any significant changes to the software's behavior that could break the system.

Is there any reason not to postpone this to 5.2? Introducing it with the next major using the exact same attribute name is fine, I just don't want to take the risk of adding it to the current stable version.

Morik:
To be hornets, mainly because i'm currently developing a plugin to add distinct icons to every menu icon and it would make life so much easier^^
And the data attribute was added for exactly such purposes in the first place.
But oviously i can overcome this by ugly querys to the href attribute -.-

dtdesign:
How about using `data-menu-identifier` instead, it wouldn't be the same as the attribute used in the `__menu` template, but still similar enough.

Your case perfectly outlines why it would be dangerous to only make modifications based on the attribute and therefore you would be required to include a more specific selector anyway. At the same time that means that you need to write two different selectors anyway, e. g. because of the differences in the markup, and therefore there is no real downside of using a slightly different attribute.

Just add a comment above the line that explains this difference:
```
{* Does not use `data-identifier` to prevent compatibility issues. See https://github.com/WoltLab/WCF/pull/2813 *}
```

Morik:
I have no problems with a slightly modified attribute name (as you mentioned, i'll need to add a special case anyway for the mobile menu) it's just so much nicer to not be forced to use the href attribute :D
If you are ok with such a change, i'm happy to modify the code accordingly.
But how about `data-mobile-identifier`, seams to be a little bit more intuitive to me.

dtdesign:
I'm fine with the suggested name and please include the comment above, it makes it much more clearer when visiting this part of the code somewhat later.

SoftCreatR:
@dtdesign is this going to be merged into 3.0 as well?

dtdesign:
No, 3.0 does not expose those identifiers at all, this feature is available with 3.1 and later only.

Issue #2815, opened 2019-01-04
Use API_VERSION for update servers and xsd paths


Gravatronics:
Replaces #2796.

Instead of ambiguous code names, use `API_VERSION` in the links of update servers and XSD paths:
- `https://update.woltlab.com/2019/`
- `https://store.woltlab.com/2019/`
- `https://www.woltlab.com/XSD/2019/foo.xsd`

PR #2821, opened 2019-01-05
Use API_VERSION for update servers and xsd paths

Gravatronics:
See #2815

### DIFF
--- a/wcfsetup/setup/db/install.sql
+++ b/wcfsetup/setup/db/install.sql
@@ -2216,8 +2216,11 @@ INSERT INTO wcf1_user_group_option_value (groupID, optionID, optionValue) VALUES
 INSERT INTO wcf1_user_group_option_value (groupID, optionID, optionValue) VALUES (4, 3, '1');	-- Administrators
 
 -- default update servers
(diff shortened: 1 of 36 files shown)


wcfsetup/setup/db/install.sql
```diff
@@ -2216,8 +2216,11 @@ INSERT INTO wcf1_user_group_option_value (groupID, optionID, optionValue) VALUES
 INSERT INTO wcf1_user_group_option_value (groupID, optionID, optionValue) VALUES (4, 3, '1');	-- Administrators
 
 -- default update servers
+-- TODO: change 2019 servers to `online`
```
mutec: Seriously 6 servers meaning 6 single requests per list-update? That's aweful - otherwise it's not my server.

dtdesign: It should be only one update server type using the API version number.

dtdesign: The servers for `vortex` and `tornado` will be removed later in the development cycle, they remain for testing purposes to validate that the update server API remains compatible with any change being made.

Issue #2808, opened 2018-12-22
Remove Google+ before March 7th, 2019


TimWolla:
see https://www.blog.google/technology/safety-security/expediting-changes-google-plus/

- [x] Check whether Google Login relies on G+ scopes (G+ API will shut down on March 7th).
- [x] Remove G+ share button.
- [x] Remove G+ profile field.

TimWolla:
Note: The G+ API will already shut down on March 7th.

TimWolla:
ref: aa7557e0d16d6b60d672cfa7fe9ac1218b782fe6

PR #2823, opened 2019-01-06
Update language files

SoftCreatR:
After moving from G+ (see https://github.com/WoltLab/WCF/issues/2808), the corresponding language variables are outdated.


SoftCreatR:
The whole PR is f'd up. I'll recreate it :P

PR #2824, opened 2019-01-06
Replace (auth) references to Google+ to Google

SoftCreatR:
After moving from G+ (see #2808), some references are still remaining.


TimWolla:
`s/to/from/` in the commit message, added additional references and applied to 3.0 instead of master, thanks.

PR #2828, opened 2019-01-10
Fix installation

SoftCreatR:
597f120 broke the WSC installation process by not checking, if `ENABLE_DEVELOPER_TOOLS` is defined.


Morik:
Well, i believe the right way would be to include the constant in the `WCFSetup.class.php`.

SoftCreatR:
> Well, i believe the right way would be to include the constant in the `WCFSetup.class.php`.

Due to the fact, that the same check exists a few lines below the errornous one, i thought it would be a better idea to keep this approach. Adding this constant to the `WCFSetup.class.php` might also do the trick, but that would be slightly inconsistent.

However, at the end, both approaches do the trick, so it's up to them, what they prefer.

PR #2829, opened 2019-01-10
Improve information in the system tab on the admin panel index even more

SoftCreatR:
Displaying some important PHP settings directly on the index page (d473e26) is a nice idea. However, i would recommend to display the value of `upload_max_filesize` as well, next to `post_max_size`, which is already there.

Furthermore, the list of developers is incomplete. The latest developer is missing here.

### DIFF
--- a/wcfsetup/install/files/acp/templates/index.tpl
+++ b/wcfsetup/install/files/acp/templates/index.tpl
@@ -159,6 +159,13 @@
 			</dl>
 			
 			<dl>
+				<dt>upload_max_filesize</dt>
+				<dd>
+					{$server[uploadMaxFilesize]}
+				</dd>
+			</dl>
+			
+			<dl>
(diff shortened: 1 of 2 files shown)


wcfsetup/install/files/acp/templates/index.tpl
```diff
@@ -195,6 +195,7 @@
 						<li>Alexander Ebert</li>
 						<li>Joshua R&uuml;sweg</li>
 						<li>Matthias Schmidt</li>
+						<li>Max Mader</li>
```
azerusDE: maybe wrong order?

SoftCreatR: I haven't seen any order.

azerusDE: Okay...
First letter of the surname!?

SoftCreatR: That could be :)

PR #2834, opened 2019-01-23
accept .gz, .tar and .tgz only for package installation/upload

mutec:
Prevent new user's from selecting the zip file containing the first-installation-tool


dtdesign:
I think it would make much more sense to check if the uploaded file is a zip archive _and_ contains the `WCFSetup.tar.gz` file at the top level. This would enable us to tell users exactly what went wrong and that they need to pick a different file instead.

mutec:
For me it's also the difference between 
image and 
image

dtdesign:
Thanks!

PR #2849, opened 2019-02-14
WIP: Font Awesome 5

TimWolla:
This adds a first proof of concept for Font Awesome 5 support via a compatibility layer in StyleCompiler that fixes up any incorrect variables names that are directly used in the CSS. Classes are already handled by the v4-shim of Font Awesome itself.

What this pull request does right now:

- [x] Show FA 5 icons
- [x] Not break backwards compatibility

What it does not yet:

- [ ] Look awesome in all places, e.g. icons in `<a>` have the color of the link instead of black; because of the differing icon sizes icons sometimes look a bit off (e.g. the scroll to the top button).
- [ ] Use clean code for the compatibility layer (I will clean it up if this POC proofs successful).
- [ ] Minify CSS (I commented out that line)
- [ ] Use `getFont.php` (I made the smallest possible change to `_variables.scss` to get it working for this POC).


wcfsetup/install/files/lib/system/style/StyleCompiler.class.php
```diff
@@ -356,7 +393,7 @@ protected function compileStylesheet($filename, array $files, array $variables,
 		}
 		
 		try {
-			$this->compiler->setFormatter('Leafo\ScssPhp\Formatter\Crunched');
+		//	$this->compiler->setFormatter('Leafo\ScssPhp\Formatter\Crunched');
```
TimWolla:
```suggestion
			$this->compiler->setFormatter('Leafo\ScssPhp\Formatter\Crunched');
```

wcfsetup/install/files/style/icon/vendor/_variables.scss
```diff
@@ -0,0 +1,1368 @@
+// Variables
+// --------------------------
+
+$fa-font-path:                "../font/" !default;
```
TimWolla: Check whether it is easy to support `getFont.php` using this variable (it might require a change to `getFont.php`).

wcfsetup/install/files/lib/system/style/StyleCompiler.class.php
```diff
@@ -51,7 +51,44 @@ class StyleCompiler extends SingletonFactory {
 	 */
 	protected function init() {
 		require_once(WCF_DIR.'lib/system/style/scssphp/scss.inc.php');
-		$this->compiler = new Compiler();
+		$this->compiler = new class extends Compiler {
```
TimWolla: XXX This violates every formatting rule we have. Clean up and move into separate class files.

wcfsetup/install/files/style/icon/vendor/_variables.scss
TimWolla: Check whether it's possible to define these variables somewhere else (so that we can just copy over the `.scss` files whenever we want to update).

wcfsetup/install/files/style/icon/icon.scss
```diff
@@ -70,6 +53,17 @@
 	}
 }
 
+@mixin fas() {
+	@include fa-icon();
+	font-family: 'Font Awesome 5 Free';
+	font-weight: 900;
+}
+@mixin far() {
```
TimWolla: Possibly create `mixin`s for the other styles (to more easily support FA Pro).

wcfsetup/install/files/style/icon/vendor/regular.scss
```diff
@@ -0,0 +1,19 @@
+@import 'variables';
+
+@font-face {
+  font-family: 'Font Awesome 5 Free';
```
TimWolla: Consider calling the font just `Font Awesome` for better backwards compatibility with code that expects the old name.

dtdesign:
The Font Awesome 5 Free package is not a proper substitute in its current state:
* The icons do not scale well at 16px × 16px and are somewhat blurry on non-4K displays.
* In comparison to Font Awesome 4, the free icons are much more bolder than before, making them more difficult to comprehend at smaller sizes.
* Icons are sometimes differently aligned both vertically and horizontally, causing "fixes" to misalign these again.

**Font Awesome 4:**
image

**Font Awesome 5 Free:**
image

**Font Awesome 4:**
image

**FontAwesome 5 Free:**
image


The fragmentation of Font Aweosme 5 makes it difficult to create a product that looks good equally good in the Core and in the commerical apps. This does not only apply to the icon styles, but also affects the lack of predictability of available icons: Developers cannot predict if one of those thousands of fancy paid icons will be available, unless it requires one of our commercial apps as its dependencies.

We'll continue to monitor the changes to Font Awesome 5 and look forward to reach an agreement with the creators which would allow us to solve this in a sane fashion.

PR #2851, opened 2019-02-19
Delete firstVisit = true, userID = 0 sessions after processing backgr…

TimWolla:
…ound queue

These sessions most likely stem from a script / cronjob that is set-up
to regularly request the background queue to ensure it does not fill up
in periods of low user activity. They “fill up” the session table, skew
the user online statistics and are not going to be re-used.

### DIFF
--- a/wcfsetup/install/files/lib/system/session/SessionHandler.class.php
+++ b/wcfsetup/install/files/lib/system/session/SessionHandler.class.php
@@ -990,6 +990,25 @@ class SessionHandler extends SingletonFactory {
 	}
 	
 	/**
+	 * Deletes this session if:
+	 * - it is newly created in this request, and
+	 * - it belongs to a guest.
+	 * 
+	 * This method is useful if you have controllers that are likely to be
+	 * accessed by a user agent that is not going to re-use sessions (e.g.
+	 * curl in a cronjob). It immediately remove the session that was created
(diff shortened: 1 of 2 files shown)


dtdesign:
Wouldn't it make much more sense to add a specific parameter that enables the automatic disposal instead of going for a shot in the dark?

TimWolla:
Where exactly would you place this parameter? Something like `WCF::getSession()->disposeIfNew()`?

dtdesign:
`deleteIfNew()`

wcfsetup/install/files/lib/system/session/SessionHandler.class.php
```diff
@@ -989,6 +989,25 @@ public function delete() {
...
+	 * - it belongs to a guest.
+	 * 
+	 * This method is useful if you have controllers that are likely to be
+	 * accessed by a user agent that is not going to re-use sessions (e.g.
+	 * curl in a cronjob). It immediately remove the session that was created
+	 * just for that request and that is not going to be used ever again.
+	 * 
+	 * @since 5.2
+	 */
+	public function deleteIfNew() {
```
dtdesign: This is a very short function and the use of two ifs and an implicit else make it much harder to read. I would like to see a single if instead that evaluates to true and then deletes the session.

PR #2861, opened 2019-03-03
Fix the Sitemap Bug

Griborim:


joshuaruesweg:
Thanks for the pull request, I have fixed the bug with a8dbf8699ff023ff490ff898a9f2586f1d5dac7f, because the method `initObjectList()` is called in `countObjects()` for example, too. The pushed solution covers all possible cases.

PR #2864, opened 2019-03-08
Stop Forum Spam integration

### LINKED ISSUE #2692: Improve the anti spam measures
Evaluate options to combat spam even more, including but not limited to external services such as StopForumSpam. Whatever solution(s) we pick, they must be legal beyond doubt (GDPR/privacy shield) and should not require awkward contracts to be signed by each site owner.

ref https://community.woltlab.com/thread/264735-spam-problems/

dtdesign:
Closes #2692


wcfsetup/install/files/lib/data/blacklist/entry/BlacklistEntry.class.php
```diff
@@ -0,0 +1,132 @@
...
+			if (self::isMatch($row['type'], $row['occurrences'])) {
+				$matches[] = ($row['type'] === 'ipv4' || $row['type'] === 'ipv6') ? 'ip' : $row['type'];
+			}
+		}
+		
+		return $matches;
+	}
+	
+	protected static function getHash($string) {
+		return hex2bin(hash('sha256', $string));
```
TimWolla:
```suggestion
		return hash('sha256', $string, true);
```
The `$raw` parameter is equivalent to `hex2bin`.

com.woltlab.wcf/option.xml
```diff
@@ -1683,5 +1733,11 @@ DESC:wcf.global.sortOrder.descending</selectoptions>
 	</import>
 	<delete>
 		<option name="module_attachment"/>
+		<option name="blacklist_sfs_username_time_period" />
```
TimWolla: Are these part of the development process? Maybe they should be removed if they never were part of a publicly released version.

dtdesign: I've removed them, I've gone through a lot of iterations during the development, especially testing out different approaches. That's also the reason there is no confidence rating, because it doesn't reflect the different reliability of the values and just using arbitrary multipliers doesn't cut it.

wcfsetup/install/files/lib/data/blacklist/entry/BlacklistEntry.class.php
```diff
@@ -24,9 +24,9 @@ class BlacklistEntry extends DatabaseObject {
 	 * @param string $username
 	 * @param string $email
 	 * @param string $ipAddress
-	 * @return boolean
+	 * @return string[]
 	 */
-	public static function shouldReject($username, $email, $ipAddress) {
+	public static function getMatches($username, $email, $ipAddress) {
 		if (BLACKLIST_SFS_USERNAME === 'skip' && BLACKLIST_SFS_EMAIL_ADDRESS === 'skip' && BLACKLIST_SFS_IP_ADDRESS === 'skip') {
 			return false;
```
jens1o: Shouldn't this return `[]` then to match the function signature.

dtdesign: Good catch, thanks!

dtdesign:
@TimWolla I made some minor changes since you marked it as reviewed. Please double check those and if they're fine, merge the PR.

wcfsetup/install/files/lib/form/RegisterForm.class.php
```diff
@@ -164,6 +174,13 @@ public function validate() {
 		if (!$this->isExternalAuthentication && (!WCF::getSession()->getVar('registrationStartTime') || (TIME_NOW - WCF::getSession()->getVar('registrationStartTime')) < self::$minRegistrationTime)) {
 			throw new UserInputException('registrationStartTime', []);
 		}
+		
+		if (BLACKLIST_SFS_ENABLE) {
+			$this->blacklistMatches = BlacklistEntry::getMatches($this->username, $this->email, UserUtil::getIpAddress());
+			if (BLACKLIST_SFS_ACTION === 'block') {
+				throw new PermissionDeniedException();
```
joshuaruesweg: Perhaps a better message should be displayed here if it is a false positive. For example, that the registration has been blocked and you should contact an administrator.

dtdesign: Good point, I've added a phrase that informs the user without giving away the offending value.

PR #2865, opened 2019-03-08
Add possiblity to force deactivation of opcache

jens1o:
This quite simple, suggested change would make it possible to support this Suite being hosted by hosting companies that aren't able to isolate the environments, but still claim to provide the `Zend Opcache`.

I decided against an option in the ACP to prevent misuse by inexperienced users. 

ref https://community.woltlab.com/thread/275439-fehler-bei-der-installation-opcache-invalidate-has-been-disabled/

### DIFF
--- a/wcfsetup/install/files/lib/system/WCF.class.php
+++ b/wcfsetup/install/files/lib/system/WCF.class.php
@@ -973,7 +973,7 @@ class WCF {
 		if (self::$zendOpcacheEnabled === null) {
 			self::$zendOpcacheEnabled = false;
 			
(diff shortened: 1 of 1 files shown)


dtdesign:
Unfortunately, this isn't a solution, because it only hides the symptom of a broken system configuration, but doesn't resolve the issue at all. In these kind of environments, PHP is unable to flush the files at any point, eventually causing outdated `options.inc.php` or `cache.*.php` files. Updates are very highly to break and system upgrades are entirely impossible.

Morik pointed out these issues very well: https://community.woltlab.com/thread/275439-fehler-bei-der-installation-opcache-invalidate-has-been-disabled/?postID=1733533#post1733533

PR #2866, opened 2019-03-11
External lib bump

TimWolla:
Untested so far.

- [x] Verified jQuery changes
- [x] Verified enquire changes
- [x] Verified symfony/css-selector (a.k.a. emogrifier dependency) changes

These libraries have updates, but are most likely incompatible and thus not (yet) included:

- [x] perfect-scrollbar (is: 0.6.2, has: 0.8.1|1.4.0)

These libraries are huge, but only slightly outdated:

- [ ] Codemirror (is: 5.43.0, has: 5.44.0)


dtdesign:
`enquire` is a big mess due to the entirely different code, so we can't really verify it by looking at the code. I suggest that we stick with the old version, deprecate it, and replace all uses with calls to `Ui/Screen` (native API, consistent breakpoints).

TimWolla:
> replace all uses

That would be:

https://github.com/WoltLab/WCF/blob/590cafbe734202ca00e779dfa39f6241caf55b8c/wcfsetup/install/files/js/WCF.Location.js#L134-L137

https://github.com/WoltLab/WCF/blob/590cafbe734202ca00e779dfa39f6241caf55b8c/wcfsetup/install/files/js/WCF.ImageViewer.js#L926-L929

and this unused import:

https://github.com/WoltLab/WCF/blob/590cafbe734202ca00e779dfa39f6241caf55b8c/wcfsetup/install/files/js/WoltLabSuite/Core/Ui/Dialog.js#L17

dtdesign:
These calls to `enquire.register` should be rewritten to `UiScreen.on('screen-md-down', …)`.

TimWolla:
I've upgraded perfect-scrollbar to the latest patch version in the 0.6 series. The 0.7 series is incompatible because of utatti/perfect-scrollbar#602

TimWolla:
Blocked by #2878.

PR #2868, opened 2019-03-13
Performance Optimizations

dtdesign:
These changes improve the overall site performance by around 5-15 percent, depending on the actual content, with performance gains increasing with the code complexity. The changes were tested with both the built in benchmark (100 runs each) and using Xdebug's profiler.

However, these are rather conservative changes that aim to improve the performance of existing code, rather than rewriting things in order to gain more performance. All changes yield a performance delta of above 1%, the size of the compiled templates is reduced by 10% on average.

### DIFF
--- a/wcfsetup/install/files/lib/util/StringUtil.class.php
+++ b/wcfsetup/install/files/lib/util/StringUtil.class.php
@@ -119,20 +119,8 @@ final class StringUtil {
 	 * @return	string
 	 */
 	public static function encodeJS($string) {
-		// unify newlines
 		$string = self::unifyNewlines($string);
-		
-		// escape backslash
-		$string = str_replace("\\", "\\\\", $string);
-		
-		// escape singe quote
-		$string = str_replace("'", "\'", $string);
-		
-		// escape new lines
-		$string = str_replace("\n", '\n', $string);
-		
-		// escape slashes
-		$string = str_replace("/", '\/', $string);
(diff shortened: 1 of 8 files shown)


wcfsetup/install/files/lib/util/StringUtil.class.php
```diff
@@ -119,20 +119,8 @@ public static function encodeHTML($string) {
...
-		
-		// escape singe quote
-		$string = str_replace("'", "\'", $string);
-		
-		// escape new lines
-		$string = str_replace("\n", '\n', $string);
-		
-		// escape slashes
-		$string = str_replace("/", '\/', $string);
+		$string = str_replace(["\\", "'", "\n", "/"], ["\\\\", "\'", '\n', '\/'], $string);
```
TimWolla: Consider: [`addcslashes($string, "\\'\n/")`](http://php.net/manual/en/function.addcslashes.php).

dtdesign: Does it actually make any difference? `str_replace()` is both binary safe and does not care for the null byte, unlike `addcslashes()`.

TimWolla: I just tried: For a string with 10000 bytes in length `str_replace` is faster. For 100000 bytes they are about the same.

You'd expect that the specialized function is the faster one …

So: Disregard that comment.

PR #2869, opened 2019-03-13
Add .bin file extension to attachments

### LINKED ISSUE #2840: Prevent attachments being damaged when transfered in FTP auto mode
Attachments are stored as simple filenames without a file extension, causing some FTP clients in auto mode to assume ASCII for whatever reason. This results in severly damaged files when an attachment is actually a binary file, e. g. an image.

We cannot use the original file extension for security reasons, it might make sense to use the `.bin` extension to give those bad FTP clients something to chew on, while preventing the files being executed on the webserver.

This should be implemented in a backwards compatible manner, e. g. trying to read the file should check for `<hash>.bin` first and then fall back to `<hash>`. The performance penalty of accessing the filesystem twice for a legacy file should be neglible.

* [x] Save new attachments with the `.bin` file extension.
* [x] Rebuilding attachments should rename the files to include `.bin`.
* [x] Lookups will be performed against `<hash>.bin` first, but just `<hash>` is a supported fallback.

TimWolla:
Closes #2840

### DIFF
--- a/wcfsetup/install/files/lib/data/attachment/Attachment.class.php
+++ b/wcfsetup/install/files/lib/data/attachment/Attachment.class.php
@@ -123,7 +123,7 @@ class Attachment extends DatabaseObject implements IRouteController, IThumbnailF
 	 * @inheritDoc
 	 */
 	public function getLocation() {
-		return self::getStorage() . substr($this->fileHash, 0, 2) . '/' . $this->attachmentID . '-' . $this->fileHash;
+		return $this->getLocationHelper(self::getStorage() . substr($this->fileHash, 0, 2) . '/' . $this->attachmentID . '-' . $this->fileHash);
 	}
 	
 	/**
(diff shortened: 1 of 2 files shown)


wcfsetup/install/files/lib/data/attachment/Attachment.class.php
```diff
@@ -140,10 +141,44 @@ public function getTinyThumbnailLocation() {
...
+	}
+	
+	/**
+	 * Returns the appropriate location with or without extension.
+	 * 
+	 * @param	string $location
+	 * @return	string
+	 */
+	protected final function getLocationHelper($location) {
+		// Check location with extension
```
dtdesign: I would change this comment and get rid of the one before the `else if`.
```
// Files are suffixed with `.bin` since 5.2, but they are recognized without the extension for backward compatibility.
```
This explains _why_ rather than _what_ the code is doing.

wcfsetup/install/files/lib/data/attachment/Attachment.class.php
```diff
@@ -140,10 +141,44 @@ public function getTinyThumbnailLocation() {
...
+		if (is_readable($location.'.bin')) {
+			return $location.'.bin';
+		}
+		// Check legacy location
+		else if (is_readable($location)) {
+			return $location;
 		}
 		
-		return self::getStorage() . substr($this->fileHash, 0, 2) . '/' . $this->attachmentID . '-thumbnail-' . $this->fileHash;
+		// Assume that the attachment is not yet uploaded.
```
dtdesign: "… has not been uploaded yet."

wcfsetup/install/files/lib/data/attachment/Attachment.class.php
TimWolla: I moved this into the PhpDoc comment.

PR #2878, opened 2019-03-24
Update to erusev/parsedown 1.7.*

TimWolla:
This is a backport of 4457f82ce54e0072f8cf770683ffab40b1e16af0 to 3.0.

Parsedown is used within WoltLab Suite only for a single importer. Recently
however plugins requiring markdown support (e.g. for syncing Discord messages)
have been submitted. The current parsedown version 1.6.0 is not sufficient
for them, presumably due to the lack of the safe mode.

Upgrade parsedown to avoid the need for them to ship a separate version
(causing issues due to duplicate classes) or an entirely different markdown
parser.

### DIFF
--- a/wcfsetup/install/files/lib/system/api/composer.json
+++ b/wcfsetup/install/files/lib/system/api/composer.json
@@ -5,7 +5,7 @@
     },
     "require": {
         "ezyang/htmlpurifier": "4.9.*",
-        "erusev/parsedown": "1.6.*",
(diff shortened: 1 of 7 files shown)


TimWolla:
cc @SoftCreatR

SoftCreatR:
If this causes troubles with the importer, we can live without an updated versions. We've performed the changes, that are needed to use Parsedown 1.6 anyways.

PR #2882, opened 2019-03-29
Detection of plain links, CSS for embedded content

dtdesign:
Plain links are `<a>` elements that use their `href` value as text content, the detection also determines if the link is "standalone", which means that it is the only content in a dedicated line.

### DIFF
--- a/wcfsetup/install/files/lib/system/html/input/node/HtmlInputNodeProcessor.class.php
+++ b/wcfsetup/install/files/lib/system/html/input/node/HtmlInputNodeProcessor.class.php
@@ -3,6 +3,7 @@ namespace wcf\system\html\input\node;
 use wcf\system\bbcode\BBCodeHandler;
 use wcf\system\event\EventHandler;
 use wcf\system\html\node\AbstractHtmlNodeProcessor;
+use wcf\system\html\node\HtmlNodePlainLink;
 use wcf\system\html\node\IHtmlNode;
 use wcf\util\DOMUtil;
 use wcf\util\StringUtil;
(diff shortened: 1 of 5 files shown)


wcfsetup/install/files/lib/system/html/input/node/HtmlInputNodeProcessor.class.php
```diff
@@ -541,4 +551,77 @@ public function createMetacodeElement(\DOMNode $node, $name, array $attributes)
...
+							// These elements are perfectly fine.
+							break;
+							
+						default:
+							$mayContainOtherContent = true;
+							break 2;
+					}
+				}
+				
+				if (!$mayContainOtherContent || $linebreaks > 1) {
```
TimWolla: Should this read `$linebreaks <= 1`?

wcfsetup/install/files/lib/system/html/node/HtmlNodePlainLink.class.php
```diff
@@ -0,0 +1,174 @@
...
+	 * Replaces the entire link, including any formatting, with the provided bbcode. This is
+	 * available for standalone links only.
+	 * 
+	 * @param BBCode $bbcode
+	 */
+	public function replaceWithBBCode(BBCode $bbcode) {
+		$this->markAsTainted();
+		
+		if ($this->objectID === 0) {
+			throw new \UnexpectedValueException('The objectID must not be null.');
```
TimWolla: Did you mean this:
```suggestion
			throw new \UnexpectedValueException('The objectID must not be zero.');
```

PR #2890, opened 2019-04-11
New group type "Owner"

dtdesign:
The high amount of flexibility of the user group system allows for very sophisticated setups, but the concept of an administrator group has been quite an issue for a lot of users. In particular, it's somewhat difficult to set up groups that act as day-to-day maintenace, without giving away too many permissions. Additionally, the administrator group could be crippled by revoking permissions, despite several safe-guards being in place.

The owner group is a special type (`groupType = 9` / `UserGroup::OWNER` ) that combines a few key concepts:
1. The owner group has several hardcoded permissions that cannot be revoked and are always present, effectively granting the ability to obtain any permission they want.
2. The owner group can only be ever edited by members of the group itself, no other group can ever be granted the permission to edit them.
3. Members of the owner group cannot be banned, a ban has no effect on them.
4. Members of the owner group cannot remove themselves from the owner group, only other members of this group can do this. This is a security measure to prevent users from accidentially removing themselves.
5. The owner group cannot be the target of automatic assignments and cannot be targeted through the bulk processing. In fact, the user bulk processing will skip every user that is a member of the owner group, regardless of the conditions.

For new installations, the built-in administrators group is defined as the owner group. Installations upgraded from earlier versions are prompted to pick a group that is then promoted to be the owner group. There can only be one owner group.

### DIFF
--- a/wcfsetup/install/files/acp/templates/userGroupOption.tpl
+++ b/wcfsetup/install/files/acp/templates/userGroupOption.tpl
@@ -55,7 +55,6 @@
 			{if $userGroupOption->optionName === 'admin.user.accessibleGroups'}
 				elBySelAll('dl[data-group-id]', container, function(dl) {
 					var groupId = parseInt(elData(dl, 'group-id'), 10);
-					console.log("Group", groupId);
 					
 					elBySelAll('input[name="values[' + groupId + '][]"', undefined, function(input) {
 						if (groupId === {@$ownerGroupID}) {
--- a/wcfsetup/install/files/acp/templates/userGroupOption.tpl
+++ b/wcfsetup/install/files/acp/templates/userGroupOption.tpl
@@ -50,6 +50,42 @@
 				elHide(bbcodeHtml);
 			});
 		});
+		
+		{if $ownerGroupID}
+			{if $userGroupOption->optionName === 'admin.user.accessibleGroups'}
+				elBySelAll('dl[data-group-id]', container, function(dl) {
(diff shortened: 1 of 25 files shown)


wcfsetup/install/files/lib/data/user/group/UserGroup.class.php
```diff
@@ -56,6 +56,12 @@ class UserGroup extends DatabaseObject implements ITitledObject {
 	 */
 	const OTHER = 4;
 	
+	/**
+	 * the owner group is always an administrator group
+	 * @var int
+	 */
+	const OWNER = 5;
```
TimWolla: `5` to `7` are in use by `com.woltlab.wcf.moderatedUserGroup`.

dtdesign: Fixed it.

wcfsetup/install/files/acp/templates/userGroupOption.tpl
```diff
@@ -50,6 +50,42 @@
 				elHide(bbcodeHtml);
 			});
 		});
+		
+		{if $ownerGroupID}
+			{if $userGroupOption->optionName === 'admin.user.accessibleGroups'}
+				elBySelAll('dl[data-group-id]', container, function(dl) {
+					var groupId = parseInt(elData(dl, 'group-id'), 10);
+					console.log("Group", groupId);
```
Gravatronics:
```suggestion
```

PR #2908, opened 2019-04-26
Fix now invalid usages of CryptoException

jens1o:
The \random_{bytes,int} functions are now called directly instead through
wcf\util\CryptoUtil,  which means that the exceptions may not be caught
with CryptoException anymore. Instead, they must be caught with the
generic \Throwable interface.
This commit also fixes usages of those functions to prevent malicious code
to overwrite these critical functions in userland code.


TimWolla:
Honestly I wonder if the exception catching should just outright be removed. The previous openssl implementation could fail for a bunch of reasons, but I expect `random_*` to be pretty reliable now.

wcfsetup/install/files/lib/util/StringUtil.class.php
```diff
@@ -47,7 +47,7 @@ public static function getHash($value) {
 	 * @return	string
 	 */
 	public static function getRandomID() {
-		return bin2hex(random_bytes(20));
+		return bin2hex(\random_bytes(20));
```
dtdesign: If you're really making this change in order to improve the security, then you should also adjust `bin2hex()`, otherwise this can be abused too.

jens1o: Throughout the code-base?

dtdesign: Nah, just at this point is sufficient, after all this is about guaranteeing a reliable source.

The performance improvement is neglible. If you're not using opcache, you don't care for performance at all and an extra "namespace lookup" is by far the least of your problems.

dtdesign:
@TimWolla Do we want to keep the try/catch or get rid of it? Is there any relevant downside of leaving it in place?

TimWolla:
> @TimWolla Do we want to keep the try/catch or get rid of it? Is there any relevant downside of leaving it in place?

The one in the SendPasswordWorker makes me unhappy, because users potentially receive a broken email (not that it would happen in practice). But: As the catch-block is modified anyway it can further be modified to be completely removed, without making much of a difference regarding the “dangers” of breakage.

dtdesign:
@jens1o Would you be so kindly to remove the mentioned try/catch block?

PR #2913, opened 2019-04-30
Remove unnecessary match against an Regex

jens1o:
see https://community.woltlab.com/thread/218145-smalltalk-labern-reden-diskutieren/?postID=1740957#post1740957


TimWolla:
Thanks.

Morik:
This should be backported to all trees as it exists sind wcf 2.0^^

TimWolla:
> This should be backported to all trees as it exists sind wcf 2.0^^

It does no harm (apart from a super slight performance reduction), thus I don't see a need for the extra effort to do so.

PR #2914, opened 2019-05-03
Fix password comparison

SoftCreatR:
The password comparison for WCF 2 / WSC does not work, from what i've experienced. Instead of providing the (empty) `$salt`, `$dbHash` needs to be provided to the `getDoubleSaltedHash` method.

To test this, simply prepend `wcf2:` to the password in the DB and try to log in. It will fail, while it works after applying the proposed change.

### DIFF
--- a/wcfsetup/install/files/lib/util/PasswordUtil.class.php
+++ b/wcfsetup/install/files/lib/util/PasswordUtil.class.php
@@ -588,7 +588,7 @@ final class PasswordUtil {
 	 * @return	boolean
 	 */
 	protected static function wcf2($username, $password, $salt, $dbHash) {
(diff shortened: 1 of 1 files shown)


SoftCreatR:
We've switched to Argon2.

TimWolla:
> We've switched to Argon2.

Can you clarify whether this is just something you noticed while doing the switch or whether this is actually causing you issues? I currently don't see that Argon2 can be supported without modifications to core files (`crypt()` does not appear to support Argon2).

SoftCreatR:
We had to modify the core files to make this work, yes. And the issue above was something we've experienced while prefixing the passwords in the db with `wcf2:`. After the suggested modification, it worked as expected.

PR #2920, opened 2019-05-05
Update RegisterForm.class.php

SoftCreatR:
Currently, users are unable to register, if `BLACKLIST_SFS_ACTION` is set to `block`, because of a missing `if` statement (`!empty($this->blacklistMatches)`).


joshuaruesweg:
Thanks!

PR #2927, opened 2019-05-09
added container element to javascript selector

Hanashi:
Please add the container to the javascript selector. Actually it will select the first element on page. If there are more elements with data-type="save", it will select the wrong element.


TimWolla:
Applied manually to 3.0. Thanks.

Note: I suggest that you create a dedicated branch for every pull request instead of using “master”. This makes it easier for you to update your fork.

Issue #2923, opened 2019-05-06
Allow multiple calls of the link to set a new password


BurntimeX:
n/t

PR #2930, opened 2019-05-15
Allow multiple calls of the link to set a new password

joshuaruesweg:
See #2923

### DIFF
--- a/wcfsetup/install/files/lib/form/NewPasswordForm.class.php
+++ b/wcfsetup/install/files/lib/form/NewPasswordForm.class.php
@@ -2,7 +2,6 @@
 namespace wcf\form;
 use wcf\data\user\User;
 use wcf\data\user\UserAction;
-use wcf\data\user\UserEditor;
 use wcf\system\exception\IllegalLinkException;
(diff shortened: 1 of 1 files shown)


wcfsetup/install/files/lib/form/NewPasswordForm.class.php
```diff
@@ -79,10 +78,6 @@ public function readParameters() {
 				$this->throwInvalidLinkException();
 			}
 			
-			(new UserEditor($this->user))->update([
-				'lastLostPasswordRequestTime' => 0,
-				'lostPasswordKey' => null
-			]);
 			WCF::getSession()->register('lostPasswordRequest', $this->user->userID);
 		}
 		else {
```
TimWolla: In the `else` case it should be verified that there still is a `lostPasswordKey`. Consider this situation:

a) Browser 'A' opens the form.
b) Browser 'B' opens the form.
c) 'A' submits.
d) 'B' can still submit, but it must not.

The minimal change would be:

1. Store userID + key in the `lostPasswordRequest` session variable (instead of just userID).
2. Verify both the userID + key in the `else` part.

Issue #2508, opened 2017-12-27
Reaction System


Gravatronics:
> **Reaction System** – Based on the like system (with internal `+1/0/-1` states), but way more flexible and user-configurable.

**Disclaimer**: This issue, at this point in time, contains early ideas about a reaction system. Nothing mentioned in this issue is guaranteed to be implemented at all or in the described way.

## Questions, Challenges, and Possible Solutions

1. **How are the different reactions implemented?**
`wcf1_like_type` table that maps a FontAwesome icon class, color and value (`-1`, `0` or `1`) to the like type; `wcf1_like` gets additional column referencing the like type, but keeps the local `likeValue` value as a copy of the `wcf1_like_type.likeValue` column (→ `likeValue` needs to be synchronized between the two tables)
1. **What happens to existing likes and dislikes?**
 Existing likes and dislikes are converted to reactions “like” and “dislike” that are delivered by the system as defaults. (Should they be deletable?)
1. **What will the reaction interface look like?**
 Keep like button (no dislike button anymore) and open a quickly accessible/reachable ui element to select (or de-select) a reaction. Possible ui element: tooltip-like element with a table-like listing of the possible reactions (the tooltip is right near the button, requiring only minimal cursor movement)
1. **What will the reaction list look like?**
 Replace the default “text” element listing of the latest users liking the object with a inline list with some like types/reactions (how are these “some reactions” determined?) and the number of times they have been used. If more than X reactions have been used, a link to open a dialog appears that shows all reactions and the number of times they have been triggered. Clicking on a reaction icon opens a dialog with all the users using that reaction. (For the dialog with all reactions, may-be find a different solution to avoid a secondary dialog.)
1. **What is the order reactions are displayed in?**
Reaction types could be sortable or their name or last time it was awarded could be used for sorting.
1. **Should all users be allowed to use all reactions?**
 If not, simple ACL could be used to handle reaction-specific permission settings.
1. **Should all reactions be available for all object types?**
 If not, when adding/editing a reaction, the relevant object types should be selectable. (What happens to existing reactions if their association with a specific object type is removed?)
1. **Should users be able to use multiple reactions for one object?**
Multiple reactions by the same user for one object could result in issues with the assumption that there can only one `Like` object per object per user.
1. **Currently, notifications only exist for likes, thus a positive reaction. How should notifications for a general reaction be handled?**
[...]

PR #2931, opened 2019-05-15
Backport reaction related interfaces

joshuaruesweg:
This backport ensures the compatibility for version 3.0 (and above) and version 5.2. With this commit, plugins can fully support the reaction system and the like system, previously this was only possible to a limited extent.
See #2508

### DIFF
--- a/wcfsetup/install/files/lib/system/user/notification/event/TReactionUserNotificationEvent.class.php
+++ b/wcfsetup/install/files/lib/system/user/notification/event/TReactionUserNotificationEvent.class.php
@@ -1,0 +1,56 @@
+<?php
+namespace wcf\system\user\notification\event;
+use wcf\data\user\UserProfile;
+use wcf\system\user\notification\object\IUserNotificationObject;
+
+/**
+ * This trait can be used in likeable user notification events to determine the reactionType counts
+ * for a specific user notification object.
+ * 
+ * HEADS UP: This trait is only included in version 3.0 and 3.1 for compatibility reasons. The method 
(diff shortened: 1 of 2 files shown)


wcfsetup/install/files/lib/system/user/notification/event/TReactionUserNotificationEvent.class.php
```diff
@@ -0,0 +1,55 @@
+<?php
+namespace wcf\system\user\notification\event;
+use wcf\data\user\UserProfile;
+use wcf\system\user\notification\object\IUserNotificationObject;
+
+/**
+ * This trait can be used in likeable user notification events to determine the reactionType counts
+ * for a specific user notification object.
+ * 
+ * HEADS UP: This trait is only included in version 3.0 and 3.1 for compatibility reasons. This
```
TimWolla: I believe this backport is primarily to prevent using an non-existant trait (which will fail hard). As the methods are not usable anyway, consider either:

1. Removing the methods from the trait. They should not be used anywhere anyway.
2. Throwing an Exception whenever the methods are called to make it fail hard in case something erroneously relies on them.

joshuaruesweg: I wouldn't want to remove the methods. However, I think the idea of an exception when the method is used is good. 
Which exception is best in this case? The `BadMethodCallException`? `NotImplementedException` is unfortunately not compatible with version 3.0.

TimWolla: `BadMethodCallException` is fine to me.

PR #2933, opened 2019-05-15
Add on(Final)?Failure callbacks to AbstractBackgroundJob

TimWolla:


jens1o:
I wonder why there is no interface for AbstractBackgroundJob?

TimWolla:
> I wonder why there is no interface for AbstractBackgroundJob?

Why should there?

PR #2936, opened 2019-05-22
Show DatePicker clear button on setDate

derpierre65:
after use e.g. $('#myDatePicker').datepicker('setDate', ...); the clear button doesnt appears, only on initialization


TimWolla:
> $('#myDatePicker').datepicker('setDate', ...);

Using the DatePicker through jQuery is deprecated. This does not mean there isn't a bug here, though.

derpierre65:
is that better now?

TimWolla:
> is that better now?

This is better. Thank you.

I'm not merging this yet, because I want to make a real functional test in addition to just reading the code.

joshuaruesweg:
Thanks!

PR #2939, opened 2019-06-08
Update LanguageAddForm.class.php

YoureCom:
Added aditionalFields


TimWolla:
Patch looks good to me when the request in #2940 is resolved, but as this form does not even use a database object action, but an *editor*:

What's your use-case for extending the `wcf1_language` database table?

YoureCom:
For one of my plugins, I wanted to provide additional information for creating / editing a language. In my particular case, it would have been a boolean to "choose this language as a mandatory language for the descriptions".

PR #2942, opened 2019-06-15
Enterprise mode

Gravatronics:


Morik:
I like the Idea (as far as i understand it).
But wouldn't it make sense to also limit the ability to edit or add update Servers ?

wcfsetup/install/files/lib/system/template/TemplateScriptingCompiler.class.php
```diff
@@ -38,6 +38,69 @@ class TemplateScriptingCompiler {
...
+		'strtotime',
+		'strtoupper',
+		'str_repeat',
+		'str_replace',
+		'str_ireplace',
+		'substr',
+		'trim',
+		'ucfirst',
+		'urlencode',
+		'wcfDebug'
```
joshuaruesweg: `base64_encode` is used in some plugins, maybe add the function to the list, too.

wcfsetup/install/files/lib/system/template/TemplateScriptingCompiler.class.php
```diff
@@ -1181,8 +1244,26 @@ public function compileVariableTag($tag, $replaceQuotes = true) {
...
-						if (strpos($values[$i], '$') !== false) $result .= '{'.$this->compileSimpleVariable($values[$i], $variableType).'}';
-						else $result .= $values[$i];
+						if (strpos($values[$i], '$') !== false) {
+							$result .= '{'.$this->compileSimpleVariable($values[$i], $variableType).'}';
+						}
+						else {
+							$result .= $values[$i];
+						}
+						
+						// `WCF::getDB()` and `WCF::getTPL()` are not supported in enterprise mode
```
joshuaruesweg: It is very easy to bypass this security feature by using the following code: 

```
{assign var='wcf' value=$__wcf}
{$wcf->getDB()|wcfDebug}
```

So you may disable assigning `$__wcf` to other variables.

joshuaruesweg: Also the following is possible: 

```
{$__wcf->__get('DB')|wcfDebug}
```

joshuaruesweg:
```
{$__wcf->__call('getDB', $__wcf->getUser()->getData())|wcfDebug}
```

dtdesign: We can resolve these by using a `EnterpriseModeWCF` class that extends the `WCF` class and is used in the template. It will restrict the access, while mapping properties and methods to the original object:
```php
class EnterpriseModeWCF extends WCF {
  private $wcf;

  public function __construct(WCF $wcf) {
    $this->wcf = $wcf;
  }

  // magic
}
```

wcfsetup/install/files/lib/system/template/TemplateScriptingCompiler.class.php
dtdesign: `max` is also missing, but `min` is supported.

wcfsetup/install/files/lib/system/template/TemplateScriptingCompiler.class.php
```diff
@@ -38,6 +38,71 @@ class TemplateScriptingCompiler {
...
+	/**
+	 * PHP functions and modifiers that can be used in enterprise mode
+	 * @var	string[]
+	 */
+	protected $enterpriseFunctions = [
+		'array_keys',
+		'array_pop',
+		'array_slice',
+		'array_values',
+		'base64_encode',
```
jens1o: `base64_decode`?

wcfsetup/install/files/acp/templates/index.tpl
```diff
@@ -136,7 +136,7 @@
 			<dl>
 				<dt>{lang}wcf.acp.index.system.php.version{/lang}</dt>
 				<dd>
-					{if $__wcf->session->getPermission('admin.configuration.package.canInstallPackage') && $__wcf->session->getPermission('admin.configuration.package.canUpdatePackage')}
+					{if (!ENABLE_ENTERPRISE_MODE || $__wcf->getUser()->hasOwnerAccess()) && $__wcf->session->getPermission('admin.configuration.package.canInstallPackage') && $__wcf->session->getPermission('admin.configuration.package.canUpdatePackage')}
```
joshuaruesweg: Should this information even be available in Enterprise mode? So PHP version, server load, etc.

dtdesign: I agree, none of this is relevant for the "average" user. Please remove the "Server" and "PHP" section using the same condition as in the change above.

PR #2946, opened 2019-06-20
New templatelistener events

YoureCom:
Added some new template listener events

### DIFF
--- a/com.woltlab.wcf/templates/boxSignedInAs.tpl
+++ b/com.woltlab.wcf/templates/boxSignedInAs.tpl
@@ -1,4 +1,6 @@
 <div class="box96">
+	{event name='signedInAs'}
+	
 	{@$__wcf->getUserProfileHandler()->getAvatar()->getImageTag(96)}
 	
 	<div>
--- a/com.woltlab.wcf/templates/boxStaffOnline.tpl
+++ b/com.woltlab.wcf/templates/boxStaffOnline.tpl
@@ -1,6 +1,8 @@
 <ul class="sidebarItemList">
 	{foreach from=$usersOnlineList item=userOnline}
 		<li class="box32">
+			{event name='staffOnlineListItem'}
+			
 			<a href="{link controller='User' object=$userOnline}{/link}">{@$userOnline->getAvatar()->getImageTag(32)}</a>
 			
 			<div class="sidebarItemTitle">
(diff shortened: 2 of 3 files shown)


com.woltlab.wcf/templates/boxSignedInAs.tpl
```diff
@@ -1,4 +1,6 @@
 <div class="box96">
+	{event name='signedInAs'}
```
dtdesign: This event messes with the `box96` layout, which targets the first child.

com.woltlab.wcf/templates/boxStaffOnline.tpl
```diff
@@ -1,6 +1,8 @@
 <ul class="sidebarItemList">
 	{foreach from=$usersOnlineList item=userOnline}
 		<li class="box32">
+			{event name='staffOnlineListItem'}
```
dtdesign: See above, this one conflicts with `box32`.

com.woltlab.wcf/templates/recentActivityListItem.tpl
```diff
@@ -5,6 +5,8 @@
 			
 			<div>
 				<div class="containerHeadline">
+					{event name='containerHeadline'}
```
dtdesign: What should go in here? The top line is already stuffed with details at this point.

com.woltlab.wcf/templates/boxSignedInAs.tpl
YoureCom: All the events I have submitted, I would like to have for my "Clantag" plugin, which sets a clantag before or behind the username. In this template the placement of the event does not matter. It could be at <div class="box96"> > <div> for example.

com.woltlab.wcf/templates/boxStaffOnline.tpl
YoureCom: Here it would be nice to have an event between `{foreach}` and `{/foreach}`.

com.woltlab.wcf/templates/recentActivityListItem.tpl
YoureCom: Here the same, it would be nice to have an event between `{foreach}` and `{/foreach}`.

dtdesign:
@YoureCom I have added the template event at the correct locations and used the same name for all of them. The changes took place in the commit 7dddf95.

PR #2962, opened 2019-06-28
3rd party auth improvements

joshuaruesweg:
* [x] The administrator is unable to review the authentication method, it is simply not visible that a user has registered themselves through a 3rd party site.
* [x] Offer an option to decouple the account at the place of the password (if the user, for example, no longer has access to the account, for whatever obscure reason).
* [x] Remove the password field for administrators for accounts registered through third parties.
* [x] Requesting a password yields an error message that denies the request, because the account is attached to a third party service. While this is correct, it is a poor UX in terms of the workflow. Instead, the message should tell the user that their account is connected with service X and present the user with a button to initiate the process.

### DIFF
--- a/wcfsetup/install/files/lib/acp/form/UserEditForm.class.php
+++ b/wcfsetup/install/files/lib/acp/form/UserEditForm.class.php
@@ -422,7 +422,7 @@ class UserEditForm extends UserAddForm {
 		// reset password
 		$this->password = $this->confirmPassword = '';
 		
-		// reload user when deleting the cover photo or disconneting from 3rd party auth provider
+		// reload user when deleting the cover photo or disconnecting from 3rd party auth provider
 		if ($this->deleteCoverPhoto || $this->disconnect3rdParty) $this->user = new User($this->userID);
 		
 		// show success message
--- a/wcfsetup/install/files/lib/acp/form/UserEditForm.class.php
+++ b/wcfsetup/install/files/lib/acp/form/UserEditForm.class.php
@@ -128,6 +128,12 @@ class UserEditForm extends UserAddForm {
 	public $deleteCoverPhoto = 0;
 	
 	/**
+	 * true to delete the current auth data
+	 * @var	boolean
+	 */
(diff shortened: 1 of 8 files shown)


wcfsetup/install/files/lib/acp/form/UserEditForm.class.php
```diff
@@ -410,8 +422,8 @@ public function save() {
 		// reset password
 		$this->password = $this->confirmPassword = '';
 		
-		// reload user when deleting the cover photo
-		if ($this->deleteCoverPhoto) $this->user = new User($this->userID);
+		// reload user when deleting the cover photo or disconneting from 3rd party auth provider
```
TimWolla:
```suggestion
		// reload user when deleting the cover photo or disconnecting from 3rd party auth provider
```

wcfsetup/install/lang/de.xml
```diff
@@ -4845,6 +4845,7 @@ Die E-Mail-Adresse des neuen Benutzers lautet: {@$user->email}
 		<item name="wcf.user.3rdparty.google.connect.error.inuse"><![CDATA[{if LANGUAGE_USE_INFORMAL_VARIANT}Dein{else}Ihr{/if} Google-Konto ist bereits mit einem anderen Benutzerkonto verknüpft.]]></item>
 		<item name="wcf.user.3rdparty.google.disconnect"><![CDATA[Verknüpfung mit Google trennen]]></item>
 		<item name="wcf.user.3rdparty.google.disconnect.success"><![CDATA[Die Verknüpfung mit {if LANGUAGE_USE_INFORMAL_VARIANT}deinem{else}Ihrem{/if} Google-Konto wurde erfolgreich getrennt.]]></item>
+		<item name="wcf.user.3rdparty.connect.info"><![CDATA[Der Benutzer ist mit {lang}wcf.user.3rdparty.{$user->getAuthProvider()}{/lang} verknüpft.]]></item>
```
TimWolla: `{lang}` in language items is a no-no.

PR #2986, opened 2019-07-17
New concept for reactions

### LINKED ISSUE #2975: Reactions should emphasis on the content rather than the author
Reactions are very versatile and enable a whole lot of different user cases, all covered by a single system. However, in it's current state the focus lies too much on the idea of "rating" content rather than expressing thoughts and emotions through reactions.

* [x] There are only plain reactions, there is no positive/neutral/negative rating attached to them.
* [x] `cumulativeLikes` is kept in the code and the database (ALTER TABLE in large communities is a real pain), but will track the total number of reactions.
* [x] Users will only have a counter that trackes the number of received reactions, again using the legacy `cumulativeLikes` column.
* [x] Remove the ability to react to the own content, the constant `LIKE_ALLOW_FOR_OWN_CONTENT` is always `0`.
* [x] Remove `LIKE_ENABLE_DISLIKE` and set it to always `0`.

dtdesign:
Closes #2975 

Major changes include:
* Reactions no longer have any rating attached to them.
* The `cumulativeLikes` column is retained, but will now track the number of reactions.
* The user's reputation is replaced by the number of received reactions.
* Users can no longer react to their own content.


wcfsetup/setup/db/install.sql
```diff
@@ -2430,8 +2422,8 @@ INSERT INTO wcf1_contact_option (optionID, optionTitle, optionDescription, optio
 INSERT INTO wcf1_contact_recipient (recipientID, name, email, isAdministrator, originIsSystem) VALUES (1, 'wcf.contact.recipient.name1', '', 1, 1);
 
 -- default reaction type
-INSERT INTO wcf1_reaction_type (title, type, showOrder, iconFile) VALUES ('wcf.reactionType.title1', 1, 1, 'like.svg');
-INSERT INTO wcf1_reaction_type (title, type, showOrder, iconFile) VALUES ('wcf.reactionType.title2', 1, 2, 'haha.svg');
-INSERT INTO wcf1_reaction_type (title, type, showOrder, iconFile) VALUES ('wcf.reactionType.title3', -1, 3, 'sad.svg');
-INSERT INTO wcf1_reaction_type (title, type, showOrder, iconFile) VALUES ('wcf.reactionType.title4', 0, 4, 'confused.svg');
-INSERT INTO wcf1_reaction_type (title, type, showOrder, iconFile) VALUES ('wcf.reactionType.title5', 1, 5, 'thanks.svg');
+INSERT INTO wcf1_reaction_type (reactionTypeID, title, showOrder, iconFile) VALUES (1, 'wcf.reactionType.title1', 1, 'like.svg');
+INSERT INTO wcf1_reaction_type (reactionTypeID, title, showOrder, iconFile) VALUES (2, 'wcf.reactionType.title5', 2, 'thanks.svg');
```
TimWolla: Unhappy with the mismatch of language item and ID: If you want to update the order maybe just move around the `showOrder` value?

dtdesign: I missed those ones, thanks!

PR #3002, opened 2019-07-28
It is now possible to specify the sort order of comments.

darkwood-studios:
See suggestion at https://community.woltlab.com/thread/277652-comment-add-insertmessage-auch-mit-before-erm%C3%B6glichen/?postID=1753004#post1753004

### DIFF
--- a/wcfsetup/install/files/js/WoltLabSuite/Core/Ui/Comment/Add.js
+++ b/wcfsetup/install/files/js/WoltLabSuite/Core/Ui/Comment/Add.js
@@ -304,9 +304,14 @@ function(
 		 * @protected
 		 */
 		_insertMessage: function(data) {
+			var insertMethod = 'after';
+			if (this._container.parentElement.dataset.sortOrder !== undefined && this._container.parentElement.dataset.sortOrder === 'ASC') {
+				insertMethod = 'before';
+			}
+			
 			// insert HTML
 			//noinspection JSCheckFunctionSignatures
-			DomUtil.insertHtml(data.returnValues.template, this._container, 'after');
+			DomUtil.insertHtml(data.returnValues.template, this._container, insertMethod);
 			
 			UiNotification.show(Language.get('wcf.global.success.add'));
 			


dtdesign:
The comment system displays the comments from latest to newest, placing the comment box at the bottom would break the user expectation and goes against any interface specs. Also, this commit does not change any of the internals, including but not limited to denormalized data that is stored for either comments or responses. Furthermore, your commit entirely skips responses that use the same UI logic to represent the order of items.

darkwood-studios:
@dtdesign  Thank you for your detailed answer.

How else should the developer implement the required case if the framework does not allow the necessary flexibility?

The fact that the reply field in the template can be placed either above or below by the developer indicates that it's left to the developer how he wants to sort the comments in different cases. In addition, it is already possible to change the sqlSortOrder for the CommentList.

So in conclusion, if there is any special case, that the user benefits from a reversed sort order, why shouldn't the framework make this case possible?

The alternative is that the developer has to duplicate tons of code from the comment system to clone the behavior of the comments (in reversed order)

dtdesign:
> The fact that the reply field in the template can be placed either above or below by the developer indicates that it's left to the developer how he wants to sort the comments in different cases.

Unfortunately, no, it doesn't. I've explained it in my previous comment that this extra template was a side effect of the implementation of WYSIWYG support in WSC 3.1. Changing the position of the template in relation to the list of comments is not a supported feature.

> So in conclusion, if there is any special case, that the user benefits from a reversed sort order, why shouldn't the framework make this case possible?

Again: Consistency. It is important to use a consistent UI for the same type of elements, otherwise you break with the expectations, eventually decreasing the usability rather than improving it. From a user perspective, two comment lists with opposing sort orders will look overall the same, but act in a completely different way, which is outright confusing and bad UX.

darkwood-studios:
Okay. Then the comment system is probably not the right choice for my project.

So I have to create my own table for entries and answers, implement LikeHandler, create action and data classes and write the UI myself to replicate everything the comment system offers. 

Then I have exactly the described UX with much more effort for which the customer has to pay. 

Too bad, because one line of code within the framework could have saved so much time.

Do you have any other suggestion to solve the request without tons of effort?

dtdesign:
First things first, this isn't just "one line", in fact your commit doesn't even make a half of the changes of what needs to be done to fully support a reversed listing of comments. Right now, this would require a change to the comment handler itself in order to transparently provide this reversal throughout the entire process of loading, storing and creating of comments and responses.

I'm a bit surprised that you have been putting that much emphasis on the implementation details, but have not yet responded to my actual concern: Lack of consistency.

I will exaggerate things beyond anything reasonable for the sake of making a point: You're suggesting the ability to change the edit icon into a red cross (normally used to delete things), just because you feel that it's better this way for your particular feature. I'm more than confident that you agree with me that this would be a less ideal decision, because it would completely screw up the typical workflow of your users.

Let me get straight to the point: If you want the ability to reverse the order of comment lists, then please provide me with arguments that justify the need to do this. Explain to me, why it makes sense to make it behave differently and explain, why the default behavior doesn't work for you. This isn't about the technical aspects of the implementation!

darkwood-studios:
> First things first, this isn't just "one line", in fact your commit doesn't even make a half of the changes of what needs to be done to fully support a reversed listing of comments.

It wasn't my intention to implement the reverse comment lists as default for all comments, but to give developers the possibility to do so depending on the use case. The developer controls the sorting in his own CommentHandler. He already has this possibility now. Just the UI prohibits it. (to change this it needs just this oneliner)

> I'm a bit surprised that you have been putting that much emphasis on the implementation details, but have not yet responded to my actual concern: Lack of consistency.

My special case will not end up in inconsistency or in bad UX. It only saves much time and code for the developer, because he can use the comment system with all its features for more use cases.

Maybe you'll understand better with my example case:

My shoutbox plugin uses its own entry table for the shoutbox entries. Because the shoutbox hasn't much features, that was fine for now. But our users are begging for more features:
- direct response to entries
- get notification on responses
- like entries
- use the editor for entries
- report spam entries
- edit entries

The comment system offers all of them and the UI is quite similar to my current shoutbox entry UI. 

So my idea is, that the admin can create multiple shoutboxes, and the users can comment them directly through the comment system of the framework. 

This results in less effort for the developer, less code within the system and much more features for the user.

But the shoutbox entries have to be sorted ASC and DESC - thats the only point why the comment system can not be used for such a case.

So there will not be any bad UX because the UX is similar to the current. (within the shoutbox)

But maybe you are right. Maybe my case is to special for a framework default.

If you will not accept the change, my fallback solution is to write my own WCF.Comment.Handler and Ui/Comment/Add and inherit from the framework classes to take the advantages from above. But that will result in some duplicate code I wanted to save.

dtdesign:
> […] He already has this possibility now. Just the UI prohibits it. (to change this it needs just this oneliner)

This not true, the caching strategy for comment response ids (that's the denormalization I mentioned earlier) uses the same sort order to determine the affected ids. I guess it is easy to miss that part, because the cache can hold up to 5 ids per comment. However, the comments go from "new" to "old", while responses are ordered from "old" to "new".

That's what I meant with that there is a whole lot more attached to it than it appears at a first glance. I can see that you might just be looking at the comment bit, but being the framework developer, we have to look at the bigger picture - if we want to make a change here, it needs to be consistent throughout the entire feature.

> My shoutbox plugin uses its own entry table for the shoutbox entries. Because the shoutbox hasn't much features, that was fine for now. But our users are begging for more features […]

Thank you, this is why I asked about what you're going to do. Up until know I was given the impression that you would embed the comments somewhere and simply flip the order, but you're heading an entirely different direction.

However, I'm not exactly sure if the comment system is the right tool for the job. It is designed with restricted limits on the number of items, supports infinite amounts of responses and ships with a full-fledged WYSIWYG editor. Additionally, from what I can tell, shoutboxes are pretty much just short lived chat messages that quickly disappear in the void, making notifications seem a bit odd to me. I'm not sure if responses fit into the linear structure of the message flow either, but I guess that is something that you're more into.

My point is that the comment system does cover the features that your users demand for, but it goes beyond that. Eventually you could reach a point where you have to start to fight against in order to get unwanted features out of the way.

The bottom line is, I'm not against making changes to core components at all, I'm just looking to see if this is really the right approach and if there is any other route that would better suit your requirements.

darkwood-studios:
Thanks for your feedback.

> My point is that the comment system does cover the features that your users demand for, but it goes beyond that. 

I'm going to rethink the concept. Maybe i'll get some other idea to speed up the work. 

As a last word to the topic I want to leave some feedback or idea regarding to the development process.

I have written much more than 50 plugins and applications for the WoltLab Suite so far, and I always have to write plenty of the same code (entry classes and tables, response classes and tables, LikeHandler, CommentHandler, AttachmentObjectTypes, QuoteHandler etc...) again and again. Thats the reason why I'm looking for a more efficient way to cover most of the use cases with default resources.

I'm really happy about the new form builder. I think there is so much more potential with new dev tools to make it faster for developers to implement basic behavior (for example: create basic classes and object types for Comment, Like, Attachment, Quote)

dtdesign:
> I'm going to rethink the concept. Maybe i'll get some other idea to speed up the work.

Very well. Please don't hesitate to ask me if you would like to discuss your idea, as I said earlier, I'm more than willing to work something out with you to get this done.

> […] and I always have to write plenty of the same code (entry classes and tables, response classes and tables, LikeHandler, CommentHandler, AttachmentObjectTypes, QuoteHandler etc...) again and again.

You're very welcome to bring these kind of issues up, either on the forums or here on GitHub.

PR #3008, opened 2019-07-31
Update instructions from WSC 3.1.9 pl 2 to WSC 5.2.0 Alpha/Beta

mutec:
I just wrote the update instructions in order to test the upgrade in my projects. Feel free to be inspired or merge some time. ;)
You might want to read the comments if you don't want to debug the same issues than I did. :)
It should get every difference made until the moment of me typing here.

PS: Even if you're not going to use them, you might want to explain why do'd so things different - maybe I'll learn something.


com.woltlab.wcf/package.xml
```diff
@@ -62,4 +62,55 @@
 		
 		<instruction type="objectType" />
 	</instructions>
+	
+	<instructions type="update" fromversion="3.1.10 pl 2">
+		<instruction type="sql">update_5.2_pre_sql_api.sql</instruction>
+		
+		<!-- contains 3 PIP-scripts, \wcf\data\option\OptionEditor, \wcf\system\cache\builder\RoutingCacheBuilder, \wcf\system\database\table\ -->
```
dtdesign: What is the rationale behind adding `OptionEditor` and `RoutingCacheBuilder` into it? Neither of those include any changes that are required for the upgrade itself?

PR #3017, opened 2019-08-08
New events in UserNotificationHandler

### LINKED ISSUE #3016: New events in UserNotificationHandler
At the moment there's the event-point `fireEvent` at `wcf\system\user\notification\UserNotificationHandler::fireEvent()`.
The 'problem' with this event is, that `$parameters['recipients']` might include users who aren't going to be recipients in the end, because they don't want that kind of notification.
It might be better to have an additional event (since moving it would break compatibility) within the stackable-update-block:
https://github.com/WoltLab/WCF/blob/ecf02509ca3658c24fdc3efe85ad2d8b45d0256c/wcfsetup/install/files/lib/system/user/notification/UserNotificationHandler.class.php#L167-L184
and one directly before the notifications are created:
https://github.com/WoltLab/WCF/blob/ecf02509ca3658c24fdc3efe85ad2d8b45d0256c/wcfsetup/install/files/lib/system/user/notification/UserNotificationHandler.class.php#L226-L251
always containing the same parameters than the existing event, but updated/additional values.

For the moment I'll change my plugin to use `UserNotificationAction::create()` instead - if there are all necessary parameters available.

PS: It could be helpful too, if there were events in `markAsConfirmed` and `markAsConfirmedByIDs`.

mutec:
This might be needed for integrations of other notification-systems/providers or marking notifications as read in an application for mobile devices without re-fetching the whole notification-list.
Currently it's not possible to handle confirmed or removed notifications correctly.
Performance should not be reduced (a lot) since those positions are not called for every request on the site.

Closes #3016

### DIFF
--- a/wcfsetup/install/files/lib/data/user/notification/UserNotificationEditor.class.php
+++ b/wcfsetup/install/files/lib/data/user/notification/UserNotificationEditor.class.php
@@ -1,7 +1,7 @@
 <?php
 namespace wcf\data\user\notification;
 use wcf\data\DatabaseObjectEditor;
-use wcf\system\WCF;
+use wcf\system\user\notification\UserNotificationHandler;
 
 /**
  * Provides functions to edit user notifications.
@@ -23,17 +23,12 @@ class UserNotificationEditor extends DatabaseObjectEditor {
 	
 	/**
 	 * Marks this notification as confirmed.
+	 *
+	 * @deprecated	5.2 This method currently only is a wrapper for markAsConfirmedByIDs in UserNotificationHandler.
+	 * 		Please use \wcf\system\user\notification\UserNotificationHandler::markAsConfirmedByIDs()
+	 * 		from now on, as this method may be removed in the future.
 	 */
 	public function markAsConfirmed() {
-		$this->update([
-			'confirmTime' => TIME_NOW,
-			'mailNotified' => 1
-		]);
-		
-		// delete notification_to_user assignment (mimic legacy notification system)
-		$sql = "DELETE FROM	wcf".WCF_N."_user_notification_to_user
-			WHERE		notificationID = ?";
-		$statement = WCF::getDB()->prepareStatement($sql);
-		$statement->execute([$this->notificationID]);
+		UserNotificationHandler::getInstance()->markAsConfirmedByIDs([$this->notificationID]);
 	}
 }
--- a/wcfsetup/install/files/lib/system/user/notification/UserNotificationHandler.class.php
+++ b/wcfsetup/install/files/lib/system/user/notification/UserNotificationHandler.class.php
@@ -181,6 +181,10 @@ class UserNotificationHandler extends SingletonFactory {
 					]);
 				}
 				WCF::getDB()->commitTransaction();
+				
+				$parameters['notificationIDs'] = $notificationIDs;
+				EventHandler::getInstance()->fireAction($this, 'updateStackableNotifications', $parameters);
+				unset($parameters['notificationIDs']);
 			}
 		}
 		
@@ -239,6 +243,10 @@ class UserNotificationHandler extends SingletonFactory {
 				],
 				'recipients' => $recipients
 			];
+			
+			$parameters['notifications'] = $notifications;
+			$parameters['recipientIDs'] = $recipientIDs;
+			EventHandler::getInstance()->fireAction($this, 'createNotification', $parameters);
 			
 			if ($event->isStackable()) {
 				$data['notifications'] = $notifications;
@@ -772,6 +780,14 @@ class UserNotificationHandler extends SingletonFactory {
 			$statement->execute($conditions->getParameters());
 			$userIDs = $statement->fetchAll(\PDO::FETCH_COLUMN);
 			
+			$parameters = [
+				'objectType' => $objectType,
+				'eventIDs' => $eventIDs,
+				'objectIDs' => $objectIDs,
+				'userIDs' => $userIDs
+			];
+			EventHandler::getInstance()->fireAction($this, 'removeNotifications', $parameters);
+			
 			// reset number of notifications
 			if (!empty($userIDs)) {
 				UserStorageHandler::getInstance()->reset(array_unique($userIDs), 'userNotificationCount');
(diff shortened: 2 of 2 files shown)


wcfsetup/install/files/lib/system/user/notification/UserNotificationHandler.class.php
```diff
@@ -171,7 +171,7 @@ public function fireEvent($eventName, $objectType, IUserNotificationObject $noti
 						guestTimesTriggered = guestTimesTriggered + ?
 					WHERE	notificationID = ?";
 				$statement = WCF::getDB()->prepareStatement($sql);
-				
+
```
joshuaruesweg: Missing indentation

wcfsetup/install/files/lib/system/user/notification/UserNotificationHandler.class.php
```diff
@@ -771,7 +779,15 @@ public function removeNotifications($objectType, array $objectIDs) {
 			$statement = WCF::getDB()->prepareStatement($sql);
 			$statement->execute($conditions->getParameters());
 			$userIDs = $statement->fetchAll(\PDO::FETCH_COLUMN);
-			
+
+			$params = [
```
joshuaruesweg: Should be `$parameters` for consistency reasons

wcfsetup/install/files/lib/system/user/notification/UserNotificationHandler.class.php
```diff
@@ -816,6 +832,15 @@ public function markAsConfirmed($eventName, $objectType, array $recipientIDs, ar
 		$parameters = $conditions->getParameters();
 		array_unshift($parameters, TIME_NOW);
 		$statement->execute($parameters);
+
+		$params = [
```
joshuaruesweg: Should be `$parameters` for consistency reasons

wcfsetup/install/files/lib/system/user/notification/UserNotificationHandler.class.php
```diff
@@ -866,6 +891,11 @@ public function markAsConfirmedByIDs(array $notificationIDs) {
...
```
joshuaruesweg: Should be `$parameters` for consistency reasons

wcfsetup/install/files/lib/data/user/notification/UserNotificationEditor.class.php
```diff
@@ -29,6 +30,8 @@ public function markAsConfirmed() {
 			'confirmTime' => TIME_NOW,
 			'mailNotified' => 1
 		]);
+
+		EventHandler::getInstance()->fireAction($this, 'markAsConfirmed');
```
joshuaruesweg: Is this method even used? Maybe this method should be deprecated, if not and no event should be added.

wcfsetup/install/files/lib/system/user/notification/UserNotificationHandler.class.php
```diff
@@ -181,6 +181,10 @@ public function fireEvent($eventName, $objectType, IUserNotificationObject $noti
 					]);
 				}
 				WCF::getDB()->commitTransaction();
+
+				$parameters['notificationIDs'] = $notificationIDs;
+				EventHandler::getInstance()->fireAction($this, 'updateStackableNotifications', $parameters);
```
joshuaruesweg: You have the `fireEvent` event right above. Why is there another event required? You have all informations which are required to read the `$notificationID` by yourself.

wcfsetup/install/files/lib/system/user/notification/UserNotificationHandler.class.php
```diff
@@ -239,6 +243,10 @@ public function fireEvent($eventName, $objectType, IUserNotificationObject $noti
 				],
 				'recipients' => $recipients
 			];
+
+			$parameters['notifications'] = $notifications;
+			$parameters['recipientIDs'] = $recipientIDs;
+			EventHandler::getInstance()->fireAction($this, 'createNotification', $parameters);
```
joshuaruesweg: You have the fireEvent event right above. Why is there another event required? In addition, the creation can be intercepted via the DBOAction.

wcfsetup/install/files/lib/system/user/notification/UserNotificationHandler.class.php
mutec: Yes, but that's causing duplications (or more) of the queries and php-logic.

wcfsetup/install/files/lib/data/user/notification/UserNotificationEditor.class.php
mutec: Currently it's not used by the WSC itself - neither by my own packages nor others I know.
I'd appreciate a deprecation/removal using the new deprecation exceptions.

wcfsetup/install/files/lib/system/user/notification/UserNotificationHandler.class.php
mutec: I checked this out, but in order to get the notification's message, title and link properly, I'll have to process the notifications (performance yaay).
But yes I care more about the update stackable and confirmed/removal events since there is no way at the moment.

wcfsetup/install/files/lib/data/user/notification/UserNotificationEditor.class.php
joshuaruesweg: The exception is postboned to WSC 3.2. But, there should not be the event. A plugin should use the "correct" API.

mutec: Did you mean WSC 5.3? Because this PR is targeting WSC 5.2. ;)

wcfsetup/install/files/lib/data/user/notification/UserNotificationEditor.class.php
```diff
@@ -23,6 +23,8 @@ class UserNotificationEditor extends DatabaseObjectEditor {
 	
 	/**
 	 * Marks this notification as confirmed.
+	 *
+	 * @deprecated
```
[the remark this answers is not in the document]
mutec: I already expected that. I'm going to separate that when everything else passes.

wcfsetup/install/files/lib/system/user/notification/UserNotificationHandler.class.php
```diff
@@ -239,6 +243,10 @@ public function fireEvent($eventName, $objectType, IUserNotificationObject $noti
 				],
 				'recipients' => $recipients
 			];
+
```
dtdesign: Please update the indentation, empty lines use the same amount of tabs, see https://github.com/WoltLab/WCF/blob/master/CONTRIBUTING.md#formatting.

This applies to all other changes in your PR.

mutec:
Are you going to merge this if I remove those two unresolved events?
Removing the `updateStackableNotifications` wouldn't make any sense since it would result in code duplication.

dtdesign:
I've went with slightly different set of events that do accomplish the same thing but provide more consistency in regards to the code flow. Especially the separation between stackable and non-stackable events didn't made much sense, since you can easily detect their difference through the `$event->isStackable()` flag.

Issue #2847, opened 2019-02-12
More resilent upgrade process


dtdesign:
The upgrade process can fail far too easily and not only give little information regarding to where it fails, it also stops at simple things like existing tables. There are a few things that can and should be improved to decrease the likelihood of failures during an upgrade:

* [x] **Recursively validate if all relevant files are writable.** (#3042)
Directories with known large amounts of unrelated files, e. g. uploaded images, should be excluded. This applies to the Core only, which is by far the most sensitive part of the upgrade process.
* [x] **Create a dedicated log file, e. g. `log/YYYY-MM-DD-update-TIMESTAMP.txt`.** (#3043)
`TIMESTAMP` equals the start of the update, any step that is _to be_ performed is dumped to this log. This should make it a bit easier to backtrack the source of an issue, or at the very least give an idea of what has failed. It should include selected meta data, such as the package that is currently being processed.
* [x] **PHP based SQL PIP that manages database upgrades.** (#3065)
The SQL PIP lacks a lot of flexibility, fails too easily and in particular cannot deal with unexpected situations, examples include already existing tables or keys. Ideally, it should be possible to specify which state a database table should have and the API figures out what _actually_ needs to be done.
Bonus points for being able to allow customers with large sites to make certain database modifications, e. g. adding columns or indices, ahead of time before running the upgrade.

PR #3042, opened 2019-08-18
Check if main directories are writable during Core updates

Gravatronics:
See #2847

### DIFF
--- a/wcfsetup/install/files/lib/system/WCF.class.php
+++ b/wcfsetup/install/files/lib/system/WCF.class.php
@@ -1200,7 +1200,7 @@ class WCF {
 		
 		if (!empty($nonWritablePaths)) {
 			$maxPaths = 10;
-			throw new \RuntimeException('The following paths are not writable: ' . implode(',' ,array_slice($nonWritablePaths, 0, $maxPaths)) . (count($nonWritablePaths) > $maxPaths ? ',' . StringUtil::HELLIP : ''));
+			throw new \RuntimeException('The following paths are not writable: ' . implode(',', array_slice($nonWritablePaths, 0, $maxPaths)) . (count($nonWritablePaths) > $maxPaths ? ',' . StringUtil::HELLIP : ''));
 		}
 	}
 }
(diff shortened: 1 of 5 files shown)


wcfsetup/install/files/lib/system/WCF.class.php
```diff
@@ -1129,4 +1130,77 @@ protected function initCronjobs() {
...
+			
+			if (!is_writable($path)) {
+				$nonWritablePaths[] = $path;
+				continue;
+			}
+			
+			DirectoryUtil::getInstance($path)->executeCallback(function($file, \SplFileInfo $fileInfo) use ($path, &$nonWritablePaths) {
+				if ($fileInfo instanceof \DirectoryIterator) {
+					if (!is_writable($fileInfo->getPath())) {
+						$nonWritablePaths[] = $fileInfo->getPath();
```
dtdesign: The paths should be relative to the document root.

wcfsetup/install/files/lib/system/WCF.class.php
```diff
@@ -1129,4 +1130,77 @@ protected function initCronjobs() {
...
+					}
+				}
+				else if (!is_writable($fileInfo->getRealPath())) {
+					$nonWritablePaths[] = $fileInfo->getRealPath();
```
dtdesign: See line 1192 above.

mutec:
Please extend this to 3rdparty packages too. At least for apps.

wcfsetup/install/files/lib/system/package/PackageInstallationDispatcher.class.php
```diff
@@ -396,6 +396,10 @@ public function getArchive() {
 	protected function installPackage(array $nodeData) {
 		$installationStep = new PackageInstallationStep();
 		
+		if ($this->getAction() === 'update' && $this->getPackageName() === 'com.woltlab.wcf') {
```
dtdesign: This doesn't work, because its check accomplishes nothing. `installPackage()` is the first node in the installation process, but the last one for an update. We're looking to validate the file permissions *before* the Core update takes place.

PR #3043, opened 2019-08-18
Log individual package installation steps

Gravatronics:
See #2847

### DIFF
--- a/wcfsetup/install/files/lib/system/package/PackageInstallationDispatcher.class.php
+++ b/wcfsetup/install/files/lib/system/package/PackageInstallationDispatcher.class.php
@@ -38,6 +38,7 @@ use wcf\system\user\storage\UserStorageHandler;
 use wcf\system\WCF;
 use wcf\util\FileUtil;
 use wcf\util\HeaderUtil;
+use wcf\util\JSON;
 use wcf\util\StringUtil;
 
 /**
@@ -132,6 +133,7 @@ class PackageInstallationDispatcher {
 		$step = null;
 		foreach ($nodes as $data) {
 			$nodeData = unserialize($data['nodeData']);
+			$this->logInstallationStep($data);
 			
 			switch ($data['nodeType']) {
 				case 'package':
@@ -152,6 +154,7 @@ class PackageInstallationDispatcher {
 			}
 			
 			if ($step->splitNode()) {
+				$this->logInstallationStep($data, 'split node');
 				$this->nodeBuilder->cloneNode($node, $data['sequenceNo']);
 				break;
 			}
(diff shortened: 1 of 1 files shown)


wcfsetup/install/files/lib/system/package/PackageInstallationDispatcher.class.php
```diff
@@ -350,11 +355,46 @@ public function install($node) {
...
+		$time = TIME_NOW;
+		$logEntry = "[" . TIME_NOW . "]\n";
+		if (!empty($node)) {
+			$logEntry .= 'sequenceNo: ' . $node['sequenceNo'] . "\n";
+			$logEntry .= 'nodeType: ' . $node['nodeType'] . "\n";
+			$logEntry .= "nodeData:\n";
+			
+			$nodeData = unserialize($node['nodeData']);
+			foreach ($nodeData as $index => $value) {
+				$logEntry .= "\t" . $index . ': ' . (!is_object($value) && !is_array($value) ? $value : JSON::encode($value)) . "\n";
```
TimWolla: Are you positive that JSON encoding objects does not lose information (such as private fields)?

wcfsetup/install/files/lib/system/package/PackageInstallationDispatcher.class.php
```diff
@@ -350,11 +355,46 @@ public function install($node) {
...
 	
+	/**
+	 * Logs an installation step.
+	 * 
+	 * @param	array		$node	data of the executed node
+	 * @param	string		$log	optional additional log text
+	 */
+	protected function logInstallationStep(array $node = [], $log = '') {
+		$time = TIME_NOW;
+		$logEntry = "[" . TIME_NOW . "]\n";
```
TimWolla: Consider using `time()` here, the update process might take longer than a second.

wcfsetup/install/files/lib/system/package/PackageInstallationDispatcher.class.php
```diff
@@ -350,11 +355,46 @@ public function install($node) {
...
 	}
 	
+	/**
+	 * Logs an installation step.
+	 * 
+	 * @param	array		$node	data of the executed node
+	 * @param	string		$log	optional additional log text
+	 */
+	protected function logInstallationStep(array $node = [], $log = '') {
+		$time = TIME_NOW;
```
TimWolla: never read

wcfsetup/install/files/lib/system/package/PackageInstallationDispatcher.class.php
Gravatronics: Objects are actually irrelevant here because only arrays are present here as a "non-printable" data type. I just added the object check for completeness.

wcfsetup/install/files/lib/system/package/PackageInstallationDispatcher.class.php
Gravatronics: I am unsure if that information is really useful in practice. The goal of this change in general is just to have a log of the last step when a problem occurs during an update so that the _order_ of the entries is of relevance.

PR #3060, opened 2019-08-27
Reaction single item fix

Cyperghost:
When the summary not in the 'containerSelector' but markt as single item, the reaction will not append after save.


joshuaruesweg:
Fixed with: 1603924e556d81c2f837a3755e600b58d26d3b8b 

Thanks for reporting and troubleshooting!

PR #3064, opened 2019-09-01
Add method to send email synchronously

ChristopherWalz:


TimWolla:
LGTM from a purely technical side. What's your use case for this?

ChristopherWalz:
One of my sites has a low rate of daily visitors (<30) and multiple cronjobs throughout the day and night send emails to users. Those should not be delayed for 1-2 hours until a new visitor triggers the background queue

TimWolla:
Depending on the type of emails I can easily see a synchronous delivery become an issue once the site gets more activity / more users. If this is only for your personal site I would recommend to either a) just copy over your implementation into your PR or b) add a system cronjob hitting the queue URL until it returns `0`. I don't feel like encouraging synchronous sends is a good thing generally.

dtdesign:
I can see that there are cases where it is tempting to use such a method, however, there are two major arguments that stand against this proposal:
1. Sending emails synchronously has the potential of breaking things, because sending data over a network is always an unreliable operation. Sending the emails directly does not only open up the potential for all sort of weird aborts, you also lose the ability to track which jobs were successful and which onces went down the drain.
2. Implementing this method in the Core without using it isn't a great idea. It won't be used in the regular code flow and thus becomes untested code (read: not used in production), opening up the possibility of becoming out of sync over time. Also, when things exists, third party developers are tempted to make use of it, without realizing the implications that come with it. We've introduced the queue for a reason and we were fully aware of the downsides of this approach, yet it turned out to be the lesser evil.

The bottom line is: The potential downsides of this change outweight the gains. You can implement such a code bit at your own discrection for your own use, but we will not ship it in the Core. Please see the comment by @TimWolla for a stable solutions to solve your problem.

PR #3065, opened 2019-09-01
Add PHP API to update database tables

Gravatronics:
See #2847


wcfsetup/install/files/lib/system/database/table/column/AbstractDatabaseTableColumn.class.php
```diff
@@ -0,0 +1,203 @@
...
+		
+		if ($column instanceof IDecimalsDatabaseTableColumn) {
+			$column->decimals($data['decimals'] ?: null);
+		}
+		
+		if ($column instanceof IEnumDatabaseTableColumn) {
+			$values = explode(',', $data['enumValues'] ?? []);
+			
+			$values = array_map(function($value) {
+				// trim one leading and one trailing `'`
```
TimWolla: Is `$value` guaranteed to be quoted at this place? I'd rather `preg_replace("/^'(.*)'$", '\\1', $value)`.

wcfsetup/install/files/lib/system/database/table/column/AbstractIntDatabaseTableColumn.class.php
```diff
@@ -0,0 +1,24 @@
...
+ * 
+ * @author	Matthias Schmidt
+ * @copyright	2001-2019 WoltLab GmbH
+ * @license	GNU Lesser General Public License <http://opensource.org/licenses/lgpl-license.php>
+ * @package	WoltLabSuite\Core\System\Database\Table\Column
+ * @since	5.2
+ */
+abstract class AbstractIntDatabaseTableColumn extends AbstractDatabaseTableColumn implements IAutoIncrementDatabaseTableColumn, ILengthDatabaseTableColumn, IUnsignedDatabaseTableColumn {
+	use TAutoIncrementDatabaseTableColumn;
+	use TLengthDatabaseTableColumn;
```
TimWolla: FWIW: The display length of integer columns is deprecated as of MySQL 8 and never made any sense. It does not restrict the values that can be stored. It's just a hint to the MySQL command line client (and possibly other applications) to properly pad the values.

see: https://dev.mysql.com/doc/refman/8.0/en/numeric-type-attributes.html

I'd take this opportunity to remove support for specifying the length of integer fields.

wcfsetup/install/files/lib/system/database/table/column/AbstractDatabaseTableColumn.class.php
Gravatronics: Yes, `$data` is from `DatabaseEditor::getColumns()`.

wcfsetup/install/files/lib/system/database/table/index/DatabaseTableForeignKey.class.php
```diff
@@ -0,0 +1,271 @@
...
+	 * name of referenced table
+	 * @var	string
+	 */
+	protected $referencedTable;
+	
+	/**
+	 * valid on delete/update actions
+	 * @var	string[]
+	 */
+	const VALID_ACTIONS = [
```
joshuaruesweg: Missing `RESTRICT` and `SET DEFAULT`

TimWolla: - `RESTRICT` is a MySQL specific alias of `NO ACTION`.
> NO ACTION: A keyword from standard SQL. In MySQL, equivalent to RESTRICT.
- `SET DEFAULT` is supported by the grammar, but not actually supported:
> SET DEFAULT: This action is recognized by the MySQL parser, but both InnoDB and NDB reject table definitions containing ON DELETE SET DEFAULT or ON UPDATE SET DEFAULT clauses.

Gravatronics: Additionally, these actions are the same as the ones supported the [SQL parser](https://github.com/WoltLab/WCF/blob/ba2c8f93b2dcb0df8c3281654fb3bbb2102a0163/wcfsetup/install/files/lib/system/database/util/SQLParser.class.php#L122).

wcfsetup/install/files/lib/system/database/table/column/AbstractDatabaseTableColumn.class.php
```diff
@@ -0,0 +1,203 @@
...
+		if ($column instanceof IAutoIncrementDatabaseTableColumn) {
+			$column->autoIncrement($data['autoIncrement'] ?: null);
+		}
+		
+		if ($column instanceof IDecimalsDatabaseTableColumn) {
+			$column->decimals($data['decimals'] ?: null);
+		}
+		
+		if ($column instanceof IEnumDatabaseTableColumn) {
+			$values = explode(',', $data['enumValues'] ?? []);
```
dtdesign: `explode()` expects the 2nd parameter to be a string, you need to change `?? []` into `?? ''`.

wcfsetup/install/files/lib/system/database/table/column/CharDatabaseTableColumn.class.php
```diff
@@ -0,0 +1,34 @@
...
+	 */
+	public function getMaximumLength() {
+		return 255;
+	}
+	
+	/**
+	 * @inheritDoc
+	 */
+	public function getMinimumLength() {
+		return 0;
```
dtdesign: Does `0` even make any sense for this type? Regardless of wether or not MySQL permits this, it makes no sense and thus implies a programming error. This applies to other column types with a length attribute (again, regardless if MySQL actually cares about it!).

wcfsetup/install/files/lib/system/database/table/column/IntDatabaseTableColumn.class.php
```diff
@@ -0,0 +1,25 @@
...
+	/**
+	 * @inheritDoc
+	 */
+	protected $type = 'int';
+	
+	/**
+	 * @inheritDoc
+	 */
+	public function getMaximumLength() {
+		return 11;
```
dtdesign: `INT` is a 32 bit number and thus limited to a length of `10`.

wcfsetup/install/files/lib/system/database/table/column/AbstractDatabaseTableColumn.class.php
TimWolla: > `explode()` expects the 2nd parameter to be a string, you need to change `?? []` into `?? ''`.

This probably wouldn't result in the correct result either, because explode always returns at least one items. I guess a good old `if (!empty($data['enumValues']))` is required here.

PR #3076, opened 2019-09-20
Hide stacktraces in "Exception in Exception"

WebFreak001:
Previously the Database->__construct filter might not match if the message is truncated such as in
```
#1 /var/www/html/intranet/lib/core.functions.php(154): file_put_contents('/var/www/html/i...', '<<<<<<<__construct('mysql:host=loca...', 'root', 'password', Array)
```
which caused our DB password to be leaked to users because the drive filled up and it couldn't log the SQL connection exception to the log file anymore.

If such a corrupt state occurs, I think it's better to rather check if debug mode is enabled and only then print the stack traces instead of always printing stack traces to prevent this leak.

Credentials and such could be in more functions than just `Database->__construct` so I think only having a case for this on something which will be facing the user is not a good solution and therefore I instead moved the whole stacktrace into the ENABLE_DEBUG_MODE flag.

Instead of using the class member function which can check if debug mode is enabled, just a simple `defined && constant` check is being done to prevent crashing in that class member method if it has some bug or gets changed.

I think this change should be included in the latest version, which is why I am merging this into the `master` branch according to the contributing guidelines, but then also be cherry-picked into the `3.0` branch to avoid this issue on the 3.0.x version.

### DIFF
--- a/wcfsetup/install/files/lib/system/WCF.class.php
+++ b/wcfsetup/install/files/lib/system/WCF.class.php
@@ -307,25 +307,43 @@ class WCF {
 				}
 			}
 		}
-		
+
+		// Debug variable only used in exception of exception handler.
+		// Kept minimal to prevent any exceptions from determining this.
+		$fatalDebug = defined('ENABLE_DEBUG_MODE') && ENABLE_DEBUG_MODE;
+
 		@header('HTTP/1.1 503 Service Unavailable');
 		try {
 			\wcf\functions\exception\printThrowable($e);
 		}
 		catch (\Throwable $e2) {
-			echo "<pre>An Exception was thrown while handling an Exception:\n\n";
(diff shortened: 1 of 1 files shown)


dtdesign:
This very specific scenario isn't ideal, but suppressing stacktraces at all cost has a significant downside in terms of troubleshooting. The downsides of such a change outweigh the benefits, which is why we chose not to make any change at this point.

WebFreak001:
I believe in a production environment it is very rare that you want to print any stacktraces of internal routines to the user. You would rather look at the error log inside the application.

Additionally this only suppresses exceptions in stacktrace printers, not the actual stacktraces that are still printed fine.

I would argue that this should be reopened under these considerations as it only affects a very very small amount of errors. Exceptions while printing exceptions is exceptionally rare and is usually not guarded against while writing any code.

PR #3088, opened 2019-10-07
corrected sql instructions WSC 3.1 > WSC 5.2

mutec:
Do not execute the foreign key for `wcf1_like.reactionTypeID` before filling the column with real values, this would be crash for sure! This is done by another PHP-PIP afterwards.


dtdesign:
Thanks!

PR #3090, opened 2019-10-07
fix apache rewrite generation

mutec:


mutec:
@TimWolla is it written in a clearly way or should I mention something special?

TimWolla:
The commit message is proper now, thanks. I'll let another developer merge to get a second pair of eyes on it, as I'm not super proficient with Apache Config.

mutec:
It might be good if you're going to test it with some use cases, too.
I checked it with my productive setup (every app in a different directory on the same level) and our management (root and sub-directory) which should handle the most cases. Or did I miss some special uses you might know about?
It's a bit tricky since we have to decide based on the absolute path and the relative path (to the domain's root). :)

PR #3093, opened 2019-10-14
template events for reaction and trophy list (dialogs)

mutec:
we might inject some JS or content like reaction-moderation


joshuaruesweg:
For which reasons are the events necessary?

mutec:
I need to load/execute some JavaScript within those templates to get every include of those for an extension of the reaction-management.

PR #3094, opened 2019-10-14
show a toggle button for reaction types in acp

mutec:
this gives administrators a quick overview which reaction types are enabled and which are not

### DIFF
--- a/wcfsetup/install/files/acp/templates/reactionTypeList.tpl
+++ b/wcfsetup/install/files/acp/templates/reactionTypeList.tpl
@@ -12,6 +12,7 @@
 		
 		$(function() {
 			new WCF.Action.Delete('wcf\\data\\reaction\\type\\ReactionTypeAction', '.reactionTypeRow');
+			new WCF.Action.Toggle('wcf\\data\\reaction\\type\\ReactionTypeAction', '.reactionTypeRow');
 		});
 	</script>
 {/if}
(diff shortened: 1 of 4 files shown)


wcfsetup/install/files/lib/data/reaction/type/ReactionTypeAction.class.php
```diff
@@ -222,4 +226,15 @@ public function delete() {
 		
 		return $returnValues;
 	}
+	
+	/**
+	 * Toggles the "isDisabled" status of the relevant objects.
```
joshuaruesweg:
```suggestion
	 * @inheritDoc
```

wcfsetup/install/files/acp/templates/reactionTypeList.tpl
```diff
@@ -46,6 +47,7 @@
 						
 						<span class="statusDisplay sortableButtonContainer">
 							<span class="icon icon16 fa-arrows sortableNodeHandle"></span>
+							<span class="jsOnly icon icon16 fa-{if $reactionType->isAssignable}check-{/if}square-o jsToggleButton jsTooltip pointer" title="{lang}wcf.global.button.{if !$reactionType->isAssignable}enable{else}disable{/if}{/lang}" data-object-id="{@$reactionType->reactionTypeID}"></span>
```
joshuaruesweg: The description text does not fit. We are talking about "disable" or "activate", but we should be talking about "assignable" or "not assignable", so that the user knows what the button does, just because in the detail editing we are not talking about disable/enable.

joshuaruesweg:
@mutec Are you planning on adjusting your pull request?

mutec:
@joshuaruesweg I do. I'm just busy at the moment.

wcfsetup/install/files/acp/templates/reactionTypeList.tpl
mutec: I'm unsure about the wording
`Verwendbar machen` / `Set assignable` and `Unverwendbar machen` / `Set nonassignable` sounds strange.

joshuaruesweg: Please use `Verwendung erlauben` and `Verwendung verbieten` and `Allow use` and `Disallow use` in english. I will adjust the detailed page soon to `Verwendung erlauben` and `Allow use` (or, if you like, you can also change the language variable with the pull request, too).

mutec: I hope I got everything done.

joshuaruesweg:
Thanks for the Initial work. Merged with: f8fa26d59c0f31d20c53cf637bbed5eb37a33f98

I've done the following in addition: 
* Squashed the commits 
* Merging the language variables, unnecessary duplicate variables were present
* Adding `data-(enable|disable)-title`, so that the language variable is correct, after enabling/disabling items

PR #3095, opened 2019-10-24
Add mixins for modern desktop screen resolutions

SoftCreatR:
Add support for bigger desktop resolutions than 1024px (target 1280px and up).

### DIFF
--- a/wcfsetup/install/files/style/bootstrap/mixin/mediaQuery.scss
+++ b/wcfsetup/install/files/style/bootstrap/mixin/mediaQuery.scss
@@ -56,7 +56,25 @@
 
 /* desktop */
 @mixin screen-lg() {
+	@media screen and (min-width: 1025px) and (max-width: 1280px) {
+		@content;
+	}
+}
+
+@mixin screen-lg-down() {
+	@media screen and (max-width: 1280px) {
+		@content;
+	}
+}
+
+@mixin screen-lg-up() {
 	@media screen and (min-width: 1025px), print {
+		@content;
+	}
+}
+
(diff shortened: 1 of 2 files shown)


wcfsetup/install/files/style/bootstrap/mixin/mediaQuery.scss
```diff
@@ -56,7 +56,25 @@
 
 /* desktop */
 @mixin screen-lg() {
+	@media screen and (min-width: 1025px) and (max-width: 1280px) {
```
dtdesign: This change would completely break backwards compatibility, because `screen-lg` would no longer apply to screen widths beyond `1280px`.

- `screen-lg` implicitly works like `screen-lg-up`
- `screen-lg-only` works between `1025px` and `1280px`
- `screen-lg-down` works as proposed
- `screen-xl` works as proposed

This does break with the naming scheme, but this is the lesser evil.

SoftCreatR: Applied the requested changes. I was unsure, if i should risk a breaking change in favor of the naming convention, or not.

dtdesign:
I was about to merge it, but then realized that `WoltLabSuite/Core/Ui/Screen` is missing the new presets in the `_mqMap` dictionary. Please add the new rules there and then its ready to be included in `next`.

SoftCreatR:
Done that. However, i've ommitted the comment, because the newly added resolutions are still desktop resolutions.

Want me to squash all commits?

PR #3113, opened 2019-11-30
Enable ability for applications to add their own custom fields in their own namespace

earlyhost:
Currently all field templates have to be placed in the wcf folder. If an application wants to create their own custom field, the template has to placed in the wcf folder as well. This change enables custom components to use their own template folder.


Gravatronics:
Please rename `$application` to `$templateApplication` and I will gladly merge the pull request.

PR #3128, opened 2020-01-03
Prevent newline after the userBBCodeTag

TimWolla:
The newline renders as a single space, which looks bad when punctuation
is being used directly after a mention.

Previously a sentence such as:

> Good to see you @Foo!

would have resulted in:

> Good to see you <a>@Foo</a>
> !

which is being rendered as:

> Good to see you @Foo !


TimWolla:
see: https://community.woltlab.com/thread/280402-nach-erw%C3%A4hnung-das-leerzeichen-entfernen/

PR #3131, opened 2020-01-05
make CoreRewriteTestAction available during offline mode

mutec:
Configuring SEO urls during offline mode is currently not possible in case the user is not logged into the frontend. This might lead to confused administrators and us searching for the reason.


jacboy:
lol wtf

dtdesign:
@mutec This affects 3.1 (currently `master`) too, could you please base your changes against that branch instead?

mutec:
@dtdesign I changed the base and recommited.

dtdesign:
Thanks!

PR #3132, opened 2020-01-05
allow manipulation of email object before background job get's enqueued

mutec:


TimWolla:
What does this allows what method calls on the `Email` object wouldn't allow?

mutec:
I need to replace the whole email body and I can't do that at the moment since the original `$mail` will be sent instead of the object I am able to modify though `$parameters`.
At the moment I'm able to convert [any email to an encrypted email](https://community.woltlab.com/thread/218145-smalltalk-labern-reden-diskutieren/?postID=1777073#post1777073) but I can't send it to the user.*

* okay, I could skip adding the job in favour of calling `MyClass->send()`, but I'm not sure about the performance-loss of this strategy.

TimWolla:
> I need to replace the whole email body and I can't do that at the moment since the original `$mail` will be sent instead of the object I am able to modify though `$parameters`.

Doesn't `$email->setBody()` work for you?

mutec:
I didn't try, but since the object get's cloned it should be senseless to modify `$eventObj` after the clone has been created.
https://github.com/WoltLab/WCF/blob/32dfe6c1c90ffac973cecfccb0502bb119b6d58d/wcfsetup/install/files/lib/system/email/Email.class.php#L500
https://github.com/WoltLab/WCF/blob/32dfe6c1c90ffac973cecfccb0502bb119b6d58d/wcfsetup/install/files/lib/system/email/Email.class.php#L519

TimWolla:
> I didn't try, but since the object get's cloned it should be senseless to modify `$eventObj` after the clone has been created.

You can use `$parameters['mail']->setBody(…)`. I would argue that passing `$this` to the EventHandler instead of `$mail` is a bug, though.

mutec:
I wanted to use `$parameters['mail']->setBody()`, but this is currently useless since that object in `$parameters` is not used. ;)
Changing to `EventHandler::getInstance()->fireAction($mail, 'getJobs', $data);` might be good too, I didn't think about that. Might there be any case where you'd want to be able to access the email including all recipients AND the one including only one of them?

TimWolla:
> I wanted to use `$parameters['mail']->setBody()`, but this is currently useless since that object in `$parameters` is not used. ;)

The `$parameters` array does not contain an object. It contains a reference to an object. Simply calling the method works just fine, because it modifies the object via the reference.

> Might there be any case where you'd want to be able to access the email including all recipients AND the one including only one of them?

The recipients are not stripped after cloning the object. BCC is handled during header generation. The only modification that cannot be reverted currently is the addition of the `X-WoltLab-Suite-Recipient` header.

mutec:
> The `$parameters` array does not contain an object. It contains a reference to an object. Simply calling the method works just fine, because it modifies the object via the reference.

Ahh. I need some sleep, thanks. It's working now. :)

PR #3138, opened 2020-01-16
Add &display=swap to Google Font request

netzhuffle:
This requests Google Fonts to deliver fonts with the `font-display: swap;` CSS option, reducing time until meaningful display epecially on slow connections. Including this option is suggested by default now from Google Fonts.


dtdesign:
Thanks!

PR #3141, opened 2020-01-22
Update googleMapsJavaScript.tpl

SoftCreatR:
Prevent loading of multiple instances of Google maps related scripts.


TimWolla:
Have you checked what happens if yet another GMaps is loaded via AJAX within an overlay?

SoftCreatR:
Yes, i did (using the post preview). Works fine so far.

The problem is, that Google yields an error, if the API is included multiple times:

> You have included the Google Maps API multiple times on this page. This may cause unexpected errors.

There are plugins (such as our GPX maps extension), that rely on the existence of the API on every page. And there are plugins and applications (e.g. the Gallery), which are loading the API (respectively the file googleMapsJavaScript.tpl) on pages, where it's needed. As of now, this would produce the result above.

This PR aims to fix this problem.

TimWolla:
Technically the files could also be loaded using require.js, which would avoid the issue as well:

```
require(['js/3rdParty/oms.min.js'], (function (oms) { // do nothing, not an actual module }))
```

This comes with bigger changes, but avoids the need of another template variable. 

In any case, I'll leave this up for review by a developer that has more hands-on knowledge of the GMap integration.

Gravatronics:
While this pull request does not fix the error when loading *additional* maps via AJAX, it fixes the potential problem for multiple maps in the same request.

Fixing the potential problem for additional maps via AJAX requires additional changes that are out of scope for a bug fix release.

PR #3142, opened 2020-01-24
Fix day in wcf.date.interval.full.past.inSentence (en.xml)

Krymonota:
e. g. "day, an hour and 49 minutes ago" -> "a day, an hour and 49 minutes ago"


Krymonota:
This line was quite difficult to edit. However, two heads are better than one 😉 It should be alright now, but please have another look at my commit.

TimWolla:
Rebased onto 3.1, slightly adjusted the commit message of the second commit and merged. Thanks.

PR #3143, opened 2020-01-31
Event added for changing variable $registerVia3rdParty in RegisterForm

Hanashi:
Most of SSO providers don't need a mail activation, because the mail is already confirmed. 3rd party developers currently can not manipulate the variable `$registerVia3rdParty` in method `wcf\form\RegisterForm::save()`. Because of that, I added an event handler for 3rd party developers.
Another solution would be to change the variable `$registerVia3rdParty` to a property of `RegisterForm`. 3rd party developers have the same problem with the variable `$saveOptions`.

### DIFF
--- a/wcfsetup/install/files/lib/form/RegisterForm.class.php
+++ b/wcfsetup/install/files/lib/form/RegisterForm.class.php
@@ -17,6 +17,7 @@ use wcf\system\email\Email;
 use wcf\system\email\Mailbox;
 use wcf\system\email\UserMailbox;
 use wcf\system\exception\NamedUserException;
+use wcf\system\event\EventHandler;
 use wcf\system\exception\PermissionDeniedException;
 use wcf\system\exception\SystemException;
 use wcf\system\exception\UserInputException;
@@ -403,6 +404,7 @@ class RegisterForm extends UserAddForm {
 			// create fake password
 			$this->password = StringUtil::getRandomID();
 		}
+		EventHandler::getInstance()->fireAction($registerVia3rdParty, 'registerVia3rdParty');
 		
 		$this->additionalFields['languageID'] = $this->languageID;
 		if (LOG_IP_ADDRESS) $this->additionalFields['registrationIpAddress'] = WCF::getSession()->ipAddress;
--- a/wcfsetup/install/files/lib/form/RegisterForm.class.php
+++ b/wcfsetup/install/files/lib/form/RegisterForm.class.php
@@ -404,7 +404,12 @@ class RegisterForm extends UserAddForm {
 			// create fake password
 			$this->password = StringUtil::getRandomID();
 		}
-		EventHandler::getInstance()->fireAction($registerVia3rdParty, 'registerVia3rdParty');
+		
+		$parameters = [
+			'saveOptions' => $saveOptions,
+			'registerVia3rdParty' => $registerVia3rdParty
+		];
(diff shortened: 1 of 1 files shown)


wcfsetup/install/files/lib/form/RegisterForm.class.php
```diff
@@ -403,6 +404,7 @@ public function save() {
 			// create fake password
 			$this->password = StringUtil::getRandomID();
 		}
+		EventHandler::getInstance()->fireAction($registerVia3rdParty, 'registerVia3rdParty');
```
TimWolla: This does not work like this. You need to specify a valid object or class name as the first parameter. On top of that the only parameter that is passed by reference is the third one. Pass an array as the third parameter and after the event update the variable with the current contents of that array.

While doing so also have a look whether there is anything else you'd like to modify, possibly creating a more generally useful event.

Hanashi:
@TimWolla what's the current status of this PR?

TimWolla:
> Why is this not merged or an answer, yet?

Even if I usually perform a timely technical review I leave the final decision, especially with regard to the target branch, to someone else.

wcfsetup/install/files/lib/form/RegisterForm.class.php
```diff
@@ -403,6 +404,14 @@ public function save() {
 			// create fake password
 			$this->password = StringUtil::getRandomID();
 		}
+
+		$data = [
+			'saveOptions' => $saveOptions,
+			'registerVia3rdParty' => $registerVia3rdParty
+		];
+		EventHandler::getInstance()->fireAction($this, 'externalAuthentication', $data);
```
TimWolla, line 412: I believe that re-using the existing `3rdParty` terminology makes this more clear:
```suggestion
		EventHandler::getInstance()->fireAction($this, 'registerVia3rdParty', $eventParameters);
```

wcfsetup/install/files/lib/form/RegisterForm.class.php
```diff
@@ -403,6 +404,14 @@ public function save() {
 			// create fake password
 			$this->password = StringUtil::getRandomID();
 		}
+
+		$data = [
```
TimWolla, line 408: `$data` is awfully generic and can accidentally clash with future changes to the method.

```suggestion
		$eventParameters = [
```

wcfsetup/install/files/lib/form/RegisterForm.class.php
```diff
@@ -17,6 +17,7 @@
 use wcf\system\email\Mailbox;
 use wcf\system\email\UserMailbox;
 use wcf\system\exception\NamedUserException;
+use wcf\system\event\EventHandler;
```
dtdesign, line 20: Incorrect location, please keep the import sorted.

wcfsetup/install/files/lib/form/RegisterForm.class.php
```diff
@@ -403,6 +404,14 @@ public function save() {
 			// create fake password
 			$this->password = StringUtil::getRandomID();
 		}
+
+		$data = [
+			'saveOptions' => $saveOptions,
+			'registerVia3rdParty' => $registerVia3rdParty
```
dtdesign, line 410: Missing trailing comma.

TimWolla:
Squashed the commits, made the necessary adjustments due to the deleted source repository and applied to 5.2. Thanks.

PR #3149, opened 2020-02-06
Update GitHub authorization to use headers instead of query parameters

Krymonota:
Using query parameters for access_token has been deprecated.

For further information, check out:
- https://community.woltlab.com/thread/281366-github-api-deprecation-notice-for-authentication-via-url-query-parameters/
- https://developer.github.com/changes/2019-11-05-deprecated-passwords-and-authorizations-api/#authenticating-using-query-parameters


Krymonota:
This change has been tested by myself. Feel free to test it as well to confirm that nothing breaks.

TimWolla:
I believe you testing this change and I verified the changes against the documentation.

LGTM, thanks.

PR #3152, opened 2020-02-06
Add missing alt attribute to reaction type image

Krymonota:
See: https://community.woltlab.com/thread/281417-reaktions-icons-alt-attribut-fehlt/

### DIFF
--- a/com.woltlab.wcf/templates/reactionTypeImage.tpl
+++ b/com.woltlab.wcf/templates/reactionTypeImage.tpl
@@ -1,5 +1,6 @@
 <img
 	src="{@$__wcf->getPath()}images/reaction/{$reactionType->iconFile}"
+	alt="{$reactionType->getTitle()}"
 	class="reactionType"
 	data-reaction-type-id="{$reactionType->reactionTypeID}"
 >


Krymonota:
I'm aware that the surrounding `li` element contains an `aria-label` attribute with exactly the same content when you add a reaction. However, this template is also embedded in other places (e. g. reaction list) where there's no `aria-label` attribute to describe the reaction type.

Krymonota:
Well, maybe we should take a similar approach to the [`renderTrophy` method](https://github.com/WoltLab/WCF/blob/c839bd496deec55ffc05dc9ad38165d84b6b8f41/wcfsetup/install/files/lib/data/trophy/Trophy.class.php#L77) to allow the attribute to be set where necessary? However, I would still suggest using the `alt` attribute in favor of the `aria-label` attribute (or simply both when applied to the same element) for best compatibility with all screen readers.

dtdesign:
I think that we could use an empty `alt` attribute instead. The title itself doesn't offer any meaningful description because it lacks the context and most important, the counter itself.

Essentially there are two goals when using a non empty `alt` attribute:
1. Show a description when the resource could not be loaded: Unreadable in major browsers due to the relatively small dimensions of the image.
2. Provide a description for visually impaired visitors: The `aria-label` already takes care of this by providing a superiod readout.

That said, an empty `alt` attribute might be the best choice?!

Krymonota:
Unfortunately, using the `aria-label` attribute doesn't always work.

Let's take a look at adding reactions to content:
React

I've added an `alt` attribute to the `img` element for reaction type id 1, 4 and 5 that describes the reaction type. There's also an empty `alt` attribute for reaction type id 2.
```html
<ul>
    <li class="reactionTypeButton" data-reaction-type-id="1" data-title="Gefällt mir" data-is-assignable="1" data-tooltip="Gefällt mir" aria-label="Gefällt mir"><img src="https://5.2.wsc.niklas.id/images/reaction/like.svg" alt="Gefällt mir" class="reactionType" data-reaction-type-id="1"><span class="reactionTypeButtonTitle">Gefällt mir</span></li>
    <li class="reactionTypeButton" data-reaction-type-id="2" data-title="Danke" data-is-assignable="1" data-tooltip="Danke" aria-label="Danke"><img src="https://5.2.wsc.niklas.id/images/reaction/thanks.svg" class="reactionType" data-reaction-type-id="2" alt=""><span class="reactionTypeButtonTitle">Danke</span></li>
    <li class="reactionTypeButton active" data-reaction-type-id="3" data-title="Haha" data-is-assignable="1" data-tooltip="Haha" aria-label="Haha"><img src="https://5.2.wsc.niklas.id/images/reaction/haha.svg" class="reactionType" data-reaction-type-id="3"><span class="reactionTypeButtonTitle">Haha</span></li>
    <li class="reactionTypeButton" data-reaction-type-id="4" data-title="Verwirrend" data-is-assignable="1" data-tooltip="Verwirrend" aria-label="Verwirrend"><img src="https://5.2.wsc.niklas.id/images/reaction/confused.svg" alt="Verwirrend" class="reactionType" data-reaction-type-id="4"><span class="reactionTypeButtonTitle">Verwirrend</span></li>
    <li class="reactionTypeButton" data-reaction-type-id="5" data-title="Traurig" data-is-assignable="1" data-tooltip="Traurig" aria-label="Traurig"><img src="https://5.2.wsc.niklas.id/images/reaction/sad.svg" alt="Traurig" class="reactionType" data-reaction-type-id="5"><span class="reactionTypeButtonTitle">Traurig</span></li>
</ul>
```
Of course the `li` element has an `aria-label` attribute. However, some screen readers (like NVDA) don't recognize it. Whether you set no attribute or an empty `alt` attribute the screen reader produces no output at all. This is only the case with my change.

However, an empty `alt` attribute seems more useful in other places like this one:
Reaction List

dtdesign:
Thanks.

PR #3153, opened 2020-02-07
Remove Typehint

SoftCreatR:
`handleCLIException` fails, because it expects an Exception, but there are cases, where it receives an Error instead (e.g. phar extension missing).


dtdesign:
Thanks!

PR #3155, opened 2020-02-08
Improve WhatsApp Message Sharing

SoftCreatR:
Currently, the WhatsApp sharing button requires `whatsapp` to be a registered protocol. For quite some time, they provide an intermediate web page, that is also compatible with desktops.


TimWolla:
Thanks.

PR #3156, opened 2020-02-08
Improve the system check during WCFSetup in regards to the graphics l…

### LINKED ISSUE #3154: Improve the system check during WCFSetup in regards to the graphics library
The GD library is currently a requirement and ImageMagick is listed as an optional component. However, some users are confused by having ImageMagick being missing, despite being flagged as optional. In reality, we don't really care if there is GD or ImageMagick, as long as at least one of them is available, therefore, the check should be modified to probe against any of these libs and pass the check if one or both are found.

* [x] Change the section on GD to the more general "Graphics Library".
* [x] Pass the test if either `GD > 2.0.0` or `ImageMagick > 6.3.6` (it's ancient anyway, so we don't care for older libs) are available.
* [x] Fail if none are present.

See the changes made by #2684.

BurntimeX:
…ibrary

Closes #3154

### DIFF
--- a/wcfsetup/install/files/lib/system/WCFSetup.class.php
+++ b/wcfsetup/install/files/lib/system/WCFSetup.class.php
@@ -13,6 +13,7 @@ use wcf\system\database\MySQLDatabase;
 use wcf\system\devtools\DevtoolsSetup;
 use wcf\system\exception\SystemException;
 use wcf\system\exception\UserInputException;
+use wcf\system\image\adapter\GDImageAdapter;
 use wcf\system\image\adapter\ImagickImageAdapter;
 use wcf\system\io\File;
 use wcf\system\io\Tar;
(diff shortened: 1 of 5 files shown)


wcfsetup/install/files/lib/system/WCFSetup.class.php
```diff
@@ -360,26 +361,20 @@ protected function showSystemRequirements() {
...
-			$temp = gd_info();
-			$match = [];
-			if (preg_match('!([0-9]+\.[0-9]+(?:\.[0-9]+)?)!', $temp['GD Version'], $match)) {
-				if (preg_match('/^[0-9]+\.[0-9]+$/', $match[1])) $match[1] .= '.0';
-				$system['gdLib']['value'] = $match[1];
-			}
+		// graphics library
+		$system['graphicsLibrary']['result'] = false;
+		$system['graphicsLibrary']['value'] = '';
+		if (ImagickImageAdapter::isSupported() && ImagickImageAdapter::supportsAnimatedGIFs(ImagickImageAdapter::getVersion())) {
```
TimWolla: I believe the check for animated GIFs is incorrect here. GD does not support animated GIFs at all, thus the support within Imagick is a benefit, but not a requirement.

dtdesign: This check is correct, animated GIFs in ImageMagick are supported since mid 2007 (ImageMagick 6.3.6). It's a (mostly obsolete) safeguard against bogus setups.

PR #3159, opened 2020-02-12
Patch jQuery against Prototype Pollution

SoftCreatR:
See https://github.com/advisories/GHSA-6c3j-c64m-qhgq

I think, patching the used lib is the best way to prevent any compatibility issues.

WSC 3.0 & 3.1 are also affected, but they're using older versions of jQuery. However, i can patch them too and create separate PRs, if you like.


dtdesign:
Thanks, but I've manually backported the original fix to 3.0 and applied the changes to all subsequent trees.

PR #3167, opened 2020-02-22
Fix int length handling during DB updates for MySQL >= 8.0.19

### LINKED ISSUE #3160: MySQL 8.0.19: Incompatible changes to integer column definitions
> Display width specification for integer data types was deprecated in MySQL 8.0.17, and now statements that include data type definitions in their output no longer show the display width for integer types […]

See https://dev.mysql.com/doc/relnotes/mysql/8.0/en/news-8-0-19.html#mysqld-8-0-19-feature

This change causes `SHOW COLUMNS FROM …` to report columns `int`, `smallint` and `tinyint` (unless it is `tinyint(1)`) without any length whatsoever. In MySQL < 8.0.19 these columns would be reported as `int(10)`, `smallint(5)`, etc. The `DatabaseTableChangeProcessor` is unaware of this change, eventually causing an infinite loop during the upgrade 3.1 -> 5.2 when dealing with the `wcf1_media` table.

Gravatronics:
Close #3160

### DIFF
--- a/wcfsetup/install/files/lib/system/database/editor/DatabaseEditor.class.php
+++ b/wcfsetup/install/files/lib/system/database/editor/DatabaseEditor.class.php
@@ -7,7 +7,7 @@ use wcf\system\exception\NotImplementedException;
  * Abstract implementation of a database editor.
  * 
  * @author	Marcel Werk
- * @copyright	2001-2020 WoltLab GmbH
+ * @copyright	2001-2019 WoltLab GmbH
  * @license	GNU Lesser General Public License <http://opensource.org/licenses/lgpl-license.php>
  * @package	WoltLabSuite\Core\System\Database\Editor
  */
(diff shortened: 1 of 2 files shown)


dtdesign:
@TimWolla Technically, we can skip the version check and always ignore _differences_ in the length of integer values, right?

TimWolla:
Yes, that's my understanding and that's what I wanted to suggest with my review. The “display length” of integer columns is only relevant when used in combination with the `ZEROFILL` option which is super obscure, non-standard and deprecated together with the display length in MySQL 8.0.17+.

dtdesign:
@Gravatronics Please remove the version check and ignore the length unless it's `TINYINT(1)`.

TimWolla:
And FWIW: Dumping a database with mysqldump on 8.0.19 will drop the integer widths as well. So anyone that ever was on a modern MySQL 8 will have “incorrect” values in there:

```
# mysqldump -p XXXX |grep AUTO_INCREMENT |head -n5
Enter password: 
  `blogID` int NOT NULL AUTO_INCREMENT,
  `coverPhotoID` int NOT NULL AUTO_INCREMENT,
  `entryID` int NOT NULL AUTO_INCREMENT,
  `coverPhotoID` int NOT NULL AUTO_INCREMENT,
  `eventID` int NOT NULL AUTO_INCREMENT,
```

PR #3173, opened 2020-03-04
Fixed a bug where the packageurl was not loaded correctly when editing a project in the dev tools

RouL:
In DevtoolsProjectEditForm.class.php the packageurl in $packageArchive should've been referenced as 'packageURL' not packageurl, that prevented loading it at all. Also the form field was referenced as 'packageurl' while the correct field name is 'packageUrl'. If you save a project again those 2 bugs will empty your packageurl field in the package.xml.

This pull request fixes that behavior.

### DIFF
--- a/wcfsetup/install/files/lib/acp/form/DevtoolsProjectEditForm.class.php
+++ b/wcfsetup/install/files/lib/acp/form/DevtoolsProjectEditForm.class.php
@@ -17,7 +17,7 @@ use wcf\system\WCF;
 
 /**
  * Shows the devtools project edit form.
- * 
+ *
  * @author	Alexander Ebert, Matthias Schmidt
  * @copyright	2001-2019 WoltLab GmbH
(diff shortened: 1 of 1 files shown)


RouL:
Whooopsi whitespaces removed...making new PR. m(

TimWolla:
> making new PR. m(

You could also edit the commit and force push into this pull request. A follow up commit would work as well, those can easily be squashed when pressing the “Merge” button.

RouL:
> You could also edit the commit and force push into this pull request. A follow up commit would work as well, those can easily be squashed when pressing the “Merge” button.

It removed _all_ trailing whitespaces. Was easier this way. ;)
But in general you're right of course.

TimWolla:
You even used the same branch. I don't see how creating a new PR can be easier than just using the existing one, because by pushing this one would simply have updated.

PR #3177, opened 2020-03-07
Provide auth credentials via the instance configuration

dtdesign:
n/t


wcfsetup/install/files/lib/data/package/update/server/PackageUpdateServer.class.php
```diff
@@ -1,5 +1,6 @@
 <?php
 namespace wcf\data\package\update\server;
+use blog\data\entry\Entry;
```
Gravatronics:
```suggestion
```

PR #3179, opened 2020-03-10
Make $keywordHighlighting public

ChristopherWalz:
Make attribute public so it can be altered with an EL beforeProcess@HtmlOutputNodeProcessor


ChristopherWalz:
Nevermind, I didn't see HtmlOutputNodeProcessor::enableKeywordHighlighting

PR #3193, opened 2020-03-22
Strict filter for files deployed through styles

dtdesign:
n/t


wcfsetup/install/files/lib/data/style/StyleEditor.class.php
```diff
@@ -42,6 +42,7 @@
 class StyleEditor extends DatabaseObjectEditor implements IEditableCachedObject {
 	const EXCLUDE_WCF_VERSION = '6.0.0 Alpha 1';
 	const INFO_FILE = 'style.xml';
+	const VALID_IMAGE_EXTENSIONS = ['.gif', '.jpg', '.jpeg', '.png'];
```
TimWolla: svg and possibly webp?

dtdesign: WebP isn't supported by us yet.

jens1o:
Out of curiosity: Why don't you use `pathinfo`?

dtdesign:
I was primarily reusing existing code, also using `pathinfo()` doesn't make a significant difference here.

PR #3194, opened 2020-03-23
Size of multi select fields too small

fogl3000:
If you have 10 or less options in a multi select condition field and using option groups, the field size is always 1.


Gravatronics:
After messing up this PR, I cherry-picked the relevant commit for 3.1 (6642399255ec9921dcb304d5412c31a380f40f08).

PR #3196, opened 2020-03-25
Remove duplicate German language item in en.xml

Krymonota:


TimWolla:
Picked manually into 5.2, please check the target branch for future contributions.

Thanks.

PR #3203, opened 2020-04-08
Specify HTML.Doctype as HTML 4 Strict for HTML Purifier

TimWolla:
Not setting a Doctype defaults to HTML 4 Transitional, allowing legacy
tags. HTML 4 Strict will attempt to convert them into a standards compliant
format.

HTML Purifier does not support the HTML 5 Doctype as a readily available
configuration.

Before:

    string '<p><font size="6">This is some text!</font></p>' (length=47)

After:

    string '<p><span style="font-size:xx-large;">This is some text!</span></p>' (length=66)

### DIFF
--- a/wcfsetup/install/files/lib/system/html/input/filter/MessageHtmlInputFilter.class.php
+++ b/wcfsetup/install/files/lib/system/html/input/filter/MessageHtmlInputFilter.class.php
@@ -52,6 +52,7 @@ class MessageHtmlInputFilter implements IHtmlInputFilter {
 			
 			$config->set('CSS.AllowedProperties', ['color', 'font-family', 'font-size']);
 			$config->set('HTML.ForbiddenAttributes', ['*@lang', '*@xml:lang']);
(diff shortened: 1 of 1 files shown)


wcfsetup/install/files/lib/system/html/input/filter/MessageHtmlInputFilter.class.php
```diff
@@ -31,6 +31,7 @@ public function apply($html) {
 		require_once(WCF_DIR . 'lib/system/html/input/filter/HTMLPurifier_URIScheme_ts3server.php');
 		
 		$html = $this->getPurifier()->purify($html);
+		var_dump($html);exit;
```
Gravatronics, line 34:
```suggestion
```

TimWolla: Already gone. I noticed myself.

TimWolla: Actually *not* gone. Now is.

PR #3206, opened 2020-04-11
Restrict the ability to modify some permissions when running in enterprise mode

dtdesign:
This is the revision of #3172 which simplified the calculation of exceeding permissions. The previous change turned out to be too restrictive in environments where the site owner is not a member of the owner group.

### DIFF
--- a/wcfsetup/install/files/lib/system/option/user/group/UserGroupOptionHandler.class.php
+++ b/wcfsetup/install/files/lib/system/option/user/group/UserGroupOptionHandler.class.php
@@ -170,7 +170,7 @@ class UserGroupOptionHandler extends OptionHandler {
 			return;
 		}
 		
-		if (ENABLE_ENTERPRISE_MODE && $this->isAdmin() && !in_array($option->optionName, $this->enterpriseBlacklist)) {
+		if ($this->isAdmin() && (!ENABLE_ENTERPRISE_MODE || !in_array($option->optionName, $this->enterpriseBlacklist))) {
 			return;
 		}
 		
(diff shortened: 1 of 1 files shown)


wcfsetup/install/files/lib/system/option/user/group/UserGroupOptionHandler.class.php
```diff
@@ -148,13 +166,17 @@ protected function isOwner() {
...
-			$typeObj = $this->getTypeObject($option->optionType);
-			
-			if ($typeObj->compare($this->optionValues[$option->optionName], WCF::getSession()->getPermission($option->optionName)) == 1) {
-				throw new UserInputException($option->optionName, 'exceedsOwnPermission');
-			}
+		if ($this->isOwner()) {
+			return;
+		}
+		
+		if (ENABLE_ENTERPRISE_MODE && $this->isAdmin() && !in_array($option->optionName, $this->enterpriseBlacklist)) {
```
TimWolla, line 173: Should this condition be:
```suggestion
		if ($this->isAdmin() && (!ENABLE_ENTERPRISE_MODE || !in_array($option->optionName, $this->enterpriseBlacklist)) {
```

→ If the user is admin _and_ either the enterprise mode is disabled or the option is not blacklisted then the user may modify the option.

dtdesign: (I used a separate commit, because you lacked a closing bracket and GitHub exploded or something when trying to edit the suggestion.)

PR #3215, opened 2020-04-17
Fix return values in the UserStorageHandler for modified fields unloa…

joshuaruesweg:
…ded users

Currently the UserStorageHandler returns obsolete values if the value is deleted/updated before the user object is cached.

### DIFF
--- a/wcfsetup/install/files/lib/system/user/storage/UserStorageHandler.class.php
+++ b/wcfsetup/install/files/lib/system/user/storage/UserStorageHandler.class.php
@@ -79,12 +79,12 @@ class UserStorageHandler extends SingletonFactory {
 				$this->cache[$row['userID']] = [];
 			}
 			
-			// Skip the field, if the value is reset before the storage is loaded.
+			// Skip the field, if the value was reset before the storage had been loaded.
 			if (isset($this->resetFields[$row['userID']]) && in_array($row['field'], $this->resetFields[$row['userID']])) {
 				continue;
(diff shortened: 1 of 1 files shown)


wcfsetup/install/files/lib/system/user/storage/UserStorageHandler.class.php
```diff
@@ -79,7 +79,18 @@ public function loadStorage(array $userIDs) {
 				$this->cache[$row['userID']] = [];
 			}
 			
-			$this->cache[$row['userID']][$row['field']] = $row['fieldValue'];
+			// Skip the field, if the value is reset before the storage is loaded.
```
dtdesign, line 82:
```suggestion
			// Skip the field, if the value was reset before the storage had been loaded.
```

wcfsetup/install/files/lib/system/user/storage/UserStorageHandler.class.php
```diff
@@ -79,7 +79,18 @@ public function loadStorage(array $userIDs) {
...
+			if (isset($this->resetFields[$row['userID']]) && in_array($row['field'], $this->resetFields[$row['userID']])) {
+				continue;
+			}
+			
+			// Use new value, if the field is updated before the storage is loaded.
```
dtdesign, line 87:
```suggestion
			// Use the new value, if the field was updated before the storage had been loaded.
```

PR #3218, opened 2020-04-21
Add contentItemMetaIcons template event

Krymonota:
n/t

### DIFF
--- a/com.woltlab.wcf/templates/articleListItems.tpl
+++ b/com.woltlab.wcf/templates/articleListItems.tpl
@@ -67,6 +67,8 @@
 							{$article->getDiscussionProvider()->getDiscussionCount()}
 						</span>
 					</div>
+					
+					{event name='contentItemMetaIcons'}
 				</div>
 			</div>
 		</article>


Krymonota:
Is there any chance that we'll ever see this template event in 5.2?

Hanashi:
Why is this not merged, yet?

dtdesign:
What is the purpose of this event, or to be more precise, what do you want to insert at this location? Events had been introduced sparsely with the redesigned layout in an effort to keep the design clean.

Although the framework offers a lot of flexibility, it is _heavily opinionated_. We see third party developers quite often adding things soley for the reason that they _can_ ,without reflecting if they _should_. Therefore, please briefly explain what you want to achieve by providing a meaningful reason.

Krymonota:
The reason for this PR is my plugin [Reading Time Estimate for Articles](https://pluginstore.woltlab.com/file/3174-reading-time-estimate-for-articles/). Several users have requested to be able to see the reading time estimate in the article list. This would be the most appropriate place from my point of view and that of the users.

Personally, I don't think it's a good idea to limit flexibility in this way. After all, plugin developers simply offer additional functionality, and the user can decide whether to enable it or install the plugin at all.

dtdesign:
Thank you for taking your time to explain it.

PR #3221, opened 2020-04-26
Work around bug in createImageBitmap in Chrome

TimWolla:
Google Chrome returns inconsistent sizing between `img.(width|height)` and
the (width|height) of the bitmap returned by `createImageBitmap()`.

See: https://bugs.chromium.org/p/chromium/issues/detail?id=1069965

This mismatch causes Pica to corrupt the resulting image.

This commit disables resizing when the bug is detected. The behavior might
need to be re-adjusted depending on what behavior Chrome's fix will result
in.

### DIFF
--- a/wcfsetup/install/files/js/WoltLabSuite/Core/Image/Resizer.js
+++ b/wcfsetup/install/files/js/WoltLabSuite/Core/Image/Resizer.js
@@ -165,6 +165,10 @@ define([
 			
 			var canvas = document.createElement('canvas');
 			
+			var chromeBug = createImageBitmap(image).then(function (bitmap) {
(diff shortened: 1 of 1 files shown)


wcfsetup/install/files/js/WoltLabSuite/Core/Image/Resizer.js
```diff
@@ -165,6 +165,10 @@ define([
 			
 			var canvas = document.createElement('canvas');
 			
+			const chromeBug = createImageBitmap(image).then(bitmap => {
```
TimWolla, line 168:
```suggestion
			var chromeBug = createImageBitmap(image).then(bitmap => {
```

PR #3290, opened 2020-05-06
Fix mistake in profile comment like notification

Krymonota:
https://community.woltlab.com/thread/283453-jemandem-hat-auf-reagiert/


Krymonota:
Fixed in commit 6857af93ecaf7a4b121bfe05799fec61a5f213b2.

PR #3294, opened 2020-05-08
Hide empty user option categories in user menu

### LINKED ISSUE #3280: Hide empty user option categories in frontend
see https://community.woltlab.com/thread/272800-anpassung-der-useroption-pip-resp-ausgabe/

Gravatronics:
Close #3280


wcfsetup/install/files/lib/system/menu/user/UserMenu.class.php
```diff
@@ -31,6 +41,14 @@ protected function checkMenuItem(ITreeMenuItem $item) {
 		
 		if (!parent::checkMenuItem($item)) return false;
 		
+		// Hide links to user option categories without an accessible options.
```
dtdesign, line 44:
```suggestion
		// Hide links to user option categories without accessible options.
```

PR #3299, opened 2020-05-10
Prevent infinite loop for nested language items

### LINKED ISSUE #3163: Accessing phrases of the same category inside a phrase can cause an infinite loop
Accessing phrases inside a phrase using the `{lang}…{/lang}` syntax can cause an infinite loop when the phrase belongs to the same category and the category has not been loaded yet. This can (and should(!)) be avoided by using the `{…|language}` modifier, however, this is often overlooked and can cause strange infinite loops that yield misleading PHP errors.

Reproduction scenario:
1. The language category `foo.bar` is not loaded and compiled yet.
2. The phrase `foo.bar.baz` is requested and contains template scripting, including `{lang}foo.bar.foobar{/lang}`.
3. The category is requested and the compilation of the category is starting.
4. The compiler comes across `foo.bar.baz` and sees the reference to the phrase `foo.bar.foobar`.
5. The lookup will fail, because `foo.bar` has not been loaded yet and therefore the compiler starts to compile that category. Infinite loop achieved.

This isn't limited to phrases of the same category, any scenario that introduces circle dependencies will cause the same issues.

Timeline:
- 5.3: Track categories that are currently processed and throw an exception if a pending category is attempted to be loaded again (circular dependency).
- 6.x: Disallow `{lang}…{/lang}` inside phrases entirely, because they are evaluated at compile time rather than runtime and are subject to strange side effects. `{…|language}` is resolved at runtime and is not affected by any of these issues.

Gravatronics:
Close #3163

### DIFF
--- a/wcfsetup/install/files/lib/data/language/Language.class.php
+++ b/wcfsetup/install/files/lib/data/language/Language.class.php
@@ -217,7 +217,7 @@ class Language extends DatabaseObject {
 		$filename = WCF_DIR.'language/'.$this->languageID.'_'.$category.'.php';
 		if (!@file_exists($filename)) {
 			if (isset($this->categoriesBeingLoaded[$category])) {
-				throw new \LogicException("Cannot load category '{$category}' while it is already being loaded.");
+				throw new \LogicException("Circular dependency detected! Cannot load category '{$category}' while it is already being loaded.");
 			}
 			
 			if ($this->editor === null) {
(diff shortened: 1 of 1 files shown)


wcfsetup/install/files/lib/data/language/Language.class.php
```diff
@@ -210,6 +216,10 @@ protected function loadCategory($category) {
 		// search language file
 		$filename = WCF_DIR.'language/'.$this->languageID.'_'.$category.'.php';
 		if (!@file_exists($filename)) {
+			if (isset($this->categoriesBeingLoaded[$category])) {
+				throw new \LogicException("Cannot load category '{$category}' while it is already being loaded.");
```
dtdesign, line 220:
```suggestion
				throw new \LogicException("Circular dependency detected! Cannot load category '{$category}' while it is already being loaded.");
```

PR #3300, opened 2020-05-11
Fix typo in iOS platform detection

Fabii547:
`window.navigator.platform` outputs `MacIntel` on iOS / iPadOS when requesting the desktop version of a site.


dtdesign:
Thanks!

PR #3302, opened 2020-05-12
Validate file extension before validating size

Krymonota:
I just wasted some time compressing a file I wanted to upload because I was told it was too large. After compressing the file, I was told that the extension was invalid/not allowed. If a file can never be uploaded due to its extension, the user should be informed directly.


TimWolla:
Thanks.

PR #3146, opened 2020-01-31
Implement a dynamic event listener

Cyperghost:
#2952

PR #3317, opened 2020-05-20
Dynamic event listener with extra support for DBOActions

dtdesign:
This is a bit more streamlined implementation of the original proposal by @Cyperghost, aiming to reduce the call stack and to promote the use as a general purpose implementation.

Closes #3146

### DIFF
--- a/wcfsetup/install/files/lib/system/event/listener/AbstractEventListener.class.php
+++ b/wcfsetup/install/files/lib/system/event/listener/AbstractEventListener.class.php
@@ -1,0 +1,48 @@
+<?php
+namespace wcf\system\event\listener;
+use wcf\data\AbstractDatabaseObjectAction;
+
+/**
+ * Preferred implementation for event listeners that dynamically invokes methods based on a predictable
+ * naming schema derived from the event name. In addition, `AbstractDatabaseObjectAction` supports deep
+ * method invocations for the generic action events. However, when there is no specific method, it will
+ * attempt to invoke the regular method instead.
+ * 
+ * Examples:
+ * Regular classes
+ *   eventName: makeSnafucated
+ *   derived method: onMakeSnafucated()
+ * 
+ * Classes deriving from AbstractDatabaseObjectAction
+ *   eventName: initializeAction
+ *   actionName: makeSnafucated
+ *   derived method: onInitializeActionMakeSnafucated()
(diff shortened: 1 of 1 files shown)


wcfsetup/install/files/lib/system/event/listener/AbstractEventListener.class.php
```diff
@@ -0,0 +1,48 @@
...
+ * Examples:
+ * Regular classes
+ *   eventName: makeSnafucated
+ *   derived method: onMakeSnafucated()
+ * 
+ * Classes deriving from AbstractDatabaseObjectAction
+ *   eventName: initializeAction
+ *   actionName: makeSnafucated
+ *   derived method: onInitializeActionMakeSnafucated()
+ *   ATTENTION: If this method does not exist, it will attempt to invoke `onMakeSnafucated()` instead.
```
TimWolla, line 20:
```suggestion
 *   ATTENTION: If this method does not exist, it will attempt to invoke `onInitializeAction()` instead. 
```

wcfsetup/install/files/lib/system/event/listener/AbstractEventListener.class.php
```diff
@@ -0,0 +1,48 @@
...
+ * @copyright   2001-2020 WoltLab GmbH
+ * @license     GNU Lesser General Public License <http://opensource.org/licenses/lgpl-license.php>
+ * @package     WoltLabSuite\Core\System\Event\Listener
+ */
+abstract class AbstractEventListener implements IParameterizedEventListener {
+	/**
+	 * @inheritDoc
+	 */
+	public final function execute($eventObj, $className, $eventName, array &$parameters) {
+		static $genericDboActionNames = ['finalizeAction', 'initializeAction', 'validateAction'];
```
TimWolla, line 32: I would make this a class constant.

dtdesign: I thought about this, but stepped back from it. The current approach encapsulates the entire logic inside this method without exposing a (public, PHP 7.0) constant. This also keeps the class as clean as possible without interfering with derived classes, making it an easier drop-in replacement for existing code.

TimWolla: I don't particularly care either way.

PR #3340, opened 2020-05-29
Show note in pollVote.tpl if poll is public

### LINKED ISSUE #3258: Explicitly mark public polls
… for example, with a badge/label.

see https://community.woltlab.com/thread/277763-umfragen-klare-kennzeichnung-wenn-%C3%B6ffentlich/

TimWolla:
Resolves #3258

------

Not terribly happy that possibly three info texts are shown below the voting buttons. However it would at least be consistent with the existing notes, while a label at the top would not.

Bildschirmfoto vom 2020-05-29 15-55-02

Making the notes a list (not part of this PR) would improve readability, but is not exactly looking great, either:

image

Opinions?


dtdesign:
The poll UI isn't pretty and should get a face lift. I'll figure something out.

PR #3344, opened 2020-06-03
Allow 'current' function in templates in enterprise mode

earlyhost:


TimWolla:
Picked manually into 5.2, because switching the target branch on GitHub would have pulled in the entirety of master into 5.2.

Thanks.

PR #3346, opened 2020-06-03
Add doubleval and floatval to enterprise functions

Krymonota:
See https://community.woltlab.com/thread/284026-seit-steam-plugin-fehler-profil-integration-f%C3%BCr-steam-2020/


TimWolla:
Thanks.

PR #3347, opened 2020-06-03
Also show code collapse button at top of code box

### LINKED ISSUE #3235: Also show code collapse button at top of code box
After clicking the `Display More` button in a code box, it becomes a `Display Less` button shown at the very end of the code box. In such a case (for long code boxes only?), a second `Display Less` button should be shown at the top of the code box to make it easier to collapse a long code box.

See https://community.woltlab.com/thread/274586-weniger-anzeigen-in-der-codebox-auch-oben/

TimWolla:
Resolves #3235 

Example:

Bildschirmfoto vom 2020-06-03 18-24-27


TimWolla:
> I assume that keeping the HTML structure intact was not an option?

I believe I could re-introduce the intermediate `<div>` without issues, but I would still need to move the `overflow: hidden` as well as a largish number of other declarations around to make the `sticky` header work. So in all cases there would be some kind of breakage.

PR #3362, opened 2020-06-07
Support copying all user groups

### LINKED ISSUE #3238: Support copying non-other user groups
Currently, only user groups with `groupType == UserGroup::OTHER` can be copied. It should be at least possible to copy the `UserGroup::USERS` user group. 

see https://community.woltlab.com/thread/279218-standard-gruppen-kopieren/

Gravatronics:
Close #3238

### DIFF
--- a/wcfsetup/install/files/lib/data/user/group/UserGroupAction.class.php
+++ b/wcfsetup/install/files/lib/data/user/group/UserGroupAction.class.php
@@ -94,7 +94,7 @@ class UserGroupAction extends AbstractDatabaseObjectAction {
 		$this->readBoolean('copyUserGroupOptions');
 		
 		$this->groupEditor = $this->getSingleObject();
-		if (!$this->groupEditor->isAccessible() || $this->groupEditor->groupType != UserGroup::OTHER) {
+		if (!$this->groupEditor->canCopy()) {
 			throw new PermissionDeniedException();
 		}
 	}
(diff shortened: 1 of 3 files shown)


wcfsetup/install/files/lib/data/user/group/UserGroupAction.class.php
```diff
@@ -120,18 +120,25 @@ public function copy() {
 		
 		$optionValues = $statement->fetchMap('optionID', 'optionValue');
 		
-		$groupAction = new UserGroupAction([], 'create', [
+		$groupType = $this->groupEditor->groupType;
+		// When copying special user groups of which only one may exist,
+		// change the group type to 'other'.
+		if (in_array($groupType, [UserGroup::EVERYONE, UserGroup::GUESTS, UserGroup::USERS, UserGroup::OWNER])) {
+			$groupType = UserGroup::OTHER;
+		}
```
dtdesign, line 128:
```suggestion
```

(See the next change for context)

wcfsetup/install/files/lib/data/user/group/UserGroupAction.class.php
```diff
@@ -120,18 +120,25 @@ public function copy() {
...
+		/** @var UserGroup $group */
+		$group = (new UserGroupAction([], 'create', [
 			'data' => [
 				'groupName' => $this->groupEditor->groupName,
 				'groupDescription' => $this->groupEditor->groupDescription,
 				'priority' => $this->groupEditor->priority,
 				'userOnlineMarking' => $this->groupEditor->userOnlineMarking,
-				'showOnTeamPage' => $this->groupEditor->showOnTeamPage
+				'showOnTeamPage' => $this->groupEditor->showOnTeamPage,
+				'groupType' => $groupType,
```
dtdesign, line 138:
```suggestion
				// Prevent the duplication of built-in groups by forcing new groups to be of the "other" type.
				'groupType' => UserGroup::OTHER,
```

Gravatronics: I used my approach to keep the group type the same for moderated user groups (even if the leaders would are not copied).

dtdesign: Good catch, I didn't thought about those. Okay, in this case, please disregard my comments.

PR #3374, opened 2020-06-23
Deduplicate Exceptions in ExceptionLogView

### LINKED ISSUE #3044: Deduplication of similar exceptions on exception log page
Deduplicate similar Exceptions (e.g. by collapsing Exceptions with an identical message + stack trace). Possibly update the calculation of the exceptionID: https://github.com/WoltLab/WCF/blob/489106fa026f06563971108cd9dd59bd98950b9e/wcfsetup/install/files/lib/core.functions.php#L151

TimWolla:
image

-------

Resolves #3044

### DIFF
--- a/wcfsetup/install/files/lib/acp/page/ExceptionLogViewPage.class.php
+++ b/wcfsetup/install/files/lib/acp/page/ExceptionLogViewPage.class.php
@@ -141,15 +141,22 @@ class ExceptionLogViewPage extends MultipleLinkPage {
 		$this->calculateNumberOfPages();
 		
 		$i = 0;
-		
+		$seenHashes = [];
 		foreach ($this->exceptions as $key => $val) {
 			$i++;
+			
+			$parsed = ExceptionLogUtil::parseException($val);
+			if (isset($seenHashes[$parsed['stackHash']])) {
+				$parsed['collapsed'] = true;
+			}
+			$seenHashes[$parsed['stackHash']] = true;
+			
 			if ($i < $this->startIndex || $i > $this->endIndex) {
 				unset($this->exceptions[$key]);
 				continue;
 			}
 			try {
-				$this->exceptions[$key] = ExceptionLogUtil::parseException($val);
+				$this->exceptions[$key] = $parsed;
 			}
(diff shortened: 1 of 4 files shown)


wcfsetup/install/files/lib/acp/page/ExceptionLogViewPage.class.php
```diff
@@ -141,15 +141,22 @@ public function readData() {
...
-		
+		$seenHashes = [];
 		foreach ($this->exceptions as $key => $val) {
 			$i++;
+			
+			$parsed = ExceptionLogUtil::parseException($val);
+			if (isset($seenHashes[$parsed['stackHash']])) {
+				$parsed['collapsed'] = true;
+			}
+			$seenHashes[$parsed['stackHash']] = true;
```
Gravatronics, line 152: Did you intentionally put this before the page filtering in the following lines?

TimWolla: Yes, because I want to deduplicate across pages. If you have a repeated error you don't need to see it on every single page of the log, just once is sufficient.

Gravatronics: I was just wondering because this approach would imply a sequential reading of all pages without skipping any page. But this isn't a real concern as the relevant exceptions can easily be de-collapsed.

TimWolla: This was already happening previously. In fact the entire log file was always read into memory and then parsed into an array. If you have megabytes of logs you have bigger issues than a broken log viewer and should probably use FTP anyway.

Gravatronics: I meant literally reading/accessing the rendered pages by the user not reading the file via PHP.

TimWolla: Ah, yes. I guess most admins will either read the log sequentially or use the exception ID search. And if not decollapsing is readily available.

Issue #3315, opened 2020-05-20
Investigate whether <figure> + <figcaption> are usuable within mediaBBCodeTag.tpl


TimWolla:
The `<figure>` HTML element would be the semantically correct solution for that template.

joshuaruesweg:
Unfortunately this is not possible, otherwise invalid HTML will be generated, as the element can be located within <p> tags (which is also a problem in the current implementation).

PR #3397, opened 2020-06-26
Limit media images' caption size to image width

### LINKED ISSUE #3228: Limit media images' caption size to image width
… causing line breaks to occur instead of the caption being wider than the image.

https://community.woltlab.com/thread/270080-fehlender-zeilenumbruch-bei-bildunterschrift/

TimWolla:
The main difference is that the `.mediaBBCode` will now be rendered like a block element (seen in the last post on the page). But according to `bbcode.xml` the `wsm` BBCode already is `<isBlockElement>1</isBlockElement>`. I am a bit confused why it didn't previously render as such and why it is a valid child of a `<p>` element. **This probably warrants filing a follow up issue.**

<details><summary>Before (click me)</summary>

before

</details>
<details><summary>After (click me)</summary>

after

</details>

-----

Resolves #3228

### DIFF
--- a/com.woltlab.wcf/templates/mediaBBCodeTag.tpl
+++ b/com.woltlab.wcf/templates/mediaBBCodeTag.tpl
@@ -8,11 +8,13 @@
 	
 	{if $media->caption}
 		<span class="mediaBBCodeCaption">
-			{if $media->captionEnableHtml}
-				{@$media->caption}
-			{else}
-				{$media->caption}
-			{/if}
+			<span class="mediaBBCodeCaptionAlignment">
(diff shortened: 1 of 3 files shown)


TimWolla:
> But according to `bbcode.xml` the `wsm` BBCode already is `<isBlockElement>1</isBlockElement>`. I am a bit confused why it didn't previously render as such and why it is a valid child of a `<p>` element.

Semirelated: https://github.com/WoltLab/WCF/issues/3315#issuecomment-647498485

TimWolla:
> **This probably warrants filing a follow up issue.**

The follow up for that is in #3410.

PR #3403, opened 2020-06-27
Support replacing existing media files

### LINKED ISSUE #2922: Option to replace the file of an existing media
n/t

Gravatronics:
Close #2922

I am open for better solutions for https://github.com/WoltLab/WCF/pull/3403/files#diff-b293f4e4dbf3acf0a27e7e999a091c9aR191-R195.

### DIFF
--- a/wcfsetup/install/files/js/WoltLabSuite/Core/Media/Manager/Base.js
+++ b/wcfsetup/install/files/js/WoltLabSuite/Core/Media/Manager/Base.js
@@ -288,10 +288,28 @@ define(
 			var listItem = this._listItems.get(~~media.mediaID);
 			var p = elByClass('mediaTitle', listItem)[0];
 			if (media.isMultilingual) {
-				p.textContent = media.title[LANGUAGE_ID] || media.filename;
+				if (media.title && media.title[LANGUAGE_ID]) {
+					p.textContent = media.title[LANGUAGE_ID];
+				}
+				else {
+					p.textContent = media.filename;
+				}
 			}
 			else {
-				p.textContent = media.title[media.languageID] || media.filename;
+				if (media.title && media.title[media.languageID]) {
+					p.textContent = media.title[media.languageID];
+				}
+				else {
+					p.textContent = media.filename;
+				}
+			}
+			
+			var thumbnail = elByClass('mediaThumbnail', listItem)[0];
+			thumbnail.innerHTML = media.elementTag;
+			// Bust browser cache by adding additional parameter.
+			var imgs = elByTag('img', thumbnail);
+			if (imgs.length) {
+				imgs[0].src += '&refresh=1';
 			}
 		},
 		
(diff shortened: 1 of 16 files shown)


wcfsetup/install/files/lib/page/MediaPage.class.php
```diff
@@ -92,9 +92,9 @@ public function readData() {
 			'filesize' => $filesize,
 			'showInline' => in_array($mimeType, self::$inlineMimeTypes),
 			'enableRangeSupport' => $this->thumbnail ? true : false,
-			'lastModificationTime' => $this->media->uploadTime,
+			'lastModificationTime' => $this->media->fileUpdateTime ?? $this->media->uploadTime,
 			'expirationDate' => TIME_NOW + 31536000,
-			'maxAge' => 31536000
+			'maxAge' => 31536000,
```
TimWolla, line 97: This long of a max age is dangerous if files are able to be replaced.

Likewise the ETag must change after replacement as well.

wcfsetup/install/files/js/WoltLabSuite/Core/Media/Manager/Base.js
```diff
@@ -288,10 +288,28 @@ define(
...
+					p.textContent = media.filename;
+				}
+			}
+			
+			var thumbnail = elByClass('mediaThumbnail', listItem)[0];
+			thumbnail.innerHTML = media.elementTag;
+			// Bust browser cache by adding additional parameter.
+			var imgs = elByTag('img', thumbnail);
+			if (imgs.length) {
+				imgs[0].src += '&refresh=1';
```
TimWolla, line 312:
```suggestion
				imgs[0].src += '&refresh=' + Date.now();
```

wcfsetup/install/files/lib/data/media/MediaAction.class.php
```diff
@@ -182,6 +188,11 @@ protected function getMediaData($media) {
 				'id' => $media->userID,
 				'title' => $media->username
 			]) : '',
+			// TODO: find better solution
```
TimWolla, line 191: TODO. This is really ugly.

wcfsetup/install/files/lib/data/media/MediaAction.class.php
```diff
@@ -675,4 +686,87 @@ public function setCategory() {
...
+	}
+	
+	/**
+	 * Replaces the actual file of a media file.
+	 * 
+	 * @return      array
+	 * @since       5.3
+	 */
+	public function replaceFile() {
+		$saveStrategy = new DefaultUploadFileSaveStrategy(self::class, [
```
TimWolla, line 713:
```suggestion
		$saveStrategy = new DefaultUploadFileSaveStrategy(static::class, [
```

?

wcfsetup/install/files/lib/data/media/MediaAction.class.php
Gravatronics: That is why I also explicitly added to the message of this pull request that I am open for better solutions :)

May-be we should add a method to UserProfile which relies on an actual template with `{user object=$userProfile}`.

dtdesign: I'm not happy about it, but adding bogus extra methods is only moving the issue somewhere else. Moving it (into an actual template) is only masquerading the issue, not solving it.

For the lack of a truly better solution: Keep it.

PR #3416, opened 2020-06-30
Allow specifying allowed file extensions for form builder's UploadFormField

### LINKED ISSUE #3414: Allow specifying allowed file extensions for form builder's UploadFormField
… making it easier for the user, because the file selection already is filtered.

TimWolla:
I'm intentionally using a fairly elaborate wording “acceptable” instead of “valid” to make it clearer that no validation is happening automatically.

Resolves #3414

### DIFF
--- a/com.woltlab.wcf/templates/uploadFieldComponent.tpl
+++ b/com.woltlab.wcf/templates/uploadFieldComponent.tpl
@@ -42,7 +42,12 @@
 		new Upload("{$uploadFieldId}UploadButtonDiv", "{$uploadFieldId}uploadFileList", {
 			internalId: '{$uploadField->getInternalId()}',
 			{if $uploadField->getMaxFiles()}maxFiles: {$uploadField->getMaxFiles()},{/if}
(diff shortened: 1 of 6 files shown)


TimWolla:
> The file type should then be checked during upload. As it is implemented now, many developers will forego a direct check of the file, because they think that the whole thing is already validated.

Contrary to the image validation a proper validation is not possible here in the general case. `setAcceptableFiles(['video/*'])` would be a valid call (e.g. for use within the video upload of the gallery), but you can't really validate `video/*` against the uploaded file. Possibly a valid video will not be detected by the installed magic db or a video is valid, but still contains garbage you don't want to have depending on the exact threat model.

On the other side using `setAcceptableFiles(['.xml'])` would allow anything ending `*.xml`, no matter the contents. So you'll need to add proper validation yourself anyway. Artificially restricting the file extension just serves to annoy the end user, without any measurable security benefit.

While I'm all for a “safe by default” API design this is not going to work out here. The comment in combination with the method name IMO is sufficiently clear that one must not rely on it to get exactly the requested files.

PR #3417, opened 2020-06-30
Add option for HTML based email signature

### LINKED ISSUE #3224: Support HTML in MAIL_SIGNATURE
see https://community.woltlab.com/thread/261572-e-mail-in-signatur-html-verwenden-k%C3%B6nnen/

TimWolla:
Resolves #3224

### DIFF
--- a/com.woltlab.wcf/templates/email_html.tpl
+++ b/com.woltlab.wcf/templates/email_html.tpl
@@ -155,11 +155,13 @@
 	{capture assign='footer'}
 	{hascontent}
 	<span style="font-size: 0;">-- <br></span>
-	{content}
-	{@MAIL_SIGNATURE|language|newlineToBreak}
-	{/content}
-	{/hascontent}
-	{/capture}
+{content}
+{if MAIL_SIGNATURE_HTML|language}
+{@MAIL_SIGNATURE_HTML|language}
+{else}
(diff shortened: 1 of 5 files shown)


com.woltlab.wcf/templates/email_html.tpl
```diff
@@ -155,11 +155,13 @@
...
-	{/content}
-	{/hascontent}
-	{/capture}
+{content}
+{if MAIL_SIGNATURE_HTML|language}
+{@MAIL_SIGNATURE_HTML|language}
+{else}
+{@MAIL_SIGNATURE|language|newlineToBreak}
+{/if}
+{/content}{/hascontent}{/capture}
```
dtdesign, line 164:
```suggestion
	{content}
	{if MAIL_SIGNATURE_HTML|language}
	{@MAIL_SIGNATURE_HTML|language}
	{else}
	{@MAIL_SIGNATURE|language|newlineToBreak}
	{/if}
	{/content}{/hascontent}{/capture}
```

TimWolla: This *was* intentional to get rid of the initial tabs in case significant whitespace was used (e.g. `<pre>`). But that doesn't actually make sense, because the entirety of the signature is generated by the constant, I probably was confused by the `{if}` logic.

Thanks.

PR #3420, opened 2020-07-01
Set display from browser to standalone

TitusKirch:
In my opinion this change increases the UX and UI on mobile devices (if you have saved a page on the desktop). It may not quite fulfil the wish of this potential customer (https://community.woltlab.com/thread/284454-pwa-als-button-speichern-app-simulieren/), but it is at least a step in the direction of his wish.

> Opens the web application to look and feel like a standalone native application. This can include the application having a different window, its own icon in the application launcher, etc. In this mode, the user agent will exclude standard browser UI elements such as an URL bar, but can include other system UI elements such as a status bar and/or system back button.
The fallback display mode for standalone is minimal-ui.

Source: https://www.w3.org/TR/appmanifest/#dom-displaymodetype-standalone

### DIFF
--- a/wcfsetup/install/files/images/favicon/default.manifest.json
+++ b/wcfsetup/install/files/images/favicon/default.manifest.json
@@ -14,5 +14,5 @@
     ],
     "theme_color": "#ffffff",
     "background_color": "#ffffff",
-    "display": "browser"
+    "display": "standalone"
 }


TimWolla:
Can you add a screenshot of the old and the new behavior, please?

TitusKirch:
> Can you add a screenshot of the old and the new behavior, please?

Sure


Here I have two pages that run with WCF/WSC on the desktop, one is the page of WoltLab (with `display: browser`) and one is my page (with `display: standalone`)


**WoltLab (with `display: browser`)**


**My page (with `display: standalone`)**

TimWolla:
Thanks. I must admit I'm not sure if I'm happy with the missing URL bar. It makes it hard to e.g. share stuff, because the software does not contain any native sharing buttons.

So there *are* some drawbacks without any real benefit, apart from getting getting a few additional pixels from the URL bar (which will automatically be hidden when scrolling, at least on Google Chrome).

TitusKirch:
> I must admit I'm not sure if I'm happy with the missing URL bar. It makes it hard to e.g. share stuff, because the software does not contain any native sharing buttons.
In the end you would use it like a native app and you don't have a url bar in it (of course everybody has different opinions about it). Concerning sharing: That's what the share buttons are for, e.g. at the end of an article, and it works great.

Maybe it would be a compromise to give the website operator the possibility to set these per style (apparently there will be one manifest file per style anyway).

dtdesign:
I'm adding this simply to give it a try and see how it works out. If there is a significant rejection rate, we can still revert to the previous setting. We'll not add an option for this either way.

PR #3428, opened 2020-07-02
Fix Url::is()

### LINKED ISSUE #3391: Url validation fails on non-URLs
Due to the nature of how `parse_url()` works, `Url::is()` (and other places, where this function is used) will never return the correct result on non-URLs. Every non-URL is considered as `path`, so the return value will never be false or empty.

Instead, `filter_var($url, FILTER_VALIDATE_URL)` or similar should be used. Alternatively, use the `PHP_URL_HOST` component parameter in combination with `parse_url()` (which leads to `null` on non-URLs, not `false`!).

---

```php
<?php
var_dump(parse_url('foo'));

/*
array(1) {
  ["path"]=>
  string(3) "foo"
}
*/
```

TimWolla:
`parse_url()` cannot be used to validate an URL, because it will accept
roughly everything. In fact this is documented in the parse_url() docs:

> This function is not meant to validate the given URL, it only breaks
> it up into the above listed parts. Partial URLs are also accepted,
> parse_url() tries its best to parse them correctly.

Fixes #3391

/cc @SoftCreatR @TitusKirch

### DIFF
--- a/wcfsetup/install/files/lib/util/Url.class.php
+++ b/wcfsetup/install/files/lib/util/Url.class.php
@@ -38,13 +38,15 @@ final class Url implements \ArrayAccess {
 	];
 	
 	/**
-	 * Tests if provided url appears to be an url and can be processed by `parse_url()`.
+	 * Tests if provided $url appears to be an URL.
+	 * 
+	 * This method is a wrapper around filter_var with FILTER_VALIDATE_URL.
 	 * 
 	 * @param       string          $url
 	 * @return      boolean
 	 */
 	public static function is($url) {
-		return parse_url($url) !== false;
+		return filter_var($url, FILTER_VALIDATE_URL) !== false;
 	}
 	
 	/**


SoftCreatR:
Maybe introduce an optional parameter for idn domains? Or directly implement the encoding mechanism? Btw. it's not just the is method. The next method also uses parse_url and performs an invalid check.

TimWolla:
> Maybe introduce an optional parameter for idn domains? Or directly implement the encoding mechanism?

Just performing domain punycoding would not be sufficient. If you do that then by extension you must also percent-encode any non-ASCII characters elsewhere in the URL (e.g. the path or query string). And you can't safely encode something that you don't properly understand. So the method might receive something that is clearly not an URL, but after encoding it looks like one, passing the `Url::is()` check, possibly introducing a security issue. You also would not really benefit from encoding just for use within the check, because later processing might assume to only see valid URLs which any URLs with non-ASCII characters are not.

In short: Perform the proper encoding for your use case at a central location somewhere in your code. What the proper encoding would be depends on your use case. This is deep in business logic territory and cannot be handled generically.

> The next method also uses parse_url and performs an invalid check.

I'm assuming you are referring to `Url::parse()` here? The usage of `parse_url` within there is correct:

[php.net](https://www.php.net/manual/en/function.parse-url.php):
> On seriously malformed URLs, parse_url() may return FALSE.

SoftCreatR:
> On seriously malformed URLs, parse_url() may return FALSE.

I know that, but what does "seriously malformed" mean? I've tested it and I wasn't able to get false returned.

TimWolla:
> I know that, but what does "seriously malformed" mean? I've tested it and I wasn't able to get false returned.

While I've not written the documentation I've [simply checked PHP's tests](https://github.com/php/php-src/blob/49a4e695845bf55e059e7f88e54b1111fe284223/ext/standard/tests/url/bug55399.phpt):

```
php > var_dump(parse_url(':'));
php shell code:1:
bool(false)
```

PR #3435, opened 2020-07-03
Remove unused Packageservers for 5.2

xopez:
As context this post: https://community.woltlab.com/thread/284532-update-server-paket-server/
It seems that only the current version of update.woltlab.com is needed as packageserver. So removing it.


dtdesign:
Thanks!

PR #3444, opened 2020-07-10
Synchronize attachments across instances

dtdesign:
Attachment handlers using the same `tmpHash` are now synchronized to share the same uploaded attachments, including removal. In addition, the option to enable thumbnails for attachments is now force enabled and cannot be turned off.


com.woltlab.wcf/templates/__wysiwygAttachmentFormField.tpl
```diff
@@ -1,7 +1,6 @@
 <ul id="{@$field->getPrefixedID()}_attachmentList" {*
 	*}class="formAttachmentList"{*
 	*}{if !$field->getAttachmentHandler()->getAttachmentList()|count} style="display: none"{/if} {*
```
Gravatronics, line 3: Additional trailing whitespace (same below).

PR #3445, opened 2020-07-11
Save changes to the sitemap objects in the registry

### LINKED ISSUE #3376: Store sitemaps in own database table
Currently, sitemaps are stored in "wcf1_object_type". Changes to the Sitemaps by the user are stored there as "additionalData". This value is potentially overwritten during a package update.

ref https://community.woltlab.com/thread/283340-sitemapreaktivierung-nach-upgrade-auf-5-2/

joshuaruesweg:
Since changes to the sitemap are saved directly in the object type, there is the problem that when you update the WCF, these changes are potentially overwritten again. This commit fixes this problem by saving changes in the registry.
Fixes #3376

### DIFF
--- a/wcfsetup/install/files/lib/system/worker/SitemapRebuildWorker.class.php
+++ b/wcfsetup/install/files/lib/system/worker/SitemapRebuildWorker.class.php
@@ -31,6 +31,10 @@ class SitemapRebuildWorker extends AbstractRebuildDataWorker {
 	 */
 	const SITEMAP_OBJECT_LIMIT = 50000;
 	
+	/**
+	 * Prefix for stored data in the registry.
+	 * @since 5.3
+	 */
 	const REGISTRY_PREFIX = 'sitemapData_';
 	
 	/**
@@ -430,6 +434,7 @@ class SitemapRebuildWorker extends AbstractRebuildDataWorker {
 	 * Reads the columns changed by the user for this sitemap object from the registry and modifies the object accordingly.
 	 * 
 	 * @param       ObjectType      $object
+	 * @since       5.3
 	 */
 	public static function prepareSitemapObject(ObjectType &$object) {
 		$sitemapData = RegistryHandler::getInstance()->get('com.woltlab.wcf', self::REGISTRY_PREFIX . $object->objectType);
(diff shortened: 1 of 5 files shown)


joshuaruesweg:
Writing a script to convert existing sitemaps into the registry is of course possible and relatively simple. Moving all parameters into the registry without fallback is not possible, because plugins always provide entries and I would have to listen with an EL (or directly in the class), when such objects are registered and would have to create registry entries there. Which I think is unnecessary.  Yes, it is currently not nice, but it is quite practicable to use the values from the object-type as fallback. In general, very few users will change anything there.

wcfsetup/install/files/lib/system/worker/SitemapRebuildWorker.class.php
```diff
@@ -421,4 +429,25 @@ private function changeUserToGuest() {
...
 		WCF::getSession()->changeUser($this->actualUser, true);
 	}
+	
+	/**
+	 * Reads the columns changed by the user for this sitemap object from the registry and modifies the object accordingly.
+	 * 
+	 * @param       ObjectType      $object
+	 * @since       5.3
+	 */
+	public static function prepareSitemapObject(ObjectType &$object) {
```
TimWolla, line 439:
```suggestion
	public static function prepareSitemapObject(ObjectType $object) {
```
Objects are always passed reference by value. No need to pass the reference by reference.

wcfsetup/install/files/lib/system/worker/SitemapRebuildWorker.class.php
```diff
@@ -421,4 +429,25 @@ private function changeUserToGuest() {
...
+		$sitemapData = RegistryHandler::getInstance()->get('com.woltlab.wcf', self::REGISTRY_PREFIX . $object->objectType);
+		
+		if ($sitemapData !== null) {
+			$sitemapData = @unserialize($sitemapData);
+			
+			if (is_array($sitemapData)) {
+				$object->priority = $sitemapData['priority'];
+				$object->changeFreq = $sitemapData['changeFreq'];
+				$object->rebuildTime = $sitemapData['rebuildTime'];
+				$object->isDisabled = $sitemapData['isDisabled'];
```
TimWolla, line 449: This is unsafe, because it adds new public properties to the ObjectType object, instead of syncing the internal `data` member.

joshuaruesweg: Yes, but the data attribute is `protected`. I could of course create a new object with the data, but that would also be _wrong_ somehow. I think the current solution is sufficient, especially because you only access the data via the Magic-Getter.

dtdesign: I don't really see an issue here, the implicit properties are not persistent and are valid for the lifetime of the request only. Also, these values are in fact correct and thus will not break any code that for whatever reason accesses them.

PR #3447, opened 2020-07-14
Add GuzzleHttp

TimWolla:
I've carefully made sure to reimplement all the various details of HTTPRequest. Specifically I tested all of the following:

- Retrieving a maximum length of the body works properly.
- The response header arrays are identical.
- Any Exception thrown is wrapped in an appropriate legacy Exception with a proper message and code.
- Authenticated requests send the same credentials.
- Redirects are being limited with a proper Exception.
- Requests with a body look identical.

Additionally I made sure that I was able to install a package, including dependencies using the package servers.

All in all I don't expect any surprises and if there's anything then the flexibility of Guzzle should allow to easily fix those.

### DIFF
--- a/wcfsetup/install/files/lib/system/io/HttpFactory.class.php
+++ b/wcfsetup/install/files/lib/system/io/HttpFactory.class.php
@@ -1,0 +1,52 @@
+<?php
+namespace wcf\system\io;
+use GuzzleHttp\Client;
+use GuzzleHttp\ClientInterface;
+
+/**
+ * Factory for HTTP Clients.
+ * 
+ * As of right now the factory returns objects that implement Guzzle's ClientInterface.
+ * Even if Guzzle will remain the HTTP client of choice for the foreseeable future you should
+ * strive to use PSR-7 objects instead of relying on Guzzle's shortcuts for best compatibility.
+ * 
+ * @author	Tim Duesterhus
+ * @copyright	2001-2020 WoltLab GmbH
+ * @license	GNU Lesser General Public License <http://opensource.org/licenses/lgpl-license.php>
+ * @package	WoltLabSuite\Core\System\Io
+ * @since	5.3
+ */
+final class HttpFactory {
+	/**
+	 * @var Client
+	 */
+	private $defaultClient;
+	
+	/**
+	 * Returns a reference to the default HTTP client.
+	 * 
+	 * @return ClientInterface
+	 */
+	public static function getDefaultClient() {
+		if (self::$defaultClient === null) {
+			self::$defaultClient = static::makeClient();
+		}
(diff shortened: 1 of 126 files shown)


Hanashi:
I believe there could be problems after this change if a developer calls `getReply` twice before the change.

TimWolla:
> I believe there could be problems after this change if a developer calls `getReply` twice before this change.

While I did not actually test this I am fairly sure that the implementation is correct, because I explicitly read the body into a string. I'll make sure to test this before merging, though.

TimWolla:
> I'll make sure to test this before merging, though.

I just have verified that calling `getReply()` multiple times will always return identical information.

wcfsetup/install/files/lib/system/io/HttpFactory.class.php
```diff
@@ -0,0 +1,66 @@
...
+	 * 
+	 * @return ClientInterface
+	 * @see Client
+	 */
+	public static function makeClient(array $options = []) {
+		return new Client(array_merge([
+			'proxy' => PROXY_SERVER_HTTP,
+			'headers' => [
+				'user-agent' => self::getDefaultUserAgent(),
+			]
```
Gravatronics, line 63:
```suggestion
			],
```

wcfsetup/install/files/lib/util/HTTPRequest.class.php
```diff
@@ -206,385 +152,102 @@ public function __construct($url, array $options = [], $postParameters = [], arr
...
+			'allow_redirects' => [
+				'max' => $this->options['maxDepth'],
+				'track_redirects' => true,
+				'on_redirect' => $redirectHandler,
+			],
+		];
+		if (isset($this->options['auth'])) {
+			$options['auth'] = [
+				$this->options['auth']['username'],
+				$this->options['auth']['password']
```
Gravatronics, line 179:
```suggestion
				$this->options['auth']['password'],
```

wcfsetup/install/files/lib/util/HTTPRequest.class.php
```diff
@@ -595,10 +258,40 @@ private function parseReply() {
...
+		
 		return [
-			'statusCode' => $this->statusCode, 
-			'headers' => $this->legacyHeaders,
-			'httpHeaders' => $this->replyHeaders,
+			'statusCode' => (string) $this->response->getStatusCode(), 
+			'headers' => $legacyHeaders,
+			'httpHeaders' => $headers,
 			'body' => $this->replyBody,
 			'url' => $this->url
```
Gravatronics, line 296:
```suggestion
			'url' => $this->url,
```

PR #3455, opened 2020-07-20
Performance improvements of the search for packages

dtdesign:
The previous implementation requested the database to insert a row and immediately retrieve its id. This causes the performance scaling to be linear due to the sequential write and read operations. The changes will write new rows in bulk and retrieve them in a secondary query, yielding a significant speedup - more than 2x with MariaDB on NVMe.

### DIFF
--- a/wcfsetup/install/files/acp/js/WCF.ACP.js
+++ b/wcfsetup/install/files/acp/js/WCF.ACP.js
@@ -1003,8 +1003,12 @@ WCF.ACP.Package.Update.Search = Class.extend({
 	
 	/**
 	 * Handles clicks on the search button.
+	 * 
+	 * @param {Event} event
 	 */
-	_click: function() {
+	_click: function(event) {
+		event.preventDefault();
+		
 		if (this._dialog === null) {
 			new WCF.Action.Proxy({
 				autoSend: true,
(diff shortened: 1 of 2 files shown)


dtdesign:
I could wrap the set of queries after the `package_update_version` rows into one transaction, although I'm not sure about the actual gains. This part is heavily optimized for write performance and I have never seen the data to show any sort of inconsistencies over a period of like 10 years?

TimWolla:
> I have never seen the data to show any sort of inconsistencies over a period of like 10 years?

I don't expect it to make any meaningful difference, because the transaction contents are very simple, most likely always succeeding.

But I'd like to always clean up a bit of technical debt when I need to revisit old code anyway. Using one large transaction would be the “most correct” solution here, because updating the list of packages is a single logical change to the database. Of course if it comes with a significant (performance) drawback the current solution is absolutely fine.

dtdesign:
I've decided to keep it the way it currently is. The reason is simple: I don't trust MySQL not to mess up the performance by having SELECTs in between INSERTs. The current transactions span pure INSERTs on the same table and MySQL reproducibly yields an excellent overall performance.

We all know that MySQL's query planer cannot be trusted 😅

PR #3448, opened 2020-07-15
Serve Google Fonts locally

TimWolla:
Resulting CSS compiles down to something like:

```
@font-face{font-family:'Slackey';font-style:normal;font-weight:400;src:local('Slackey Regular'), local('Slackey-Regular'), url("../font/getFont.php?family=Slackey&filename=Slackey-Regular.woff2") format('woff2')}
```

Bildschirmfoto vom 2020-07-15 12-15-41
Bildschirmfoto vom 2020-07-15 12-15-34
Bildschirmfoto vom 2020-07-15 12-15-50

Blocked on #3447, will be a draft PR until #3447 is merged.

-------

Resolves #3394

PR #3456, opened 2020-07-21
Clean up asset handling of styles

### LINKED ISSUE #3451: Clean up asset handling of styles
As already mentioned in #3438 the whole file handling for styles is pretty unclean and in order for a refactoring. We currently have the following:

1. The style's logo. Searched in the style's image folder with a fallback to the global `images/` folder. Not really cleaned up when deleting a style.
2. The style's preview. Stored in the global `images/` folder.
3. The style's default coverPhoto. Stored in the `images/coverPhotos/` folder and named using the styleID.
4. The style's favicon. Stored in the `images/favicon/` folder and named using the styleID.

On top of that the style's images folder does not necessarily exist and has an arbitrary name. During a style import if the style's image folder is already in use it will automatically be adjusted by appending a number.

So all in all the files for a style are all over the place. Going forward I propose the following:

1. _Always_ create a style specific asset folder.
2. Do not allow an arbitrary name for the style specific asset folder. Use the `styleID`.
3. Store all assets within this asset folder:
    - The logo.
    - The mobile logo.
    - The preview image.
    - The default coverphoto.
    - The favicon.
    - The Android / Windows manifests.
    - Any images uploaded as part of this very issue.
    - **Maybe:** The style's compiled CSS.

During style export the checkbox to include the style's image folder will be removed. The folder will *always* be exported.

_Originally posted by @TimWolla in https://github.com/WoltLab/WCF/issues/3364#issuecomment-660115162_

TimWolla:
I've adjusted the handling of:

- Style Preview
- Style Logo
- Style Coverphoto
- Style Favicon
- Style copying
- Style export
- Style import

I believe that I've adjusted all the necessary places. I've tested that uploading and deleting all types of files works when both adding as well as editing styles. I've verified that the export now exports the full image folder, including the Android Manifest and Windows Browser Config. I've also verified that importing “old” styles from the Plugin-Store works fine.

I've completely removed the legacy JavaScript files and PHP methods. While this technically is a backwards compatibility break I don't expect and programmatic third party access there and making them 100% compatible coming with quite a bit of work, because they make a large number of assumptions that no longer hold true.

----------------------

Resolves #3451


TimWolla:
I believe I now made all the necessary changes to the asset handling. I did not yet adjust the previous remarks about the code style. I'd like to first have a functional test to make sure I did not break anything or miss anything while performing the adjustments. If everything is working properly we can think of the best way forward regarding code cleanup.

TimWolla:
Merged master due to the recent merge of #3448.

wcfsetup/install/files/lib/acp/form/StyleEditForm.class.php
```diff
@@ -170,6 +170,18 @@ public function readData() {
 					$file,
 				]);
 			}
+			if ($this->style->hasFavicon) {
+				foreach (['png', 'jpg', 'gif'] as $extension) {
+					if (file_exists($this->style->getAssetPath()."favicon.template.".$extension)) {
+						$file = new UploadFile($this->style->getAssetPath()."favicon.template.".$extension, "favicon.template.".$extension, true, true, false);
```
dtdesign, line 176:
```suggestion
					$filename = "favicon.template.".$extension;
					if (file_exists($this->style->getAssetPath().$filename)) {
						$file = new UploadFile($this->style->getAssetPath().$filename, $filename, true, true, false);
```

wcfsetup/install/files/lib/data/style/Style.class.php
```diff
@@ -232,14 +245,20 @@ public function getCoverPhoto() {
...
+		return WCF_DIR . 'images/coverPhotos/default.jpg';
 	}
 	
 	/**
 	 * @return string
 	 * @since 5.2
 	 */
 	public function getCoverPhotoUrl() {
+		if ($this->coverPhotoExtension) {
+			return FileUtil::getRelativePath(WCF_DIR, $this->getAssetPath()).'coverPhoto.'.$this->coverPhotoExtension;
```
dtdesign, line 260: This is incorrect, `WCF::getPath()` returns an absolute URL to the specified app (implicit: the core), but your code only yields a path relative to the Core's directory.

TimWolla: You are entirely correct here, of course.

TimWolla:
Merged master for #3467

PR #3460, opened 2020-07-24
Ship Open Sans by default

### LINKED ISSUE #3458: Bake Open Sans into default installation
This is a follow up issue for #3394. After #3448 was merged the font will no longer be embedded from Google. As Open Sans is being used by the default style it will need to be shipped during set-up.

TimWolla:
Resolves #3458


wcfsetup/install/files/lib/system/package/plugin/StylePackageInstallationPlugin.class.php
```diff
@@ -32,7 +32,7 @@ public function install() {
 		$styleEditor = StyleHandler::getInstance()->getStyleByName($this->installation->getPackageName(), false);
 		
 		// import style
-		$style = StyleEditor::import($filename, $this->installation->getPackageID(), $styleEditor);
+		$style = StyleEditor::import($filename, $this->installation->getPackageID(), $styleEditor, !PACKAGE_ID);
```
TimWolla, line 35: Not sure whether font download should ever run during package installation. For now I'm only excluding it during WCFSetup.

PR #3462, opened 2020-07-26
Add detection for Chromium based Edge browser

SoftCreatR:
See https://community.woltlab.com/thread/284939-edge-cromium-und-edge-f%C3%BCr-android-werden-das-chrome-erkannt/?postID=1811803#post1811803

### DIFF
--- a/wcfsetup/install/files/lib/data/user/online/UserOnline.class.php
+++ b/wcfsetup/install/files/lib/data/user/online/UserOnline.class.php
@@ -215,8 +215,13 @@ class UserOnline extends UserProfile {
 		}
 		
 		// edge
-		if (preg_match('~edge/(\d{2}\.\d+)~i', $this->userAgent, $match)) {
+		if (preg_match('~edge?/([\d\.]+)~i', $this->userAgent, $match)) {
 			return 'Microsoft Edge '.$match[1];
+		}
+		
+		// edge mobile
+		if (preg_match('~edga/([\d\.]+)~i', $this->userAgent, $match)) {
+			return 'Microsoft Edge Mobile '.$match[1];
 		}
 		
 		// vivaldi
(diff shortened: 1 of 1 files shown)


wcfsetup/install/files/lib/data/user/online/UserOnline.class.php
```diff
@@ -215,9 +215,14 @@ public function getBrowser() {
 		}
 		
 		// edge
-		if (preg_match('~edge/(\d{2}\.\d+)~i', $this->userAgent, $match)) {
+		if (preg_match('~edge/(\d{2}\.\d+)|edg/([\d\.]+)~i', $this->userAgent, $match)) {
 			return 'Microsoft Edge '.$match[1];
```
TimWolla, line 219: I believe this will not return a proper version number for the Chromium based edge, as the version will be captured by a distinct capture group.

wcfsetup/install/files/lib/data/user/online/UserOnline.class.php
```diff
@@ -215,9 +215,14 @@ public function getBrowser() {
...
 		}
+
+		// edge mobile (new)
+		if (preg_match('~edga/([\d\.]+)~i', $this->userAgent, $match)) {
+			return 'Microsoft Edge Mobile '.(isset($match[2]) ? $match[2] : $match[1]);
```
TimWolla, line 224: Likewise there won't  ever be a `$match[2]` here. Did you mix up the return statements?

wcfsetup/install/files/lib/data/user/online/UserOnline.class.php
xopez: Yeah, there stands just "Microsoft Edge" without any version

dtdesign:
Thanks!

PR #3466, opened 2020-07-28
Fixes exception when not using primary PayPal email

Sir-Will:
PayPal allows adding multiple emails, if you are using one of the alternative emails instead of the primary email then WoltLab will throw the exception. If the email is being validated then both should be checked to allow using none primary emails.

The `receiver_email` is always the primary email while `business` is the secondary in the IPN.

### DIFF
--- a/wcfsetup/install/files/lib/action/PaypalCallbackAction.class.php
+++ b/wcfsetup/install/files/lib/action/PaypalCallbackAction.class.php
@@ -56,7 +56,7 @@ class PaypalCallbackAction extends AbstractAction {
 			
 			// Check that receiver_email is your Primary PayPal email
 			if (strtolower($_POST['receiver_email']) != strtolower(PAYPAL_EMAIL_ADDRESS) && strtolower($_POST['business']) != strtolower(PAYPAL_EMAIL_ADDRESS)) {
-				throw new SystemException("Mismatching receiver_email and business '" . $_POST['receiver_email'] . "', expected '".PAYPAL_EMAIL_ADDRESS."'.");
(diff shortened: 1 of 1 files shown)


wcfsetup/install/files/lib/action/PaypalCallbackAction.class.php
```diff
@@ -55,8 +55,8 @@ public function execute() {
 			}
 			
 			// Check that receiver_email is your Primary PayPal email
-			if (strtolower($_POST['receiver_email']) != strtolower(PAYPAL_EMAIL_ADDRESS)) {
-				throw new SystemException("Mismatching receiver_email '" . $_POST['receiver_email'] . "', expected '".PAYPAL_EMAIL_ADDRESS."'.");
+			if (strtolower($_POST['receiver_email']) != strtolower(PAYPAL_EMAIL_ADDRESS) && strtolower($_POST['business']) != strtolower(PAYPAL_EMAIL_ADDRESS)) {
+				throw new SystemException("Mismatching receiver_email and business '" . $_POST['receiver_email'] . "', expected '".PAYPAL_EMAIL_ADDRESS."'.");
```
TimWolla, line 59: I can't comment on whether the change is correct, because I don't have the necessary PayPal expertise, but in any case the `receiver_email` in the exception message should also be adjusted.

Sir-Will: This part existed before but was removed: https://github.com/WoltLab/WCF/commit/4b200b5e9a1534d34465d03afe509df98a18a496
While it's not a requirement it's still possible to use it.

I'm using a seperate email for the forum to seperate it with other payments which resulted in this.

Sir-Will: I adjusted the exception message to include the business email

Sir-Will:
Should be good now.

PR #3475, opened 2020-08-04
Add checkMemoryLimit() method to ImageAdapter

### LINKED ISSUE #3229: Add method to check for sufficient memory to process images
We have many places with code like this

```
FileUtil::checkMemoryLimit($width * $height * ($extension == 'png' ? 4 : 3) * 2.1)
```

which could be moved to a central method (or to the image adapter?).

see https://community.woltlab.com/thread/272061-einf%C3%BChren-von-attachment-canmodifyimage/

TimWolla:
Resolves #3229

### DIFF
--- a/wcfsetup/install/files/lib/system/image/adapter/IMemoryAwareImageAdapter.class.php
+++ b/wcfsetup/install/files/lib/system/image/adapter/IMemoryAwareImageAdapter.class.php
@@ -1,0 +1,25 @@
+<?php
+namespace wcf\system\image\adapter;
+
+/**
+ * A memory aware image adapter is able to determine whether it is
+ * likely able to process an image within the process' memory limit.
+ * 
+ * @author	Tim Duesterhus
+ * @copyright	2001-2020 WoltLab GmbH
+ * @license	GNU Lesser General Public License <http://opensource.org/licenses/lgpl-license.php>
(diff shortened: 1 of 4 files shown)


wcfsetup/install/files/lib/system/image/adapter/IMemoryAwareImageAdapter.class.php
```diff
@@ -0,0 +1,25 @@
...
+	/**
+	 * Returns whether it is believed that sufficient memory
+	 * is available to process an image with the given properties.
+	 * 
+	 * @param	integer		$width
+	 * @param	integer		$height
+	 * @param	string		$mimeType
+	 * @return	boolean
+	 */
+	public function checkMemoryLimit($width, $height, $mimeType);
```
jens1o, line 24: Couldn't you use the PHP 7 return type `bool` in the function signature?

dtdesign: Return types in PHP 7.0 are a mess, it supports neither `void` nor nullable types, let alone union types. It would only make sense to make use of them if the usage is consistent, including but not limited to parameter typings.

PHP 7.2 is the earliest version that supports a somewhat sane level of features in terms of strong typing.

PR #3476, opened 2020-08-04
Failsafe Style SCSS

joshuaruesweg:
Integrates a SCSS validation for 

- Style(Add|Edit)Form
- StyleGlobalValuesForm

For performance reasons, we can only compile the default style in StyleGlobalValuesForm, otherwise the PHP script needs to much time (with 5 styles it was above 30 seconds).


wcfsetup/install/lang/de.xml
```diff
@@ -2839,6 +2839,7 @@ Kein Abschnitt darf leer sein und alle Abschnitten dürfen nur folgende Zeichen
 		<item name="wcf.acp.style.advanced.overrideScss.error.invalid"><![CDATA[Eingabe „{$error[text]}“ ungültig]]></item>
 		<item name="wcf.acp.style.advanced.overrideScss.error.predefined"><![CDATA[Variable „{$error[text]}“ wird bereits durch den Stil-Editor gesetzt]]></item>
 		<item name="wcf.acp.style.advanced.overrideScss.error.unknown"><![CDATA[Variable „{$error[text]}“ unbekannt]]></item>
+		<item name="wcf.acp.style.advanced.individualScss.error"><![CDATA[Der SCSS-Code ist invalid: {$errorType.message}.]]></item>
```
TimWolla, line 2842:
```suggestion
		<item name="wcf.acp.style.advanced.individualScss.error"><![CDATA[Der SCSS-Code enthält einen Fehler: {$errorType.message}.]]></item>
```

wcfsetup/install/lang/en.xml
```diff
@@ -2766,6 +2766,7 @@ If you have <strong>already bought the licenses for the listed apps</strong>, th
 		<item name="wcf.acp.style.advanced.overrideScss.error.invalid"><![CDATA[The input for “{$error[text]}” invalid.]]></item>
 		<item name="wcf.acp.style.advanced.overrideScss.error.predefined"><![CDATA[The variable “{$error[text]}” is already defined by the style editor.]]></item>
 		<item name="wcf.acp.style.advanced.overrideScss.error.unknown"><![CDATA[The variable “{$error[text]}” is not recognized.]]></item>
+		<item name="wcf.acp.style.advanced.individualScss.error"><![CDATA[The SCSS code is invalid: {$errorType.message}.]]></item>
```
TimWolla, line 2769:
```suggestion
		<item name="wcf.acp.style.advanced.individualScss.error"><![CDATA[The SCSS code contains an error: {$errorType.message}.]]></item>
```

wcfsetup/install/files/lib/system/style/StyleCompiler.class.php
```diff
@@ -60,38 +66,148 @@ protected function init() {
...
+			$this->compileStylesheetToString(
+				$files,
+				$variables,
+				$individualScss . (!empty($parameters['scss']) ? "\n" . $parameters['scss'] : ''),
+				function($content) {
+					return $content;
+				}
+			);
+		}
+		catch (SystemException $e) {
```
TimWolla, line 157:
```suggestion
		catch (\Exception $e) {
```

wcfsetup/install/lang/de.xml
```diff
@@ -2904,6 +2905,7 @@ Kein Abschnitt darf leer sein und alle Abschnitten dürfen nur folgende Zeichen
 		<item name="wcf.acp.style.globalValues"><![CDATA[Stilunabhängiges CSS und SCSS]]></item>
 		<item name="wcf.acp.style.globalValues.description"><![CDATA[Das unten stehende CSS und SCSS wird auf alle Stile angewandt, die individuelle Anpassungen in den Stilen sind jedoch höherwertig.]]></item>
 		<item name="wcf.acp.style.globalValues.input"><![CDATA[Individuelles CSS und SCSS]]></item>
+		<item name="wcf.acp.style.globalValues.input.error"><![CDATA[Der SCSS-Code ist invalid: {$errorType.message}.]]></item>
```
joshuaruesweg, line 2908:
```suggestion
		<item name="wcf.acp.style.globalValues.input.error"><![CDATA[Der SCSS-Code enthält einen Fehler: {$errorType.message}.]]></item>
```

wcfsetup/install/lang/en.xml
```diff
@@ -2832,6 +2833,7 @@ If you have <strong>already bought the licenses for the listed apps</strong>, th
 		<item name="wcf.acp.style.globalValues"><![CDATA[Global CSS and SCSS]]></item>
 		<item name="wcf.acp.style.globalValues.description"><![CDATA[The CSS and SCSS entered below is applied to all styles, but custom values in styles take precedence.]]></item>
 		<item name="wcf.acp.style.globalValues.input"><![CDATA[Individual CSS and SCSS]]></item>
+		<item name="wcf.acp.style.globalValues.input.error"><![CDATA[The SCSS code is invalid: {$errorType.message}.]]></item>
```
joshuaruesweg, line 2836:
```suggestion
		<item name="wcf.acp.style.globalValues.input.error"><![CDATA[The SCSS code contains an error: {$errorType.message}.]]></item>
```

wcfsetup/install/files/lib/system/style/StyleCompiler.class.php
```diff
@@ -333,9 +449,30 @@ protected function prepareFile($filename) {
...
+	}
+	
+	/**
+	 * Compiles SCSS stylesheets and returns them as an array for the `ltr` and the `rtl` version.
+	 *
+	 * @param	string[]	$files
+	 * @param	string[]	$variables
+	 * @param	string		$individualScss
+	 * @param	callable	$callback
+	 * @return      String[]
```
TimWolla, line 472:
```suggestion
	 * @return      string[]
```

wcfsetup/install/files/lib/system/style/StyleCompiler.class.php
```diff
@@ -60,38 +66,148 @@ protected function init() {
...
+				$files,
+				$variables,
+				$individualScss . (!empty($parameters['scss']) ? "\n" . $parameters['scss'] : ''),
+				function($content) {
+					return $content;
+				}
+			);
+		}
+		catch (\Exception $e) {
+			return $e->getMessage();
```
TimWolla, line 158: I think I would be happier with returning the full Exception or `null`. It makes the return value handling less ugly, because you clearly get a truthy or falsy value (unfortunately with inverted meaning). Also getting the full Exception allows the consumer inspect the stack trace or additional values, this might be helpful for future extensions.

joshuaruesweg:
> What are the performance characteristics of this change? How much longer does it take to save the style after this PR is applied?

Around 6 seconds on my local machine for both (saving styles and saving global style values).

TimWolla:
> Around 6 seconds on my local machine for both (saving styles and saving global style values).

This is fairly slow. On my machine I can reproduce times between 4 and 6 seconds. This makes iterating a style very slow for experienced users why tend to not make typos or test in their local environment. Maybe the user should have the options “Check and save” and “Save” or the validation should be disabled in debug mode?

joshuaruesweg:
> Maybe the user should have the options “Check and save” and “Save”

I don't see, that there should be two buttons for saving. This only confuses the user (and in case of doubt he clicks on the wrong button). 

> or the validation should be disabled in debug mode?

This validation is intended to protect inexperienced users from destroying your style and making the page inaccessible. Pages are usually not run in debug mode. I would do it the other way around if I had to. In debug mode, don't check the whole thing (where development environments are running), but in a productive system (debug mode is switched off), always check the whole thing.

TimWolla:
> I would do it the other way around if I had to. In debug mode, don't check the whole thing (where development environments are running), but in a productive system (debug mode is switched off), always check the whole thing.

Yes, that's what I (attempted to) say.

dtdesign:
I'm wondering if we could do it a bit differently, e.g. trying to compile the final CSS into a temporary file and simply report errors if that fails? Then, if it i succeeds, perform an atomic replacement of the compiled style file, so we only have to compile the style once.

This would have the benefit that for testing/"development" purposes the time required to make modifications and view them is constant, because there is no on the fly compilation step in accessing the frontend.

joshuaruesweg:
I think that is the best idea. I will implement it.

wcfsetup/install/files/lib/acp/form/StyleAddForm.class.php
```diff
@@ -447,13 +454,35 @@ public function validate() {
...
+	public function validateIndividualScss() {
+		$variables = array_merge(StyleCompiler::getDefaultVariables(), $this->variables);
+		
+		$this->styleTestFileDir = FileUtil::getTemporaryFilename('style_');
+		FileUtil::makePath($this->styleTestFileDir);
+		
+		$result = StyleCompiler::getInstance()->testStyle($this->styleTestFileDir, $this->apiVersion, false, $variables);
+		
+		if ($result !== null) {
+			throw new UserInputException('individualScss', [
```
TimWolla, line 480: Should you delete the temporary directory here?

wcfsetup/install/files/lib/acp/form/StyleEditForm.class.php
```diff
@@ -60,6 +64,33 @@ public function validate() {
...
+		
+		$variables = array_merge(StyleCompiler::getDefaultVariables(), $variables);
+		
+		$this->styleTestFileDir = FileUtil::getTemporaryFilename('style_');
+		FileUtil::makePath($this->styleTestFileDir);
+		
+		$result = StyleCompiler::getInstance()->testStyle($this->styleTestFileDir, $this->apiVersion, false, $variables);
+		
+		if ($result !== null) {
+			throw new UserInputException('individualScss', [
```
TimWolla, line 88: ditto

wcfsetup/install/files/lib/acp/form/StyleGlobalValuesForm.class.php
```diff
@@ -58,6 +65,38 @@ public function readData() {
...
+			// whether there are syntax issues. 
+			$defaultStyle = StyleHandler::getInstance()->getDefaultStyle();
+			if ($defaultStyle !== null) {
+				$this->styleTestFileDir = FileUtil::getTemporaryFilename('style_');
+				FileUtil::makePath($this->styleTestFileDir);
+				
+				$errorMessage = StyleCompiler::getInstance()->testStyle($this->styleTestFileDir, $defaultStyle->apiVersion, $defaultStyle->imagePath, $defaultStyle->getVariables(), $tmpFile);
+				
+				if ($errorMessage !== null) {
+					throw new UserInputException('styles', [
```
TimWolla, line 88: ditto

wcfsetup/install/files/lib/acp/form/StyleAddForm.class.php
```diff
@@ -768,6 +797,15 @@ public function save() {
...
 		]);
 		
+		// save compiled style
+		if ($this->styleTestFileDir && file_exists($this->styleTestFileDir . '/style.css') && file_exists($this->styleTestFileDir . '/style-rtl.css')) {
+			$styleFilename = StyleCompiler::getFilenameForStyle($style);
+			rename($this->styleTestFileDir . '/style.css', $styleFilename . '.css');
+			rename($this->styleTestFileDir . '/style-rtl.css', $styleFilename . '-rtl.css');
+			
+			rmdir($this->styleTestFileDir);
+		}
```
TimWolla, line 807: The CSS must not be written in StyleAddForm, because the imagePath is not yet known.

wcfsetup/install/files/lib/system/style/StyleCompiler.class.php
```diff
@@ -60,38 +66,150 @@ protected function init() {
...
+	protected function getFiles() {
+		if (!$this->files) {
+			$files = $this->getCoreFiles();
+			
+			// read stylesheets in dependency order
+			$sql = "SELECT		filename, application
 			FROM		wcf".WCF_N."_package_installation_file_log
 			WHERE           CONVERT(filename using utf8) REGEXP ?
 					AND packageID <> ?
 			ORDER BY	packageID";
```
TimWolla, line 181: Indentation messed up?

joshuaruesweg: Yes

dtdesign:
@joshuaruesweg Do you have some peformance numbers at hand?

joshuaruesweg:
@dtdesign Saving the standard design without any modifications is between 5 and 7 seconds. 
Saving the style global values is between 5 and 6 seconds. 

Tested on a 2.4 GHz 8-Core Intel Core i9 MacOS machine.

PR #3483, opened 2020-08-05
Overhauled the page action buttons' behavior

### LINKED ISSUE #3461: "Go to page top" button on mobile
ref https://community.woltlab.com/thread/284911-navigation-nach-oben-auf-jeder-seite-einblenden/

dtdesign:
The new implementation introduces a few key changes to the floating page action buttons:
* Removed the translate effect to reduce visual noise, using only opacity now.
* Buttons next to the "to top" button are now shown next to each other on mobile with overflow support.
* The "to top" button is now available on all devices.
* Page action buttons disappear on mobile when scrolling down, but appear again when scrolling up.
* Moved the buttons up on mobile to account for the touch sensitive space in browsers.

Closes #3461


TimWolla:
I'm noticing strange behavior on Chrome on Android:

1. I start at the end of the page.
2. I give the page a strong swipe.
3. The page scrolls to the very top in a single motion.
4. The “scroll to top” button shows up for a split second.

I assume this is due to the event being queued when starting the scroll and after the scroll is finished the “should I show the button” condition is not being re-checked.

TimWolla:
Similarly I am seeing the following behavior. It might have the same root cause.

If I start scrolling the page downwards and without lifting the finger I scroll upwards a bit and then back down the button will also appear, but almost immediately disappear again.

Maybe the debouncing on mobile should be adjusted to take touch events into account, not considering the scrolling to be finished until the finger is lifted? In fact I am not even sure whether *debouncing* is what you want. Should the debouncing rather be a throttling?

TimWolla:
Okay, the behavior is much better now. However I am not sure, if it is intended that all page action buttons disappear when scrolling down on mobile (vs. just when the scroll up button is alone). It's pretty confusing that the “Quotes” button does not always show up.

Also I'm seeing the following behavior on mobile:
- Scroll up: Quotes + To Top appear
- Click Quotes: Both disappear
- Close the overlay: To Top still is gone and the Quotes button is faded out.

dtdesign:
The overlay will prevent the scrolling of the page, however, this will also reset the scroll position and eventually restoring it.

1. Scrolled the page down and a bit up to 400 pixel from top (`pageYOffset = 400`).
2. Opened the quotes overlay. Overflow handling causes the position to be set to 0 (`pageYOffset = 0`).
3. Closing the overlay, the position is restored, now the browser is at 400 pixel from top again.

This is really tricky, because in all steps a `scroll` event is fired and the implementation is unable to tell this apart from a regular scroll event where the user drags from the top to 400 pixel.

I'm not sure how we can address this, especially because there are potentially a lot of sources that could manipulate the page overflow and scroll offset.

TimWolla:
> Opened the quotes overlay. Overflow handling causes the position to be set to 0 (`pageYOffset = 0`).

So at this point to my understanding the following happens:
- A scroll up is detected, the buttons technically are visible.
- The “To Top” button is hidden, because we are at the top of the page. 

But happened to the quote button at this point? Is it outside of the viewport?

dtdesign:
If the editor or any dialog is visible, the page action buttons are removed from the view. This is necessary because they are using `z-index` of 400 which places it atop most other things. This is an entirely different mechanic that simply forces a `display: none` upon the buttons.

TimWolla:
Okay, the behavior after the last commit is feeling good to me now. What about this concern [I mentioned earlier](https://github.com/WoltLab/WCF/pull/3483#issuecomment-670379457):

> However I am not sure, if it is intended that all page action buttons disappear when scrolling down on mobile (vs. just when the scroll up button is alone). It's pretty confusing that the “Quotes” button does not always show up.

dtdesign:
We'll keep an eye on it, for now, it should be fine imho. The quote button behavior was always a bit subpar on mobile, because page action buttons are forced hidden whenever the editor receives focus.

PR #3485, opened 2020-08-06
Update value of default cookie prefix to 5.3

Krymonota:
n/t


TimWolla:
This one is obviously correct, but as the set-up uses a randomized prefix in ~100% of the cases anyway it might make sense to juse use a generic value here, that won't need to be adjusted again.

PR #3486, opened 2020-08-06
Use generic value for default cookie prefix

Krymonota:
... so that it doesn't have to be adjusted for new versions. See https://github.com/WoltLab/WCF/pull/3485#issuecomment-669991914.


dtdesign:
Thanks!

PR #3500, opened 2020-08-11
Add missing phrases to English 3rd party register language items

Krymonota:
These phrases exist in the German language file, but not in the English one.


dtdesign:
Thanks!

PR #3502, opened 2020-08-12
Use timing safe comparison to validate `state` parameter for social login

### LINKED ISSUE #3501: Use timing safe comparison to validate `state` parameter for social login

Krymonota:
The Twitter social login is left out because the implementation still uses OAuth 1.0, which does not support the `state` parameter.

Closes #3501

### DIFF
--- a/wcfsetup/install/files/lib/action/FacebookAuthAction.class.php
+++ b/wcfsetup/install/files/lib/action/FacebookAuthAction.class.php
@@ -68,7 +68,7 @@ class FacebookAuthAction extends AbstractAction {
 			}
 			
 			// validate state, validation of state is executed after fetching the access_token to invalidate 'code'
-			if (!isset($_GET['state']) || $_GET['state'] != WCF::getSession()->getVar('__facebookInit')) throw new IllegalLinkException();
+			if (!isset($_GET['state']) || !\hash_equals(WCF::getSession()->getVar('__facebookInit'), $_GET['state'])) throw new IllegalLinkException();
 			WCF::getSession()->unregister('__facebookInit');
 			
 			try {
--- a/wcfsetup/install/files/lib/action/GithubAuthAction.class.php
+++ b/wcfsetup/install/files/lib/action/GithubAuthAction.class.php
@@ -64,7 +64,7 @@ class GithubAuthAction extends AbstractAction {
 			}
 			
 			// validate state, validation of state is executed after fetching the access_token to invalidate 'code'
-			if (!isset($_GET['state']) || $_GET['state'] != WCF::getSession()->getVar('__githubInit')) throw new IllegalLinkException();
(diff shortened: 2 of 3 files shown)


wcfsetup/install/files/lib/action/FacebookAuthAction.class.php
```diff
@@ -68,7 +68,7 @@ public function execute() {
 			}
 			
 			// validate state, validation of state is executed after fetching the access_token to invalidate 'code'
-			if (!isset($_GET['state']) || $_GET['state'] != WCF::getSession()->getVar('__facebookInit')) throw new IllegalLinkException();
+			if (!isset($_GET['state']) || !\hash_equals($_GET['state'], WCF::getSession()->getVar('__facebookInit'))) throw new IllegalLinkException();
```
TimWolla, line 71: You need to switch the parameter order. The first parameter is defined to be the known string, the second is defined to be the user string.

It does not matter with the *current* implementation of `hash_equals` within the C source of PHP, but it might matter in case of later changes.

```suggestion
			if (!isset($_GET['state']) || !\hash_equals(WCF::getSession()->getVar('__facebookInit'), $_GET['state'])) throw new IllegalLinkException();
```

wcfsetup/install/files/lib/action/GithubAuthAction.class.php
```diff
@@ -64,7 +64,7 @@ public function execute() {
 			}
 			
 			// validate state, validation of state is executed after fetching the access_token to invalidate 'code'
-			if (!isset($_GET['state']) || $_GET['state'] != WCF::getSession()->getVar('__githubInit')) throw new IllegalLinkException();
+			if (!isset($_GET['state']) || !\hash_equals($_GET['state'], WCF::getSession()->getVar('__githubInit'))) throw new IllegalLinkException();
```
TimWolla, line 67:
```suggestion
			if (!isset($_GET['state']) || !\hash_equals(WCF::getSession()->getVar('__githubInit'), $_GET['state'])) throw new IllegalLinkException();
```

wcfsetup/install/files/lib/action/GoogleAuthAction.class.php
```diff
@@ -67,7 +67,7 @@ public function execute() {
 			}
 			
 			// validate state, validation of state is executed after fetching the access_token to invalidate 'code'
-			if (!isset($_GET['state']) || $_GET['state'] != WCF::getSession()->getVar('__googleInit')) throw new IllegalLinkException();
+			if (!isset($_GET['state']) || !\hash_equals($_GET['state'], WCF::getSession()->getVar('__googleInit'))) throw new IllegalLinkException();
```
TimWolla, line 70:
```suggestion
			if (!isset($_GET['state']) || !\hash_equals(WCF::getSession()->getVar('__googleInit'), $_GET['state'])) throw new IllegalLinkException();
```

TimWolla:
Thanks.

PR #3504, opened 2020-08-12
Fix type of `options` parameter in HTTPRequest PHPDoc

Krymonota:


Krymonota:
Yup, but it bothered me to use an `@var` comment in my IDE 😛

PR #3513, opened 2020-08-15
Added possiblity to disable the preview-button

ilouHD:
Added possibility to disable the preview-button of the wysiwyg form container, that - in default - will be generated automatically. In some cases you don't need the preview, because you implement another preview yourself (e.g. sending mails with wysiwyg input rendered directly in mail-template)

### DIFF
--- a/wcfsetup/install/files/lib/system/form/builder/container/wysiwyg/WysiwygFormContainer.class.php
+++ b/wcfsetup/install/files/lib/system/form/builder/container/wysiwyg/WysiwygFormContainer.class.php
@@ -485,7 +485,7 @@ class WysiwygFormContainer extends FormContainer {
 					->objectType($this->messageObjectType)
 					->wysiwygId($this->getWysiwygId())
 					->objectId($this->getObjectId())
-				);
+			);
 		}
 		
 		EventHandler::getInstance()->fireAction($this, 'populate');
--- a/wcfsetup/install/files/lib/system/form/builder/container/wysiwyg/WysiwygFormContainer.class.php
+++ b/wcfsetup/install/files/lib/system/form/builder/container/wysiwyg/WysiwygFormContainer.class.php
@@ -47,7 +47,7 @@ class WysiwygFormContainer extends FormContainer {
 	protected $attachmentData;
 	
 	/**
-	 * `true` if the preview-button should be generated automatically and `false` otherwise
+	 * `true` if the preview button should be shown and `false` otherwise
 	 * @var         bool
 	 * @since       5.3
 	 */
@@ -214,9 +214,9 @@ class WysiwygFormContainer extends FormContainer {
 	}
 	
 	/**
-	 * Sets whether the preview-button should be generated or not and returns this form container.
+	 * Sets whether the preview button should be shown or not and returns this form container.
 	 * 
-	 * By default, the preview-button is enabled will be generated.
+	 * By default, the preview button is shown.
 	 * 
 	 * @param       bool                    $enablePreviewButton
 	 * @return      WysiwygFormContainer    this form container
@@ -339,8 +339,9 @@ class WysiwygFormContainer extends FormContainer {
 	}
 	
 	/**
-	 * Returns `true` if the preview-button should be generated automatically and returns `false` otherwise.
-	 * By default, the preview-button is generated automatically.
+	 * Returns `true` if the preview button will be shown and returns `false` otherwise.
+	 * 
+	 * By default, the preview button is shown.
 	 * 
 	 * @return      bool
 	 * @since       5.3
--- a/wcfsetup/install/files/lib/system/form/builder/container/wysiwyg/WysiwygFormContainer.class.php
+++ b/wcfsetup/install/files/lib/system/form/builder/container/wysiwyg/WysiwygFormContainer.class.php
@@ -47,6 +47,12 @@ class WysiwygFormContainer extends FormContainer {
 	protected $attachmentData;
 	
 	/**
+	 * `true` if the preview-button should be generated automatically and `false` otherwise
+	 * @var bool
+	 */
+	protected $enablePreviewButton = true;
+	
+	/**
 	 * name of the relevant message object type
 	 * @var	string
 	 */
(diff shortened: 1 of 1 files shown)


wcfsetup/install/files/lib/system/form/builder/container/wysiwyg/WysiwygFormContainer.class.php
```diff
@@ -46,6 +46,12 @@ class WysiwygFormContainer extends FormContainer {
 	 */
 	protected $attachmentData;
 	
+	/**
+	 * `true` if the preview-button should be generated automatically and `false` otherwise
+	 * @var bool
```
Gravatronics, line 51:
```suggestion
	 * @var         bool
	 * @since       5.3
```

wcfsetup/install/files/lib/system/form/builder/container/wysiwyg/WysiwygFormContainer.class.php
```diff
@@ -206,6 +212,20 @@ public function attachmentData($objectType = null, $parentObjectID = 0) {
 		return $this;
 	}
 	
+	/**
+	 * Sets whether the preview-button should be generated or not and returns this form container.
+	 * 
+	 * By default, the preview-button is enabled will be generated.
+	 * 
+	 * @param	boolean		$enablePreviewButton
+	 * @return	WysiwygFormContainer		this form container
```
Gravatronics, line 221:
```suggestion
	 * @param       bool                    $enablePreviewButton
	 * @return      WysiwygFormContainer    this form container
	 * @since       5.3
```

wcfsetup/install/files/lib/system/form/builder/container/wysiwyg/WysiwygFormContainer.class.php
```diff
@@ -316,6 +336,16 @@ public function isRequired() {
 		return $this->required;
 	}
 	
+	/**
+	 * Returns `true` if the preview-button should be generated automatically and returns `false` otherwise.
+	 * By default, the preview-button is generated automatically.
+	 *
+	 * @return	bool
```
Gravatronics, line 343:
```suggestion
	 * 
	 * @return      bool
	 * @since       5.3
```

wcfsetup/install/files/lib/system/form/builder/container/wysiwyg/WysiwygFormContainer.class.php
```diff
@@ -448,12 +478,14 @@ public function populate() {
...
-				->wysiwygId($this->getWysiwygId())
-				->objectId($this->getObjectId())
-		);
+		if ($this->enablePreviewButton === true) {
+			$this->getDocument()->addButton(
+				WysiwygPreviewFormButton::create($this->getWysiwygId() . 'PreviewButton')
+					->objectType($this->messageObjectType)
+					->wysiwygId($this->getWysiwygId())
+					->objectId($this->getObjectId())
+				);
```
Gravatronics, line 487:
```suggestion
			);
```

wcfsetup/install/files/lib/system/form/builder/container/wysiwyg/WysiwygFormContainer.class.php
```diff
@@ -46,6 +46,13 @@ class WysiwygFormContainer extends FormContainer {
 	 */
 	protected $attachmentData;
 	
+	/**
+	 * `true` if the preview-button should be generated automatically and `false` otherwise
```
Gravatronics, line 50:
```suggestion
	 * `true` if the preview button should be shown and `false` otherwise
```

wcfsetup/install/files/lib/system/form/builder/container/wysiwyg/WysiwygFormContainer.class.php
```diff
@@ -206,6 +213,21 @@ public function attachmentData($objectType = null, $parentObjectID = 0) {
 		return $this;
 	}
 	
+	/**
+	 * Sets whether the preview-button should be generated or not and returns this form container.
+	 * 
+	 * By default, the preview-button is enabled will be generated.
```
Gravatronics, line 219:
```suggestion
	 * Sets whether the preview button should be shown or not and returns this form container.
	 * 
	 * By default, the preview button is shown.
```

wcfsetup/install/files/lib/system/form/builder/container/wysiwyg/WysiwygFormContainer.class.php
```diff
@@ -316,6 +338,17 @@ public function isRequired() {
 		return $this->required;
 	}
 	
+	/**
+	 * Returns `true` if the preview-button should be generated automatically and returns `false` otherwise.
+	 * By default, the preview-button is generated automatically.
```
Gravatronics, line 343:
```suggestion
	 * Returns `true` if the preview button will be shown and returns `false` otherwise.
	 * 
	 * By default, the preview button is shown.
```

wcfsetup/install/files/lib/system/form/builder/container/wysiwyg/WysiwygFormContainer.class.php
```diff
@@ -448,12 +481,14 @@ public function populate() {
 			$this->setAttachmentHandler();
 		}
 		
-		$this->getDocument()->addButton(
-			WysiwygPreviewFormButton::create($this->getWysiwygId() . 'PreviewButton')
-				->objectType($this->messageObjectType)
-				->wysiwygId($this->getWysiwygId())
-				->objectId($this->getObjectId())
-		);
+		if ($this->enablePreviewButton === true) {
```
Gravatronics, line 484:
```suggestion
		if ($this->enablePreviewButton) {
```

ilouHD: Why should we remove the `=== true` check here?

The only reason I can imagine is to also support 1 and 0 for valid values (and other equivalents). They will be false in any case in my suggestion, espacially `1 === true` will result to false.

In this case I would suggest to use only `$this->enablePreviewButton == true`, which doesn't take care of the actual datatype , but that will then be equivalent to your suggestion.

TimWolla: > Why should we remove the `=== true` check here?

An `if` already carries an implicit `== true` within its definition. Writing an explicit `== true` is redundant.

PR #3514, opened 2020-08-16
Forms: Provide a link to edit an item that was just created

### LINKED ISSUE #3289: Forms: Provide a link to edit an item that was just created
Forms are generally designed to reset themselves after submitting and providing a clean form afterwards. There is usually a generic success message at the top, which we can extend to provide a link to edit the just created item.

We should check if it is possible to generate these links automatically without requiring any manual edits.

Partly inspired by https://community.woltlab.com/thread/272406-boxen-speichern-und-nicht-zur%C3%BCck-zur-liste-wechseln/.

BurntimeX:
i
Closes #3289


BurntimeX:
> Use object over id? I expect that most if not all of those objects support that.
This requires that the objects implement `IRouteController`, which is often not the case.

dtdesign:
How about this view instead, it is still pretty dominant because of the link style, but fits into the slim appearance:

image

dtdesign:
LGTM, expect for the adjustments required to use a phrase instead of the button.

PR #3525, opened 2020-08-19
Prevent rich metacode being dysfunctional in new messages

dtdesign:
The `isset` strategy does not work properly, because in some cases the HTML is evaluated before being placed in the final template, causing the variable to be set already.

This changes modifies the behavior to always attempt to initialize the JavaScript, but made changes to ensure that each subsequent call is idempotent. One side effect of this change is the increase of generated HTML if an affected metacode appears more than once. We should revisit the entire system in future iterations by introducing a central class that performs the on-the-fly initialization of such components.


TimWolla (approved):
Diff LGTM

PR #3538, opened 2020-08-28
fix updating the styles to 5.3

mutec:
If `Style::$imagePath` is empty, it resulted in copying the shole WSC-folder into a style-asset-folder - and doing this until we reache the timeout limits.
Including `global.php` ended up in trying to define the option-constants (especially the fixed-value-options like `MODULE_ATTACHMENT`) again while they are already defined - which causes a Fatal Error-Message.

### DIFF
--- a/wcfsetup/install/files/acp/update_com.woltlab.wcf_5.3_style.php
+++ b/wcfsetup/install/files/acp/update_com.woltlab.wcf_5.3_style.php
@@ -1,5 +1,4 @@
 <?php
-require('global.php');
 
 use wcf\data\style\StyleEditor;
 use wcf\data\style\StyleList;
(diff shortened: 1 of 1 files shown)


wcfsetup/install/files/acp/update_com.woltlab.wcf_5.3_style.php
```diff
@@ -32,7 +31,7 @@
 	
 	// 3) If the file had a custom image folder we need to copy all files into the asset folder.
 	// Moving the files is unsafe, because multiple styles can share a single image folder.
-	if ($style->imagePath != 'images/') {
+	if (!empty($style->imagePath) && $style->imagePath != 'images/') {
 		$srcPath = FileUtil::addTrailingSlash(WCF_DIR.$style->imagePath);
```
TimWolla, line 35:
```suggestion
	$srcPath = FileUtil::addTrailingSlash(WCF_DIR.$style->imagePath);
	if ($srcPath !== WCF_DIR && $srcPath !== WCF_DIR.'images/') {
```

I hope this suggestion applies well. My suggestion also avoids issues with differences in trailing slashes.

mutec: Good idea. I changed the commit like that.

TimWolla (approved):

TimWolla:
Thanks.

PR #3539, opened 2020-08-28
WIP: Prepare upgrade instructions for 5.2 to 5.3

mutec:
We need to provide `/lib/core.functions.php` and `/lib/system/api/*` before executing the complete file-PIP; otherwise we're getting issues with Guzzle being not found and causing an abortion.
Put the style and email-confirmation scripts into standalone-instructions since they might cause long execution times. In my test-environment I updated without that parameter.


com.woltlab.wcf/update_5.3.sql
```diff
@@ -0,0 +1,16 @@
...
+DROP TABLE IF EXISTS wcf1_devtools_missing_language_item;
+CREATE TABLE wcf1_devtools_missing_language_item (
+	itemID INT(10) NOT NULL AUTO_INCREMENT PRIMARY KEY,
+	languageID INT(10),
+	languageItem VARCHAR(191) NOT NULL,
+	lastTime INT(10) NOT NULL,
+	stackTrace MEDIUMTEXT NOT NULL,
+	
+	UNIQUE KEY (languageID, languageItem)
+);
```
TimWolla, line 14: Is there a reason this can't be done with the update script?

wcfsetup/install/files/acp/update_com.woltlab.wcf_5.3.php
```diff
@@ -0,0 +1,49 @@
...
+];
+
+(new DatabaseTableChangeProcessor(
+/** @var ScriptPackageInstallationPlugin $this */
+	$this->installation->getPackage(),
+	$tables,
+	WCF::getDB()->getEditor())
+)->process();
+
+// Delete mail form during the upgrade to make it 100% non-functional.
```
TimWolla, line 47: Should this delete the file from the file log? This probably should also be split into a dedicated update script.

com.woltlab.wcf/update_5.3.sql
mutec: I thought so first, but I didn't read the exception carefully and forgot reverting that change.

dtdesign:
Thanks!

PR #3540, opened 2020-08-28
Fixed editing styles using an url instead of relative path to logo files

mutec:
Before WSC 5.3 it was possible (and used as a default when uploading an image) to use http(s)-urls for `pageLogo` and `pageLogoMobile` which cannot be handled by `UploadFile`.
This change downloads the files to the style's assets-folder and uses that new file.

### DIFF
--- a/wcfsetup/install/files/lib/acp/form/StyleEditForm.class.php
+++ b/wcfsetup/install/files/lib/acp/form/StyleEditForm.class.php
@@ -2,10 +2,8 @@
 namespace wcf\acp\form;
 use wcf\data\style\Style;
 use wcf\data\style\StyleAction;
-use wcf\data\user\cover\photo\UserCoverPhoto;
 use wcf\form\AbstractForm;
 use wcf\system\exception\IllegalLinkException;
-use wcf\system\exception\SystemException;
 use wcf\system\exception\UserInputException;
 use wcf\system\file\upload\UploadFile;
 use wcf\system\file\upload\UploadHandler;
@@ -13,6 +11,8 @@ use wcf\system\language\I18nHandler;
 use wcf\system\style\StyleCompiler;
 use wcf\system\WCF;
 use wcf\util\FileUtil;
+use wcf\util\HTTPRequest;
+use wcf\util\Url;
 
 /**
  * Shows the style edit form.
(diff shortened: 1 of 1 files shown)


wcfsetup/install/files/lib/acp/form/StyleEditForm.class.php
```diff
@@ -139,19 +139,54 @@ protected function readStyleVariables() {
...
+	 * @param       string        $url
+	 * @param       string        $target
+	 */
+	protected function getLogoFromUrl($url, $target) {
+		$request = new HTTPRequest($url);
+		$request->execute();
+		$reply = $request->getReply();
+		
+		file_put_contents($target, $reply['body']);
+	}
```
TimWolla, line 188: https://docs.woltlab.com/migration_wsc-52_libraries.html#guzzle

wcfsetup/install/files/lib/acp/form/StyleEditForm.class.php
```diff
@@ -139,19 +139,54 @@ protected function readStyleVariables() {
...
+				$this->variables['pageLogo'] = '';
+			}
 		}
 		if ($this->variables['pageLogoMobile']) {
-			$file = new UploadFile($this->style->getAssetPath().$this->variables['pageLogoMobile'], basename($this->variables['pageLogoMobile']), true, true, true);
-			UploadHandler::getInstance()->registerFilesByField('pageLogoMobile', [
-				$file,
-			]);
+			if (Url::is($this->variables['pageLogoMobile'])) {
+				$filename = basename(Url::parse($this->variables['pageLogoMobile'])['path']);
```
TimWolla, line 159: You must not trust the remote extension (and by extension the remote filename). Please validate the path against a sane whitelist (gif, png, jpeg, jpg, svg) and ignore the logo otherwise.

TimWolla (commented):
Good catch. I'm not sure whether this is the best place to download the files, though. Especially since this happens until the style is actually submitted once.

Ideally the StyleEditForm would not contain any compatibility code. Unfortunately downloading files from the network during the upgrade is a no-no as well.

dtdesign:
I'm not sure if this really is that much of an issue. Using a remote image is rather obscure and I'm fairly confident that pretty much all occurences are caused by the style upload. We can run this from within a script that is looped for every style during the upgrade.

The solution could look like this:
1. If the link resolves to a path within the software, treat it as a local file reference and copy the file.
2. If it is a remote file, try to download it, but use a very short timeout, 5-10 seconds should be sufficient.
3. If the download fails for whatever reason: Discard it and move on. No error is displayed, the upgrade must not be aborted prematurely.

TimWolla:
@mutec We decided to handle this as part of the update process by queueing maintenance background jobs. I'll handle this and thus close this PR. Thanks for bringing the issue to our attention.

PR #3565, opened 2020-09-15
Extend automated guessing logic of highlighters

TimWolla:
The automated detection is very old and was not updated for Prism. I've added a few additional languages that can be detected pretty reliably and are somewhat commonly used. A notable language that is missing still is C#, any suggestions for that?


dtdesign:
> A notable language that is missing still is C#, any suggestions for that?

```cs
using System;
```

This reference is often found and generated by most IDEs by default.

TimWolla:
Thanks. I've added that and tested it with some random example code I found on the Internet.

dtdesign (approved):

PR #3573, opened 2020-09-24
Refactor Session Handling: Integration branch

### LINKED ISSUE #3527: Refactor session handling
The current session handling comes with quite a few limitations, due to historical (or rather: archaic) design choices. These limitations include a sharing of session variables across all of a single user's devices and an auto login that relies on the knowledge of the user's password as well as relying on storing a hash of the user's password on the user's device.

In short: The session system needs to be moved into 2020. Here's the plan:

- Each of the user's devices gets a dedicated, long-lived, session.

This ticks all of the above requirements.

In more detail the implementation could look something like the following. It should maintain compatibility as much as possible, while drastically improving the logic.

- A new database table `wcf1_session_new` is introduced (name is up for discussion):
    ```
    CREATE TABLE wcf1_session_new (
     sessionID CHAR(40) NOT NULL,
     userID INT DEFAULT NULL,
     sessionVariables MEDIUMBLOB,
     time INT NOT NULL,
     KEY (userID),
     KEY (time)
    );
    ```
- `wcf1_session_new.userID` is *not* a `UNIQUE KEY`.
- `wcf1_session_virtual` is removed.
- `wcf1_session` is being maintained for compatibility with the online list which directly pokes around within the session. Unnecessary columns (`sessionVariables`) will be dropped. It is not being used for user identification any longer.

- Direct access to `wcf1_session_new` is strictly forbidden. No compatibility guarantees are given whatsoever. The only supported interface is the PHP API.

- The concept of “auto login” will be removed. There will only be a _persistent_ login. The terminology change is intentional.
    - There will no longer be a checkbox to select whether you want to persistently login.
    - Instead every session for a user will automatically be long-lived.
    - User sessions will be removed after X days of inactivity. Guest sessions will be removed after Y minutes of inactivity.
        - Proposal: X = 14, Y = 60.
        - As long as the user regularly visits the community their session will not expire.
        - Stored values will no longer magically vanish.
            - Either the user is logged in and all their data is there.
            - Or the user is actually logged out and their (session) data is gone.
    - The `_userID` and `_password` cookies will go away.
    - There will only be a single `_session` cookie that contains the signed `wcf1_session_new.sessionID` value.
        - `SameSite=lax` (or possibly `=none`), `HttpOnly`.

- The CSRF token will no longer be stored within the session, but instead within a cookie.
    - This cookie will be `SameSite=strict`, but _not_ `HttpOnly`.

Advantages:
- It allows us to sanely migrate away from the current non-standard password hashing scheme. Argon2 is looking pretty good.
- The UX during social logins can be improved, because the user can be persistently logged in without one of their other devices being logged out due to the password changing.
    - Similarly a login using a single-use token within an email *could* be implemented (security is no worse than simply using the lost password feature), without messing up other sessions.
    - Or a login by scanning a QR code on an authenticated device.
- A **proper**, non-hackish two factor solution can be implemented, because state is cleanly separated by device.
- The user can be shown a list of authenticated sessions, enabling them to easily kill off a single session without needing to change their password.
- The user will no longer see “your session expired, please resubmit”, improving UX quite a bit.

Disadvantages:
- It breaks existing third party login schemes that rely on generating cookies themselves.
- It possibly breaks stuff that directly pokes around in `wcf1_session`, even if that table still is maintained on a best-effort basis.

TimWolla:
Resolves #3527 

Includes:

- #3574
- #3580 
- #3583 
- #3586 
- #3587 
- #3593


wcfsetup/install/files/lib/system/user/authentication/password/PasswordAlgorithmManager.class.php
```diff
@@ -0,0 +1,62 @@
...
+	/**
+	 * Returns the password algorithm with the given name.
+	 *
+	 * @throws InvalidArgumentException If the password algorithm does not exist.
+	 * @throws ImplementationException If the password algorithm does not implement IPasswordAlgorithm.
+	 */
+	public function getAlgorithmFromName(string $name): IPasswordAlgorithm {
+		// The wcf1e algorithm can be recognized with the following regular expression.
+		// The algorithm is handled by the Wcf1e password algorithm class. 
+		if (preg_match('~^wcf1e[cms][01][ab][01]$~', $name)) {
```
TimWolla, line 27:
```suggestion
		if (\preg_match('~^wcf1e[cms][01][ab][01]$~', $name)) {
```

wcfsetup/install/files/lib/system/package/PackageInstallationDispatcher.class.php
```diff
@@ -230,16 +231,15 @@ public function install($node) {
...
-					}
-					catch (\Throwable $e) {
-						// ignore, the secret will stay empty and crypto operations
-						// depending on it will fail
-					}
+					$statement->execute([
+						$signatureSecret = \bin2hex(\random_bytes(20)),
+						'signature_secret'
+					]);
+					define('SIGNATURE_SECRET', $signatureSecret);
```
TimWolla, line 238:
```suggestion
					\define('SIGNATURE_SECRET', $signatureSecret);
```

wcfsetup/setup/db/install.sql
```diff
@@ -1839,6 +1819,18 @@ CREATE TABLE wcf1_user_rank (
 	hideTitle TINYINT(1) NOT NULL DEFAULT 0
 );
 
+DROP TABLE IF EXISTS wcf1_user_session;
+CREATE TABLE wcf1_user_session (
+	sessionID CHAR(40) NOT NULL PRIMARY KEY,
```
TimWolla, line 1824: Test whether a `BIGINT` `auto_increment` column for the `PRIMARY KEY` with `sessionID` being a `UNIQUE KEY` improves `INSERT` and `UPDATE` performance.

wcfsetup/install/files/lib/system/session/SessionHandler.class.php
```diff
@@ -339,9 +305,54 @@ protected function defineConstants() {
...
+				$cookieDomain = HeaderUtil::getCookieDomain();
+				$cookieDomain = ($cookieDomain !== null ? '; domain='.$cookieDomain : '');
+			}
+			else {
+				// SameSite=strict is not supported in a multi domain set-up, because
+				// it breaks cross-application requests.
+				$sameSite = '; SameSite=strict';
+			}
+			
+			header('set-cookie: XSRF-TOKEN='.rawurlencode($xsrfToken).'; path=/'.$cookieDomain.(RouteHandler::secureConnection() ? '; secure' : '').$sameSite, false);
```
TimWolla, line 352:
```suggestion
			\header('set-cookie: XSRF-TOKEN='.\rawurlencode($xsrfToken).'; path=/'.$cookieDomain.(RouteHandler::secureConnection() ? '; secure' : '').$sameSite, false);
```

wcfsetup/install/files/lib/system/session/SessionHandler.class.php
```diff
@@ -425,97 +429,91 @@ public function getUser() {
...
+			($this->isACP   ? self::ACP_SESSION_LIFETIME  :
+			($row['userID'] ? self::USER_SESSION_LIFETIME :
+			(                 self::GUEST_SESSION_LIFETIME)));
+		if ($row['lastActivityTime'] < (TIME_NOW - $lifetime)) {
+			return false;
 		}
-		else {
-			$this->virtualSession = SessionVirtual::getExistingSession($sessionID);
+		
+		$variables = @unserialize($row['sessionVariables']);
```
TimWolla, line 458:
```suggestion
		$variables = @\unserialize($row['sessionVariables']);
```

wcfsetup/install/files/lib/system/session/SessionHandler.class.php
```diff
@@ -425,97 +429,91 @@ public function getUser() {
...
+		// Check whether the session variables became corrupted.
+		if (!is_array($variables)) {
```
TimWolla, line 460:
```suggestion
		if (!\is_array($variables)) {
```

wcfsetup/install/files/lib/system/session/SessionHandler.class.php
```diff
@@ -526,92 +524,54 @@ protected function validate() {
...
+				(sessionID, ipAddress, userAgent, lastActivityTime, sessionVariables)
+			VALUES
+				(?, ?, ?, ?, ?)";
+		$statement = WCF::getDB()->prepareStatement($sql);
+		$statement->execute([
+			$this->sessionID,
+			UserUtil::getIpAddress(),
+			UserUtil::getUserAgent(),
+			TIME_NOW,
+			serialize([]),
```
TimWolla, line 540:
```suggestion
			\serialize([]),
```

wcfsetup/install/files/lib/system/session/SessionHandler.class.php
```diff
@@ -891,78 +770,62 @@ protected function changeUserVirtual(User $user) {
...
-			'lastActivityTime' => TIME_NOW
-		];
 		if ($this->variablesChanged) {
-			$data['sessionVariables'] = serialize($this->variables);
+			$sql = "UPDATE	wcf".WCF_N."_".($this->isACP ? 'acp' : 'user')."_session
+				SET	sessionVariables = ?
+				WHERE	sessionID = ?";
+			$statement = WCF::getDB()->prepareStatement($sql);
+			$statement->execute([
+				serialize($this->variables),
```
TimWolla, line 779:
```suggestion
				\serialize($this->variables),
```

wcfsetup/install/files/lib/system/session/SessionHandler.class.php
```diff
@@ -891,78 +770,62 @@ protected function changeUserVirtual(User $user) {
...
+			$data = [
+				'ipAddress' => $this->ipAddress,
+				'userAgent' => $this->userAgent,
+				'requestURI' => $this->requestURI,
+				'requestMethod' => $this->requestMethod,
+				'lastActivityTime' => TIME_NOW,
+				'userID' => $this->user->userID,
+				'sessionID' => $this->sessionID,
+			];
+			if (!class_exists('wcf\system\CLIWCF', false) && !$this->disableTracking) {
```
TimWolla, line 797:
```suggestion
			if (!\class_exists('wcf\system\CLIWCF', false) && !$this->disableTracking) {
```

wcfsetup/setup/db/install.sql
TimWolla: It most likely would improve `INSERT` performance, but tank `SELECT` (and probably also `UPDATE`) performance, because of the additional indirection. It's not entirely clear which type of query dominates here, because that depends on the cookies the user-agent sends. The safest option should be using what worked in the past: `sessionID` being the `PRIMARY KEY`.

wcfsetup/install/files/lib/system/session/SessionHandler.class.php
```diff
@@ -339,9 +312,54 @@ protected function defineConstants() {
...
 	protected function initSecurityToken() {
-		if ($this->getVar('__SECURITY_TOKEN') === null) {
-			$this->register('__SECURITY_TOKEN', bin2hex(\random_bytes(20)));
+		$xsrfToken = '';
+		if (!empty($_COOKIE['XSRF-TOKEN'])) {
+			// We intentionally do not extract the signed value and instead just verify the correctness.
+			//
+			// The reason is that common JavaScript frameworks can use the contents of the `XSRF-TOKEN` cookie as-is,
+			// without performing any processing on it, improving interoperability. Leveraging that JavaScript framework
+			// feature requires the author of the controller to check the value within the `X-XSRF-TOKEN` request header
```
Gravatronics, line 321: "Leveraging that JavaScript framework feature requires" There is at least an article missing.

wcfsetup/install/files/lib/system/session/SessionHandler.class.php
```diff
@@ -339,9 +312,54 @@ protected function defineConstants() {
...
+				$cookieDomain = HeaderUtil::getCookieDomain();
+				$cookieDomain = ($cookieDomain !== null ? '; domain='.$cookieDomain : '');
+			}
+			else {
+				// SameSite=strict is not supported in a multi domain set-up, because
+				// it breaks cross-application requests.
+				$sameSite = '; SameSite=strict';
+			}
+			
+			\header('set-cookie: XSRF-TOKEN='.\rawurlencode($xsrfToken).'; path=/'.$cookieDomain.(RouteHandler::secureConnection() ? '; secure' : '').$sameSite, false);
```
Gravatronics, line 359: Doesn't this code implies that an appropriate method in `HeaderUtil` is missing to performances the same action?

wcfsetup/install/files/lib/system/session/SessionHandler.class.php
```diff
@@ -425,97 +436,91 @@ public function getUser() {
...
+			UserUtil::getIpAddress(),
+			UserUtil::getUserAgent(),
+			TIME_NOW,
+			$this->sessionID,
+		]);
+		
+		// Refresh cookie.
+		if ($this->user->userID && !$this->isACP) {
+			HeaderUtil::setCookie(
+				($this->isACP ? 'acp' : 'user')."_session",
```
Gravatronics, line 491:
```suggestion
				($this->isACP ? 'acp' : 'user') . '_session',
```

wcfsetup/install/files/lib/system/session/SessionHandler.class.php
```diff
@@ -425,97 +436,91 @@ public function getUser() {
...
 		}
 		
-		$this->user = new User($this->session->userID);
-		if ($this->isACP) {
-			$this->virtualSession = ACPSessionVirtual::getExistingSession($sessionID);
+		// Check whether the session technically already expired.
+		$lifetime =
+			($this->isACP   ? self::ACP_SESSION_LIFETIME  :
+			($row['userID'] ? self::USER_SESSION_LIFETIME :
+			(                 self::GUEST_SESSION_LIFETIME)));
```
Gravatronics, line 460: I think an explicit, lengthier piece of code would make it easier to comprehend than nested ternary operators.

wcfsetup/install/files/lib/system/session/SessionHandler.class.php
```diff
@@ -526,92 +531,54 @@ protected function validate() {
...
-		}
+		$this->sessionID = \bin2hex(\random_bytes(20));
 		
-		// create new session hash
-		$sessionID = bin2hex(\random_bytes(20));
+		// Create new session.
+		$sql = "INSERT INTO wcf".WCF_N."_".($this->isACP ? 'acp' : 'user')."_session
+				(sessionID, ipAddress, userAgent, lastActivityTime, sessionVariables)
+			VALUES
+				(?, ?, ?, ?, ?)";
```
Gravatronics, line 540:
```suggestion
		$sql = "INSERT INTO     wcf".WCF_N."_".($this->isACP ? 'acp' : 'user')."_session
			                (sessionID, ipAddress, userAgent, lastActivityTime, sessionVariables)
			VALUES          (?, ?, ?, ?, ?)";
```

wcfsetup/install/files/lib/system/session/SessionHandler.class.php
TimWolla: see: https://github.com/WoltLab/WCF/pull/3593#discussion_r503777640

wcfsetup/install/files/lib/system/session/SessionHandler.class.php
TimWolla: It reads correctly to me. Is the following phrasing better: “To leverage that JavaScript framework feature the author needs to …”?

wcfsetup/install/files/lib/data/user/User.class.php
```diff
@@ -142,49 +144,46 @@ public function __construct($id, $row = null, DatabaseObject $object = null) {
...
-				$rebuild = true;
-			}
-			
-			// password is correct
-			if (\hash_equals($this->password, PasswordUtil::getDoubleSaltedHash($password, $this->password))) {
-				$isValid = true;
-			}
+		$manager = PasswordAlgorithmManager::getInstance();
+		
+		// Compatibility for WoltLab Suite < 5.4.
```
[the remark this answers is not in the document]
TimWolla: No, this is required forever (or at least until all password have a prefix).

wcfsetup/install/files/lib/data/user/User.class.php
```diff
@@ -142,49 +144,46 @@ public function __construct($id, $row = null, DatabaseObject $object = null) {
...
+		$isValid = $algorithm->verify($password, $hash);
+		
+		if (!$isValid) {
+			return false;
+		}
+		
+		$defaultAlgorithm = $manager->getDefaultAlgorithm();
+		if (\get_class($algorithm) !== \get_class($defaultAlgorithm) ||
+			$algorithm->needsRehash($hash)
+		) {
```
TimWolla, line 170:
```suggestion
		if (\get_class($algorithm) !== \get_class($defaultAlgorithm) || $algorithm->needsRehash($hash)) {
```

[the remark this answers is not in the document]
TimWolla: Indeed. I wonder what that previously looked like.

wcfsetup/install/files/lib/data/acp/session/ACPSessionEditor.class.php
```diff
@@ -46,9 +47,7 @@ public function update(array $parameters = []) {
 	}
 	
 	/**
-	 * Deletes active sessions of the given users.
-	 * 
-	 * @param	integer[]	$userIDs
+	 * @deprecated 5.4 - Sessions are managed via the SessionHandler.
```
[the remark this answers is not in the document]
TimWolla: Good call. I'm not sure whether I should make this a no-op as well.

wcfsetup/install/files/lib/form/LoginForm.class.php
```diff
@@ -62,8 +41,6 @@ public function assignVariables() {
 		parent::assignVariables();
 		
 		WCF::getTPL()->assign([
-			'useCookies' => $this->useCookies,
-			'supportsPersistentLogins' => UserAuthenticationFactory::getInstance()->getUserAuthentication()->supportsPersistentLogins(),
```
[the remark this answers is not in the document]
TimWolla: Yes, that's probably a good idea. It should be `false`, though.

wcfsetup/install/files/lib/system/session/SessionHandler.class.php
TimWolla: Technically yes. This cookie is somewhat special, though (lack of prefix).

IMO cookie handling in general is in the need for a refactoring, because the method in HeaderUtil takes a gazillion parameters.

wcfsetup/install/files/lib/system/session/SessionHandler.class.php
Gravatronics: Oh, I now understand the sentence. It should read "Leveraging *this* JavaScript framework feature requires", then.

wcfsetup/install/files/lib/system/session/SessionHandler.class.php
```diff
@@ -339,9 +312,54 @@ protected function defineConstants() {
...
 	 */
 	protected function initSecurityToken() {
-		if ($this->getVar('__SECURITY_TOKEN') === null) {
-			$this->register('__SECURITY_TOKEN', bin2hex(\random_bytes(20)));
+		$xsrfToken = '';
+		if (!empty($_COOKIE['XSRF-TOKEN'])) {
+			// We intentionally do not extract the signed value and instead just verify the correctness.
+			//
+			// The reason is that common JavaScript frameworks can use the contents of the `XSRF-TOKEN` cookie as-is,
+			// without performing any processing on it, improving interoperability. Leveraging that JavaScript framework
```
TimWolla, line 320:
```suggestion
			// without performing any processing on it, improving interoperability. Leveraging this JavaScript framework
```

wcfsetup/install/files/lib/system/session/SessionHandler.class.php
Gravatronics:
```php
if ($this->isACP) {
    $lifetime = self::ACP_SESSION_LIFETIME;
}
else if ($row['userID']) {
    $lifetime = self::USER_SESSION_LIFETIME;
}
else {
    $lifetime = self::GUEST_SESSION_LIFETIME;
}
```

I do not see why this formatting would have any downside.

TimWolla: - It's more than twice as long, causing less of the method to fit on screen.
- It repeats the `$lifetime =` variable for each case, putting more words on the screen.
- It requires diagonal scans to check the condition against the value, whereas the ternary is formatted in a nice columnar laylout.

PR #3574, opened 2020-09-24
Remove legacy autologin functionality

TimWolla:
After this pull request is merged the legacy autologin functionality will be removed:

- Checkbox during login is gone.
- New cookies won't be created.
- Old cookies won't be read.


wcfsetup/install/files/lib/system/user/authentication/DefaultUserAuthentication.class.php
```diff
@@ -15,18 +15,17 @@
  */
 class DefaultUserAuthentication extends AbstractUserAuthentication {
 	/**
-	 * @inheritDoc
+	 * @deprecated 5.4 - This method always returns true, as user sessions are long-lived now.
 	 */
 	public function supportsPersistentLogins() {
 		return true;
 	}
```
TimWolla, line 22: Should this rather `return false;`? Reviewing this PR myself makes me wonder, because `storeAccessData` and `loginAutomatically` are noops anyway.

dtdesign: From a strictly logically perspective this method should return `true`. From a backwards compatibility standpoint it should be `false`, because it prevents legacy code from accessing certain code paths that might no longer be applicable.

If we can safely use `false` here, then it should be that value, simply for best compatibility.

dtdesign (approved):

wcfsetup/install/files/lib/system/user/authentication/DefaultUserAuthentication.class.php
TimWolla: Yes, `false` should be safe here, as the method is deprecated now (thus it should not longer be used within our code).

PR #3583, opened 2020-09-30
Remove session validation options

TimWolla:
- Remove session_validate_ip_address option
- Remove session_validate_user_agent option
- Remove SessionHandler::validate()


TimWolla (commented):
Should a compatibility `define()` be added to `WCF::loadOptions()` for the super slim chance that anyone uses these or do we not care here?

dtdesign:
Yes, please hardwire those constants with sane values.

TimWolla:
Okay, I've made the change and will squash them into the proper commits during the final rebase.

dtdesign (approved):

PR #3584, opened 2020-09-30
Set alt attribut for user avatars

TitusKirch:
Set alt attribut for user avatars to wcf.user.avatar


TitusKirch:
Currently it is correct according to https://validator.w3.org, but several SEO tools report that you should set the alt attribute. This might be a suitable solution here.

TimWolla:
The `alt` attribute was intentionally removed in commit 27509f81777b69aa4d5393cc4407f3b4b95a795e. Avatars are not essential content for users with impaired vision. See also: https://community.woltlab.com/thread/276983-accessibility-verbessern/?postID=1746616#post1746616

PR #3589, opened 2020-10-02
Check for the "jpeg" and "png" support of GD

### LINKED ISSUE #3568: Detect GD's jpeg support in SystemCheckPage

dtdesign:
Closes #3568

### DIFF
--- a/wcfsetup/install/files/acp/templates/systemCheck.tpl
+++ b/wcfsetup/install/files/acp/templates/systemCheck.tpl
@@ -104,6 +104,25 @@
 			<small>{lang}wcf.acp.systemCheck.php.sha256.description{/lang}</small>
 		</dd>
 	</dl>
+	
+	<dl{if !$results[php][gd][result]} class="formError"{/if}>
+		<dt>{lang}wcf.acp.systemCheck.php.gd{/lang}</dt>
+		<dd>
+			{if $results[php][gd][result]}
+				{@$statusOk} {lang}wcf.acp.systemCheck.pass{/lang}
+			{else}
+				<ul class="nativeList">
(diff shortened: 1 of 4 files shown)


wcfsetup/install/files/acp/templates/systemCheck.tpl
```diff
@@ -104,6 +104,25 @@
...
+	
+	<dl{if !$results[php][gd][result]} class="formError"{/if}>
+		<dt>{lang}wcf.acp.systemCheck.php.gd{/lang}</dt>
+		<dd>
+			{if $results[php][gd][result]}
+				{@$statusOk} {lang}wcf.acp.systemCheck.pass{/lang}
+			{else}
+				<ul class="nativeList">
+					{if !$results[php][gd][jpeg]}
+						<li>{@$statusInsufficient} <kbd>jpeg</kbd></li>
```
Gravatronics, line 116: Why are you using `<kdb>` here but quotation marks in `wcf.acp.systemCheck.php.gd.description`? (Same applies to the next `<li>`.)

Gravatronics (commented):
By manually changing the code in the PHP file to test different scenarios, I found that there is no visual difference between `gd_info()` not being available and it being available but support for both `jpeg` and `png` being missing (not sure if this is a relevant case).

wcfsetup/install/files/acp/templates/systemCheck.tpl
dtdesign: I kept it consistent to the list of missing PHP extensions.

dtdesign:
I'm not overly concerned by this, it will report the missing `gd` extension anyway, so this is a bit redundant at worst.

PR #3593, opened 2020-10-05
Implement the new session handling 

TimWolla:
First draft of the new session logic. There are still a few TODOs and it needs extensive testing, but the tests done so far look very good.


wcfsetup/install/files/lib/system/WCF.class.php
```diff
@@ -590,8 +593,8 @@ protected function initApplications() {
 				$application->__run();
 			}
 			
-			// refresh the session 1 minute before it expires
-			self::getTPL()->assign('__sessionKeepAlive', SESSION_TIMEOUT - 60);
+			/** @deprecated The below variable is deprecated. */
+			self::getTPL()->assign('__sessionKeepAlive', 60 * 50);
```
dtdesign, line 597:
```suggestion
			self::getTPL()->assign('__sessionKeepAlive', 59 * 60);
```

This matches the previous default value for consistency. Also changed the order to represent `<minutes> * <seconds>`.

wcfsetup/install/files/lib/system/session/SessionHandler.class.php
```diff
@@ -147,18 +152,30 @@ class SessionHandler extends SingletonFactory {
...
 	 */
 	public function __get($key) {
-		if (isset($this->environment[$key])) {
+		if ($key === 'sessionID') {
+			return $this->sessionID;
+		}
+		else if ($key === 'userID') {
+			return $this->user->userID;
+		}
+		// TODO: pageID, pageObjectID, parentPageID, parentPageObjectID
```
dtdesign, line 172: Create a tracking issue _if_ this will not be part of this PR.

wcfsetup/install/files/lib/system/session/SessionHandler.class.php
```diff
@@ -249,39 +255,20 @@ public function initSession() {
...
 	/**
 	 * Initializes environment variables.
 	 */
 	protected function initEnvironment() {
 		$this->environment = [
-			'lastRequestURI' => $this->session->requestURI,
-			'lastRequestMethod' => $this->session->requestMethod,
+		//	TODO:
+		//	'lastRequestURI' => $this->session->requestURI,
+		//	'lastRequestMethod' => $this->session->requestMethod,
```
dtdesign, line 265: See above, create a tracking issue _if_ this will not be addressed in this PR.

wcfsetup/install/files/lib/system/session/SessionHandler.class.php
```diff
@@ -405,46 +370,114 @@ public function getUser() {
...
+		$row = $statement->fetchSingleRow();
+		
+		if (!$row) {
+			return false;
+		}
+		
+		// Check whether the session technically already expired.
+		$lifetime = ($this->isACP ? self::ACP_SESSION_LIFETIME :
+			($row['userID'] ? self::USER_SESSION_LIFETIME :
+			self::GUEST_SESSION_LIFETIME));
```
dtdesign, line 393:
```suggestion
		$lifetime = $row['userID'] ? self::USER_SESSION_LIFETIME : self::GUEST_SESSION_LIFETIME;
		if ($this->isACP) $lifetime = self::ACP_SESSION_LIFETIME;
```
This is only a suggestion, not an actual issue. I usually try to avoid nested shorthand, because I feel they are easily misinterpreted at a first glance.

wcfsetup/install/files/lib/system/session/SessionHandler.class.php
```diff
@@ -405,46 +370,114 @@ public function getUser() {
...
+		if ($row['lastActivityTime'] < (TIME_NOW - $lifetime)) {
+			return false;
+		}
+		
+		$this->sessionID = $sessionID;
+		$this->user = new User($row['userID']);
+		$this->variables = unserialize($row['sessionVariables']);
```
dtdesign, line 400:
```suggestion
		$this->variables = @unserialize($row['sessionVariables']);
		if (!is_array($this->variables)) $this->variables = [];
```

Rationale: Serialized values become magically corrupted in shared hosting environments, better discard the value than bricking this session.

wcfsetup/install/files/lib/system/session/SessionHandler.class.php
```diff
@@ -405,46 +370,114 @@ public function getUser() {
...
+		$statement->execute([
+			UserUtil::getIpAddress(),
+			UserUtil::getUserAgent(),
+			TIME_NOW,
+			$this->sessionID,
+		]);
+		
+		// Refresh cookie.
+		if ($this->user->userID) {
+			HeaderUtil::setCookie(($this->isACP ? 'acp' : 'user')."_session", $this->sessionID, TIME_NOW + 86400 * 14);
```
dtdesign, line 417: This appears to send a cookie with every response?

wcfsetup/install/files/lib/system/session/SessionHandler.class.php
```diff
@@ -405,46 +370,114 @@ public function getUser() {
...
 		}
 		
-		$this->user = new User($this->session->userID);
+		// Fetch legacy session.
+		if (!$this->isACP) {
+			$sql = "SELECT	*
+				FROM	wcf".WCF_N."_session
+				WHERE	(userID IS NULL AND sessionID = ?)
+				OR	(userID IS NULL AND spiderID = ?)
+				OR	(userID IS NOT NULL AND userID = ?)";
```
dtdesign, line 426: This query is quite awkward.

wcfsetup/install/files/lib/system/session/SessionHandler.class.php
```diff
@@ -801,17 +753,58 @@ public function delete() {
...
+			TIME_NOW - self::GUEST_SESSION_LIFETIME,
+		]);
+		
+		$sql = "DELETE FROM	wcf".WCF_N."_user_session
+			WHERE		lastActivityTime < ?
+				AND	userID IS NOT NULL";
+		$statement = WCF::getDB()->prepareStatement($sql);
+		$statement->execute([
+			TIME_NOW - self::USER_SESSION_LIFETIME,
+		]);
```
dtdesign, line 799: Please combine these two queries to improve the performance. In situations with extreme amounts of session this query has caused significant bottlenecks through locks.

dtdesign (changes requested):

wcfsetup/install/files/lib/system/session/SessionHandler.class.php
TimWolla: Yes, because with long lived sessions of 14 days after the last activity we need to take care that the cookie also lives for these 14 days. Storing some timestamp to only update the cookie once per hour or so feels fragile.

wcfsetup/install/files/lib/system/session/SessionHandler.class.php
TimWolla: Elaborate, please? Is it the formatting? Is it the `WHERE` condition?

dtdesign: The query is quite generic because it tries all kinds of conditions, even though we know much more ahead of time. If `$row['userID']` does not contain a user id, then this is either a search engine or a regular guest. If it contains an user id, we don't care for the first two conditions at all.

TimWolla: So I basically add an `if ($row['userID'])` to the PHP code and send a different query in the *then* and *else* part, yes?

wcfsetup/install/files/lib/system/session/SessionHandler.class.php
dtdesign: That's quite extreme and I have never seen a site behave like this. Why do we even care, just let the cookie live until eternity. The session id is either valid or not, it makes no difference what the browser presents to us.

wcfsetup/install/files/lib/system/session/SessionHandler.class.php
dtdesign: The logic goes like this:

1. If `$row['userID']` then only check for sessions with an exact match for `userID`.
2. If `$this->getSpiderID(…)` returns an id, then only check by `spiderID`.
3. Else check for the session id.

It might make sense to use a condition builder here for ease of logic.

wcfsetup/install/files/lib/system/session/SessionHandler.class.php
TimWolla: > That's quite extreme and I have never seen a site behave like this.

GitHub sends an updated session cookie with every request. Laravel does as well (by default?).

> Why do we even care, just let the cookie live until eternity.

I wouldn't be surprised if browsers start to restrict the maximum age of a cookie. IMO it's also questionable to store an “eternity cookie” on a users device when the user might not visit the community any longer. Also even an eternity cookie will expire one day, leading to surprise logouts. Personally I'm always annoyed when I'm suddenly logged out, because one year passed since I logged in last and my autologin cookies expired. This is especially annoying on my phone.

dtdesign: Restricting the maximum age? That move would easily break half of the internet, so even if there is an intention to limit them, it will take place over a long period.

I don't really like the idea of sending cookies with every request, it increases the payload, especially for requests that have (close to) no response body. If you are really concerned, we can simply set the cookie to 2x the session lifetime and embed a timestamp/date which refreshes the cookie if it is older than 7 days or something. This makes sure that as long as there is a session, the cookie is still around and we only refresh the cookie once per week.

wcfsetup/install/files/lib/system/session/SessionHandler.class.php
TimWolla: I've split this into user / guest only. The query optimizer will handle the `spiderID = NULL` case just fine.

wcfsetup/install/files/lib/system/session/SessionHandler.class.php
TimWolla: I agree that the nested if-expression is not super readable. I find your suggestion harder to understand, because the logic  is reversed. The winning case (ACP) comes last.

I just tried a full-blown if-else if-else, but that's super bloated. A dedicated method *could* work, but that requires the reader to jump around in the file. I'm trying to keep this code simple and straight forward for auditing reasons.

I've now committed a somewhat unorthodox suggestion with regard to the code style. But I find it very readable. What do you think?

dtdesign (approved):

wcfsetup/install/files/lib/system/session/SessionHandler.class.php
TimWolla: I honestly don't see this happening. Especially since the `sessionVariables` are now properly stored in a `BLOB` column. But I'd rather drop the whole session instead of pretending everything is fine, even if it is not.

wcfsetup/install/files/lib/system/session/SessionHandler.class.php
```diff
@@ -147,18 +125,40 @@ class SessionHandler extends SingletonFactory {
...
+		if ($key === 'sessionID') {
+			return $this->sessionID;
 		}
+		else if ($key === 'userID') {
+			return $this->user->userID;
+		}
+		// TODO: pageID, pageObjectID, parentPageID, parentPageObjectID
 		
-		return $this->session->{$key};
+		/** @deprecated	5.4 - These values can be retrieved more efficiently by directly using the methods in e.g. UserUtil. */
```
TimWolla, line 147: Would it be helpful for IDE use if I define these as public properties and immediately mark them as deprecated?

dtdesign: You could have just kept the `protected $enviroment` variable for this purpose, I don't see a benefit of moving it here?

wcfsetup/install/files/lib/system/session/SessionHandler.class.php
```diff
@@ -147,18 +125,40 @@ class SessionHandler extends SingletonFactory {
...
+			'requestMethod' => !empty($_SERVER['REQUEST_METHOD']) ? substr($_SERVER['REQUEST_METHOD'], 0, 7) : '',
+			'spiderID' => $this->getSpiderID(UserUtil::getUserAgent()),
+			'lastActivityTime' => TIME_NOW,
+		];
+		
+		if (array_key_exists($key, $environment)) {
+			return $environment[$key];
+		}
+		
+		throw new \BadMethodCallException("Unknown key '".$key."'");
```
dtdesign, line 161:
```suggestion
		return $environment[$key] ?? null;
```

Your change modified the behavior of the method which would previously returned `null` for unknown keys.

dtdesign (changes requested):

wcfsetup/install/files/lib/system/session/SessionHandler.class.php
TimWolla: I moved the stuff to slim the whole property list down. I don't believe marking the `$environment` property as deprecated would have helped with the following:

```
WCF::getSession()->userAgent // accesses the magic getter
```

The deprecation of `->environment` cannot magically extend to `->userAgent`. However I realize that adding public properties prevents the getter from being called :thinking:

TimWolla: Hacking around with a trait works according to: https://stackoverflow.com/a/51890536/782822

wcfsetup/install/files/lib/system/session/SessionHandler.class.php
```diff
@@ -136,29 +143,36 @@ final class SessionHandler extends SingletonFactory {
...
+				return !empty($_SERVER['REQUEST_METHOD']) ? substr($_SERVER['REQUEST_METHOD'], 0, 7) : '';
+			case 'lastActivityTime':
+				return TIME_NOW;
+			
+			default:
+				return null;
 		}
 		
-		throw new \BadMethodCallException("Unknown key '".$key."'");
+		return null;
```
dtdesign: Just remove these two lines, the `default` causes this code to be unreachable.

dtdesign (changes requested):

wcfsetup/install/files/lib/system/template/plugin/CsrfTokenPrefilterTemplatePlugin.class.php
```diff
@@ -0,0 +1,26 @@
...
+	/**
+	 * @inheritDoc
+	 */
+	public function execute($templateName, $sourceContent, TemplateScriptingCompiler $compiler) {
+		$getToken = '$__wcf->session->getSecurityToken()';
+		
+		return strtr($sourceContent, [
+			'{csrfToken type=raw}' => sprintf('{%s}', $getToken),
+			'{csrfToken type=url}' => sprintf('{%s|rawurlencode}', $getToken),
+			'{csrfToken}' => sprintf('<input type="hidden" name="t" value="{%s}">', $getToken),
```
dtdesign, line 23:
```suggestion
			'{csrfToken type=raw}' => sprintf('{@%s}', $getToken),
			'{csrfToken type=url}' => sprintf('{@%s|rawurlencode}', $getToken),
			'{csrfToken}' => sprintf('<input type="hidden" name="t" value="{@%s}">', $getToken),
```

Rationale: We have full control over the value of the security token, thus we can safe some CPU cycles here.

dtdesign (changes requested):

PR #3594, opened 2020-10-05
Fix 'Undefined index: additionalData'

Cyperghost:


TimWolla:
Can you give some information under which circumstances this happens?

Cyperghost:
> Can you give some information under which circumstances this happens?

If you create an new object of Box, but with wrong id, for example
`$box = new Box(1234);` 
But we don´t have an box in our system wie the id `1234`

TimWolla:
Okay. Thank you. Can you rebase your commit onto 3.1 and add the additional explanation to the commit message or would you like me to handle this?

Cyperghost:
> Okay. Thank you. Can you rebase your commit onto 3.1 and add the additional explanation to the commit message or would you like me to handle this?

If you have time, it would be great if you could do it. Otherwise I will do it later.

Cyperghost:
As it looks like it is only a problem with WSC5.2 but not with WSC3.1. In WSC3.1 and 3.0, there is no crash.

TimWolla:
Okay, I'll leave this then for the time being and let someone else have a look.

Cyperghost:
> Okay, I'll leave this then for the time being and let someone else have a look.

In WSC3.1 the error message is suppressed.

https://github.com/WoltLab/WCF/blob/0b5ee19c8b51ea6f6c9635110d89bbfe5b180548/wcfsetup/install/files/lib/data/box/Box.class.php#L139

TimWolla:
Ah. I see. Thanks, applied the patch to 5.2 with an adjusted commit message.

PR #3595, opened 2020-10-07
Fixed regex in item lists

ilouHD:
Added escape character before `$` to match the expected purpose of the variable `{$objectId}` in `submitFieldName` in ItemList and StaticItemList

### DIFF
--- a/wcfsetup/install/files/js/WoltLabSuite/Core/Ui/ItemList.js
+++ b/wcfsetup/install/files/js/WoltLabSuite/Core/Ui/ItemList.js
@@ -110,7 +110,7 @@ define(['Core', 'Dictionary', 'Language', 'Dom/Traverse', 'EventKey', 'WoltLabSu
 							for (var i = 0, length = values.length; i < length; i++) {
 								input = elCreate('input');
 								input.type = 'hidden';
-								input.name = options.submitFieldName.replace(/\{\$objectId\}/, values[i].objectId);
+								input.name = options.submitFieldName.replace('{$objectId}', values[i].objectId);
 								input.value = values[i].value;
 								
 								form.appendChild(input);
--- a/wcfsetup/install/files/js/WoltLabSuite/Core/Ui/ItemList.js
+++ b/wcfsetup/install/files/js/WoltLabSuite/Core/Ui/ItemList.js
@@ -110,7 +110,7 @@ define(['Core', 'Dictionary', 'Language', 'Dom/Traverse', 'EventKey', 'WoltLabSu
 							for (var i = 0, length = values.length; i < length; i++) {
 								input = elCreate('input');
 								input.type = 'hidden';
-								input.name = options.submitFieldName.replace(/{$objectId}/, values[i].objectId);
+								input.name = options.submitFieldName.replace(/{\$objectId}/, values[i].objectId);
 								input.value = values[i].value;
 								
 								form.appendChild(input);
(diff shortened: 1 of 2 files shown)


wcfsetup/install/files/js/WoltLabSuite/Core/Ui/ItemList.js
```diff
@@ -110,7 +110,7 @@ define(['Core', 'Dictionary', 'Language', 'Dom/Traverse', 'EventKey', 'WoltLabSu
 							for (var i = 0, length = values.length; i < length; i++) {
 								input = elCreate('input');
 								input.type = 'hidden';
-								input.name = options.submitFieldName.replace(/{$objectId}/, values[i].objectId);
+								input.name = options.submitFieldName.replace(/{\$objectId}/, values[i].objectId);
```
TimWolla, line 113: Will the braces characters also need to be encoded? They have a special meaning in a regular expression as well.

ilouHD: Curly braces do have a special meaning as well, thats true. But they only quantify a group or equal in a regex. In this case here, the above regex matches exactly the string `{$objectId}`. But I can add an escape-character here too, so that will never be a problem. Will add it.

wcfsetup/install/files/js/WoltLabSuite/Core/Ui/ItemList.js
```diff
@@ -110,7 +110,7 @@ define(['Core', 'Dictionary', 'Language', 'Dom/Traverse', 'EventKey', 'WoltLabSu
 							for (var i = 0, length = values.length; i < length; i++) {
 								input = elCreate('input');
 								input.type = 'hidden';
-								input.name = options.submitFieldName.replace(/{$objectId}/, values[i].objectId);
+								input.name = options.submitFieldName.replace(/\{\$objectId\}/, values[i].objectId);
```
dtdesign, line 113:
```suggestion
								input.name = options.submitFieldName.replace('{$objectId}', values[i].objectId);
```

wcfsetup/install/files/js/WoltLabSuite/Core/Ui/ItemList/Static.js
```diff
@@ -88,7 +88,7 @@ define(['Core', 'Dictionary', 'Language', 'Dom/Traverse', 'EventKey', 'Ui/Simple
 							for (var i = 0, length = values.length; i < length; i++) {
 								input = elCreate('input');
 								input.type = 'hidden';
-								input.name = options.submitFieldName.replace(/{$objectId}/, values[i].objectId);
+								input.name = options.submitFieldName.replace(/\{\$objectId\}/, values[i].objectId);
```
dtdesign, line 91:
```suggestion
								input.name = options.submitFieldName.replace('{$objectId}', values[i].objectId);
```

dtdesign (changes requested):
Just use a string value instead. It behaves exactly the same, but without the need for escaping that reduces the readability.

wcfsetup/install/files/js/WoltLabSuite/Core/Ui/ItemList.js
TimWolla: Seeing this: Is there a need to perform multiple replacements? Then we'll need a regular expression + the `/g` modifier.

dtdesign: No, it is fine, this is a single `name` attribute.

ilouHD:
Haven't think on that, but that makes sense :)

dtdesign:
Changing the base didn't really work out, therefore I manually applied the changes to the other branches. Thanks for your help!

PR #3602, opened 2020-10-13
Update defaultStyle after migrating styles

TimWolla:
It appears that importing the defaultStyle causes the style specific asset folder to be created, leading to issues with the style migration.


TimWolla:
This line probably is at fault, because `->extract()` automatically creates the full path:

https://github.com/WoltLab/WCF/blob/dabe9fd236cae88b85c920be42ef1a0eaf50ac77/wcfsetup/install/files/lib/data/style/StyleEditor.class.php#L621

dtdesign (approved):

Issue #3606, opened 2020-10-14
Adjust SECURITY_TOKEN_* to {csrfToken}


TimWolla:
Depends on #3593

TimWolla:
Either fixed or created follow-up issues for everything I could find.

PR #3612, opened 2020-10-19
Backport CsrfTokenFunctionTemplatePlugin

TimWolla:
see 3f6a261b1e6a3804370eb1e2a046ea6c666dbedd
see #3606

### DIFF
--- a/wcfsetup/install/files/lib/system/template/plugin/CsrfTokenFunctionTemplatePlugin.class.php
+++ b/wcfsetup/install/files/lib/system/template/plugin/CsrfTokenFunctionTemplatePlugin.class.php
@@ -1,0 +1,39 @@
+<?php
+namespace wcf\system\template\plugin;
+use wcf\system\exception\SystemException;
+use wcf\system\template\TemplateEngine;
+use wcf\system\WCF;
+
+/**
+ * Template compiler plugin that prints the CSRF token ("Security Token").
+ * 
+ * Usage:
+ * 	{csrfToken}
+ * 	{csrfToken type=raw}
+ * 	{csrfToken type=url}
+ * 
+ * @author	Tim Duesterhus
+ * @copyright	2001-2020 WoltLab GmbH
+ * @license	GNU Lesser General Public License <http://opensource.org/licenses/lgpl-license.php>
+ * @package	WoltLabSuite\Core\System\Template\Plugin
+ */
+class CsrfTokenFunctionTemplatePlugin implements IFunctionTemplatePlugin {
+	/**
+	 * @inheritDoc
+	 */
+	public function execute($tagArgs, TemplateEngine $tplObj) {
+		$token = WCF::getSession()->getSecurityToken();
+		$type = $tagArgs['type'] ?? 'form';
+		
+		switch ($type) {
+			case 'raw':
+				return $token;
+			case 'url':
+				return \rawurlencode($token);
+			case 'form':
+				return \sprintf('<input type="hidden" name="t" value="%s">', $token);
+			default:
+				throw new SystemException("Invalid type '".$type."' given.");
+		}
+	}
+}


wcfsetup/install/files/lib/system/template/plugin/CsrfTokenFunctionTemplatePlugin.class.php
```diff
@@ -0,0 +1,38 @@
...
+ * @license	GNU Lesser General Public License <http://opensource.org/licenses/lgpl-license.php>
+ * @package	WoltLabSuite\Core\System\Template\Plugin
+ */
+class CsrfTokenFunctionTemplatePlugin implements IFunctionTemplatePlugin {
+	/**
+	 * @inheritDoc
+	 */
+	public function execute($tagArgs, TemplateEngine $tplObj) {
+		if (isset($tagArgs['type']) && $tagArgs['type'] === 'raw') {
+			return \wcf\system\WCF::getSession()->getSecurityToken();
```
Gravatronics, line 26: Why are you not importing the class? (also below)

wcfsetup/install/files/lib/system/template/plugin/CsrfTokenFunctionTemplatePlugin.class.php
```diff
@@ -0,0 +1,38 @@
...
+		}
+		else if (isset($tagArgs['type']) && $tagArgs['type'] === 'url') {
+			return \rawurlencode(\wcf\system\WCF::getSession()->getSecurityToken());
+		}
+		else if (!isset($tagArgs['type']) || $tagArgs['type'] === 'form') {
+			return sprintf('<input type="hidden" name="t" value="%s">', \wcf\system\WCF::getSession()->getSecurityToken());
+		}
+		else {
+			throw new SystemException("Invalid type '".$tagArgs['type']."' given.");
```
Gravatronics, line 35: Replace `SystemException` with a build-in exception?

wcfsetup/install/files/lib/system/template/plugin/CsrfTokenFunctionTemplatePlugin.class.php
```diff
@@ -0,0 +1,38 @@
...
+		}
```
Gravatronics, line 36: Untested alternative code structure making the code easier to read IMO.

```suggestion
		$securityToken = WCF::getSession()->getSecurityToken();
		
		$type = $tagArgs['type'] ?? 'form';
		switch ($type) {
			case 'raw':
				return $securityToken;
				
			case 'url':
				return \rawurlencode($securityToken);
				
			case 'form':
				return sprintf('<input type="hidden" name="t" value="%s">', $securityToken);
				
			default:
				throw new SystemException("Invalid type '{$type}' given.");
		}
```

Gravatronics (changes requested):

wcfsetup/install/files/lib/system/template/plugin/CsrfTokenFunctionTemplatePlugin.class.php
TimWolla: Historic reasons. I've initially used a BlockTemplatePlugin which allowed me to directly emit PHP code, but it came with a few drawbacks. That's why I opted for Function + Prefilter and didn't adjust this.

I'll fix this in session-meta and then update this PR.

wcfsetup/install/files/lib/system/template/plugin/CsrfTokenFunctionTemplatePlugin.class.php
TimWolla: For consistency with other template stuff (and because I'm not sure how the system would react) I'm using SystemException here.

wcfsetup/install/files/lib/system/template/plugin/CsrfTokenFunctionTemplatePlugin.class.php
TimWolla: Yes, that looks good to me, thanks.

dtdesign (approved):

Gravatronics (approved):

PR #3615, opened 2020-10-19
Fix endless loop in removeExifData when the APP1 signature does not match

TimWolla:


TimWolla:
Ah crap. I initially committed to 5.2 before deciding to create a branch and subsequently forgot to remove the commit from 5.2 … Let's just hope that the fix is sufficient and works fine :sweat_smile:

Issue #3623, opened 2020-10-21
Adjust HttpFactory for PSR-18 / Guzzle 7


TimWolla:

TimWolla:
Not going to do this. The PSR-17 API is *much* less fun to use, because it does not follow redirects by itself: https://github.com/guzzle/guzzle/blob/34bc199aef77ff1c13f7126c0464f8fa9ee55f9f/src/Client.php#L134

It's not expected that Guzzle is replaced by anything else anyway.

PR #3624, opened 2020-10-21
Use Guzzle instead of FileUtil::downloadFileFromHttp

joshuaruesweg:


wcfsetup/install/files/lib/system/cronjob/RefreshSearchRobotsCronjob.class.php
```diff
@@ -15,15 +16,23 @@
  * @package	WoltLabSuite\Core\System\Cronjob
  */
 class RefreshSearchRobotsCronjob extends AbstractCronjob {
+	/**
+	 * URL to the spider list.
+	 */
+	const SPIDER_LIST_URL = 'http://assets.woltlab.com/spiderlist/typhoon/list.xml';
```
TimWolla, line 22: Consider adding an access modifier here. I suggest `private`:
```suggestion
	private const SPIDER_LIST_URL = 'http://assets.woltlab.com/spiderlist/typhoon/list.xml';
```

wcfsetup/install/files/lib/system/cronjob/RefreshSearchRobotsCronjob.class.php
```diff
@@ -15,15 +16,23 @@
...
 		parent::execute($cronjob);
 		
-		$filename = FileUtil::downloadFileFromHttp('http://assets.woltlab.com/spiderlist/typhoon/list.xml', 'spiders');
+		$client = HttpFactory::getDefaultClient();
+		$request = new Request('GET', self::SPIDER_LIST_URL);
+		$response = $client->send($request);
+		
 		$xml = new XML();
-		$xml->load($filename);
+		$xml->loadXML('list.xml', $response->getBody()->getContents());
```
TimWolla, line 35:
```suggestion
		$xml->loadXML('list.xml', (string) $response->getBody());
```
Reason: The PSR-7 API is not well-designed in this regard. `->getContents()` returns the *remaining* contents. This most likely is all of the contents here, but a bit fragile and misleading. `->__toString()` effectively is a `->rewind()` + `->getContents()` and thus is safer.

wcfsetup/install/files/lib/system/package/PackageArchive.class.php
```diff
@@ -426,13 +429,21 @@ protected function filterUpdateInstructions() {
...
 	public function downloadArchive() {
-		$prefix = 'package';
+		$tmpFile = FileUtil::getTemporaryFilename('package_');
 		
-		// file transfer via hypertext transfer protocol.
-		$this->archive = FileUtil::downloadFileFromHttp($this->archive, $prefix);
+		$client = HttpFactory::getDefaultClient();
+		$request = new Request('GET', $this->archive, [
+			'stream' => true,
+		]);
```
TimWolla, line 437:
```suggestion
		$request = new Request('GET', $this->archive);
```

Reason: Guzzle automatically reads the response into a `fopen('php://temp')` stream. This one automatically spills over to disk anyway and thus will not overflow the memory.

wcfsetup/install/files/lib/system/package/PackageArchive.class.php
```diff
@@ -426,13 +429,21 @@ protected function filterUpdateInstructions() {
 	 * @return	string		path to the dowloaded file
 	 */
 	public function downloadArchive() {
-		$prefix = 'package';
+		$tmpFile = FileUtil::getTemporaryFilename('package_');
```
TimWolla, line 432: Should you delete this file in an error case?

TimWolla (changes requested):

wcfsetup/install/files/lib/data/user/avatar/Gravatar.class.php
```diff
@@ -98,13 +99,20 @@ public function getURL($size = null) {
...
 	public static function test($email) {
 		$gravatarURL = sprintf(self::GRAVATAR_BASE, md5(mb_strtolower($email)), 80, GRAVATAR_DEFAULT_TYPE);
+		$client = HttpFactory::getDefaultClient();
+		$request = new Request('GET', $gravatarURL);
+		
 		try {
-			$tmpFile = FileUtil::downloadFileFromHttp($gravatarURL, 'gravatar');
-			@unlink($tmpFile);
-			return true;
+			$response = $client->send($request);
```
TimWolla, line 106:
```suggestion
			$response = $client->sendRequest($request);
```

`sendRequest` is the standardized PSR-7 method. Apply to the other locations as well, please.

wcfsetup/install/files/lib/system/cronjob/RefreshSearchRobotsCronjob.class.php
joshuaruesweg: What is the advantage of this, except that it can no longer be read from the outside? This is not a critical value.

TimWolla: > except that it can no longer be read from the outside?

Exactly that. Anything that is `public` or `protected` is part of the public API. For better flexibility we should strive to not expose stuff that does not need to be exposed.

wcfsetup/install/files/lib/data/user/avatar/Gravatar.class.php
joshuaruesweg: I will postpone this change to: https://github.com/WoltLab/WCF/issues/3623

wcfsetup/install/files/lib/system/package/PackageArchive.class.php
joshuaruesweg: No, I would not like to change that in the pull-request, but only with a new one. Up to now the file was not deleted either (possibly also for debugging reasons).

TimWolla (approved):
LGTM

PR #3626, opened 2020-10-22
Remove PHP 7.0 compatibility code for Reflection

TimWolla:


dtdesign (approved):

wcfsetup/install/files/lib/system/form/builder/data/processor/CustomFormDataProcessor.class.php
```diff
@@ -64,12 +64,8 @@ public function __construct($id, callable $formDataProcessor = null, callable $o
 			/** @var \ReflectionType $parameterType */
 			$parameterType = $parameters[0]->getType();
 			if (!(
-				// PHP 7.1+
 				($parameterType instanceof \ReflectionNamedType &&
-					($parameterType->getName() === IFormDocument::class || is_subclass_of($parameterType->getName(), IFormDocument::class))) ||
-				// PHP 7.0
-				(get_class($parameterType) === \ReflectionType::class &&
-					($parameterType->__toString() === IFormDocument::class || is_subclass_of($parameterType->__toString(), IFormDocument::class)))
+					($parameterType->getName() === IFormDocument::class || is_subclass_of($parameterType->getName(), IFormDocument::class)))
```
Gravatronics, line 68: You should be able to remove a pair of parentheses from the overall condition. (Same applies to other changes.)

Gravatronics (changes requested):

wcfsetup/install/files/lib/system/form/builder/data/processor/CustomFormDataProcessor.class.php
TimWolla: This is intentional for 2 reasons:
- The possible addition of union types in the check.
- The formatting across multiple lines which invite the reader to add something to it, accidentally breaking the logic.

PR #3627, opened 2020-10-22
Update composer dependencies

TimWolla:
- Update to ezyang/htmlpurifier 4.13.*
- Tighten up minimum version of scssphp/scssphp

pelago/emogrifier is missing for now, because it introduces backwards incompatible changes by removing the Emogrifier class.


TimWolla:
The removal of the Purifier files is caused by: https://github.com/ezyang/htmlpurifier/pull/257. Someone needs to check whether we are using these “extras” or not.

TimWolla:
> Someone needs to check whether we are using these “extras” or not.

I eyeballed the diff and did a basic functional test. Everything appears to work fine.

PR #3630, opened 2020-10-23
Remove DatabaseTableUtil

TimWolla:


Gravatronics (approved):
The file should also be explicitly deleted during the update process.

TimWolla:
> The file should also be explicitly deleted during the update process.

I created a follow-up issue for that.

PR #3633, opened 2020-10-23
Add flood control API

### LINKED ISSUE #3033: New API to track certain actions utilizing a rolling window
Several components already feature some sort of a flood control, but they are limited by the time since the last content was created. A new API should be introduced that can be used to track these dates and define multiple levels of rolling windows:
- Number of messages within a certain time frame.
- Time since the last message of the same type.

Gravatronics:
These changes only add the new API but do not employ it yet.

Close #3033

### DIFF
--- a/wcfsetup/install/files/lib/system/flood/FloodControl.class.php
+++ b/wcfsetup/install/files/lib/system/flood/FloodControl.class.php
@@ -17,6 +17,7 @@ use wcf\util\UserUtil;
  */
 final class FloodControl extends SingletonFactory {
 	private const PRUNE_TIME = TIME_NOW - 31 * 86400;
+	private const IDENTIFIER_PREFIX = WCF_UUID . ':' . self::class . ':';
 	
 	/**
 	 * Returns the number of contents by a certain identifier type within a certain
@@ -90,7 +91,7 @@ final class FloodControl extends SingletonFactory {
 		return \hash_hmac(
 			'md5',
 			'guest:' . $ipAddress,
-			'wcf' . WCF_N. '_flood_log' . $objectType,
+			self::IDENTIFIER_PREFIX . $objectType,
 			true
 		);
 	}
@@ -163,7 +164,7 @@ final class FloodControl extends SingletonFactory {
 		return \hash_hmac(
 			'md5',
 			'user:' . $userID,
-			'wcf' . WCF_N. '_flood_log' . $objectType,
+			self::IDENTIFIER_PREFIX . $objectType,
 			true
 		);
 	}
--- a/wcfsetup/install/files/lib/system/flood/FloodControl.class.php
+++ b/wcfsetup/install/files/lib/system/flood/FloodControl.class.php
@@ -20,8 +20,8 @@ final class FloodControl extends SingletonFactory {
 	private const IDENTIFIER_PREFIX = WCF_UUID . ':' . self::class . ':';
 	
 	/**
-	 * Returns the number of contents by a certain identifier type within a certain
-	 * period of time `($time-$interval, $time]` and the earliest time within the period content
+	 * Returns the number of contents by a certain identifier type within a certain period of
+	 * time `($time-$interval, $time]` and the earliest time within the period at which content
 	 * was created.
 	 */
 	private function countContentByIdentifier(string $objectType, string $identifier, \DateInterval $interval, int $time): array {
@@ -97,8 +97,8 @@ final class FloodControl extends SingletonFactory {
 	}
 	
 	/**
-	 * Returns the last time a guest/user created content of the given type or `null` if they
-	 * have not created such content.
+	 * Returns the last time a guest/user with the given identifier created content of the given
+	 * type or `null` if they have not created such content within the stored period of time.
 	 */
 	private function getLastTimeByIdentifier(string $objectType, string $identifier): ?int {
 		$sql = "SELECT          time
--- a/wcfsetup/install/files/lib/system/flood/FloodControl.class.php
+++ b/wcfsetup/install/files/lib/system/flood/FloodControl.class.php
@@ -43,8 +43,8 @@ final class FloodControl extends SingletonFactory {
 	
 	/**
 	 * Returns the number of contents the active user created of the given type within a certain
-	 * period of time `($time-$interval, $time]` and the earliest time within the period content
-	 * was created.
+	 * period of time `($time-$interval, $time]` and the earliest time within the period at which
+	 * content was created.
 	 */
 	public function countContent(string $objectType, \DateInterval $interval, int $time = TIME_NOW): array {
 		if (WCF::getUser()->userID) {
@@ -57,8 +57,8 @@ final class FloodControl extends SingletonFactory {
 	
 	/**
 	 * Returns the number of contents a guest created of the given type within a certain period
-	 * of time `($time-$interval, $time]` and the earliest time within the period content was
-	 * created.
+	 * of time `($time-$interval, $time]` and the earliest time within the period at which
+	 * content was created.
 	 */
 	public function countGuestContent(string $objectType, string $ipAddress, \DateInterval $interval, int $time = TIME_NOW): array {
 		return $this->countContentByIdentifier(
@@ -71,8 +71,8 @@ final class FloodControl extends SingletonFactory {
 	
 	/**
 	 * Returns the number of contents a user created of the given type within a certain period
-	 * of time `[$time-$interval, $time]` and the earliest time within the period content was
-	 * created.
+	 * of time `[$time-$interval, $time]` and the earliest time within the period at which
+	 * content was created.
 	 */
 	public function countUserContent(string $objectType, int $userID, \DateInterval $interval, int $time = TIME_NOW): array {
 		return $this->countContentByIdentifier(
--- a/wcfsetup/install/files/lib/system/flood/FloodControl.class.php
+++ b/wcfsetup/install/files/lib/system/flood/FloodControl.class.php
@@ -115,8 +115,8 @@ final class FloodControl extends SingletonFactory {
 	}
 	
 	/**
-	 * Returns the last time a guest created content of the given type or `null` if they have not
-	 * created such content.
+	 * Returns the last time a guest with the given ip address created content of the given type
+	 * or `null` if they have not created such content within the stored period of time.
 	 */
 	public function getGuestLastTime(string $objectType, string $ipAddress): ?int {
 		return $this->getLastTimeByIdentifier(
@@ -169,8 +169,8 @@ final class FloodControl extends SingletonFactory {
 	}
 	
 	/**
-	 * Returns the last time a user created content of the given type or `null` if they have not
-	 * created such content.
+	 * Returns the last time a user with the given id created content of the given type or
+	 * `null` if they have not created such content within the stored period of time.
 	 */
 	public function getUserLastTime(string $objectType, int $userID): ?int {
 		return $this->getLastTimeByIdentifier(
--- a/wcfsetup/install/files/lib/system/flood/FloodControl.class.php
+++ b/wcfsetup/install/files/lib/system/flood/FloodControl.class.php
@@ -3,6 +3,7 @@ namespace wcf\system\flood;
 use wcf\data\object\type\ObjectTypeCache;
 use wcf\system\SingletonFactory;
 use wcf\system\WCF;
+use wcf\util\DateUtil;
 use wcf\util\UserUtil;
 
 /**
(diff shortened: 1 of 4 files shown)


wcfsetup/install/files/lib/system/flood/FloodControl.class.php
```diff
@@ -0,0 +1,241 @@
...
+ * Flood control tracks the times contents are created by users.
+ * 
+ * @author      Matthias Schmidt
+ * @copyright   2001-2020 WoltLab GmbH
+ * @license     GNU Lesser General Public License <http://opensource.org/licenses/lgpl-license.php>
+ * @package     WoltLabSuite\Core\System\Flood
+ * @since       5.4
+ */
+class FloodControl extends SingletonFactory {
+	const PRUNE_TIME = TIME_NOW - 31 * 86400;
```
TimWolla, line 19:
```suggestion
	private const PRUNE_TIME = TIME_NOW - 31 * 86400;
```

wcfsetup/install/files/lib/system/flood/FloodControl.class.php
```diff
@@ -0,0 +1,241 @@
...
+/**
+ * Flood control tracks the times contents are created by users.
+ * 
+ * @author      Matthias Schmidt
+ * @copyright   2001-2020 WoltLab GmbH
+ * @license     GNU Lesser General Public License <http://opensource.org/licenses/lgpl-license.php>
+ * @package     WoltLabSuite\Core\System\Flood
+ * @since       5.4
+ */
+class FloodControl extends SingletonFactory {
```
TimWolla, line 18:
```suggestion
final class FloodControl extends SingletonFactory {
```

Prefer making classes `final` where no clear need for extension exists. Removing the `final` is easily possible, the inverse is not.

When making the class `final` also adjust `protected` to `private`.

wcfsetup/install/files/lib/system/flood/FloodControl.class.php
```diff
@@ -0,0 +1,241 @@
...
+	const PRUNE_TIME = TIME_NOW - 31 * 86400;
+	
+	/**
+	 * Returns the number of contents by a certain identifier type within a certain
+	 * period of time `[$time-$interval, $time]` and the earliest time within the period content
+	 * was created.
+	 */
+	protected function countContentByIdentifier(string $objectType, string $identifier, \DateInterval $interval, int $time): array {
```
TimWolla, line 26:
```suggestion
	protected function countContentByIdentifier(string $objectType, string $identifier, \DateInterval $interval, \DateTime $time): array {
```

More strongly typed, especially since you call `getDateTimeByTimestamp` anyway?

wcfsetup/install/files/lib/system/flood/FloodControl.class.php
```diff
@@ -0,0 +1,241 @@
...
+		$sql = "SELECT  COUNT(*) AS count, MIN(time) AS earliestTime
+			FROM    wcf" . WCF_N . "_flood_control
+			WHERE   objectTypeID = ?
+			        AND identifier = ?
+			        AND time >= ?
+			        AND time <= ?";
```
TimWolla, line 32: Should this be a closed interval? I would expect an half-open interval:

I'd also like to suggest a slight reformatting to make the interval more clear:
```suggestion
			        AND (? <= time AND time < ?)";
```

wcfsetup/install/files/lib/system/flood/FloodControl.class.php
```diff
@@ -0,0 +1,241 @@
...
+		
+		return $statement->fetchSingleRow();
+	}
+	
+	/**
+	 * Returns the number of contents the active user created of the given type within a certain
+	 * period of time `[$time-$interval, $time]` and the earliest time within the period content
+	 * was created.
+	 */
+	public function countContent(string $objectType, \DateInterval $interval, int $time = TIME_NOW): array {
```
TimWolla, line 49: Talking about “contents” feels awkward and limiting with regard to use-cases (consider limiting reactions which do not really fit the term “content”). How about any of the following:

- Hits
- Triggers

wcfsetup/install/files/lib/system/flood/FloodControl.class.php
```diff
@@ -0,0 +1,241 @@
...
+	
+	/**
+	 * Returns the identifier used for a guest with the given ip address for content of the
+	 * given object type.
+	 */
+	protected function getGuestIdentifier(string $objectType, string $ipAddress): string {
+		return \hash_hmac(
+			'md5',
+			'guest:' . $ipAddress,
+			'wcf' . WCF_N. '_flood_log' . $objectType,
```
TimWolla, line 94:
```suggestion
			'wcf1_flood_log:' . $objectType,
```

I removed the `WCF_N` to keep it simple. The purpose of the key is to not allow obvious cross-correlation across multiple tables. Having `WCF_N` != 1 is pretty unlikely, making the code more complex than necessary.

wcfsetup/install/files/lib/system/flood/FloodControl.class.php
Gravatronics: With the current implementation, the `\DateTime` object is only created once in this method but with this change, each calling methods would have to change the timestamp to a `\DateTime` object.

wcfsetup/install/files/lib/system/flood/FloodControl.class.php
Gravatronics: At the moment, it is quite easy to search for `wcf1` in the code to find instances where `WCF_N` was not used by accident. Until `WCF_N` is automatically replaced in queries, I would also find it surprising to find a hardcoded `wcf1`.

TimWolla: Is `$objectType` the full object type identifier here? Then simply using the `$objectType` should be sufficiently unique. But the following would also make sense to me:

```
static::class.':'.$objectType
```

With an optional prefix of `WCF_UUID` for uniqueness across installations.

wcfsetup/install/files/lib/system/flood/FloodControl.class.php
dtdesign: I'm fine with "content", it is generic enough and to some extent imply that its UGC. There is a thin line between something being "content" and not, for instance, imagine reactions could have an optional comment?

wcfsetup/install/files/lib/system/flood/FloodControl.class.php
```diff
@@ -0,0 +1,241 @@
...
+	 * Returns the number of contents by a certain identifier type within a certain
+	 * period of time `($time-$interval, $time]` and the earliest time within the period content
+	 * was created.
+	 */
+	private function countContentByIdentifier(string $objectType, string $identifier, \DateInterval $interval, int $time): array {
+		$sql = "SELECT  COUNT(*) AS count, MIN(time) AS earliestTime
+			FROM    wcf" . WCF_N . "_flood_control
+			WHERE   objectTypeID = ?
+			        AND identifier = ?
+			        AND (time > ? AND time <= ?)";
```
TimWolla, line 32:
```suggestion
			        AND (? < time AND time <= ?)";
```
Personal preference: I consider this more readable. Feel free to ignore.

wcfsetup/install/files/lib/system/flood/FloodControl.class.php
```diff
@@ -0,0 +1,241 @@
...
+ * @since       5.4
+ */
+final class FloodControl extends SingletonFactory {
+	private const PRUNE_TIME = TIME_NOW - 31 * 86400;
+	private const IDENTIFIER_PREFIX = WCF_UUID . ':' . self::class . ':';
+	
+	/**
+	 * Returns the number of contents by a certain identifier type within a certain
+	 * period of time `($time-$interval, $time]` and the earliest time within the period content
+	 * was created.
```
TimWolla, line 25:
```suggestion
	 * period of time `($time-$interval, $time]` and the smallest logged timestamp within that period.
```
I stumbled over the “period content” part.

wcfsetup/install/files/lib/system/flood/FloodControl.class.php
```diff
@@ -0,0 +1,241 @@
...
+			'md5',
+			'guest:' . $ipAddress,
+			self::IDENTIFIER_PREFIX . $objectType,
+			true
+		);
+	}
+	
+	/**
+	 * Returns the last time a guest/user created content of the given type or `null` if they
+	 * have not created such content.
```
TimWolla, line 101:
```suggestion
	 * Returns the maximum timestamp for the given type or `null` if the identifier did not create such content within the stored timeframe.
```

wcfsetup/install/files/lib/system/flood/FloodControl.class.php
```diff
@@ -0,0 +1,241 @@
...
+	 */
+	private function getLastTimeByIdentifier(string $objectType, string $identifier): ?int {
+		$sql = "SELECT          time
+			FROM            wcf" . WCF_N . "_flood_control
+			WHERE           objectTypeID = ?
+			                AND identifier = ?
+			ORDER BY        time DESC";
+		$statement = WCF::getDB()->prepareStatement($sql, 1);
```
TimWolla, line 109: Check the query plan, but I expect the suggestion to look better or same with it being more readable.

```suggestion
		$sql = "SELECT          MAX(time)
			FROM            wcf" . WCF_N . "_flood_control
			WHERE           objectTypeID = ?
			                AND identifier = ?";
		$statement = WCF::getDB()->prepareStatement($sql);
```

wcfsetup/install/files/lib/system/flood/FloodControl.class.php
```diff
@@ -0,0 +1,241 @@
...
+			$time,
+		]);
+		
+		return $statement->fetchSingleRow();
+	}
+	
+	/**
+	 * Returns the number of contents the active user created of the given type within a certain
+	 * period of time `($time-$interval, $time]` and the earliest time within the period content
+	 * was created.
```
TimWolla, line 47: Adjust the phrasing as well. Or maybe simply refer back to `countContentByIdentifier`? Not sure which method should be documented, either `countContentByIdentifier` or the public one (this).

Same applies to all the other methods.

wcfsetup/setup/db/install.sql
```diff
@@ -1952,6 +1962,8 @@ ALTER TABLE wcf1_edit_history_entry ADD FOREIGN KEY (obsoletedByUserID) REFERENC
 
 ALTER TABLE wcf1_event_listener ADD FOREIGN KEY (packageID) REFERENCES wcf1_package (packageID) ON DELETE CASCADE;
 
+ALTER TABLE wcf1_flood_control ADD FOREIGN KEY (objectTypeID) REFERENCES wcf1_object_type (objectTypeID) ON DELETE CASCADE;
```
TimWolla, line 1965: The more I look at this key the less I like it:

1. It can easily create an deletion storm on package uninstallation.
2. It makes the `INSERT`s slower, because it needs to check `wcf1_object_type`.

As entries in `wcf1_flood_control` expiry naturally we probably can leave this FK out.

wcfsetup/setup/db/install.sql
```diff
@@ -1962,8 +1964,6 @@ ALTER TABLE wcf1_edit_history_entry ADD FOREIGN KEY (obsoletedByUserID) REFERENC
 
 ALTER TABLE wcf1_event_listener ADD FOREIGN KEY (packageID) REFERENCES wcf1_package (packageID) ON DELETE CASCADE;
 
-ALTER TABLE wcf1_flood_control ADD FOREIGN KEY (objectTypeID) REFERENCES wcf1_object_type (objectTypeID) ON DELETE CASCADE;
```
TimWolla: I'm not sure whether the comment would be better located down here.

wcfsetup/install/files/lib/system/flood/FloodControl.class.php
```diff
@@ -0,0 +1,240 @@
...
+			$this->getObjectTypeID($objectType),
+			$identifier,
+		]);
+		
+		return $statement->fetchSingleColumn();
+	}
+	
+	/**
+	 * Returns the last time a guest created content of the given type or `null` if they have not
+	 * created such content.
```
TimWolla, line 119:
```suggestion
	 * created such content within the stored period of time.
```
Consistency with the `getLastTimeByIdentifier` you adjusted. Applies below as well.

TimWolla (approved):
I don't see the need for any further functional changes. I trust that you are able to adjust comments. LGTM now.

wcfsetup/install/files/lib/system/flood/FloodControl.class.php
```diff
@@ -0,0 +1,240 @@
...
+	
+	/**
+	 * Returns the number of contents a guest created of the given type within a certain period
+	 * of time `($time-$interval, $time]` and the earliest time within the period at which
+	 * content was created.
+	 */
+	public function countGuestContent(string $objectType, string $ipAddress, \DateInterval $interval, int $time = TIME_NOW): array {
+		return $this->countContentByIdentifier(
+			$objectType,
+			$this->getUserIdentifier($objectType, $ipAddress),
```
joshuaruesweg, line 66: Wrong method?

PR #3636, opened 2020-10-23
Remove in-application gzip compression

### LINKED ISSUE #3634: Remove the gzipping logic
Compression of the output is best performed by the web server and not the application, for several reasons:

- Less code equals less bugs.
  - Web servers contain a mature implementation that is able to handle the various quirks within user-agents.
- Easier debugging
  - The gzip compression was a constant source of confusion until it learned to not gzip when an Exception was handled.
  - strace between web server and FPM is readable in plain text.
- Less CPU time burnt in PHP.
- Possibility for the web server to automagically use better compression algorithms, such as Brotli.

The biggest drawback is that it increases the amount of data exchanged between web server and FPM, increasing the number of syscalls to read / write the data on the internal socket. The benefits far outweigh this drawback, though.

I verified that the following software combinations on Ubuntu 20.04 compress the output of a simple PHP script:

- Apache 2.4 + mod_php 7.4
- Apache 2.4 + FPM 7.4
- nginx 1.18 + FPM 7.4

TimWolla:
Resolves #3634

### DIFF
--- a/com.woltlab.wcf/option.xml
+++ b/com.woltlab.wcf/option.xml
@@ -604,11 +604,6 @@ private:wcf.acp.option.exception_privacy.private</selectoptions>
 				<optiontype>boolean</optiontype>
 				<defaultvalue>1</defaultvalue>
 			</option>
-			<option name="http_enable_gzip">
-				<categoryname>general.system.http</categoryname>
-				<optiontype>boolean</optiontype>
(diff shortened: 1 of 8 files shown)


TimWolla:
I'll resolve the Merge Conflict right before merge.

dtdesign (approved):

PR #3644, opened 2020-10-24
Add like support check to comment responses

Krymonota:
If the provided comment manager doesn't support likes, it's not possible to like comments, as can be seen here:

- https://github.com/WoltLab/WCF/blob/6747fe291c584d7c8295912f8d94297588d142b2/com.woltlab.wcf/templates/commentList.tpl#L59
- https://github.com/WoltLab/WCF/blob/6747fe291c584d7c8295912f8d94297588d142b2/com.woltlab.wcf/templates/commentList.tpl#L59

However, it's never checked for comment responses whether the comment manager supports likes. This results in the like button being displayed for comment responses, even though likes aren't supported.


joshuaruesweg:
Thanks!

PR #3654, opened 2020-10-28
Use new flood control API for comments

Gravatronics:
See #3033


wcfsetup/install/files/lib/data/comment/CommentAction.class.php
```diff
@@ -13,9 +13,11 @@
 use wcf\system\captcha\CaptchaHandler;
 use wcf\system\comment\manager\ICommentManager;
 use wcf\system\comment\CommentHandler;
+use wcf\system\exception\NamedUserException;
```
TimWolla, line 16: Useless import?

TimWolla (approved):

PR #3665, opened 2020-10-29
update versions to 5.4

mutec:


com.woltlab.wcf/package.xml
```diff
@@ -49,16 +49,4 @@
 		
 		<instruction type="script">acp/post_install.php</instruction>
 	</instructions>
-	
-	<instructions type="update" fromversion="5.3.*">
```
TimWolla, line 53: This change is bogus.

TimWolla (commented):
Thanks, but the version number is intentionally not yet adjusted.

- It creates more merge conflicts after new 5.3 releases.
- It prevents easy installation of packages that exclude 5.4 for testing purposes.
- Related to the previous: The package servers for 5.4 are not yet available either.

PR #3680, opened 2020-11-02
Converted JS files to TypeScript + backwards compatibility for inheritance

dtdesign:
n/t


wcfsetup/install/files/ts/WoltLabSuite/Core/Ui/ItemList/Static.ts
```diff
@@ -0,0 +1,443 @@
...
+    if (event.target === list) {
+      element.focus();
+    }
+  });
+
+  const listItem = document.createElement("li");
+  listItem.className = "input";
+  list.appendChild(listItem);
+
+  element.addEventListener("keydown", (ev: KeyboardEvent) => keyDown(ev));
```
TimWolla, line 75: Why do you need to add the type here? TypeScript *should* be able to automatically derive it.

wcfsetup/install/files/js/WoltLabSuite/Core/Ui/ItemList/Static.js
```diff
@@ -1,424 +1,346 @@
...
- * @copyright	2001-2019 WoltLab GmbH
- * @license	GNU Lesser General Public License <http://opensource.org/licenses/lgpl-license.php>
- * @module	WoltLabSuite/Core/Ui/ItemList/Static
+ * @author  Alexander Ebert, Matthias Schmidt
+ * @copyright  2001-2019 WoltLab GmbH
+ * @license  GNU Lesser General Public License <http://opensource.org/licenses/lgpl-license.php>
+ * @module  WoltLabSuite/Core/Ui/ItemList/Static
  */
-define(['Core', 'Dictionary', 'Language', 'Dom/Traverse', 'EventKey', 'Ui/SimpleDropdown'], function (Core, Dictionary, Language, DomTraverse, EventKey, UiSimpleDropdown) {
+define(["require", "exports", "tslib", "../../Core", "../../Dom/Traverse", "../../Language", "../Dropdown/Simple"], function (require, exports, tslib_1, Core, DomTraverse, Language, Simple_1) {
```
TimWolla, line 9: I'm unable to sanely review the diff of this .js file, it's just too huge.

TimWolla (approved):
Apart from the one remark this LGTM.

wcfsetup/install/files/ts/WoltLabSuite/Core/Ui/ItemList/Static.ts
dtdesign: For some reason PhpStorm is unable to determine the type because of the type for the union type. If it move the type union into the argument list, it works, but using the `ItemListInputElement` type bricks it.

It's a bit verbose, but prevents my IDE from lighting up like a christmas tree.

TimWolla: That's a good rationale.

PR #3682, opened 2020-11-02
Menu item add

### LINKED ISSUE #3678: Direct link to menu item add form in the list of menus
n/t

ghost:
Close #3678

I moved the menu item list icon also to the end so that there is a consistent order: first the two icons related to the menu itself (editing/deleting) and then the two icons related to its items (list/adding).

Additionally, the menu name is shown on the menu item add/edit form to make it obvious to which menu the item is being added/to which menu the edited item belongs.


BurntimeX (approved):
LGTM

PR #3700, opened 2020-11-06
Convert the comment (response) modules to TypeScript

dtdesign:
n/t


wcfsetup/install/files/ts/WoltLabSuite/Core/Ui/Notification.ts
```diff
@@ -51,7 +53,7 @@ function hide() {
 /**
  * Displays a notification.
  */
-export function show(message?: string, callback?: Callback, cssClassName?: string): void {
+export function show(message?: string, callback?: Callback | null, cssClassName?: string): void {
```
TimWolla, line 56:
```suggestion
export function show(message: string = "wcf.global.success", callback: Callback | null = null, cssClassName: string = "success"): void {
```

and then remove the handling within the method body.

wcfsetup/install/files/ts/WoltLabSuite/Core/Ui/Comment/Response/Edit.ts
```diff
@@ -0,0 +1,143 @@
...
+interface AjaxResponse {
+  actionName: string;
+  returnValues: {
+    message: string;
+    template: string;
+  };
+}
+
+class UiCommentResponseEdit extends UiCommentEdit {
+  protected readonly _responses = new Set<HTMLElement>();
```
TimWolla, line 27: Should this be a WeakSet?

wcfsetup/install/files/ts/WoltLabSuite/Core/Ui/Comment/Edit.ts
```diff
@@ -0,0 +1,329 @@
...
+  actionName: string;
+  returnValues: {
+    message: string;
+    template: string;
+  };
+}
+
+class UiCommentEdit {
+  protected _activeElement: HTMLElement | null = null;
+  protected readonly _comments = new Set<HTMLElement>();
```
TimWolla, line 32: Should this be a WeakSet?

wcfsetup/install/files/ts/WoltLabSuite/Core/Ui/Comment/Edit.ts
```diff
@@ -0,0 +1,329 @@
...
+  protected readonly _container: HTMLElement;
+  protected _editorContainer: HTMLElement | null = null;
+
+  /**
+   * Initializes the comment edit manager.
+   */
+  constructor(container: HTMLElement) {
```
TimWolla, line 39:
```suggestion
  constructor(private readonly _container: HTMLElement) {
```
(even though it's probably inconsistent by now)

dtdesign: It is redundant, because `this._container` is set in the parent constructor. Also I try to avoid mixing constructor property promotion with regular properties.

wcfsetup/install/files/ts/WoltLabSuite/Core/Ui/Notification.ts
dtdesign: This doesn't work well, because legacy callees might pass `null` instead of `undefined`.

PR #3718, opened 2020-11-11
Update packageSearchResultList.tpl

xopez:
Should fix https://community.woltlab.com/thread/286683-paketsuche-zeigt-falschen-string/


TimWolla:
Picked manually, because of the wrong target branch (master instead of 5.3). Please make sure to select the proper branch (the lowest version to fix). Thanks.

PR #3719, opened 2020-11-11
Update packageSearchResultList.tpl

ilouHD:
Fixed https://community.woltlab.com/thread/286683-paketsuche-zeigt-falschen-string/


felix-d1strict:
Duplicate of #3718

TimWolla:
Thanks for the contribution. I've taken the patch in #3718, because that one came earlier and neither was to be preferred on a technical merit.

PR #3721, opened 2020-11-11
Incorrect language variable changed.

felix-d1strict:
Should fix: https://community.woltlab.com/thread/286689-sprachvariable-im-acp-von-option-inhalte-von-externen-anbietern/


TimWolla:
Thanks. Picked manually (incorrect target branch) and slightly rephrased the commit message (“Fix typo” instead of “Update” in the subject line).

PR #3732, opened 2020-11-18
Support multi-factor authentication within ACP

TimWolla:
It needs a face lift, but it's working from a purely functional POV.


wcfsetup/install/files/lib/acp/form/MultifactorAuthenticationForm.class.php
```diff
@@ -0,0 +1,180 @@
...
+/**
+ * Represents the multi-factor authentication form.
+ *
+ * @author	Tim Duesterhus
+ * @copyright	2001-2020 WoltLab GmbH
+ * @license	GNU Lesser General Public License <http://opensource.org/licenses/lgpl-license.php>
+ * @package	WoltLabSuite\Core\Acp\Form
+ * @since	5.4
+ */
+class MultifactorAuthenticationForm extends AbstractFormBuilderForm {
```
TimWolla, line 26: I really want to avoid class inheritance across frontend and ACP. During development of MFA I realized that the fact that the frontend login inherits from the ACP login causes ugly control flow.

OTOH duplicated code is not nice either. Should I look into moving the methods into a trait?

wcfsetup/install/files/acp/templates/multifactorAuthentication.tpl
```diff
@@ -0,0 +1,44 @@
...
+				<ol class="boxMenu">
+					{foreach from=$setups item='_setup'}
+						<li{if $setup->getId() == $_setup->getId()} class="active"{/if}>
+							<a class="boxMenuLink" href="{link controller='MultifactorAuthentication' object=$_setup url=$redirectUrl}{/link}"><span class="boxMenuLinkTitle">{lang}wcf.user.security.multifactor.{$_setup->getObjectType()->objectType}{/lang}</span></a>
+						</li>
+					{/foreach}
+				</ol>
+			</nav>
+		</div>
+	</section>
```
dtdesign, line 20: This indentation looks off by one level.

wcfsetup/install/files/lib/acp/form/LoginForm.class.php
```diff
@@ -213,7 +213,7 @@ public function save() {
 	 */
 	protected function performRedirect(bool $needsMultifactor = false) {
 		if ($needsMultifactor) {
-			$this->url = LinkHandler::getInstance()->getControllerLink(MultifactorAuthenticationForm::class, [
+			$this->url = LinkHandler::getInstance()->getLink('MultifactorAuthentication', [
```
dtdesign, line 216: I don't like this change, what's the motivation behind this?

wcfsetup/install/files/lib/acp/form/MultifactorAuthenticationForm.class.php
dtdesign: Deriving from the frontend class (and vice versa) isn't a bad thing. Traits always feel like one more layer of indirection and drastically reduce readability in terms of code flow.

Just derive the forms from each other, none of the solutions are fine, but that one is the least awkward. Especially security releated code should not be duplicated, because it _will_ get out of sync.

dtdesign (changes requested):

wcfsetup/install/files/lib/acp/form/LoginForm.class.php
TimWolla: Using `getLink` will implicitly preserve the current environment. `getControllerLink` forces the environment of the class (in this case: the frontend).

wcfsetup/install/files/acp/templates/multifactorAuthentication.tpl
TimWolla: Ah, yes indeed, I'll fix that. However this page needs a full relayout anyway, because I simply copied over the frontend HTML which isn't really compatible with the ACP's CSS.

dtdesign (approved):

PR #3740, opened 2020-11-19
Clean up comments when removing moderation queues

### LINKED ISSUE #3707: Deleting moderation queues does not purge comments
This cause some strange side effects on MySQL < 8.0, if the top queues are removed and MySQL is restarted. MySQL before 8.0 did not actually store the `AUTO_INCREMENT` value for InnoDB tables, instead the largest id + 1 is used after an restart. This can cause moderation queue ids to be reused, causing comments for the "original" id to show up again.

dtdesign:
Fixes #3707

### DIFF
--- a/wcfsetup/install/files/acp/update_com.woltlab.wcf_5.2.10_orphanedComments.php
+++ b/wcfsetup/install/files/acp/update_com.woltlab.wcf_5.2.10_orphanedComments.php
@@ -12,7 +12,7 @@ $commentObjectTypeID = ObjectTypeCache::getInstance()->getObjectTypeIDByName(
 	"com.woltlab.wcf.moderation.queue"
 );
 
-$sql = "SELECT  objectID
+$sql = "SELECT  DISTINCT objectID
         FROM    wcf" . WCF_N . "_comment
         WHERE   objectTypeID = ?
         AND     objectID NOT IN (
@@ -21,7 +21,7 @@ $sql = "SELECT  objectID
                 )";
 $statement = WCF::getDB()->prepareStatement($sql);
 $statement->execute([$commentObjectTypeID]);
-$objectIDs = $statement->fetchList('objectID');
+$objectIDs = $statement->fetchAll(\PDO::FETCH_COLUMN);
 
 CommentHandler::getInstance()->deleteObjects(
 	"com.woltlab.wcf.moderation.queue",
--- a/wcfsetup/install/files/acp/update_com.woltlab.wcf_5.2.10_orphanedComments.php
+++ b/wcfsetup/install/files/acp/update_com.woltlab.wcf_5.2.10_orphanedComments.php
@@ -2,6 +2,7 @@
 namespace wcf\acp;
 
 use wcf\data\object\type\ObjectTypeCache;
+use wcf\system\comment\CommentHandler;
 use wcf\system\WCF;
 
 // Earlier versions did remove moderation queues without taking care of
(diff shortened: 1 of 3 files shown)


wcfsetup/install/files/lib/system/cronjob/ModerationQueueCronjob.class.php
```diff
@@ -45,6 +46,20 @@ public function execute(Cronjob $cronjob) {
...
+				"com.woltlab.wcf.moderation.queue"
+			);
+			
+			$conditions = new PreparedStatementConditionBuilder();
+			$conditions->add("objectTypeID = ?", [$commentObjectTypeID]);
+			$conditions->add("objectID IN (?)", [$queueIDs]);
+			
+			$sql = "DELETE FROM wcf".WCF_N."_comment " . $conditions;
+			$statement = WCF::getDB()->prepareStatement($sql);
+			$statement->execute($conditions->getParameters());
```
TimWolla, line 62:
```suggestion
			CommentHandler::getInstance()->deleteObjects('com.woltlab.wcf.moderation.queue', $queueIDs);
```
?

wcfsetup/install/files/acp/update_com.woltlab.wcf_5.2.10_orphanedComments.php
```diff
@@ -0,0 +1,21 @@
...
+use wcf\system\WCF;
+
+// Earlier versions did remove moderation queues without taking care of
+// comments associated with them.
+$commentObjectTypeID = ObjectTypeCache::getInstance()->getObjectTypeIDByName(
+	"com.woltlab.wcf.comment.commentableContent",
+	"com.woltlab.wcf.moderation.queue"
+);
+
+$sql = "DELETE FROM     wcf" . WCF_N . "_comment
```
TimWolla, line 14: Possibly also adjust this to use `CommentHandler`, but a bare query within the update script is acceptable to me.

TimWolla (changes requested):

wcfsetup/install/files/acp/update_com.woltlab.wcf_5.2.10_orphanedComments.php
TimWolla: Looking into `CommentHandler` it does quite a lot of operations, so the update script should also use `CommentHandler`.

wcfsetup/install/files/acp/update_com.woltlab.wcf_5.2.10_orphanedComments.php
```diff
@@ -0,0 +1,29 @@
...
+use wcf\system\WCF;
+
+// Earlier versions did remove moderation queues without taking care of
+// comments associated with them.
+$commentObjectTypeID = ObjectTypeCache::getInstance()->getObjectTypeIDByName(
+	"com.woltlab.wcf.comment.commentableContent",
+	"com.woltlab.wcf.moderation.queue"
+);
+
+$sql = "SELECT  objectID
```
TimWolla, line 15:
```suggestion
$sql = "SELECT  DISTINCT objectID
```
This makes the result set smaller in case of multiple comments for a single queue entry.

wcfsetup/install/files/acp/update_com.woltlab.wcf_5.2.10_orphanedComments.php
```diff
@@ -0,0 +1,29 @@
...
+        FROM    wcf" . WCF_N . "_comment
+        WHERE   objectTypeID = ?
+        AND     objectID NOT IN (
+                        SELECT  queueID
+                        FROM    wcf" . WCF_N . "_moderation_queue
+                )";
+$statement = WCF::getDB()->prepareStatement($sql);
+$statement->execute([$commentObjectTypeID]);
+$objectIDs = $statement->fetchList('objectID');
```
TimWolla, line 24:
```suggestion
$objectIDs = $statement->fetchAll(\PDO::FETCH_COLUMN);
```
I wasn't even aware that `fetchList()` existed.

TimWolla (approved):
The functionality is fine now. The remarks are not necessary to implement.

wcfsetup/install/files/acp/update_com.woltlab.wcf_5.2.10_orphanedComments.php
dtdesign: We should consider deprecating it then, I wasn't aware of `fetchAll()` for single column results.

PR #3752, opened 2020-11-25
Verweide -> Vermeide

netzhuffle:
Small language fix for German informal.


TimWolla:
Applied manually due to the incorrect target branch (master instead of 5.3) and slightly rephrased the commit message.

Thanks!

PR #3768, opened 2020-11-27
Convert the clipboard handler to TypeScript

dtdesign:
n/t


wcfsetup/install/files/ts/WoltLabSuite/Core/Controller/Clipboard.ts
```diff
@@ -0,0 +1,706 @@
...
+
+interface ContainerData {
+  checkboxes: HTMLCollectionOf<HTMLInputElement>;
+  element: HTMLElement;
+  markAll: HTMLInputElement | null;
+  markedObjectIds: Set<number>;
+}
+
+interface ItemData {
+  items: { [key: string]: ClipboardActionData };
```
TimWolla, line 36: I believe this could be:
```suggestion
  items: Record<string, ClipboardActionData>;
```

wcfsetup/install/files/ts/WoltLabSuite/Core/Controller/Clipboard.ts
```diff
@@ -0,0 +1,706 @@
...
+  [key: string]: number[];
+}
+
+interface AjaxResponse {
+  actionName: string;
+  returnValues: {
+    action: string;
+    items?: {
+      // They key is the `typeName`
+      [key: string]: ItemData;
```
TimWolla, line 64: Likewise this could be a record, possibly with a `type TypeName = string` as the key.

wcfsetup/install/files/js/WoltLabSuite/Core/Controller/Clipboard.js
```diff
@@ -1,201 +1,158 @@
 /**
  * Clipboard API Handler.
  *
- * @author	Alexander Ebert
- * @copyright	2001-2019 WoltLab GmbH
- * @license	GNU Lesser General Public License <http://opensource.org/licenses/lgpl-license.php>
- * @module	WoltLabSuite/Core/Controller/Clipboard
+ * @author  Alexander Ebert
+ * @copyright  2001-2019 WoltLab GmbH
+ * @license  GNU Lesser General Public License <http://opensource.org/licenses/lgpl-license.php>
```
TimWolla, line 6: Unfortunately this diff is too large to review :disappointed:

TimWolla (approved):
LGTM. The two remarks boil down to preference in the end.

wcfsetup/install/files/js/WoltLabSuite/Core/Controller/Clipboard.js
dtdesign: It was a bit of a nightmare, but I spent quite some time testing it, therefore I'm somewhat confident it should be fine.

wcfsetup/install/files/ts/WoltLabSuite/Core/Controller/Clipboard.ts
dtdesign: `Record<K, V>` will only be used when `K` is restricted to known values. `{ [key: string]: V }` is used for data structures with an unknown set of keys.

PR #3773, opened 2020-11-30
Convert the ACP page modules to TypeScript

dtdesign:
n/t


wcfsetup/install/files/js/WoltLabSuite/Core/Acp/Ui/Page/Menu.js
```diff
@@ -1,102 +1,107 @@
 /**
```
TimWolla, line 1: This one is not reviewable.

TimWolla (approved):

PR #3778, opened 2020-12-02
Use consistent phrasing

TimWolla:
- Consistently use 'Benutzerkonto' instead of 'Account' in de.xml
- Consistently use 'Kennwort' instead of 'Passwort' in de.xml


dtdesign (commented):
Why "Kennwort" over "Passwort"? Running a quick google search yields 14,5m results for "Kennwort", but "Passwort" yields 173m results. Even Wikipedia redirects to "Passwort" and pretty much every OS calls it "Passwort".

We should use the more common term "Passwort" instead.

TimWolla:
> Why "Kennwort" over "Passwort"?

It's the more commonly used phrase within WoltLab Suite. Most of the user visible phrases already use Kennwort, including the page titles which are not easily adjusted during an update: https://github.com/WoltLab/WCF/blob/31eeece8891d36a38fc50e0e8d961f9ef8a7d81d/com.woltlab.wcf/page.xml#L164-L191

dtdesign:
Windows/Microsoft appears to favor "Kennwort" and the page titles are a good reason to stick with it.

dtdesign (approved):

PR #3781, opened 2020-12-03
Remove reCAPTCHA v1 implementation

TimWolla:


com.woltlab.wcf/templates/recaptcha.tpl
```diff
@@ -1,100 +1,7 @@
...
-								recaptcha_response_field: Recaptcha.get_response()
-							};
-						});
-					});
-				</script>
-			{/if}
-		</dl>
-	</section>
-	{else}
+	{if RECAPTCHA_PUBLICKEY !== '' && RECAPTCHA_PRIVATEKEY !== ''}
```
dtdesign, line 4:
```suggestion
	{if RECAPTCHA_PUBLICKEY && RECAPTCHA_PRIVATEKEY}
```
Makes this easier to read.

wcfsetup/install/files/acp/templates/recaptcha.tpl
```diff
@@ -1,100 +1,7 @@
...
-								recaptcha_response_field: Recaptcha.get_response()
-							};
-						});
-					});
-				</script>
-			{/if}
-		</dl>
-	</section>
-	{else}
+	{if RECAPTCHA_PUBLICKEY !== '' && RECAPTCHA_PRIVATEKEY !== ''}
```
dtdesign, line 4:
```suggestion
	{if RECAPTCHA_PUBLICKEY && RECAPTCHA_PRIVATEKEY}
```

dtdesign (changes requested):

com.woltlab.wcf/templates/recaptcha.tpl
TimWolla: Okay, I applied this. I think it might even be possible to completely remove that condition, because the template probably is never even loaded without any keys. But I didn't want to make too large of a change so I just inverted the original condition and preserved the `{else}` part.

dtdesign (approved):

PR #3783, opened 2020-12-03
Add form field for active user's password

### LINKED ISSUE #3777: Add a form field that is suitable to verify the password of an existing user
See this PR for context: https://github.com/WoltLab/WCF/pull/3775/files#diff-d20cc4db175b2d1688513f529c6b259218fef11fe26dee20760c70bcfad8246cR59-R70

The `ReauthenticationForm` within that pull request is intended to verify the user's identity within the long running sessions of WoltLab Suite 5.4 to make sure that the user themselves and not someone accessing an unlocked computer is performing the action.

If the form builder successfully validates the form the user will be marked as authenticated for a specific period of time. As of right now the only verification method is entering the user's password. Due to the lack of a proper form field the PR uses a simple TextFormField with a custom validator that calls `->checkPassword()`. Future adjustments might or might not support reauthentication by a different means (e.g. by entering a valid multi-factor code).

The functionality of that PR is complete, apart from having a proper password field (and explanatory texts). The implementation of the field requested in this issue can be tested with PR #3775. Reauthentication can be triggered by accessing the multifactor management.

- [ ] At the very least it must be possible to create a form field with the appropriate `type="password"` attribute to make sure the password is hidden when entering it.
- [ ] Anything that goes further than this is nice to have:
     - [ ] `autocomplete="current-password"` (see also: #3728)
     - [ ] Automatically attaching a proper validator to the field
     - [ ] Support for rate limiting inputs / user authentication failure logging.

If additional functionality beyond the bare minimum is added then care should be taken to make sure that this functionality is generic enough. As an example different forms might have different rate limiting requirements. In addition users attached to a third party provider (`->authData != ''`) might not have a local password. It is not entirely clear whether that's something that needs to be handled by the field itself or by the field's user.

An existing reauthentication implementation that predates the ReauthenticationForm is part of AccountManagementForm. It would probably be helpful if both implementations are able to use the same validation logic as rate limiting is missing from AccountManagement as of right now: https://github.com/WoltLab/WCF/blob/a480c62ba78c3f1271b850a799d7b8af16486c7a/wcfsetup/install/files/lib/form/AccountManagementForm.class.php#L190-L199.

Gravatronics:
Close #3777

### DIFF
--- a/wcfsetup/install/files/lib/system/form/builder/field/user/UserPasswordField.class.php
+++ b/wcfsetup/install/files/lib/system/form/builder/field/user/UserPasswordField.class.php
@@ -1,0 +1,90 @@
+<?php
+namespace wcf\system\form\builder\field\user;
+use wcf\system\form\builder\field\AbstractFormField;
+use wcf\system\form\builder\field\IAutoFocusFormField;
+use wcf\system\form\builder\field\IPlaceholderFormField;
+use wcf\system\form\builder\field\TAutoFocusFormField;
+use wcf\system\form\builder\field\TDefaultIdFormField;
+use wcf\system\form\builder\field\TPlaceholderFormField;
+use wcf\system\form\builder\field\validation\FormFieldValidationError;
+use wcf\system\WCF;
+
+/**
+ * Implementation of a form field to enter the password of the active user.
+ * 
+ * This field is only available for logged-in users not user third party providers for logging in.
+ * This field uses the `wcf.user.password` language item as the default form field label and uses
+ * `password` as the default node id.
+ * 
+ * @author      Matthias Schmidt
+ * @copyright   2001-2020 WoltLab GmbH
+ * @license     GNU Lesser General Public License <http://opensource.org/licenses/lgpl-license.php>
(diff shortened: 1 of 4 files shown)


wcfsetup/install/files/lib/system/form/builder/field/user/UserPasswordField.class.php
```diff
@@ -0,0 +1,90 @@
...
+	
+	/**
+	 * @inheritDoc
+	 */
+	public function validate() {
+		if ($this->isRequired() && !$this->getValue()) {
+			$this->addValidationError(new FormFieldValidationError('empty'));
+		}
+		else if ($this->getValue() && !WCF::getUser()->checkPassword($this->getValue())) {
+			$this->addValidationError(
```
TimWolla, line 80: It be helpful for the user if the field is being cleared on validation errors. I'm doing the same for multi-factor codes.

```suggestion
			$this->value('');
			$this->addValidationError(
```

TimWolla (approved):
Overall this LGTM.

PR #3807, opened 2020-12-12
Greatly improve the performance of GIF processing with ImageMagick

dtdesign:
See https://community.woltlab.com/thread/285609-performance-engpass-bei-der-verarbeitung-von-gifs/

### DIFF
--- a/wcfsetup/install/files/lib/system/image/adapter/ImagickImageAdapter.class.php
+++ b/wcfsetup/install/files/lib/system/image/adapter/ImagickImageAdapter.class.php
@@ -349,25 +349,17 @@ class ImagickImageAdapter implements IImageAdapter {
 			throw new SystemException("Given image is not a valid Imagick-object.");
 		}
 		
-		// circumvent writeImages() bug in version 3.1.0 RC 1
-		if (phpversion('imagick') == '3.1.0RC1' && $this->supportsWritingAnimatedGIF) {
-			$file = fopen($filename, 'w');
-			$image->writeImagesFile($file);
-			fclose($file);
+		// Greatly reduces the time required to create the image and drastically
+		// reduces the filesize to more reasonable levels without a visible
+		// quality loss.
+		//
+		// See https://github.com/Imagick/imagick/issues/360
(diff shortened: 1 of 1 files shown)


SoftCreatR:
https://stackoverflow.com/questions/62481526/imagemagick-quantizeimage-100x-slower-on-newer-versions

dtdesign:
Thanks, but I could not verify the findings with recent ImageMagick versions. The run difference between enabling or disabling the dither falls into the statistical error.

Also running `identify` on the sample image I've used reports 269 frames. Rescaling them within 8 seconds is perfectly fine, each frame took around 30ms. This includes unpacking the GIF by reconstructing each frame, scaling each one and putting it all back together. 8s appears to be slow, but 30ms per frame on a notebook CPU with disabled turbo is perfectly fine.

SoftCreatR:
Alright. I just came across this, when trying to find out, what that method does.

wcfsetup/install/files/lib/system/image/adapter/ImagickImageAdapter.class.php
```diff
@@ -356,6 +356,16 @@ public function writeImage($image, $filename) {
 			fclose($file);
 		}
 		else {
```
TimWolla, line 358: Should the `if` part of the condition be handled as well?

TimWolla (approved):
LGTM other than that question which should result in a perfectly obvious change.

wcfsetup/install/files/lib/system/image/adapter/ImagickImageAdapter.class.php
dtdesign: We could simply remove that `if`, this is a 6 year old workaround for ancient PHP versions using ImageMagick 3.1.0RC1 from early 2012.

TimWolla: I'd be fine with that.

TimWolla (approved):

PR #3809, opened 2020-12-14
Make the ACP session access log more useful

### LINKED ISSUE #3537: ACP session log is next to useless
- [x] Anything AjaxProxyAction gives absolutely no insights into what was happening there.
- [x] The URI field appears to just show the actual GET parameters, thus stripping the ID that is part of the route (mostly just showing the security token which *is* something that should be stripped.

TimWolla:
This commit intentionally overloads the existing fields to keep the change
simple. The log is not really usuable for programmatic processing, due to the
existing constraints. The cleanest solution probably would be storing all the
loggable information as a JSON blob and then parsing that one when accessing
the log and querying it with MySQL's JSON function. But the latter are not
supported yet.

All in all this change improves the ACP session access log a lot and that is
what counts.

Resolves #3537

### DIFF
--- a/wcfsetup/install/files/lib/system/event/listener/SessionAccessLogListener.class.php
+++ b/wcfsetup/install/files/lib/system/event/listener/SessionAccessLogListener.class.php
@@ -1,5 +1,6 @@
 <?php
 namespace wcf\system\event\listener;
+use wcf\action\AJAXInvokeAction;
 use wcf\data\acp\session\access\log\ACPSessionAccessLogEditor;
 use wcf\data\acp\session\log\ACPSessionLog;
 use wcf\data\acp\session\log\ACPSessionLogEditor;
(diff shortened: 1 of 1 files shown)


wcfsetup/install/files/lib/system/event/listener/SessionAccessLogListener.class.php
```diff
@@ -51,22 +52,28 @@ public function execute($eventObj, $className, $eventName, array &$parameters) {
...
+				}
+			}
 			
 			// save access
 			ACPSessionAccessLogEditor::create([
 				'sessionLogID' => $sessionLogID,
 				'ipAddress' => UserUtil::getIpAddress(),
 				'time' => TIME_NOW,
-				'requestURI' => $requestURI,
+				'requestURI' => substr($requestURI, 0, 255),
```
dtdesign, line 74: This will cut off the request ID, if the request URI is really long. I would move the `substr` call to line 57-59 and add an additional `else` case to trim the "regular" URI.

dtdesign (changes requested):

wcfsetup/install/files/lib/system/event/listener/SessionAccessLogListener.class.php
```diff
@@ -55,7 +55,9 @@ public function execute($eventObj, $className, $eventName, array &$parameters) {
 			// Fetch request URI + request ID (if available).
 			$requestURI = UserUtil::getRequestURI();
 			if ($requestId = \wcf\getRequestId()) {
-				$requestURI .= ' ('.$requestId.')';
+				$requestIdSuffix = ' ('.$requestId.')';
+				// Ensure that the request ID fits by truncating the URI.
+				$requestURI = substr($requestURI, 0, 255 - strlen($requestIdSuffix) - 1).$requestIdSuffix;
```
dtdesign, line 60:
```suggestion
				$requestURI = substr($requestURI, 0, 255 - strlen($requestIdSuffix)).$requestIdSuffix;
```
Why `-1`? If the `$requestId` equals `a`, then the suffix is 4 characters long. `255 - 4 - 1 = 250`.

dtdesign (changes requested):

wcfsetup/install/files/lib/system/event/listener/SessionAccessLogListener.class.php
TimWolla: I'd blame programming in C, but that would probably be a lie. You're correct of course.

dtdesign (approved):
You can squash the commits if you like, your call.

PR #3813, opened 2020-12-16
Session Cookie Updates

TimWolla:
- Encode additional data into the session cookie
- Refresh the session cookie only once it aged a bit
- Remove obsolete import from LoginForm
- Fix username suggestion in ACP's LoginForm

### DIFF
--- a/wcfsetup/install/files/lib/system/session/SessionHandler.class.php
+++ b/wcfsetup/install/files/lib/system/session/SessionHandler.class.php
@@ -201,38 +201,117 @@ final class SessionHandler extends SingletonFactory {
 	public function setHasValidCookie($hasValidCookie) { }
 	
 	/**
-	 * Returns the session ID stored in the session cookie or `null`.
+	 * Parses the session cookie value, returning an array with the stored fields.
+	 * 
+	 * The return array is guaranteed to have a `sessionId` key.
 	 */
-	private function getSessionIdFromCookie(): ?string {
-		$cookieName = COOKIE_PREFIX.($this->isACP ? 'acp' : 'user')."_session";
+	private function parseCookie(string $value): array {
+		$length = \mb_strlen($value, '8bit');
+		if ($length < 1) {
+			throw new \InvalidArgumentException(\sprintf(
+				'Expected at least 1 Byte, %d given.',
+				$length
+			));
+		}
+		
+		$version = \unpack('Cversion', $value)['version'];
+		if (!in_array($version, [1], true)) {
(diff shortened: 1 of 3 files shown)


wcfsetup/install/files/lib/system/session/SessionHandler.class.php
```diff
@@ -201,38 +201,116 @@ public function setCookieSuffix() { }
...
+			throw new \InvalidArgumentException(\sprintf(
+				'Unknown version %d',
+				$version
+			));
+		}
+		
+		if ($version === 1) {
+			if (\mb_strlen($value, '8bit') !== 26) {
+				throw new \InvalidArgumentException(\sprintf(
+					'Expected exactly 26 Bytes, %d given.'
```
Gravatronics, line 228: Missing value for `%d`.

TimWolla, line 228:
```suggestion
			if ($length !== 26) {
				throw new \InvalidArgumentException(\sprintf(
					'Expected exactly 26 Bytes, %d given.',
					$length
```

TimWolla: Indeed. I wanted to refactor that to simply use `$length`, but I must have gotten distracted.

wcfsetup/install/files/lib/system/session/SessionHandler.class.php
```diff
@@ -201,38 +201,116 @@ public function setCookieSuffix() { }
...
+	 * 
+	 * The return array is guaranteed to have a `sessionId` key.
+	 */
+	private function parseCookie(string $value): array {
+		$length = \mb_strlen($value, '8bit');
+		if ($length < 1) {
+			throw new \InvalidArgumentException(\sprintf(
+				'Expected at least 1 Byte, %d given.',
+				$length
+			));
```
dtdesign, line 214: This could be simplified, `mb_strlen()` returns an integer and the `if` checks for "less than 1", therefore the only possible value is zero.

dtdesign (changes requested):

wcfsetup/install/files/lib/system/session/SessionHandler.class.php
TimWolla: I want to keep this for consistency with the version specific check in line 226.

TimWolla:
Update instructions are adjusted now.

dtdesign (approved):

PR #3814, opened 2020-12-16
Delete content from multiple users simultaneously

### LINKED ISSUE #3592: Delete content from multiple users simultaneously
There is already the option to delete the contents of a user. Using the clipboard, this action could also be applied to several users at the same time.

joshuaruesweg:
Close #3592

### DIFF
--- a/wcfsetup/install/files/acp/templates/userList.tpl
+++ b/wcfsetup/install/files/acp/templates/userList.tpl
@@ -45,12 +45,14 @@
 			WCF.ACP.User.SendNewPasswordHandler.init();
 		{/if}
 		
-		require(['Language', 'WoltLabSuite/Core/Acp/Ui/User/Editor'], function (Language, AcpUiUserList) {
+		require(['Language', 'WoltLabSuite/Core/Acp/Ui/User/Editor', 'WoltLabSuite/Core/Acp/Ui/User/Content/Remove/Clipboard'], function (Language, AcpUiUserList, AcpUserContentRemoveClipboard) {
 			Language.addObject({
 				'wcf.acp.user.action.sendNewPassword.confirmMessage': '{jslang}wcf.acp.user.action.sendNewPassword.confirmMessage{/jslang}'
 			});
 			
 			new AcpUiUserList();
+			
+			new AcpUserContentRemoveClipboard();
 		});
 		
 		{event name='javascriptInit'}
--- a/wcfsetup/install/files/lib/system/worker/UserContentRemoveWorker.class.php
+++ b/wcfsetup/install/files/lib/system/worker/UserContentRemoveWorker.class.php
@@ -30,7 +30,6 @@ class UserContentRemoveWorker extends AbstractWorker {
 	protected $limit = 10;
 	
 	/**
-	 * user
 	 * @var	User[]
 	 */
 	protected $users = [];
@@ -42,10 +41,9 @@ class UserContentRemoveWorker extends AbstractWorker {
 	protected $data = null;
 	
 	/**
-	 * 
 	 * @var null 
 	 */
-	public $contentProvider = null;
+	public $contentProviders = null;
 	
 	/**
 	 * @inheritDoc
@@ -61,9 +59,7 @@ class UserContentRemoveWorker extends AbstractWorker {
 			$userList->readObjects();
 			
 			if ($userList->count() !== count($this->parameters['userIDs'])) {
-				$diff = array_diff($this->parameters['userIDs'], array_map(function (User $user) {
-					return $user->userID;
-				}, $userList->getObjects()));
+				$diff = array_diff($this->parameters['userIDs'], array_column($userList->getObjects(), 'userID'));
 				
 				throw new \InvalidArgumentException('The parameter `userIDs` contains unknown values ('. implode(', ', $diff) .').');
 			}
@@ -86,16 +82,12 @@ class UserContentRemoveWorker extends AbstractWorker {
 				throw new \InvalidArgumentException('The parameter `contentProvider` must be an array.');
 			}
 			
-			$knownContentProvider = array_map(function ($contentProvider) {
-				return $contentProvider->objectType;
-			}, ObjectTypeCache::getInstance()->getObjectTypes('com.woltlab.wcf.content.userContentProvider'));
-			
-			$unknownContentProvider = array_diff($this->parameters['contentProvider'], $knownContentProvider);
+			$unknownContentProvider = array_diff($this->parameters['contentProvider'], array_column(ObjectTypeCache::getInstance()->getObjectTypes('com.woltlab.wcf.content.userContentProvider'), 'objectType'));
 			if (!empty($unknownContentProvider)) {
 				throw new \InvalidArgumentException('The parameter `contentProvider` contains unknown objectTypes ('. implode(', ', $unknownContentProvider) .').');
 			}
 			
-			$this->contentProvider = $this->parameters['contentProvider'];
+			$this->contentProviders = $this->parameters['contentProvider'];
 		}
 		
 		if ($this->loopCount === 0) {
@@ -121,11 +113,9 @@ class UserContentRemoveWorker extends AbstractWorker {
 			'count' => 0
 		];
 		
-		$contentProviders = ObjectTypeCache::getInstance()->getObjectTypes('com.woltlab.wcf.content.userContentProvider');
-		
 		// add the required object types for the select content provider
-		if (is_array($this->contentProvider)) {
-			foreach ($this->contentProvider as $contentProvider) {
+		if (is_array($this->contentProviders)) {
+			foreach ($this->contentProviders as $contentProvider) {
 				$objectType = ObjectTypeCache::getInstance()->getObjectTypeByName('com.woltlab.wcf.content.userContentProvider', $contentProvider);
 				
 				if ($objectType->requiredobjecttype !== null) {
@@ -138,14 +128,24 @@ class UserContentRemoveWorker extends AbstractWorker {
 							throw new \RuntimeException('Unknown required object type "' . $objectTypeName . '" for object type "' . $contentProvider . '" given.');
 						}
 						
-						$this->contentProvider[] = $objectTypeName;
+						$this->contentProviders[] = $objectTypeName;
 					}
 				}
 			}
 		}
 		
+		$contentProviders = ObjectTypeCache::getInstance()->getObjectTypes('com.woltlab.wcf.content.userContentProvider');
+		
+		// sort object types
+		uasort($contentProviders, function ($a, $b) {
+			$niceValueA = ($a->nicevalue ?: 0);
+			$niceValueB = ($b->nicevalue ?: 0);
+			
+			return $niceValueA <=> $niceValueB;
+		});
+		
 		foreach ($contentProviders as $contentProvider) {
-			if ($this->contentProvider === null || (is_array($this->contentProvider) && in_array($contentProvider->objectType, $this->contentProvider))) {
+			if ($this->contentProviders === null || (is_array($this->contentProviders) && in_array($contentProvider->objectType, $this->contentProviders))) {
 				foreach ($this->users as $user) {
 					/** @var IUserContentProvider $processor */
 					$processor = $contentProvider->getProcessor();
@@ -153,33 +153,17 @@ class UserContentRemoveWorker extends AbstractWorker {
 					$count = $contentList->countObjects();
 					
 					if ($count) {
-						if (!isset($this->data['provider'][$contentProvider->objectType])) {
-							$this->data['provider'][$contentProvider->objectType] = [
-								'count' => 0,
-								'objectTypeID' => $contentProvider->objectTypeID,
-								'nicevalue' => $contentProvider->nicevalue ?: 0,
-								'user' => [],
-							];
-						}
-						
-						$this->data['provider'][$contentProvider->objectType]['user'][$user->userID] = [
-							'count' => $count,
+						$this->data['provider'][] = [
+							'userID' => $user->userID,
+							'objectTypeID' => $contentProvider->objectTypeID,
+							'count' => $count
 						];
-						$this->data['provider'][$contentProvider->objectType]['count'] += $count;
 						
 						$this->data['count'] += ceil($count / $this->limit) * $this->limit;
 					}
 				}
 			}
 		}
-		
-		// sort object types
-		uasort($this->data['provider'], function ($a, $b) {
-			$niceValueA = ($a['nicevalue'] ?: 0);
-			$niceValueB = ($b['nicevalue'] ?: 0);
-			
-			return $niceValueA <=> $niceValueB;
-		});
 	}
 	
 	/**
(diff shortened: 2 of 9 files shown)


wcfsetup/install/files/ts/WoltLabSuite/Core/Acp/Ui/User/Content/Remove/Clipboard.ts
```diff
@@ -0,0 +1,126 @@
...
+   * Executes the remove content worker.
+   */
+  private executeWorker(objectTypes: string[]): void {
+    new AcpUiWorker({
+      // dialog
+      dialogId: "removeContentWorker",
+      dialogTitle: Language.get("wcf.acp.content.removeContent"),
+
+      // ajax
+      className: "\\wcf\\system\\worker\\UserContentRemoveWorker",
```
TimWolla, line 65:
```suggestion
      className: "wcf\\system\\worker\\UserContentRemoveWorker",
```

wcfsetup/install/files/ts/WoltLabSuite/Core/Acp/Ui/User/Content/Remove/Clipboard.ts
```diff
@@ -0,0 +1,126 @@
...
+  public userIds: number[];
+  private readonly dialogId = "userContentRemoveClipboardPrepareDialog";
+
+  /**
+   * Initializes the content remove handler.
+   */
+  constructor() {
+    EventHandler.add("com.woltlab.wcf.clipboard", "com.woltlab.wcf.user", (data: EventData) => {
+      if (data.data.actionName === "com.woltlab.wcf.user.deleteUserContent") {
+        this.userIds = data.data.parameters.objectIDs;
```
TimWolla, line 48: Avoid state in the class. Pass the userIDs as the second parameter of `Ajax.api()`.

wcfsetup/install/files/ts/WoltLabSuite/Core/Acp/Ui/User/Content/Remove/Clipboard.ts
```diff
@@ -0,0 +1,126 @@
...
+
+  get dialogContent(): HTMLElement {
+    return UiDialog.getDialog(this.dialogId)!.content;
+  }
+
+  _ajaxSuccess(data: AjaxResponse): void {
+    UiDialog.open(this, data.returnValues.template);
+
+    const submitButton = this.dialogContent.querySelector('input[type="submit"]') as HTMLElement;
+    submitButton.addEventListener("click", () => this.submit());
```
TimWolla, line 100: Doesn't the Dialog itself provide callbacks for this?

wcfsetup/install/files/lib/system/worker/UserContentRemoveWorker.class.php
```diff
@@ -29,9 +31,9 @@ class UserContentRemoveWorker extends AbstractWorker {
 	
 	/**
 	 * user
-	 * @var	User
+	 * @var	User[]
 	 */
-	protected $user = null;
+	protected $user = [];
```
TimWolla, line 36: Rename this to `users`.

TimWolla (changes requested):
Some first remarks. Generally I find the logic super hard to follow here and will need to spend more time understanding it.

wcfsetup/install/files/ts/WoltLabSuite/Core/Acp/Ui/User/Content/Remove/Clipboard.ts
joshuaruesweg: No, unfortunately he does not. I wouldn't change it either, even if I would love to.

wcfsetup/install/files/lib/system/worker/UserContentRemoveWorker.class.php
```diff
@@ -175,11 +203,21 @@ public function execute() {
...
 		$objectList->readObjectIDs();
 		$processor->deleteContent($objectList->objectIDs);
 		
+		$this->data['provider'][$providerObjectType]['user'][$userID]['count'] -= $this->limit;
+		
+		if ($this->data['provider'][$providerObjectType]['user'][$userID]['count'] <= 0) {
+			unset($this->data['provider'][$providerObjectType]['user'][$userID]);
+		} 
+		
 		$this->data['provider'][$providerObjectType]['count'] -= $this->limit;
```
TimWolla, line 221: I believe this might be buggy with the refactoring now. Consider the following:

1. A limit of 10.
2. Two users with 3 posts each – a total of 6.
3. The first user is processed: This will subtract 10 from 6, resulting in -4, resulting in the second user now being processed.

wcfsetup/install/files/lib/system/worker/UserContentRemoveWorker.class.php
```diff
@@ -128,19 +146,29 @@ private function generateData() {
...
+								'objectTypeID' => $contentProvider->objectTypeID,
+								'nicevalue' => $contentProvider->nicevalue ?: 0,
+								'user' => [],
+							];
+						}
+						
+						$this->data['provider'][$contentProvider->objectType]['user'][$user->userID] = [
+							'count' => $count,
+						];
+						$this->data['provider'][$contentProvider->objectType]['count'] += $count;
```
TimWolla, line 168: This is a horrible data structure and I believe this is part of the reason the issue happened. May I suggest the following:

1. Sort the providers by niceValue before iterating them.
     - They are already in the correct order then.
2. Invert the 2 loops.
     - This processes all contents for a single user before processing the next user, feeling more natural.
3. Then make this a flat queue:
     ```
     $this->queue[] = [ 'userID' => …, 'provider' => …, 'remaining' => … ];
     ```

When processing you then simply take the first item in the queue and process it until `remaining < 0`, then removing it and proceeding with the next. No need for complex nesting. The number of remaining items then is a simple `array_sum(array_column($this->queue, 'remaining'))` away.

TimWolla (changes requested):

wcfsetup/install/files/lib/system/worker/UserContentRemoveWorker.class.php
joshuaruesweg: No, because of the following code during the initialisation: https://github.com/WoltLab/WCF/blob/8ea6fb91cb5abd62470724f54c320d4df87e8ffe/wcfsetup/install/files/lib/system/worker/UserContentRemoveWorker.class.php#L170

wcfsetup/install/files/ts/WoltLabSuite/Core/Acp/Ui/User/Content/Remove/Clipboard.ts
joshuaruesweg: No I can't, because I need it during the execution of the worker.

wcfsetup/install/files/lib/system/worker/UserContentRemoveWorker.class.php
joshuaruesweg: I think it complexifies the logic way more, then it is now. Especially because we currently only execute the object types for the users that really have content (and object types w/o any content will be skipped anyway). In a flat map, I need to test that with every object type again. Keep in mind that the logic has been in use for 2 (?) versions.

wcfsetup/install/files/lib/system/worker/UserContentRemoveWorker.class.php
```diff
@@ -29,9 +31,9 @@ class UserContentRemoveWorker extends AbstractWorker {
 	
 	/**
 	 * user
```
TimWolla, line 33:
```suggestion
```

wcfsetup/install/files/lib/system/worker/UserContentRemoveWorker.class.php
```diff
@@ -49,18 +51,34 @@ class UserContentRemoveWorker extends AbstractWorker {
...
-			throw new \InvalidArgumentException('userID is unknown.');
+		if (isset($this->parameters['userIDs']) && is_array($this->parameters['userIDs']) && !empty($this->parameters['userIDs'])) {
+			$userList = new UserList();
+			$userList->setObjectIDs($this->parameters['userIDs']);
+			$userList->readObjects();
+			
+			if ($userList->count() !== count($this->parameters['userIDs'])) {
+				$diff = array_diff($this->parameters['userIDs'], array_map(function (User $user) {
+					return $user->userID;
+				}, $userList->getObjects()));
```
TimWolla, line 66:
```suggestion
				$diff = array_diff($this->parameters['userIDs'], array_column($userList->getObjects(), 'userID'));
```

Could apply likewise to `$knownContentProvider` below.

wcfsetup/install/files/lib/system/worker/UserContentRemoveWorker.class.php
```diff
@@ -210,4 +252,16 @@ public function finalize() {
...
 		return LinkHandler::getInstance()->getLink('UserList');
 	}
+	
+	/**
+	 * Generates a key for session data saving.
+	 */
+	protected function generateKey(): string {
+		$userIDs = array_map(function (User $user) {
+			return $user->userID;
+		}, $this->users);
```
TimWolla, line 262:
```suggestion
		$userIDs = array_column($this->users, 'userID');
```

wcfsetup/install/files/lib/system/worker/UserContentRemoveWorker.class.php
TimWolla: > Especially because we currently only execute the object types for the users that really have content (and object types w/o any content will be skipped anyway).

You are free to not put the entries into the queue if `count == 0`.

> Keep in mind that the logic has been in use for 2 (?) versions.

The previous logic has been in use for several versions, but the logic to handle multiple users is new.

And quite frankly: I'm having a very hard time following the control flow here due to `$this->data` containing 5 (!) nested arrays where you need to subtract the number of processed items in various locations to make sure to correctly remove specific parts of the arrays to indicate that this part was fully processed.

A queue is as simple of a data structure as you can get. It _will_ drastically simplify the code: Process the first `(userID, objectTypeID)` pair until `count` is non-positive, repeat until no more pairs exist.

TimWolla (changes requested):

wcfsetup/install/files/lib/system/worker/UserContentRemoveWorker.class.php
```diff
@@ -126,32 +135,36 @@ private function generateData() {
...
-				$contentList = $processor->getContentListForUser($this->user);
-				$count = $contentList->countObjects();
-				
-				if ($count) {
-					$this->data['provider'][$contentProvider->objectType] = [
-						'count' => $count,
-						'objectTypeID' => $contentProvider->objectTypeID,
-						'nicevalue' => $contentProvider->nicevalue ?: 0
-					];
+				foreach ($this->users as $user) {
```
TimWolla, line 150: I would invert the order of the 2 loops:

Outer loop: Users.
Inner Loop: Content Providers.

This ensures that users after processed one after another which is the more natural order IMO.

wcfsetup/install/files/lib/system/worker/UserContentRemoveWorker.class.php
```diff
@@ -126,32 +135,36 @@ private function generateData() {
 			}
 		}
 		
+		$contentProviders = ObjectTypeCache::getInstance()->getObjectTypes('com.woltlab.wcf.content.userContentProvider');
```
TimWolla, line 138: This could be removed if `$this->contentProvider` would store the `$objectType` instead of the `$objectTypeName`. Then you could directly sort the actual list and could get rid of the condition within the loop.

wcfsetup/install/files/lib/system/worker/UserContentRemoveWorker.class.php
```diff
@@ -103,8 +114,6 @@ private function generateData() {
 			'count' => 0
 		];
 		
-		$contentProviders = ObjectTypeCache::getInstance()->getObjectTypes('com.woltlab.wcf.content.userContentProvider');
-		
 		// add the required object types for the select content provider
 		if (is_array($this->contentProvider)) {
```
TimWolla, line 118:
```suggestion
		if (is_array($this->contentProviders)) {
```
The variable should be renamed to be less confusing.

TimWolla (commented):
Thanks. This is much cleaner now. I've some remark to make this a bit more readable, but the functionality LGTM.

wcfsetup/install/files/lib/system/worker/UserContentRemoveWorker.class.php
joshuaruesweg: I think in the end it doesn't matter. I want to avoid having to check the content providers several times. That would make the code less readable, because the user doesn't notice when and how what is deleted anyway.

wcfsetup/install/files/lib/system/worker/UserContentRemoveWorker.class.php
joshuaruesweg: Keep in mind, that the class variable `$contentProvider` is a parameter, which is read in `validate()`. In the end, it doesn't matter where the whole thing is sorted (and does not clean up the code).

TimWolla: I was thinking about this (untested):

```
diff --git i/wcfsetup/install/files/lib/system/worker/UserContentRemoveWorker.class.php w/wcfsetup/install/files/lib/system/worker/UserContentRemoveWorker.class.php
index 1649009ed5..795cae338e 100644
--- i/wcfsetup/install/files/lib/system/worker/UserContentRemoveWorker.class.php
+++ w/wcfsetup/install/files/lib/system/worker/UserContentRemoveWorker.class.php
@@ -1,5 +1,6 @@
 <?php
 namespace wcf\system\worker;
+use wcf\data\object\type\ObjectType;
 use wcf\data\object\type\ObjectTypeCache;
 use wcf\data\user\User;
 use wcf\data\user\UserList;
@@ -113,6 +114,9 @@ class UserContentRemoveWorker extends AbstractWorker {
 			'count' => 0
 		];
 		
+		/** @var ObjectType[] $contentProviders */
+		$contentProviders = [];
+		
 		// add the required object types for the select content provider
 		if (is_array($this->contentProviders)) {
 			foreach ($this->contentProviders as $contentProvider) {
@@ -128,13 +132,14 @@ class UserContentRemoveWorker extends AbstractWorker {
 							throw new \RuntimeException('Unknown required object type "' . $objectTypeName . '" for object type "' . $contentProvider . '" given.');
 						}
 						
-						$this->contentProviders[] = $objectTypeName;
+						$contentProviders[] = $objectType;
 					}
 				}
 			}
 		}
-		
-		$contentProviders = ObjectTypeCache::getInstance()->getObjectTypes('com.woltlab.wcf.content.userContentProvider');
+		else {
+			$contentProviders = ObjectTypeCache::getInstance()->getObjectTypes('com.woltlab.wcf.content.userContentProvider');
+		}
 		
 		// sort object types
 		uasort($contentProviders, function ($a, $b) {
@@ -145,22 +150,20 @@ class UserContentRemoveWorker extends AbstractWorker {
 		});
 		
 		foreach ($contentProviders as $contentProvider) {
-			if ($this->contentProviders === null || (is_array($this->contentProviders) && in_array($contentProvider->objectType, $this->contentProviders))) {
-				foreach ($this->users as $user) {
-					/** @var IUserContentProvider $processor */
-					$processor = $contentProvider->getProcessor();
-					$contentList = $processor->getContentListForUser($user);
-					$count = $contentList->countObjects();
+			foreach ($this->users as $user) {
+				/** @var IUserContentProvider $processor */
+				$processor = $contentProvider->getProcessor();
+				$contentList = $processor->getContentListForUser($user);
+				$count = $contentList->countObjects();
+				
+				if ($count) {
+					$this->data['provider'][] = [
+						'userID' => $user->userID,
+						'objectTypeID' => $contentProvider->objectTypeID,
+						'count' => $count
+					];
 					
-					if ($count) {
-						$this->data['provider'][] = [
-							'userID' => $user->userID,
-							'objectTypeID' => $contentProvider->objectTypeID,
-							'count' => $count
-						];
-						
-						$this->data['count'] += ceil($count / $this->limit) * $this->limit;
-					}
+					$this->data['count'] += ceil($count / $this->limit) * $this->limit;
 				}
 			}
 		}
```

With Whitespace ignored:

```
diff --git i/wcfsetup/install/files/lib/system/worker/UserContentRemoveWorker.class.php w/wcfsetup/install/files/lib/system/worker/UserContentRemoveWorker.class.php
index 1649009ed5..795cae338e 100644
--- i/wcfsetup/install/files/lib/system/worker/UserContentRemoveWorker.class.php
+++ w/wcfsetup/install/files/lib/system/worker/UserContentRemoveWorker.class.php
@@ -1,5 +1,6 @@
 <?php
 namespace wcf\system\worker;
+use wcf\data\object\type\ObjectType;
 use wcf\data\object\type\ObjectTypeCache;
 use wcf\data\user\User;
 use wcf\data\user\UserList;
@@ -113,6 +114,9 @@ class UserContentRemoveWorker extends AbstractWorker {
 			'count' => 0
 		];
 		
+		/** @var ObjectType[] $contentProviders */
+		$contentProviders = [];
+		
 		// add the required object types for the select content provider
 		if (is_array($this->contentProviders)) {
 			foreach ($this->contentProviders as $contentProvider) {
@@ -128,13 +132,14 @@ class UserContentRemoveWorker extends AbstractWorker {
 							throw new \RuntimeException('Unknown required object type "' . $objectTypeName . '" for object type "' . $contentProvider . '" given.');
 						}
 						
-						$this->contentProviders[] = $objectTypeName;
+						$contentProviders[] = $objectType;
 					}
 				}
 			}
 		}
-		
+		else {
 			$contentProviders = ObjectTypeCache::getInstance()->getObjectTypes('com.woltlab.wcf.content.userContentProvider');
+		}
 		
 		// sort object types
 		uasort($contentProviders, function ($a, $b) {
@@ -145,7 +150,6 @@ class UserContentRemoveWorker extends AbstractWorker {
 		});
 		
 		foreach ($contentProviders as $contentProvider) {
-			if ($this->contentProviders === null || (is_array($this->contentProviders) && in_array($contentProvider->objectType, $this->contentProviders))) {
 			foreach ($this->users as $user) {
 				/** @var IUserContentProvider $processor */
 				$processor = $contentProvider->getProcessor();
@@ -164,7 +168,6 @@ class UserContentRemoveWorker extends AbstractWorker {
 			}
 		}
 	}
-	}
 	
 	/**
 	 * @inheritDoc
```

Which I'd say *is* cleaner, because the `if` could be removed.

TimWolla (approved):
Overall this LGTM now.

wcfsetup/install/files/acp/templates/userList.tpl
```diff
@@ -51,6 +51,8 @@
 			});
 			
 			new AcpUiUserList();
+			
+			new AcpUserContentRemoveClipboard();
```
TimWolla, line 55: Bad merge: The file is not listed in `require()`.

TimWolla (changes requested):

PR #3824, opened 2020-12-21
Preserve package server credentials during the upgrade from 5.2 to 5.3.

### LINKED ISSUE #3805: Preserve package server credentials during the upgrade 5.2 -> 5.3
n/t

dtdesign:
Fixes #3805

### DIFF
--- a/wcfsetup/install/files/acp/update_com.woltlab.wcf_5.3_packageServer.php
+++ b/wcfsetup/install/files/acp/update_com.woltlab.wcf_5.3_packageServer.php
@@ -11,46 +11,13 @@ $sql = "SELECT  *
 $statement = WCF::getDB()->prepareStatement($sql);
 $statement->execute();
 
-// Try to extract authentication credentials from the previous package servers.
-$newServers = [
-	"update" => [
-		"loginUsername" => "",
-		"loginPassword" => "",
-	],
-	"store" => [
-		"loginUsername" => "",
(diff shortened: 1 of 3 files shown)


wcfsetup/install/files/acp/update_com.woltlab.wcf_5.3_packageServer.php
```diff
@@ -0,0 +1,56 @@
...
+];
+
+$deleteServerIDs = [];
+while ($row = $statement->fetchArray()) {
+	$deleteServerIDs[] = $row["packageUpdateServerID"];
+	
+	// Only use values from the "2019" servers to avoid dealing with outdated
+	// credentials that have never been updated.
+	$serverURL = $row["serverURL"];
+	if (preg_match("~^https?://update\.woltlab\.com/2019/~", $serverURL)) {
```
TimWolla, line 33: Shouldn't you preserve the 5.3 update servers? They must have been added for the upgrade anyway.

Thinking about this: It should be sufficient to simply add something like `AND serverURL NOT LIKE '%5.3/'` to the query in the SQL file to fix the issue, no?

dtdesign: Yes, you are correct, that is much easier.

TimWolla (approved):

PR #3826, opened 2020-12-22
Refactor search query parse of MysqlSearchEngine

TimWolla:


TimWolla:
There's an infinite loop bug when the `@` character is used in the query (because I intentionally did not support this, but still must truncate words at an `@`).

dtdesign (commented):
Looks good to me and we can just strip the `@`. It is barely of any use, unknown to most and not supported by Elasticsearch.

TimWolla:
> Looks good to me and we can just strip the `@`. It is barely of any use, unknown to most and not supported by Elasticsearch.

Yes, that was my plan.

--------

I've fixed all the parsing issues now. It also didn't properly parse parentheses, percent signs and newlines. I've now tested the parser against all 3 character ASCII strings to make sure it does not end up in an infinite loop.

The actual query parser now supports grouping with parentheses, but the query rewriter strips those out again for user experience reasons. If we want to support this type of grouping we can easily add support to the query rewriter, because the parser already is robust enough to handle those.

dtdesign (approved):

PR #3833, opened 2020-12-22
Add support for pattern attribute in form builder

### LINKED ISSUE #3735: Add support for the pattern attribute to TextFormField
see https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/pattern
see https://github.com/WoltLab/WCF/blob/1e08b49f1c61ecb81a15f1d163d7322043a0512e/com.woltlab.wcf/templates/__multifactorTotpCodeField.tpl#L8

That specific pattern in the TOTP code field helps force a numeric keyboard on certain mobile devices and thus is partly related to #3734.

A proper integration might or might not want to validate the pattern on the server side as well.

Note: A full integration should support adding a `title=""` attribute, containing further explanation with regard to the pattern allowed, as explained in MDN.

Gravatronics:
Close #3735

### DIFF
--- a/wcfsetup/install/files/lib/system/form/builder/field/IPatternFormField.class.php
+++ b/wcfsetup/install/files/lib/system/form/builder/field/IPatternFormField.class.php
@@ -1,0 +1,28 @@
+<?php
+namespace wcf\system\form\builder\field;
+
+/**
+ * Represents a form field that supports the `pattern` attribute.
+ *
+ * @author      Matthias Schmidt
+ * @copyright   2001-2020 WoltLab GmbH
+ * @license     GNU Lesser General Public License <http://opensource.org/licenses/lgpl-license.php>
+ * @package     WoltLabSuite\Core\System\Form\Builder\Field
+ * @see         https://html.spec.whatwg.org/multipage/input.html#attr-input-pattern
+ * @since       5.4
+ */
+interface IPatternFormField {
+	/**
+	 * Returns the `pattern` attribute of the form field.
+	 *
+	 * If `null` is returned, no `pattern` attribute will be set.
+	 */
+	public function getPattern(): ?string;
+	
+	/**
+	 * Sets the `pattern` attribute of the form field.
+	 *
+	 * If `null` is given, the attribute is unset.
+	 */
+	public function pattern(?string $pattern): self;
+}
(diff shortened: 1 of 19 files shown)


wcfsetup/install/files/lib/system/form/builder/field/IPatternFormField.class.php
```diff
@@ -0,0 +1,28 @@
...
+	 * Returns the `pattern` attribute of the form field.
+	 *
+	 * If `null` is returned, no `pattern` attribute will be set.
+	 */
+	public function getPattern(): ?string;
+	
+	/**
+	 * Sets the `pattern` attribute of the form field.
+	 *
+	 * If `null` is given, the attribute is unset.
```
TimWolla, line 25: Maybe add a remark that the pattern will *not* be validated automatically.

wcfsetup/install/files/lib/system/user/multifactor/totp/CodeFormField.class.php
```diff
@@ -32,6 +32,7 @@ public function __construct() {
 		$this->addFieldClass('multifactorTotpCode');
 		$this->autoComplete('off');
 		$this->inputMode('numeric');
+		$this->pattern('[0-9\s]*');
```
TimWolla, line 35:
```suggestion
		$this->pattern('[0-9]*');
```
TOTP / Email does not allow for spaces. Only backup codes do (because they are very long and shown in chunks).

wcfsetup/install/files/lib/system/user/multifactor/email/CodeFormField.class.php
```diff
@@ -28,6 +28,7 @@ public function __construct() {
 		$this->addFieldClass('multifactorEmailCode');
 		$this->autoComplete('off');
 		$this->inputMode('numeric');
+		$this->pattern('[0-9\s]*');
```
TimWolla, line 31:
```suggestion
		$this->pattern('[0-9]*');
```
TOTP / Email does not allow for spaces. Only backup codes do (because they are very long and shown in chunks).

TimWolla (changes requested):
This LGTM is the code suggestions are resolved. The adjustment of the comment is optional.

wcfsetup/install/files/lib/system/form/builder/field/IPatternFormField.class.php
Gravatronics: The form builder API generally documents whenever an exception can be expected due to validation so that the lack thereof for this method already implies no such validation takes place.

Issue #3839, opened 2020-12-30
WysiwygFormContainer cannot be children of TabFormContainer


TitusKirch:
Currently WysiwygFormContainer can't be children of TabFormContainer because you get the following error message:
`A tab menu container may only have another tab menu container as a parent, not as an earlier ancestor. `
https://github.com/WoltLab/WCF/blob/master/wcfsetup/install/files/lib/system/form/builder/container/FormContainer.class.php#L110

Is there a reason for this behavior? I just sorted out the above linked lines for the scenario and could not find any real problems with the WysiwygFormContainer.

The reason why I ask this is: I would like to "recreate" the following with the FormBuilder:

![image](https://user-images.githubusercontent.com/16943133/103323257-4c19fa80-4a42-11eb-8d0b-45ad81826ff6.png)

Would it be possible to create an exception for this behavior at least for the WysiwygFormContainer (preferably retroactive for 5.2)?

PR #3840, opened 2020-12-30
Allow WysiwygFormContainer as children of ITabMenuFormContainer

TitusKirch:
This is a draft for #3839


TimWolla:
> This is a draft

@TitusKirch Could you explain what's currently missing / what you plan to change before it is ready for review?

TitusKirch:
I had tried another method, but it was even more inelegant than the current solution. Afterwards I had forgotten to submit this here still to the Reviwe.

Gravatronics:
See ecb2e8f33333214acf8185acc26869f27ec64803

PR #3844, opened 2021-01-02
Fix failing cronjobs

joshuaruesweg:


wcfsetup/install/files/lib/system/cronjob/CronjobScheduler.class.php
```diff
@@ -130,11 +130,17 @@ protected function resetFailedCronjobs() {
 						$data['failCount'] = $cronjob->failCount + 1;
 						
 						// The cronjob exceeded the maximum fail count.
-						// Thus it must be disabled.
-						if ($data['failCount'] > Cronjob::MAX_FAIL_COUNT) {
+						// Cronjobs which can be disabled, should be disabled.
```
Gravatronics, line 133:
```suggestion
						// Cronjobs that can be disabled, should be disabled.
```

wcfsetup/install/files/lib/system/cronjob/CronjobScheduler.class.php
```diff
@@ -130,11 +130,17 @@ protected function resetFailedCronjobs() {
...
+						if ($cronjob->canBeDisabled && $data['failCount'] >= Cronjob::MAX_FAIL_COUNT) {
 							$data['isDisabled'] = 1;
 							$data['failCount'] = 0;
 						}
+						else if ($data['failCount'] >= Cronjob::MAX_FAIL_COUNT) {
```
Gravatronics, line 138: `$data['failCount'] >= Cronjob::MAX_FAIL_COUNT` appears twice. Why not keep the previous condition (with `>=` instead of `>`) and then add a simple `if-else` below it?

Gravatronics (changes requested):

Gravatronics (approved):
Code change looks good, but did not test changes.

PR #3848, opened 2021-01-05
Convert Prism usage to TypeScript

TimWolla:


wcfsetup/install/files/ts/WoltLabSuite/Core/Bbcode/Code.ts
```diff
@@ -0,0 +1,124 @@
...
+ */
+
+import * as Language from "../Language";
+import * as Clipboard from "../Clipboard";
+import * as UiNotification from "../Ui/Notification";
+import Prism from "../Prism";
+import * as PrismHelper from "../Prism/Helper";
+import PrismMeta from "../prism-meta";
+
+const CHUNK_SIZE = 50;
```
dtdesign, line 17:
```suggestion
const chunkSize = 50;
```
Constants should look like regular variables.

wcfsetup/install/files/ts/WoltLabSuite/Core/Bbcode/Code.ts
```diff
@@ -0,0 +1,124 @@
...
+
+async function waitForIdle(): Promise<void> {
+  return new Promise(function (resolve, _reject) {
```
dtdesign, line 20:
```suggestion
  return new Promise((resolve, _reject) => {
```

wcfsetup/install/files/ts/WoltLabSuite/Core/Bbcode/Code.ts
```diff
@@ -0,0 +1,124 @@
...
+      setTimeout(resolve, 0);
+    }
+  });
+}
+
+class Code {
+  private codeContainer: HTMLElementTagNameMap["code"];
+  private language: string | undefined;
+
+  constructor(private readonly container: HTMLElement) {
```
dtdesign, line 33: Please do not mix constructor promotion with regular property definitions.

wcfsetup/install/files/ts/WoltLabSuite/Core/Bbcode/Code.ts
```diff
@@ -0,0 +1,124 @@
...
+    if ((window as any).requestIdleCallback) {
+      (window as any).requestIdleCallback(resolve, { timeout: 5000 });
+    } else {
+      setTimeout(resolve, 0);
+    }
+  });
+}
+
+class Code {
+  private codeContainer: HTMLElementTagNameMap["code"];
```
dtdesign, line 30: I usually resort to `HTMLElement` when there is no specific interface available _and_ there are no fancy properties that we want to access. I had to look up what this type is about, which kind of defeats the purpose of having obvious typings. Not sure about this one.

wcfsetup/install/files/ts/WoltLabSuite/Core/Bbcode/Code.ts
```diff
@@ -0,0 +1,124 @@
...
+    }
+
+    const button = document.createElement("span");
+    button.className = "icon icon24 fa-files-o pointer jsTooltip";
+    button.setAttribute("title", Language.get("wcf.message.bbcode.code.copy"));
+    button.addEventListener("click", () => {
+      void Clipboard.copyElementTextToClipboard(this.codeContainer).then(() => {
+        UiNotification.show(Language.get("wcf.message.bbcode.code.copy.success"));
+      });
+    });
```
dtdesign, line 70:
```suggestion
    button.addEventListener("click", async () => {
      await Clipboard.copyElementTextToClipboard(this.codeContainer);

      UiNotification.show(Language.get("wcf.message.bbcode.code.copy.success"));
    });
```
This isn't NodeJS, stop marching to the right 😛

wcfsetup/install/files/ts/WoltLabSuite/Core/Prism/Helper.ts
```diff
@@ -0,0 +1,44 @@
...
+
+export function* splitIntoLines(container: Node): Generator<Element, void> {
+  const it = document.createNodeIterator(container, NodeFilter.SHOW_TEXT, {
+    acceptNode() {
+      return NodeFilter.FILTER_ACCEPT;
+    },
+  });
+
+  let line = document.createElement("span");
+  for (let node = it.nextNode(); node; node = it.nextNode()) {
```
dtdesign, line 18: Isn't this just a `while (let node = it.nextNode())` in disguise? I can see why `for (;;)` is more concise than `while (true)`, but in this case the `for()` implies a known sequence, which it isn't. Maybe I'm just a bit nitpicky here.

dtdesign (changes requested):

wcfsetup/install/files/ts/WoltLabSuite/Core/Bbcode/Code.ts
TimWolla: eslint does not accept `async` functions for callbacks (because they do not return `void`).

wcfsetup/install/files/ts/WoltLabSuite/Core/Bbcode/Code.ts
TimWolla: I opted to make this a static class member and adjusted the casing.

wcfsetup/install/files/ts/WoltLabSuite/Core/Bbcode/Code.ts
dtdesign: We could toggle `checksVoidReturn` of [`@typescript-eslint/no-misused-promises`](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-misused-promises.md#options) to allow these uses. I see nothing wrong with this?

TimWolla: Okay, I adjusted that. Misuses in event listeners should be found during review.

dtdesign (approved):

PR #3851, opened 2021-01-06
Update to scssphp/scssphp 1.4

TimWolla:
This is primarily a test to check whether linguist-vendored works correctly,
but updates can't hurt, so ¯\_(ツ)_/¯


TimWolla:
> This is primarily a test to check whether linguist-vendored works correctly,

Apparently it doesn't, unless I am misunderstanding what the effect is. All the files were expanded within the PR. At least locally `git check-attr` returns the correct value.

PR #3854, opened 2021-01-07
Improve TypeScript safety

TimWolla:
**NOT TESTED YET**

### DIFF
--- a/wcfsetup/install/files/ts/WoltLabSuite/Core/Date/Picker.ts
+++ b/wcfsetup/install/files/ts/WoltLabSuite/Core/Date/Picker.ts
@@ -698,7 +698,7 @@ const DatePicker = {
             value = "";
           } else {
             element.dataset.value = time.toString();
-            const format = isTimeOnly ? "formatTime" : "formatDate" + (isDateTime ? "Time" : "");
+            const format = isTimeOnly ? "formatTime" : isDateTime ? "formatDateTime" : "formatDate";
             value = DateUtil[format](date);
           }
         }
--- a/wcfsetup/install/files/ts/WoltLabSuite/Core/Date/Picker.ts
+++ b/wcfsetup/install/files/ts/WoltLabSuite/Core/Date/Picker.ts
@@ -698,8 +698,15 @@ const DatePicker = {
             value = "";
           } else {
             element.dataset.value = time.toString();
-            const format = isTimeOnly ? "formatTime" : isDateTime ? "formatDateTime" : "formatDate";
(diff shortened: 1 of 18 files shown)


wcfsetup/install/files/ts/WoltLabSuite/Core/Date/Picker.ts
```diff
@@ -698,7 +698,7 @@ const DatePicker = {
             value = "";
           } else {
             element.dataset.value = time.toString();
-            const format = isTimeOnly ? "formatTime" : "formatDate" + (isDateTime ? "Time" : "");
+            const format = isTimeOnly ? "formatTime" : isDateTime ? "formatDateTime" : "formatDate";
```
dtdesign, line 701: Can't we just split this up into the separate calls? The performance is the same, but the viewer no longer barfs internally.

wcfsetup/install/files/ts/WoltLabSuite/Core/I18n/Plural.ts
```diff
@@ -9,184 +9,105 @@
...
-  },
-
-  /**
-   * `v` represents the number of digits of the fractional part (1.234 yields 3)
-   */
-  getV(n: number): number {
-    return n.toString().replace(/^[^.]*\.?/, "").length;
-  },
-
+enum Category {
```
dtdesign, line 12:
```suggestion
const enum Category {
```
This allows the compiler to inline the value and remove the enum from the final code. There is no partical use of carrying it around on runtime?

dtdesign (commented):
In general I would also like to see enforced return types on any function that isn't anonymous. Typical callbacks are fine because the return value of the function is inferred from the context anyway.

wcfsetup/install/files/ts/WoltLabSuite/Core/I18n/Plural.ts
TimWolla: I attempted that, but somehow that caused issues with defining the `Parameters` type. I've re-tested this and apparently my later changes fixed that definition issue.

TimWolla:
> In general I would also like to see enforced return types on any function that isn't anonymous.

eslint already does this at module boundaries (i.e. exported functions). I would need to check whether there's anything to require this for all named functions.

TimWolla:
I now verified the correct workings of the `Plural` module which received a bigger refactoring. The other modules just got some types, so this PR is ready now.

Gravatronics (approved):

Issue #3859, opened 2021-01-09
Deprecate multi domain setups


dtdesign:
Multi domain setups have been semi supported for quite some time, but they are a constant source of issues while bringing little to no value to the table. CORS is the most prominent case for added complexity to support these setups, this also applies to the crippled `localStorage` to some extent. This indirectly also causes the app domain management to be overly complex, instead of providing a simple interface with proper safeguards against misconfigurations.

This change does not impact the ability to assign cookies to the domain apex.

Proposed roadmap:
* 5.4: Deprecate multi domain setups and add a warning if such a setup is found.
* 5.5: Remove the support for multi domain setups.

mutec:
Are you going to cancel the support for multiple subdomains too?
I'd prefer to be able to use a subdomain for every application if I want.

dtdesign:
Entirely different domains `example.com` for app A and `example.org` for app B have never been supported, because it is impossible to share a single cookie between them.

The user of different subdomains for apps, such as `a.example.com` for app A and `b.example.com` for app B are refered to as "multi domain setups" and will be deprecated and eventually removed in the future. The support for cookie sharing using the domain apex will remain.

TitusKirch:
> 5.5: Remove the support for multi domain setups.

But this would be an API-incompatible change and I thought you guys wanted to get into [Semantic Versioning 2.0.0](https://semver.org)  now.

dtdesign:
> But this would be an API-incompatible change and I thought you guys wanted to get into [Semantic Versioning 2.0.0](https://semver.org) now.

First of all, this isn't part of the (public) API at all, it is merely a feature that allows you to utilize different subdomains.

Second, this is a pure gimmick feature that adds zero real value, but comes with a hefty tax on the development. The hidden cost of that feature is quite significant, because apart from the initial design and development phase, we and other developers have to constantly be aware of the consequences such as CORS. Not only is it something that needs to be considered for changes and new features, it also cripples the localStorage, severly holding us back on innovations that would built on top of that.

Frankly, there will never be a 100% backwards compatibility, because we have to make progress at some point and cannot postpone innovations till eternity. Just look at the code of how many sections have been deprecated for years and are still part of the API. We do take this seriously, but at times have to make sacrifices, because this is the real world and not a contained lab environment.

TitusKirch:
Please don't get me wrong, I am not against this change. I just still find it very hard to understand your versioning because you feel every two versions change it or don't follow through 100%.

[...]

PR #3860, opened 2021-01-09
Convert `WCF.Message.Quote.Handler` to TypeScript

dtdesign:
This PR will only take care of `WCF.Message.Quote.Handler` to provide a more stable interface for the text selection.

`WCF.Message.Quote.Manager` is another candidate for a rewrite, but the entire implementation is a dumpster fire, which should not be converted in its current state. The existing concept is flawed in many ways and must be redesigned from the ground up to provide a better general UX with focus on the mobile interactions. Ideally we stop relying on the server to store the quotes and use the `localStorage` with live synchronization among tabs once multi domain setups (see #3859) are finally gone.


wcfsetup/install/files/ts/WoltLabSuite/Core/Dom/Util.ts
```diff
@@ -472,6 +472,23 @@ const DomUtil = {
 
     return innerError as HTMLElement | null;
   },
+
+  /**
+   * Finds the closest element that matches the provided selector. This is a helper
+   * function because `closest()` does exist on elements only.
```
TimWolla, line 478:
```suggestion
   * function because `closest()` does exist on elements only (i.e. it is missing on text nodes).
```
?

Note: Applying this suggestion probably makes Prettier fail.

wcfsetup/install/files/ts/WoltLabSuite/Core/Ui/Message/Quote.ts
```diff
@@ -0,0 +1,553 @@
...
+}
+
+interface ElementBoundaries {
+  bottom: number;
+  left: number;
+  right: number;
+  top: number;
+}
+
+class UiMessageQuote implements AjaxCallbackObject {
```
TimWolla, line 26:
```suggestion
export class UiMessageQuote implements AjaxCallbackObject {
```
Convention is both the regular export + default export, because both have different benefits.

PR #3863, opened 2021-01-10
Dynamic WebP avatars

### LINKED ISSUE #3862: Dynamically serve avatars as WebP with fallback
Avatars have by far the biggest impact on the network payload because they're used in abudance, creating value for every community on pretty much every page accessed in the software. Using the `<picture>` element is a massive breaking change, therefore we'll use server side detection based on the `accept` header to check for the presence of `image/webp` and yield a different source for avatars.

Additionally, we need to provide a fallback for SVG and WebP images for use in emails, because email clients are stuck in the early 2000s and do not support newer formats.

If the user avatar is a …
* GIF
  * Do nothing. While GIFs are losless, they are already quite small because they use a restricted color palette of 256 colors. Also they support animations which have a mixed support for WebP and are painfully expensive to convert.
* WebP
  * Generate a JPEG with acceptable compression (60% quality?)
* JPEG
  * ~Generate a losless WebP image (JPEG usually is lossy, avoid trashing the quality even more)~
  * Generate a WebP @ 80% quality (see comment below)
* PNG
  * Generate a lossy WebP image @ 80% quality (see comment below)

We are then able to serve a WebP image whenever the browser reports support for `image/webp` and there is a WebP variant, otherwise fall back to a supported image format.

Depends on #3861

dtdesign:
User avatars are served as WebP images if the browser indicated support for it via the `accept` http header. This approach is less flexible than `<picture>`, but is fully backwards compatible and thus does not introduce any breaking changes whatsoever.

Closes #3862


wcfsetup/install/files/lib/data/user/avatar/UserAvatarEditor.class.php
```diff
@@ -58,11 +58,81 @@ public function deleteFiles() {
...
+			return false;
+		}
+		
+		if ($this->avatarExtension === "gif") {
+			// We do not touch GIFs at all.
+			return false;
+		}
+		
+		// The image adapters do not support the conversion to different
+		// file types. However, we do require full GD support for jpeg, png
```
TimWolla, line 93: This is not true. Since some version an Imagick-only set-up is supported.

wcfsetup/install/files/lib/util/ImageUtil.class.php
```diff
@@ -180,6 +180,23 @@ public static function fixOrientation($filename) {
...
+	/**
+	 * Examines the `accept` header to determine if the browser
+	 * supports WebP images.
+	 */
+	public static function browserSupportsWebP(): bool {
+		static $supportsWebP = null;
+		
+		if ($supportsWebP === null) {
+			$supportsWebP = false;
+			if (!empty($_SERVER["HTTP_ACCEPT"]) && preg_match("~(?:^|[,;])image/webp[,;]~", $_SERVER["HTTP_ACCEPT"])) {
```
TimWolla, line 192: This misparses the `HTTP_ACCEPT` header. It's not dangerous in this case, but I'd rather have this correct:

1. `explode()` HTTP_ACCEPT at commas.
2. Explode each array item at semicolons.
3. Trim the resulting values.
4. If the first item is `image/webp` then webp is supported.

Something like this should work. It can be simplified a bit (e.g. replace the ArrayUtil::trim() by a trim() on the mimeType).

```php
$acceptableMimeTypes = array_map(function ($acceptableMimeType) {
 [$mimeType] = ArrayUtil::trim(explode(';', $acceptableMimeType));
 return $mimeType;
}, explode(',', $_SERVER['HTTP_ACCEPT']));
return in_array('image/webp', $acceptableMimeTypes);
```

TimWolla (changes requested):

wcfsetup/install/files/lib/data/user/avatar/UserAvatarEditor.class.php
```diff
@@ -89,46 +90,29 @@ public function createAvatarVariant(): bool {
...
-				case "png":
-					$source = imagecreatefrompng($filename);
-					break;
-				
-				default:
-					throw new \LogicException("Unreachable");
-			}
-			
-			imagewebp($source, $this->getLocation(null, true));
+			$imageAdapter->saveImageAs($imageAdapter, $this->getLocation(null, true), "webp", 80);
```
jens1o, line 115: `$imageAdapter->saveImageAs($image, $this->getLocation(null, true), "webp", 80);`

dtdesign: Good catch, thanks!

wcfsetup/install/files/lib/data/user/avatar/UserAvatarEditor.class.php
```diff
@@ -58,11 +59,64 @@ public function deleteFiles() {
...
+		// If the uploaded avatar is already a WebP image, then create a JPEG
+		// as a fallback image and flip the image data to match the JPEG.
+		if ($this->avatarExtension === "webp") {
+			$filenameJpeg = preg_replace('~\.webp$~', '.jpeg', $filenameWebP);
+			
+			$imageAdapter->saveImageAs($image, $filenameJpeg, "jpeg", 80);
+			
+			$data = [
+				"avatarExtension" => "jpeg",
+				"fileHash" => sha1_file($filenameJpeg),
```
TimWolla, line 111: Did you check this? Changing the `fileHash` should result in the code searching the avatar in a different location, breaking it.

wcfsetup/install/files/lib/data/user/avatar/UserAvatar.class.php
```diff
@@ -52,20 +54,29 @@ class UserAvatar extends DatabaseObject implements IUserAvatar {
...
-		return WCF_DIR . 'images/avatars/' . $this->getFilename($size);
+	public function getLocation($size = null, $forceWebP = null) {
+		return WCF_DIR . 'images/avatars/' . $this->getFilename($size, $forceWebP);
 	}
 	
 	/**
 	 * Returns the file name of this avatar.
 	 * 
 	 * @param	integer		$size
+	 * @param boolean $forceWebP
```
TimWolla, line 68: This should be `?bool`. In fact it should be possible to actually use a scalar type hint here, because the parameter is new. Applies likewise above.

PR #3869, opened 2021-01-11
Image adapters now support saving to GIF/JPG/PNG/WebP

dtdesign:
n/t

### DIFF
--- a/wcfsetup/install/files/lib/system/image/adapter/GDImageAdapter.class.php
+++ b/wcfsetup/install/files/lib/system/image/adapter/GDImageAdapter.class.php
@@ -496,6 +496,47 @@ class GDImageAdapter implements IImageAdapter {
 	/**
 	 * @inheritDoc
 	 */
+	public function saveImageAs($image, string $filename, string $type, int $quality = 100): void {
+		if (!$this->isImage($image)) {
+			throw new SystemException("Given image is not a valid image resource.");
+		}
+		
+		ob_start();
+		
+		imagealphablending($image, false);
+		imagesavealpha($image, true);
+		
+		switch ($type) {
+			case "gif":
+				imagegif($image);
+				break;
+				
+			case "jpg":
+			case "jpeg":
+				imagejpeg($image, null, $quality);
+				break;
+				
+			case "png":
+				imagepng($image, null, $quality);
+				break;
+			
+			case "webp":
+				imagewebp($image, null, $quality);
+				break;
+			
+			default:
+				throw new \LogicException("Unreachable");
+		}
+		
+		$stream = ob_get_contents();
+		ob_end_clean();
+		
+		file_put_contents($filename, $stream);
+	}
+	
+	/**
+	 * @inheritDoc
+	 */
 	public static function isSupported() {
 		return function_exists('gd_info');
 	}
--- a/wcfsetup/install/files/lib/system/image/adapter/GDImageAdapter.class.php
+++ b/wcfsetup/install/files/lib/system/image/adapter/GDImageAdapter.class.php
@@ -347,8 +347,11 @@ class GDImageAdapter implements IImageAdapter {
 		
 		ob_start();
 		
+		// fix PNG alpha channel handling
+		// see http://php.net/manual/en/function.imagecopymerge.php#92787
 		imagealphablending($image, false);
 		imagesavealpha($image, true);
+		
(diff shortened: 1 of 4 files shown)


wcfsetup/install/files/lib/system/image/adapter/GDImageAdapter.class.php
```diff
@@ -493,6 +493,47 @@ public function overlayImageRelative($file, $position, $margin, $opacity) {
 		// does nothing
 	}
 	
+	/**
+	 * @inheritDoc
+	 */
+	public function saveImageAs($image, string $filename, string $type, int $quality = 100): void {
+		if (!$this->isImage($image)) {
+			throw new SystemException("Given image is not a valid image resource.");
```
TimWolla, line 501: Do not use `SystemException` for entirely new code that does not need the backwards compatibility with catch blocks.

 `InvalidArgumentException` would be fitting here.

wcfsetup/install/files/lib/system/image/adapter/GDImageAdapter.class.php
```diff
@@ -493,6 +493,47 @@ public function overlayImageRelative($file, $position, $margin, $opacity) {
...
+		}
+		
+		ob_start();
+		
+		imagealphablending($image, false);
+		imagesavealpha($image, true);
```
TimWolla, line 507: Add a comment why these are required.

wcfsetup/install/files/lib/system/image/adapter/GDImageAdapter.class.php
```diff
@@ -493,6 +493,47 @@ public function overlayImageRelative($file, $position, $margin, $opacity) {
...
+			case "png":
+				imagepng($image, null, $quality);
+				break;
+			
+			case "webp":
+				imagewebp($image, null, $quality);
+				break;
+			
+			default:
+				throw new \LogicException("Unreachable");
```
TimWolla, line 528: This is not _actually_ unreachable (because code might call this method directly), use an `InvalidArgumentException`.

wcfsetup/install/files/lib/system/image/adapter/ImagickImageAdapter.class.php
```diff
@@ -461,6 +461,46 @@ public static function getVersion() {
 		return $match['version'];
 	}
 	
+	/**
+	 * @inheritDoc
+	 */
+	public function saveImageAs($image, string $filename, string $type, int $quality = 100): void {
+		if (!($image instanceof \Imagick)) {
+			throw new SystemException("Given image is not a valid Imagick-object.");
```
TimWolla, line 469: ditto

wcfsetup/install/files/lib/system/image/adapter/ImagickImageAdapter.class.php
```diff
@@ -461,6 +461,46 @@ public static function getVersion() {
...
+			case "png":
+				$fileFormat = "png";
+				break;
+			
+			case "webp":
+				$fileFormat = "webp";
+				break;
+			
+			default:
+				throw new \LogicException("Unreachable");
```
TimWolla, line 497: ditto

wcfsetup/install/files/lib/system/image/adapter/GDImageAdapter.class.php
dtdesign: I disagree on that one. The adapters expect the validation to have happened in `ImageAdapter` which is the proper wrapper.

It is common for methods to have their input pre-validated by the wrapper and acting directly on the input without further validation. Yes, you can circumvent this wrapper by instantiiating the actual implementation, but at that point you are bypassing the intended API.

I opted for the `LogicException` because we commonly use them for case which should not happen and are the consequence of an actual error in the program flow. One such error would be the support for more file types in the wrapper without implementing them in the actual image handler.

TimWolla: FWIW: You appear to have adjusted *one* of the two Unreachable exceptions: https://github.com/WoltLab/WCF/commit/96fcf43c58aa8f87af1be30f11f5bb090fa13fcc#diff-e14386efab40bca7cfcc86fe25cb9ac966320587154c2eabd717b10ee118c120R533

PR #3884, opened 2021-01-20
Add support for phpBB's combined hash

TimWolla:


wcfsetup/install/files/lib/util/PasswordUtil.class.php
```diff
@@ -332,8 +332,116 @@ protected static function mybb1($username, $password, $salt, $dbHash) {
 	 * @param	string		$dbHash
 	 * @return	boolean
 	 */
-	protected static function phpbb3($username, $password, $salt, $dbHash) {
-		return self::phpass($username, $password, $salt, $dbHash);
+	public static function phpbb3($username, $password, $salt, $dbHash) {
+		$phpassResult = self::phpass($username, $password, $salt, $dbHash);
+		
+		if ($phpassResult === true) {
```
dtdesign, line 338:
```suggestion
		if ($phpassResult) {
```

wcfsetup/install/files/lib/util/PasswordUtil.class.php
```diff
@@ -332,8 +332,116 @@ protected static function mybb1($username, $password, $salt, $dbHash) {
...
+		// Strip the type prefix.
+		$dbHash = substr($dbHash, strlen($matches[0]));
+		
+		// The following loop only supports the multi-hash variant.
+		// Everything else should already be handled at this point.
+		if (count($algorithms) == 1) {
+			return false;
+		}
+		foreach ($algorithms as $algorithm) {
+			if (($dollar = strpos($dbHash, '$')) === false) {
```
dtdesign, line 356:
```suggestion
			$dollar = strpos($dbHash, '$')
			if ($dollar === false) {
```
The value is required later on, making it easier to find it instead of hiding it in an `if`.

dtdesign (changes requested):
LGTM

joshuaruesweg (approved):
LGTM

PR #3889, opened 2021-01-23
Dynamic WebP avatars

### LINKED ISSUE #3862: Dynamically serve avatars as WebP with fallback

dtdesign:
User avatars are served as WebP images if the browser indicated support for it via the accept http header. This approach is less flexible than <picture>, but is fully backwards compatible and thus does not introduce any breaking changes whatsoever.

Closes #3862

Replaces #3863 which was created prior to the PSR-12 adoption and recreating it was cleaner.

### DIFF
--- a/wcfsetup/install/files/lib/data/user/avatar/UserAvatar.class.php
+++ b/wcfsetup/install/files/lib/data/user/avatar/UserAvatar.class.php
@@ -84,6 +84,7 @@ class UserAvatar extends DatabaseObject implements IUserAvatar
         }
 
         $directory = \substr($this->fileHash, 0, 2);
+
         return \sprintf(
             '%s/%d-%s.%s',
             $directory,
--- a/wcfsetup/install/files/lib/data/user/avatar/UserAvatar.class.php
+++ b/wcfsetup/install/files/lib/data/user/avatar/UserAvatar.class.php
@@ -4,6 +4,7 @@ namespace wcf\data\user\avatar;
 
 use wcf\data\DatabaseObject;
 use wcf\system\WCF;
+use wcf\util\ImageUtil;
 use wcf\util\StringUtil;
 
 /**
@@ -14,13 +15,14 @@ use wcf\util\StringUtil;
  * @license GNU Lesser General Public License <http://opensource.org/licenses/lgpl-license.php>
  * @package WoltLabSuite\Core\Data\User\Avatar
  *
- * @property-read   int $avatarID       unique id of the user avatar
- * @property-read   string $avatarName     name of the original avatar file
- * @property-read   string $avatarExtension    extension of the avatar file
- * @property-read   int $width          width of the user avatar image
- * @property-read   int $height         height of the user avatar image
- * @property-read   int|null $userID         id of the user to which the user avatar belongs or null
- * @property-read   string $fileHash       SHA1 hash of the original avatar file
+ * @property-read int $avatarID unique id of the user avatar
+ * @property-read string $avatarName name of the original avatar file
+ * @property-read string $avatarExtension extension of the avatar file
+ * @property-read int $width width of the user avatar image
+ * @property-read int $height height of the user avatar image
+ * @property-read int|null $userID id of the user to which the user avatar belongs or null
+ * @property-read string $fileHash SHA1 hash of the original avatar file
+ * @property-read int $hasWebP `1` if there is a WebP variant, else `0`
  */
 class UserAvatar extends DatabaseObject implements IUserAvatar
 {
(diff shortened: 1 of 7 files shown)


wcfsetup/install/files/lib/system/worker/UserRebuildDataWorker.class.php
```diff
@@ -211,8 +211,16 @@ public function execute()
             $avatarList = new UserAvatarList();
             $avatarList->getConditionBuilder()->add('user_avatar.userID IN (?)', [$userIDs]);
             $avatarList->getConditionBuilder()->add(
-                '(user_avatar.width <> ? OR user_avatar.height <> ?)',
-                [UserAvatar::AVATAR_SIZE, UserAvatar::AVATAR_SIZE]
+                '(
+					(user_avatar.width <> ? OR user_avatar.height <> ?)
+					OR (user_avatar.hasWebP = ? AND user_avatar.avatarExtension <> ?)
+				)',
```
Gravatronics, line 217:
```suggestion
                '(
                        (user_avatar.width <> ? OR user_avatar.height <> ?)
                     OR (user_avatar.hasWebP = ? AND user_avatar.avatarExtension <> ?)
                )',
```

wcfsetup/install/files/lib/data/user/avatar/UserAvatar.class.php
```diff
@@ -14,13 +15,14 @@
...
- * @property-read   int|null $userID         id of the user to which the user avatar belongs or null
- * @property-read   string $fileHash       SHA1 hash of the original avatar file
+ * @property-read int $avatarID unique id of the user avatar
+ * @property-read string $avatarName name of the original avatar file
+ * @property-read string $avatarExtension extension of the avatar file
+ * @property-read int $width width of the user avatar image
+ * @property-read int $height height of the user avatar image
+ * @property-read int|null $userID id of the user to which the user avatar belongs or null
+ * @property-read string $fileHash SHA1 hash of the original avatar file
+ * @property-read int $hasWebP `1` if there is a WebP variant, else `0`
```
TimWolla, line 25: This might be related to the PSR-12 changes, but this should not have happened. Can you check what caused this reformatting? Maybe it should / needs to be applied to all files before.

wcfsetup/install/files/lib/data/user/avatar/UserAvatar.class.php
```diff
@@ -55,26 +57,37 @@ class UserAvatar extends DatabaseObject implements IUserAvatar
...
      */
-    public function getFilename($size = null)
+    public function getFilename($size = null, ?bool $forceWebP = null)
     {
-        return \substr(
-            $this->fileHash,
-            0,
-            2
-        ) . '/' . $this->avatarID . '-' . $this->fileHash . ($size !== null ? ('-' . $size) : '') . '.' . $this->avatarExtension;
+        if ($forceWebP === true || ($forceWebP === null && $this->hasWebP && ImageUtil::browserSupportsWebP())) {
```
TimWolla, line 77:
```suggestion
        if (
            $forceWebP === true
            || ($forceWebP === null && $this->hasWebP && ImageUtil::browserSupportsWebP())
       ) {
```
Even if the line length is not exceed this makes it more readable, because each case is handled on a separate line.

wcfsetup/install/files/lib/data/user/avatar/UserAvatarEditor.class.php
```diff
@@ -66,11 +68,81 @@ public function deleteFiles()
...
+            // location.
+            $newFileHash = sha1_file($filenameJpeg);
+
+            $tmpAvatar = clone $this;
+            $tmpAvatar->data["avatarExtension"] = "jpeg";
+            $tmpAvatar->data["fileHash"] = $newFileHash;
+            $newLocation = $tmpAvatar->getLocation(null, false);
+
+            $dir = \dirname($newLocation);
+            if (!@\file_exists($dir)) {
```
TimWolla, line 130: The `@` should not be necessary here.

wcfsetup/install/files/lib/data/user/avatar/UserAvatarEditor.class.php
```diff
@@ -66,11 +68,81 @@ public function deleteFiles()
...
+                FileUtil::makePath($dir);
+            }
+
+            @\rename($filenameJpeg, $newLocation);
```
TimWolla, line 134: And neither should it here. If this `rename` fails we must not silently fail.

wcfsetup/setup/db/install.sql
```diff
@@ -1541,7 +1541,8 @@ CREATE TABLE wcf1_user_avatar (
 	width SMALLINT(5) NOT NULL DEFAULT 0,
 	height SMALLINT(5) NOT NULL DEFAULT 0,
 	userID INT(10),
-	fileHash VARCHAR(40) NOT NULL DEFAULT ''
+	fileHash VARCHAR(40) NOT NULL DEFAULT '',
+    hasWebP TINYINT(1) NOT NULL DEFAULT 0
```
TimWolla, line 1545: Inconsistent indentation. Possible editor misconfiguration after the PSR-12 adjustments?

TimWolla, line 1545: Add this change to https://github.com/WoltLab/WCF/blob/master/wcfsetup/install/files/acp/update_com.woltlab.wcf_5.4_db.php, please.

TimWolla (changes requested):
And of course the trailing whitespace reported by phpcs should be removed.

wcfsetup/install/files/lib/data/user/avatar/UserAvatar.class.php
dtdesign: That was my work, I manually removed the extra spaces.

TimWolla (approved):
LGTM if the computer says "yes".

PR #3895, opened 2021-01-25
Use prettier for SCSS

TimWolla:


dtdesign (approved):
I glanced through the changes and could not find any obvious issues.

PR #3900, opened 2021-01-26
Fix User::hasOwnerAccess()

TimWolla:
- Remove caching from User::hasOwnerAccess()
- Check owner access after checking controller blacklist in RequestHandler


dtdesign (approved):

jens1o:
Can we expect a release in the next few hours for this patch?

PR #3906, opened 2021-01-27
Use the `input` event to detect the comma on Chromium for Android

dtdesign:
see https://community.woltlab.com/thread/287910-uiitemlist-mobil-nicht-per-komma-trennbar/


TimWolla (approved):
Verified using Chrome on Android 10 with SwiftKey.

PR #3911, opened 2021-01-28
Deprecate Master Password

### LINKED ISSUE #3698: Deprecate the master password

TimWolla:
Still up for discussion:

- [x] Skipping the master password if it is not yet set up.
    - Preserve the functionality for "old users".
    - Disable the functionality for "new users".
- [x] Skipping the master password if the current user has MFA enabled.

Resolves #3698

-------------

- Mark the master password as deprecated in the option description
- Show deprecation message on master password authentication

### DIFF
--- a/wcfsetup/install/files/acp/templates/masterPassword.tpl
+++ b/wcfsetup/install/files/acp/templates/masterPassword.tpl
@@ -10,6 +10,10 @@
 	<h1 class="contentTitle">{lang}wcf.acp.masterPassword.enter{/lang}</h1>
 </header>
 
+<p class="warning">
+	{lang}wcf.acp.masterPassword.enter.deprecated{/lang}
+</p>
+
 {include file='formError'}
 
 <form method="post" action="{link controller='MasterPassword'}{/link}">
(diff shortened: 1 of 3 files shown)


dtdesign:
Can't we just keep it for the time being and simply strongly advise to use 2FA instead? The notice could be put up both for the module setting and whenever prompted to enter the master password. It doesn't require any significant changes, but still encourages users to migrate or pick 2FA instead.

TimWolla:
> Can't we just keep it for the time being and simply strongly advise to use 2FA instead? The notice could be put up both for the module setting and whenever prompted to enter the master password. It doesn't require any significant changes, but still encourages users to migrate or pick 2FA instead.

That's exactly what the PR does in its current state. The checkboxes are “up for discussion” for the reason that they are much more invasive, which might not be desired.

dtdesign:
I see no point in further restricting the use of the master password. Maybe use a stronger wording in the notice, such as advising against it with the remark that the feature will be removed in future version without further notice.

BurntimeX (approved):

PR #3928, opened 2021-02-03
Deprecate StyleUtil::updateStyleFile()

### LINKED ISSUE #3927: Deprecate StyleUtil::updateStyleFile()
This method appears useless to me, because the ACP style is compiled when non-existent during the call to `getStylesheet`. `StyleUtil::updateStyleFile()` is only ever used during during package installation / updates and already needed a special workaround for 5.4: https://github.com/WoltLab/WCF/commit/7640fc5b5e0c56319a48e5e99f65a17f2fb27cc2#diff-c586fafd91b59c8475c03b80128652d8f94efaa46c82d736f8ceffff0d46edecR71

TimWolla:
This is a draft PR, because it was not yet tested with WCFSetup and the 5.3 → 5.4 upgrade, yet.

------------

Resolves #3927


dtdesign:
I'm not sure what I'm supposed to review here?

TimWolla:
> I'm not sure what I'm supposed to review here?

I might have missed something during my evaluation of the situation, it could be possible that there is a deeper reason behind the style compilation being part of the package installation. If you think the change is sane in general then feel free to approve and then I'll make the necessary tests and merge.

dtdesign:
You put it there in 7640fc5b5e0c56319a48e5e99f65a17f2fb27cc2, I have no idea why it was required nor what the issue was.

TimWolla:
> You put it there in [7640fc5](https://github.com/WoltLab/WCF/commit/7640fc5b5e0c56319a48e5e99f65a17f2fb27cc2), I have no idea why it was required nor what the issue was.

No. In the previous commit I added the `skipStyleUpdate` parameter: https://github.com/WoltLab/WCF/commit/ed4bfa914fbf911c7ab7fb2568a361e29a758f40 to skip the style update. In the commit you linked I then made use of that parameter, because the update to Guzzle 7 caused issues with the style update.

The style update itself exists since the start of the git history: https://github.com/WoltLab/WCF/blame/5.3/wcfsetup/install/files/lib/system/package/plugin/FilePackageInstallationPlugin.class.php#L85

I now realized that the style update here might not be necessary at all, allowing me to revert the workaround I added for the 5.3 → 5.4 upgrade and possibly make the package system more reliable, because the SCSS compiler will – tests pending – no longer run during the package installation.

dtdesign:
Frankly speaking, I don't know if we _still_ need it. The history suggest that this was possibly even carried over from WCF 1.x, who knows. Try it out and if it fails, keep the previous version, otherwise go ahead with the change. That's the best answer I can offer you here, I simply don't know it.

TimWolla:
> Try it out and if it fails, keep the previous version, otherwise go ahead with the change. That's the best answer I can offer you here, I simply don't know it.

That's sufficient for me.

TimWolla:
Tested WCFSetup and the upgrade. Merging.

PR #3929, opened 2021-02-03
Clean up StyleCompiler

TimWolla:
This PR performs a cleanup within the StyleCompiler in preparation of upcoming changes.

Specifically the compileStylesheet() monster method was simplified by moving
the SCSS preprocessing (~ SCSS file concatenation) and CSS postprocessing (~
header insertion) into the caller. For the latter the necessary code already
resided in the caller and was passed in as a callback method.

While this new style compilation logic might look a bit like copy and paste it
should be no less readable by moving common logic into helper methods and by
simplifying the control and data flow with the removal of the callback.

It also allows us much greater control over the whole process, e.g. allowing us
to insert specific CSS for the frontend styles only.

------------

- Make the StyleCompiler `final`
- Remove the StyleCompiler::$compiler property
- Add StyleCompiler::injectHeader()
- Add StyleCompiler::convertToRtl()
- Add StyleCompiler::writeCss()
- Make StyleCompiler::compileStylesheet() only compile
- Pass SCSS instead of files + invididualScss to StyleCompiler::compileStylesheet()
- Use a `/*!` comment for the CSS header

### DIFF
--- a/wcfsetup/install/files/lib/system/style/StyleCompiler.class.php
+++ b/wcfsetup/install/files/lib/system/style/StyleCompiler.class.php
@@ -552,15 +552,7 @@ EOT;
 
         $content = $callback($content);
 
-        // write stylesheet
-        \file_put_contents($filename . '.css', $content);
-        FileUtil::makeWritable($filename . '.css');
-
-        $content = $this->convertToRtl($content);
-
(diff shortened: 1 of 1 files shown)


dtdesign (approved):
LGTM

PR #3934, opened 2021-02-04
Inverted Permissions for Boxes and Pages

### LINKED ISSUE #3542: Exclude groups from viewing custom boxes
Currently it is only possible to determine which groups are allowed to see a custom box. As soon as the user is a member of one of these groups, it will be displayed. It would make sense to optional reverse the behavior here. Show the box only if the user is not in the selected groups (as it is already possible in other places).

joshuaruesweg:
Closes #3542

### DIFF
--- a/wcfsetup/install/files/acp/templates/aclSimple.tpl
+++ b/wcfsetup/install/files/acp/templates/aclSimple.tpl
@@ -1,5 +1,6 @@
 {if !$__aclSimplePrefix|isset}{assign var='__aclSimplePrefix' value=''}{/if}
 {if !$__aclInputName|isset}{assign var='__aclInputName' value='aclValues'}{/if}
+{if !$__supportsInvertedPermissions|isset}{assign var='__supportsInvertedPermissions' value=false}{/if}
 
 <div class="section">
 	<dl>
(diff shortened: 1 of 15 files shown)


wcfsetup/install/files/lib/data/box/Box.class.php
```diff
@@ -554,7 +555,7 @@ public function getConditions()
      */
     public function isAccessible()
     {
-        return SimpleAclResolver::getInstance()->canAccess('com.woltlab.wcf.box', $this->boxID);
+        return SimpleAclResolver::getInstance()->canAccess('com.woltlab.wcf.box', $this->boxID) xor $this->invertPermissions;
```
TimWolla, line 558: The `xor` is non-obvious here. Make this explicit:

```
$canAccess = …->canAccess();
if ($this->invertPermissions) {
 $canAccess = !$canAccess;
}
return $canAccess;
```

wcfsetup/install/files/lib/data/page/Page.class.php
```diff
@@ -264,7 +265,7 @@ public function isVisible()
      */
     public function isAccessible()
     {
-        return SimpleAclResolver::getInstance()->canAccess('com.woltlab.wcf.page', $this->pageID);
+        return SimpleAclResolver::getInstance()->canAccess('com.woltlab.wcf.page', $this->pageID) xor $this->invertPermissions;
```
TimWolla, line 268: ditto

wcfsetup/setup/db/install.sql
```diff
@@ -312,6 +312,7 @@ CREATE TABLE wcf1_box (
 	linkPageObjectID INT(10) NOT NULL DEFAULT 0,
 	externalURL VARCHAR(255) NOT NULL DEFAULT '',
 	isDisabled TINYINT(1) NOT NULL DEFAULT 0,
+	invertPermissions TINYINT(1) NOT NULL DEFAULT 0,
```
TimWolla, line 315: Add the update instructions to `_db.php`.

wcfsetup/install/files/acp/templates/aclSimple.tpl
```diff
@@ -1,5 +1,6 @@
 {if !$__aclSimplePrefix|isset}{assign var='__aclSimplePrefix' value=''}{/if}
 {if !$__aclInputName|isset}{assign var='__aclInputName' value='aclValues'}{/if}
+{if !$__supportsInvertedPermissions|isset}{assign var='__supportsInvertedPermissions' value='0'}{/if}
```
TimWolla, line 3:
```suggestion
{if !$__supportsInvertedPermissions|isset}{assign var='__supportsInvertedPermissions' value=false}{/if}
```
?

wcfsetup/install/files/acp/templates/boxAdd.tpl
```diff
@@ -473,7 +473,7 @@
 		</div>
 		
 		<div id="acl" class="tabMenuContent">
-			{include file='aclSimple'}
+			{include file='aclSimple' __supportsInvertedPermissions="1"}
```
TimWolla, line 476:
```suggestion
			{include file='aclSimple' __supportsInvertedPermissions=true}
```
? (ditto below)

TimWolla (changes requested):

TimWolla (approved):

Issue #3395, opened 2020-06-25
Improve list of authentication failures


BurntimeX:
- [ ] add new database column to store the failure type (unknown username, invalid password)
- [ ] show the failure type as text in the list
- [ ] add options to filter the list

TimWolla:
Multi-factor authentication failures (c|sh)ould possibly be logged as well.

PR #3950, opened 2021-02-06
Log user authentication failure type

Gravatronics:
See #3395


TimWolla (approved):
This technically is redundant with the `userID` column, but this LGTM. Maybe it should be extended to include MFA failures?

Gravatronics:
> This technically is redundant with the `userID` column, but this LGTM.
Having an explicit column allows additional errors.

> Maybe it should be extended to include MFA failures?
It would be best if you would cover it then as you have written the MFA system.

PR #3962, opened 2021-02-11
Make DevtoolsPackageArchive::openArchive() more efficient

TimWolla:


wcfsetup/install/files/lib/system/devtools/package/DevtoolsPackageArchive.class.php
```diff
@@ -39,13 +37,55 @@ public function openArchive()
...
+                        return false;
+                    }
+
+                    if ($current->isDir()) {
+                        // Check if we are in the project root.
+                        if ($it->getSubPath() === '') {
+                            // Skip acptemplates / files / templates.
+                            if (
+                                \preg_match(
+                                    '/^(acptemplates|files|templates)(_[a-z0-9]+)?$/',
```
Gravatronics, line 61: This change narrows down the previous regular expression. I assume that you checked that the second group matches the rules for abbreviations used elsewhere.

Gravatronics (approved):
What about adding return types to the callback functions?

wcfsetup/install/files/lib/system/devtools/package/DevtoolsPackageArchive.class.php
TimWolla: The abbreviations are derived from the package name (a-z0-9_-), turned into PHP constants (not allowing for the hyphen), thus the only thing that would be missing is the underscore which I expect to be reasonably not used :smiley:

TimWolla:
> What about adding return types to the callback functions?

They are not as useful as the parameter types (which are required for autocompletion), but I added them for completeness.

PR #3991, opened 2021-02-12
Add anchors to account settings

TimWolla:
- Add `id`s to sections in SettingsForm
- Add `id`s to sections in AccountSecurityPage


dtdesign (commented):
Please use a namespace for these ids to reduce the likelihood of collisions. `multifactor` -> `userSettings_multifactor` etc.

TimWolla:
> Please use a namespace for these ids to reduce the likelihood of collisions. `multifactor` -> `userSettings_multifactor` etc.

I opted for `section_` as the prefix. `userSettings_` would be misleading in `accountSecurity` and `accountSecurity_` would be pretty redundant.

dtdesign (approved):

PR #3996, opened 2021-02-15
Fixes twitch clips which contain a dash char

Sir-Will:
Twitch clips can contain a dash (´-`).
Example: https://clips.twitch.tv/NeighborlyManlyBaconMikeHogu-_9zfXbRK47vik1WO


TimWolla:
Picked manually for 5.2, thanks.

PR #3998, opened 2021-02-16
Move `<meta charset="utf-8">` from headInclude.tpl to header.tpl

### LINKED ISSUE #3802: Move `<title>` after `<meta charset="utf-8">`
The page's title might contain unicode characters and thus probably should appear after the charset tag.

TimWolla:
Resolves #3802

### DIFF
--- a/com.woltlab.wcf/templates/headInclude.tpl
+++ b/com.woltlab.wcf/templates/headInclude.tpl
@@ -1,4 +1,3 @@
-<meta charset="utf-8">
 <meta name="viewport" content="width=device-width, initial-scale=1">
 <meta name="format-detection" content="telephone=no">
 {include file='headIncludeRobotsMetaTag'}
--- a/com.woltlab.wcf/templates/header.tpl
+++ b/com.woltlab.wcf/templates/header.tpl
@@ -1,6 +1,7 @@
 {include file='documentHeader'}
 
 <head>
+	<meta charset="utf-8">
 	{if !$pageTitle|isset}
 		{assign var='pageTitle' value=''}
 		{if (!$__wcf->isLandingPage() || !USE_PAGE_TITLE_ON_LANDING_PAGE) && $__wcf->getActivePage() != null && $__wcf->getActivePage()->getTitle()}


dtdesign:
Over the time I have witnessed quite a few setups running a customized `headInclude` template for whatever reasons. I'm a bit concerned that this cause all sorts of strange issues that will eventually end up in the ticket support.

Do we _really_ need to make this change? We got away with the current setup for, like, ever and this doesn't really add any value except for keeping things tidy. I do expect browsers to be smart enough to finalize the charset for the title after they finished parsing at least the document head?

TimWolla:
> Do we _really_ need to make this change?

I am not entirely sure and I also was concerned about modified templates, that's why this one went in as a PR instead of being committed directly like the others. However a modified `headInclude` template should only result in a duplicated `charset`, not in a missing `charset`. I _expect_ that to be safe.

> I do expect browsers to be smart enough to finalize the charset for the title after they finished parsing at least the document head?

Resources on the Internet are not entirely clear regarding possible issues with an inverted order. A document on Google Code (https://code.google.com/archive/p/doctype-mirror/wikis/MetaCharsetAttribute.wiki) says:

> In order for all browsers to recognize a <meta charset> declaration, it must be
>
>    Within the <head> element,
>    Before any elements that contain text, such as the <title> element, AND
>    Within the first 512 bytes of your document, including DOCTYPE and whitespace

I expect this to be about very outdated browsers, but at the very least it's not clear.

I also found a few resources that say a belated `charset` might block the processing / rendering pipeline of the web browser. I remember something about needing to restart processing the document from the very beginning if the charset changes from the top of my head, but I might misremember.

Within WoltLab Suite we also set the `charset` within the `Content-Type` header since a long time (https://github.com/WoltLab/WCF/blob/3634e3262710f12f1cffdb78fb3c1c20841e0d30/wcfsetup/install/files/lib/util/HeaderUtil.class.php#L60). This might've hidden some or even all of the possible issues.

---------

With that said I'm leaning _slightly_ towards “merge this PR”.

dtdesign:
Uh, I made a typo, I meant the `header` template, not `headInclude`. A modified `header` template would mean that there would be no `charset` declaration in the HTML at all.

TimWolla:
Technically the `<title>` could also be moved into the `headInclude` template, though I expect that one to be modified even more often (but users will notice a missing title more easily).

I'll leave the decision up to you. Either merge this as is, request changes, or close without merging.

dtdesign:
I would suggest keeping it as-is. Even causing the top part of the document to be processed again is not an expensive operation. The include of `headInclude` appears at the very top, with the `charset` being on line 5 from the top, ignoring empty lines, nothing but the title appeared before.

Issue #3864, opened 2021-01-10
Suppress unsupported image types for avatars in emails


dtdesign:
Email clients are notorious for their incredibly bad support for modern HTML and CSS. The widespread use of email clients that are incappable of modern image formats are a constant source of confusion for administrators. We should dynamically modify the output of avatars in emails to provide a better experience:

* [x] Force disable WebP avatars and use the fallback image instead.
* [x] The default avatar must not yield a SVG, we must provide a "regular" image format. JPEG is a sufficient alternative.
* [ ] Adjust the email templates.

TimWolla:
All resolved. For posteriority I searched the affected templates via:

```
ag -Q -G 'email_.*.tpl' -- '->getAvatar()->getImageTag('
```

PR #4001, opened 2021-02-17
Add interface to request safe avatar formats

TimWolla:
- Remove deprecated methods from IUserAvatar interface
- Add ISafeFormatAvatar
- Add AvatarDecorator
- Implement ISafeFormatAvatar in UserAvatar
- Return AvatarDecorator in UserProfile::getAvatar()

Related to #3864


dtdesign (commented):
LGTM

TimWolla:
Okay, I've just picked the first commit into master and rebased this one.

--------

Now that the general design is approved:
- [ ] The methods need to be used in all email templates.
- [ ] The default avatar needs to be adjusted.

Regarding (2) the original issue says:

> The default avatar must not yield a SVG, we must provide a "regular" image format. JPEG is a sufficient alternative.

It is not really clear to me how I would efficiently retrieve a non-SVG from the default avatar that is not completely ugly. Any suggestions?

dtdesign:
> It is not really clear to me how I would efficiently retrieve a non-SVG from the default avatar that is not completely ugly. Any suggestions?

Just fall back to the "sillhoute" avatar which is a plain PNG image.

At some point we should consider just removing the avatar from emails entirely. Since images are blocked by default in many email clients anyway, they barely offer any benefits while adding a lot more hoops to jump through.

TimWolla:
> At some point we should consider just removing the avatar from emails entirely. Since images are blocked by default in many email clients anyway, they barely offer any benefits while adding a lot more hoops to jump through.

I wonder if making the change in this PR is worth it then:

- For compatibility reasons it requires this elaborate decorator with the additional interfaces, locking us into supporting an API that is not particularly good.
- All the email templates need to be adjusted, including plugins, otherwise some emails will show the avatar and some will not, leading to an inconsistent user experience.
    - This would need to be repeated after the removal of avatars.
- The format limitations does not affect all email clients. Looking at [Can I Email](https://www.caniemail.com/features/image-webp/) the Webp support does not seem to be too shabby (with Outlook on Windows being the notable example; as always).

dtdesign:
> For compatibility reasons it requires this elaborate decorator with the additional interfaces, locking us into supporting an API that is not particularly good.

Can't we just use some extra wrapper for the template engine that is used for emails only? We just need a consistent method to check if the avatar is requested from within an email context, offloading the logic straight into the avatar classes.

> All the email templates need to be adjusted, including plugins, otherwise some emails will show the avatar and some will not, leading to an inconsistent user experience.

That's not an argument against such a change. There will always be changes that require third party templates to be adjusted to match the default.

> The format limitations does not affect all email clients. Looking at Can I Email the Webp support does not seem to be too shabby (with Outlook on Windows being the notable example; as always).

If we can detect that this is an "email context", we can just deliver the fallback image (= force skip webp) and we're done.

TimWolla:
> Can't we just use some extra wrapper for the template engine that is used for emails only? We just need a consistent method to check if the avatar is requested from within an email context, offloading the logic straight into the avatar classes.

Emails already use the EmailTemplateEngine, so we already have that wrapper. However communicating this context to the avatar from the EmailTemplateEngine probably requires ugly, intransparent access to global variables. I'd like to prevent such global state if possible.

> That's not an argument against such a change. There will always be changes that require third party templates to be adjusted to match the default.

Sure. But I'd like to avoid this happening twice, especially with the intermediate solution requires adding hacky compatibility wrappers that we'll have to maintain until eternity, because someone might rely on them forever.

------

I prefer the automatic detection of the email context over the current solution implemented in this PR, because it would be easier to clean that up later-on. However I'm not terribly happy with either of them.

dtdesign:
Looking further into this, we should keep avatars, but find a way to restrict the types of images. The easiest approach with the best stability would be to modify all email templates containing avatars to generate the url using `getEmailImageTag()`.

The method `getEmailImageTag()` is responsible to generate an url that is safe for use in emails, that is: JPG/PNG/GIF. This is easily done on our end and easily done for third party authors. Bonus points for having a predictable code path without relying on any semi-global state.

TimWolla:
> The easiest approach with the best stability would be to modify all email templates containing avatars to generate the url using `getEmailImageTag()`.

Okay, then that is the solution from this PR, just using a different method name. I'll adjust the default avatar, then this can be merged and the templates be modified afterwards.

dtdesign (approved):

PR #4003, opened 2021-02-18
Allow disabling the admin contact in ContactForm

TimWolla:


wcfsetup/install/files/lib/form/ContactForm.class.php
```diff
@@ -92,7 +92,7 @@ public function readParameters()
         $this->recipientList->getConditionBuilder()->add("contact_recipient.isDisabled = ?", [0]);
         $this->recipientList->readObjects();
 
-        if (\count($this->recipientList) < 0) {
+        if (!\count($this->recipientList)) {
```
joshuaruesweg, line 95: `empty()`?

TimWolla: Doesn't work with DatabaseObjectLists  and doesn't really matter in this case, because the total number of recipients is expected to be << 10 for a seldomly accessed page.

dtdesign (approved):

PR #4004, opened 2021-02-18
Force lowercase during SFS check

SoftCreatR:
See https://community.woltlab.com/thread/288664-stopforumspam-spammer-werden-nicht-an-der-freischaltung-gehindert/

The spammer didn't register as image but as image and since the username and the email address are being checked "as is", the user was able to register without being flagged as potential spammer.

### DIFF
--- a/wcfsetup/install/files/lib/data/blacklist/entry/BlacklistEntry.class.php
+++ b/wcfsetup/install/files/lib/data/blacklist/entry/BlacklistEntry.class.php
@@ -73,7 +73,7 @@ class BlacklistEntry extends DatabaseObject {
 	}
 	
 	protected static function getHash($string) {
-		return hash('sha256', $string, true);
+		return hash('sha256', mb_strtolower($string), true);
 	}
 	
 	protected static function isMatch($type, $occurrences) {


TimWolla:
This change is not correct.

- The local part of an email address is case sensitive and thus must not be normalized ([RFC 5321#2.3.11](https://tools.ietf.org/html/rfc5321#section-2.3.11)).
    - I however acknowledge that in the real world email addresses tend to be case insensitive and that the risk of a false-positive match is appropriately low.
- For usernames, where normalization actually is viable without ignoring standards it breaks the matching, because SFS provides usernames within a case-preserving list, that is used as-is within the hash database.
    - Thus lowercasing the username will make username matching completely non-functional.
- Thinking about Unicode simply lowercasing the value to perform a match is not going to cut it either. The string should also be normalized (probably using the NFC normalization).

In addition this normalization would need to be appropriately documented / indicated, because performing this kind of normalization effectively creates data out of thin air and thus is not just an alternative representation of the SFS database, but a completely distinct database. This might or might not have legal liability implications in case of a false-positive.

In any case performing such a normalization probably would not be qualified within the stability guarantees of 5.2.

The best solution going forward probably would be if the SFS themselves would provide pre-hashed lists with a clear indication of the normalization that needs be performed to match within those pre-hashed lists.

TimWolla:
Closing this as per the comment above and the discussion within the linked thread.

PR #4021, opened 2021-02-24
Reset the internal state of the embedded object manager to its previous state

### LINKED ISSUE #4018: Nested HtmlOutputProcessors with embedded objects modify the internal state
This is kind of an obscure bug that just recently surfaced in combination with the bbcodes for threads and posts. It was observed that multiple of such bbcodes within one message will cause only the first one to be rendered. Actually, depending on the circumstances, additional bbcodes could be rendered, if they are embedded objects of the first referenced object.

The issue unfolds like this:
1. A message with support for embedded objects is being rendered.
2. `HtmlOutputProcessor::setContext()` sets the active message of `MessageEmbeddedObjectManager`.
3. During processing, a bbcode is encounted that makes use of `HtmlOutputProcessor` itself (here: Render an excerpt of a post).
4. The same logic as in step 2 happens, setting the internal state of the `MessageEmbeddedObjectManager` to the object in step 3.
5. Once the embedded message is rendered, the control flow switches back to the original message.
6. The next bbcode in this message attempts to read from the `MessageEmbeddedObjectManager`, but the active message points to the embedded message in step 3 now.
7. The bbcode is unable to render the data and its fallback logic kicks in.

dtdesign:
Fixes #4018

### DIFF
--- a/wcfsetup/install/files/lib/system/html/output/HtmlOutputProcessor.class.php
+++ b/wcfsetup/install/files/lib/system/html/output/HtmlOutputProcessor.class.php
@@ -65,11 +65,18 @@ class HtmlOutputProcessor extends AbstractHtmlProcessor {
 	public function process($html, $objectType, $objectID, $doKeywordHighlighting = true, $languageID = null) {
 		$this->languageID = $languageID;
 		$this->setContext($objectType, $objectID);
-		
-		$this->getHtmlOutputNodeProcessor()->setOutputType($this->outputType);
-		$this->getHtmlOutputNodeProcessor()->enableKeywordHighlighting($doKeywordHighlighting);
-		$this->getHtmlOutputNodeProcessor()->load($this, $html);
-		$this->getHtmlOutputNodeProcessor()->process();
+
+		MessageEmbeddedObjectManager::getInstance()->setActiveMessage($objectType, $objectID, $this->languageID);
+
+		try {
+			$this->getHtmlOutputNodeProcessor()->setOutputType($this->outputType);
+			$this->getHtmlOutputNodeProcessor()->enableKeywordHighlighting($doKeywordHighlighting);
+			$this->getHtmlOutputNodeProcessor()->load($this, $html);
+			$this->getHtmlOutputNodeProcessor()->process();
+		}
+		finally {
+			MessageEmbeddedObjectManager::getInstance()->reset();
+		}
 	}
 	
 	/**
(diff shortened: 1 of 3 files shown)


wcfsetup/install/files/lib/system/html/output/HtmlOutputProcessor.class.php
```diff
@@ -90,7 +90,11 @@ public function setOutputType($outputType) {
 	 * @inheritDoc
 	 */
 	public function getHtml() {
-		return $this->getHtmlOutputNodeProcessor()->getHtml();
+		$html = $this->getHtmlOutputNodeProcessor()->getHtml();
+
+		MessageEmbeddedObjectManager::getInstance()->reset();
```
TimWolla, line 95: Will this cause issues if `->getHtml()` is being called more than once for a single message?

wcfsetup/install/files/lib/system/message/embedded/object/MessageEmbeddedObjectManager.class.php
```diff
@@ -66,6 +66,12 @@ class MessageEmbeddedObjectManager extends SingletonFactory {
 		'insert' => [],
 		'remove' => []
 	];
+
+	/**
+	 * A list of previous active message settings used to restore
+	 * the internal state in case of nested message processing.
+	 */
+	protected $acitveMessageHistory = [];
```
TimWolla, line 74:
```suggestion
	protected $activeMessageHistory = [];
```
(adjust the other location yourself, please).

wcfsetup/install/files/lib/system/message/embedded/object/MessageEmbeddedObjectManager.class.php
```diff
@@ -283,10 +289,36 @@ public function getContentLanguageID() {
...
 		$this->activeMessageObjectTypeID = ObjectTypeCache::getInstance()->getObjectTypeIDByName('com.woltlab.wcf.message', $messageObjectType);
 		$this->activeMessageID = $messageID;
 		$this->activeMessageLanguageID = $languageID;
 	}
+
+	/**
+	 * Restores the internal state in case of nested message processing.
+	 */
+	public function reset() {
+		$newState = array_pop($this->acitveMessageHistory);
```
TimWolla, line 309: When merging this up: Don't forget to add the `\` (or simply also add it for 5.3).

TimWolla (changes requested):
Overall this looks good to me, just a few small nits.

wcfsetup/install/files/lib/system/html/output/HtmlOutputProcessor.class.php
dtdesign: Good catch, this would indeed cause issues. Duh.

TimWolla (approved):

PR #4023, opened 2021-02-25
Add `title` attribute to device name in session list

TimWolla:
Not sure if having a `title` attribute on a `<h3>` element violates some style
guidelines, but I considered this the most fitting location.


dtdesign:
This is bad UX, if you want to include this information, then show it in its own line. If you're concerned about the size, you can use `<small>`.

PR #4033, opened 2021-02-26
Visibility toggle for password inputs

### LINKED ISSUE #3392: Visibility toggle for password inputs
password-toggle2
password-toggle

BurntimeX:
Closes #3392

### DIFF
--- a/ts/WoltLabSuite/Core/Ui/Password.ts
+++ b/ts/WoltLabSuite/Core/Ui/Password.ts
@@ -1,3 +1,12 @@
+/**
+ * Visibility toggle for password input fields.
+ *
+ * @author Marcel Werk
+ * @copyright	2001-2021 WoltLab GmbH
+ * @license	GNU Lesser General Public License <http://opensource.org/licenses/lgpl-license.php>
+ * @module	WoltLabSuite/Core/Ui/Password
+ */
+
 import DomChangeListener from "../Dom/Change/Listener";
 import * as Language from "../Language";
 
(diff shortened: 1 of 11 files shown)


dtdesign:
The password strength check is currently sort of broken, because it does not expect the password input to be already wrapped.

ts/WoltLabSuite/Core/Ui/Password.ts
```diff
@@ -0,0 +1,57 @@
+import DomChangeListener from "../Dom/Change/Listener";
```
TimWolla, line 1: Add a module comment.

ts/WoltLabSuite/Core/Ui/Password.ts
```diff
@@ -0,0 +1,57 @@
...
+    if (event.key === "Enter" || event.key === " ") {
+      event.preventDefault();
+      toggle(input, button, icon);
+    }
+  });
+}
+
+function toggle(input: HTMLInputElement, button: HTMLElement, icon: HTMLElement): void {
+  icon.classList.toggle("fa-eye");
+  icon.classList.toggle("fa-eye-slash");
```
TimWolla, line 52: Avoid using `toggle` when more logic independent of changing classes happens. Otherwise you might get inconsistent states.

I suggest to rewrite this whole method into a very explicit `if` for readability (avoids the use of ternaries).

```
if (input.type === "password") {
  icon.classList.add("fa-eye");
  icon.classList.add("fa-eye-slash");
  button.dataset.tooltip = Language.get(…);
  input.type = "text";
}
```

TimWolla (approved):
LGTM as far as I'm qualified to tell.

PR #4036, opened 2021-02-26
Unfurl url

### LINKED ISSUE #3964: Support for URL unfurling
Embed a short preview of external URLs if they are placed in a dedicated line of text. The data must be fetched asynchronously and rely on OG attributes only for legal reasons, since these attributes are especially designed for sharing.

* [ ] Only consider external URLs that are on a dedicated line of next with no content before or after it.
* [ ] Asynchronously load the data, just display the plain link while the data has not been fetched yet.
* [ ] Rely on the `meta` elements `og:title`, `og:description` and `og:image` only.

See https://community.woltlab.com/thread/277017-url-unfurling/

joshuaruesweg:
TODO: 

* [x] Save Fetch time
* [x] ~~Refetch entries, if they are too old and embedded again~~
* [x] CSS
* [x] Update SQL-Update script
* [x] Check rebuilding data (URLs should not be fetched if you rebuild e.g. posts)
* [x] Support `og:image:url`
* [x] ~~Add own ua for image proxy~~

----------------

Resolves #3964


wcfsetup/install/files/lib/util/UnfurlUrlUtil.class.php
```diff
@@ -0,0 +1,254 @@
...
+                    throw new \DomainException();
+                }
+            }
+
+            // download image
+            try {
+                $client = HttpFactory::makeClient([
+                    RequestOptions::TIMEOUT => 10,
+                    RequestOptions::STREAM => true,
+                ]);
```
TimWolla, line 208: Do not create a fresh client for every download. Use a helper method:

```
public static function getHttpClient(): ???
```

wcfsetup/setup/db/install.sql
```diff
@@ -1432,6 +1432,19 @@ CREATE TABLE wcf1_trophy(
 	KEY(categoryID)
 );
 
+DROP TABLE IF EXISTS wcf1_unfurl_url;
+CREATE TABLE wcf1_unfurl_url (
```
TimWolla, line 1436: Don't forget to add this to the DB update script when you are finished.

wcfsetup/setup/db/install.sql
```diff
@@ -1432,6 +1432,19 @@ CREATE TABLE wcf1_trophy(
...
+	urlID INT(10) NOT NULL AUTO_INCREMENT PRIMARY KEY,
+	url TEXT NOT NULL,
+	urlHash VARCHAR(40) NOT NULL UNIQUE KEY,
```
TimWolla, line 1439: Make this an explicit `UNIQUE KEY (urlHash)` at the end of the table definition.

wcfsetup/setup/db/install.sql
```diff
@@ -1432,6 +1432,19 @@ CREATE TABLE wcf1_trophy(
...
+	title VARCHAR(255) NOT NULL DEFAULT '',
+	description TEXT, 
+	imageUrl TEXT, 
+	imageType ENUM('SQUARED', 'COVER', 'NOIMAGE') NOT NULL DEFAULT 'NOIMAGE',
+	imageHash VARCHAR(40) NOT NULL DEFAULT '',
+	status ENUM('PENDING', 'SUCCESSFUL', 'REJECTED') DEFAULT 'PENDING'
```
TimWolla, line 1445: I tend to avoid `ENUM`s in favor of a simple `VARCHAR`. This makes future adjustments easier.

wcfsetup/setup/db/install.sql
```diff
@@ -1432,6 +1432,19 @@ CREATE TABLE wcf1_trophy(
...
+DROP TABLE IF EXISTS wcf1_unfurl_url;
+CREATE TABLE wcf1_unfurl_url (
+	urlID INT(10) NOT NULL AUTO_INCREMENT PRIMARY KEY,
+	url TEXT NOT NULL,
+	urlHash VARCHAR(40) NOT NULL UNIQUE KEY,
+	title VARCHAR(255) NOT NULL DEFAULT '',
+	description TEXT, 
+	imageUrl TEXT, 
+	imageType ENUM('SQUARED', 'COVER', 'NOIMAGE') NOT NULL DEFAULT 'NOIMAGE',
+	imageHash VARCHAR(40) NOT NULL DEFAULT '',
```
TimWolla, line 1444: Do you need a dedicated `imageHash`? Can't you simply reuse the `urlHash` for the image? Especially since you only have one image per URL?

wcfsetup/install/files/lib/util/UnfurlUrlUtil.class.php
```diff
@@ -0,0 +1,254 @@
...
+                $request = new Request('GET', $url, [
+                    'via' => '1.1 wsc',
```
TimWolla, line 210: Note to myself: I need to check whether this is correct here.

wcfsetup/install/files/lib/util/UnfurlUrlUtil.class.php
```diff
@@ -0,0 +1,254 @@
...
+     */
+    private function fetchUrl(): void
+    {
+        try {
+            $client = HttpFactory::makeClient([
+                RequestOptions::TIMEOUT => 10,
+                RequestOptions::STREAM => true,
+            ]);
+            $request = new Request('GET', $this->url, [
+                'range' => \sprintf('bytes=%d-%d', 0, self::MAX_SIZE - 1),
```
TimWolla, line 70: Note to self: I'd like to check whether we should use a custom `user-agent` here.

TimWolla (changes requested):

wcfsetup/install/files/lib/util/UnfurlUrlUtil.class.php
TimWolla: Remove this header, it should not be here, because we are not acting as a proxy and we don't need to prevent embed loops.

wcfsetup/install/files/lib/system/background/job/UnfurlUrlJob.class.php
```diff
@@ -0,0 +1,159 @@
+<?php
+
+namespace wcf\system\background\job;
+
+use wcf\data\unfurl\url\UnfurlUrl;
+use wcf\data\unfurl\url\UnfurlUrlAction;
+use function wcf\functions\exception\logThrowable;
```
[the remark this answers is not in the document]
joshuaruesweg: Should I use the FQN in the function instead?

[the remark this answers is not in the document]
TimWolla: Yes, please. See: https://github.com/WoltLab/WCF/search?q=logThrowable

wcfsetup/install/files/lib/system/html/node/HtmlNodeUnfurlLink.class.php
```diff
@@ -0,0 +1,35 @@
...
+    public static function setUnfurl(HtmlNodePlainLink $link): void
+    {
+        if ($link->isStandalone()) {
+            $object = new UnfurlUrlAction([], 'findOrCreate', [
+                'data' => [
+                    'url' => $link->href,
+                ],
+            ]);
+            $returnValues = $object->executeAction();
+            $link->topLevelParent->firstChild->setAttribute(self::UNFURL_URL_ID_ATTRIBUTE_NAME, $returnValues['returnValues']->urlID);
```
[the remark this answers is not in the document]
joshuaruesweg: No, the ID is then overwritten when it is saved. It is also not possible to enter any ID to see what may have been linked in an internal area.

wcfsetup/install/files/lib/system/background/job/UnfurlUrlBackgroundJob.class.php
```diff
@@ -0,0 +1,160 @@
...
+
+                        // filter images which are too large or too small
+                        $isSquared = $imageData[0] === $imageData[1];
+                        if (
+                            (!$isSquared && ($imageData[0] < 300 && $imageData[1] < 150))
+                            || \min($imageData[0], $imageData[1]) < 50
+                        ) {
+                            $data['imageType'] = UnfurlUrl::IMAGE_NO_IMAGE;
+                        } else {
+                            if ($imageData[0] === $imageData[1]) {
```
jens1o, line 95:
```suggestion
                            if ($isSquared) {
```

joshuaruesweg: Thanks. But keep in mind, that these Pull Request is WIP and not cleaned at all.

wcfsetup/install/files/lib/util/UnfurlUrlUtil.class.php
```diff
@@ -0,0 +1,254 @@
...
+/**
+ * Helper class to unfurl specific urls.
+ *
+ * @author      Joshua Ruesweg
+ * @copyright   2001-2021 WoltLab GmbH
+ * @license     GNU Lesser General Public License <http://opensource.org/licenses/lgpl-license.php>
+ * @package     WoltLabSuite\Core\Util
+ * @since       5.4
+ */
+final class UnfurlUrlUtil
```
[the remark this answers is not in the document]
joshuaruesweg: I dislike the classname `Url`, because we have already a class named `Url` which is used inside this class. What about `UnfurledUrl`?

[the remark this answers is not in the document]
TimWolla: I recently attempted to somewhat avoid the [Smurf Naming Convention](https://devcards.io/smurf-naming-convention), because it tends to get overly long in many cases.

However I agree that the `Url` clash is unfortunate and that `Url` also is a super generic class name. I don't particularly like `UnfurledUrl`, though. How about `UrlResponse` or `UnfurlResponse`? These still should combine rather nicely with the proposed static `fetch()` method.

wcfsetup/setup/db/install.sql
joshuaruesweg: No, not really. But then I need a separate column to save the file type.

joshuaruesweg: But actually yes: Otherwise I don't know, whether I have saved the image, or not (because the image proxy is enable previously)…

wcfsetup/install/files/lib/util/UnfurlUrlUtil.class.php
```diff
@@ -0,0 +1,254 @@
...
+        return null;
+    }
+    
+    /**
+     * Returns the image url for the current url.
+     */
+    public function getImageUrl(): ?string
+    {
+        if (!empty($this->body)) {
+            $metaTags = $this->getDomDocument()->getElementsByTagName('meta');
```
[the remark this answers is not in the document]
joshuaruesweg: `\DomDocument` is way better, if the website responses invalid HTML.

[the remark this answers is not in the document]
TimWolla: You should be able to use XPath with `DomDocument`:

```
$x = new \DOMXPath($document);
$x->query("//meta[@property='og:image']"); // DOMNodeList
```

TimWolla (changes requested):

wcfsetup/install/files/lib/system/message/unfurl/UnfurlResponse.class.php
```diff
@@ -0,0 +1,294 @@
...
+     *
+     * @throws ParsingFailed If the body cannot be parsed (e.g. the url is an JSON file).
+     */
+    private function readDomDocument(): void
+    {
+        \libxml_use_internal_errors(true);
+        $this->domDocument = new \DOMDocument();
+        if (!$this->domDocument->loadHTML('<?xml version="1.0" encoding="UTF-8"?>' . $this->body)) {
+            throw new ParsingFailed("Could not parse body.");
+        }
```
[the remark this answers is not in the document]
joshuaruesweg: Is there any documentation, what exactly an internal error is?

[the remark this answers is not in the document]
TimWolla: The method name is a bit of a misnomer. `\libxml_use_internal_errors()` controls whether PHP errors should be generated or not. We don't want it to generate PHP errors, because that is super wonky.

wcfsetup/setup/db/install.sql
joshuaruesweg: I personally dislike to have an `VARCHAR` there. With an `ENUM` you can guarantee, that there is an correct value, even if the user edits the table by himself (for whatever reason).

But I have changed it to an VARCHAR.

wcfsetup/install/files/lib/system/background/job/UnfurlUrlBackgroundJob.class.php
```diff
@@ -0,0 +1,240 @@
...
+    }
+
+    private function downloadImage(Response $imageResponse): string
+    {
+        $image = "";
+        while (!$imageResponse->getBody()->eof()) {
+            $image .= $imageResponse->getBody()->read(8192);
+
+            if ($imageResponse->getBody()->tell() >= UnfurlResponse::MAX_IMAGE_SIZE) {
+                break;
```
TimWolla, line 135: Thinking about this: A `break;` probably is not super useful here. The image request should simply fail of the response length is too long. Otherwise we'll truncate the image, leading to a broken image.

Thus I suggest to:
```suggestion
                return "";
```

and then removing the `Range` request header. This should likewise be adjusted in ImageProxy.

wcfsetup/install/files/lib/data/unfurl/url/UnfurlUrl.class.php
```diff
@@ -0,0 +1,116 @@
...
+     */
+    public static function getByUrl(string $url): ?self
+    {
+        if (!Url::is($url)) {
+            throw new \InvalidArgumentException("Given URL is not a valid URL.");
+        }
+
+        $sql = "SELECT		unfurl_url.*
+				FROM		wcf" . WCF_N . "_unfurl_url unfurl_url
+				WHERE		unfurl_url.urlHash = ?";
```
TimWolla, line 106: Fix the formatting here.

com.woltlab.wcf/templates/unfurlUrl.tpl
```diff
@@ -0,0 +1,14 @@
+{if $object->status == "SUCCESSFUL"}
+	<div class="unfurlCard {if $object->imageType == 'COVER' && !$object->getImageUrl()|empty}unfurlLargeContentImage{elseif $object->imageType == 'SQUARED' && !$object->getImageUrl()|empty}unfurlSquaredContentImage{/if}">
+		<a href="{$object->url}"{if EXTERNAL_LINK_REL_NOFOLLOW || EXTERNAL_LINK_TARGET_BLANK} rel="{if EXTERNAL_LINK_REL_NOFOLLOW}nofollow{/if}{if EXTERNAL_LINK_REL_NOFOLLOW && EXTERNAL_LINK_TARGET_BLANK} {/if}{if EXTERNAL_LINK_TARGET_BLANK}noopener noreferrer{/if}"{/if}{if EXTERNAL_LINK_TARGET_BLANK} target="_blank"{/if}>
+			<div{if !$object->getImageUrl()|empty} style="background-image: url('{$object->getImageUrl()}')"{/if}></div>
```
TimWolla, line 4: Will this cause issues if the image URL contains a `'`?

com.woltlab.wcf/templates/unfurlUrl.tpl
```diff
@@ -0,0 +1,14 @@
...
+			<div class="unfurlInformation">
+				<div class="urlTitle">{$object->title}</div>
+				<div class="urlDescription">{$object->description}</div>
+				<div class="urlHost">{$object->getHost()}</div>
+			</div>
+		</a>
+	</div>
+{else}
+	<a href="{$object->url}" class="externalURL" {if EXTERNAL_LINK_REL_NOFOLLOW || EXTERNAL_LINK_TARGET_BLANK} rel="{if EXTERNAL_LINK_REL_NOFOLLOW}nofollow{/if}{if EXTERNAL_LINK_REL_NOFOLLOW && EXTERNAL_LINK_TARGET_BLANK} {/if}{if EXTERNAL_LINK_TARGET_BLANK}noopener noreferrer{/if}"{/if}{if EXTERNAL_LINK_TARGET_BLANK} target="_blank"{/if}>{$object->url}</a>
```
TimWolla, line 13: Use the `{anchor}` tag. The `rel="ugc"` is missing here.

TimWolla (changes requested):

wcfsetup/setup/db/install.sql
```diff
@@ -1446,6 +1446,20 @@ CREATE TABLE wcf1_trophy(
...
+DROP TABLE IF EXISTS wcf1_unfurl_url;
+CREATE TABLE wcf1_unfurl_url (
+	urlID INT(10) NOT NULL AUTO_INCREMENT PRIMARY KEY,
+	url TEXT NOT NULL,
+	urlHash VARCHAR(40) NOT NULL UNIQUE KEY (urlHash),
+	title VARCHAR(255) NOT NULL DEFAULT '',
+	description TEXT,
+	imageUrl TEXT NOT NULL DEFAULT '',
+	imageType VARCHAR(255) NOT NULL DEFAULT 'NOIMAGE',
+	imageHash VARCHAR(45) NOT NULL DEFAULT '',
```
TimWolla, line 1458: I dislike this, because it contains both the extension as well as the hash itself. May I suggest the following:

1. Use the `urlHash` as the `imageHash`.
2. Change to column to `storedImageExtension VARCHAR(255) DEFAULT NULL,`

And then using `storedImageExtension` in place of `imageHash`. If it is `null` no image is stored locally.

wcfsetup/setup/db/install.sql
```diff
@@ -1446,6 +1446,20 @@ CREATE TABLE wcf1_trophy(
 	KEY(categoryID)
 );
 
+DROP TABLE IF EXISTS wcf1_unfurl_url;
+CREATE TABLE wcf1_unfurl_url (
+	urlID INT(10) NOT NULL AUTO_INCREMENT PRIMARY KEY,
+	url TEXT NOT NULL,
+	urlHash VARCHAR(40) NOT NULL UNIQUE KEY (urlHash),
```
TimWolla, line 1453: You misunderstood the part regarding the UNIQUE KEY. It should go at the bottom (similar to `username` for `wcf1_user`).

TimWolla (changes requested):

wcfsetup/setup/db/install.sql
joshuaruesweg: I dislike using the `urlHash` as `imageHash`. 

On many pages (e.g. WoltLab.com) the same image is used as the logo on various pages. If we save the image with the `urlHash` we duplicate the image quite often. If we save the `imageHash`, we store the image only once.

TimWolla: Maybe a dedicated database table for the images is in order then? All the `image*` columns should be moved there, with this table only containing the `imageID`. This theoretically would also allow pruning the images independently after a certain time to further save disk space. The `imageType` would then be automatically derived from the information stored in that image table. If `imageID IS NULL`, then no image will be shown.

wcfsetup/setup/db/install.sql
```diff
@@ -1446,6 +1446,20 @@ CREATE TABLE wcf1_trophy(
...
+	title VARCHAR(255) NOT NULL DEFAULT '',
+	description TEXT,
+	imageUrl TEXT NOT NULL DEFAULT '',
```
TimWolla, line 1456: This can't have a real default value (Error in query (1101): BLOB, TEXT, GEOMETRY or JSON column 'imageUrl' can't have a default value ). I suggest:
```suggestion
	imageUrl TEXT DEFAULT NULL,
```

wcfsetup/setup/db/install.sql
```diff
@@ -1446,6 +1446,20 @@ CREATE TABLE wcf1_trophy(
...
+	imageType VARCHAR(255) NOT NULL DEFAULT 'NOIMAGE',
+	imageHash VARCHAR(45) NOT NULL DEFAULT '',
+	status VARCHAR(255) NOT NULL DEFAULT 'PENDING',
+	lastFetch INT(10) NOT NULL DEFAULT 0
```
TimWolla, line 1460:
```suggestion
	lastFetch INT(10) NOT NULL
```
The default just doesn't make sense. It should be explicitly provided.

TimWolla (changes requested):

wcfsetup/install/files/lib/system/background/job/UnfurlUrlBackgroundJob.class.php
```diff
@@ -0,0 +1,243 @@
...
+            if (empty(StringUtil::trim($unfurlResponse->getTitle()))) {
+                $this->save(UnfurlUrl::STATUS_REJECTED);
+
+                return;
+            }
+
+            $title = StringUtil::truncate($unfurlResponse->getTitle(), 255);
+            $description = "";
+            if ($unfurlResponse->getDescription()) {
+                $description = StringUtil::truncate($unfurlResponse->getDescription());
```
TimWolla, line 79:
```suggestion
                $description = StringUtil::truncate($unfurlResponse->getDescription(), 1000);
```
Otherwise the method will default to 80.

wcfsetup/setup/db/install.sql
```diff
@@ -1446,6 +1446,32 @@ CREATE TABLE wcf1_trophy(
...
+);
+
+DROP TABLE IF EXISTS wcf1_unfurl_url_image;
+CREATE TABLE wcf1_unfurl_url_image (
+	imageID INT(10) NOT NULL AUTO_INCREMENT PRIMARY KEY,
+	imageUrl TEXT NOT NULL,
+	imageHash VARCHAR(40) DEFAULT NULL,
+	width INT(10) NOT NULL,
+	height INT(10) NOT NULL,
+	imageExtension VARCHAR(4) DEFAULT NULL,
```
TimWolla, line 1470:
```suggestion
	imageExtension VARCHAR(255) DEFAULT NULL,
```
You are not saving space with just 4 characters and it unnecessarily restricts future extensions.

wcfsetup/install/files/acp/update_com.woltlab.wcf_5.4_db.php
```diff
@@ -222,6 +223,60 @@
...
+            ObjectIdDatabaseTableColumn::create('imageID'),
+            TextDatabaseTableColumn::create('imageUrl')
+                ->notNull(),
+            VarcharDatabaseTableColumn::create('imageHash')
+                ->notNull()
+                ->length(40),
+            NotNullInt10DatabaseTableColumn::create('width'),
+            NotNullInt10DatabaseTableColumn::create('height'),
+            VarcharDatabaseTableColumn::create('imageExtension')
+                ->length(4),
```
TimWolla, line 238: see remark in install.sql

wcfsetup/install/files/lib/data/unfurl/url/UnfurlUrl.class.php
```diff
@@ -0,0 +1,171 @@
...
+        return $this->getImageType() == self::IMAGE_SQUARED && !empty($this->getImageUrl());
+    }
+
+    private function getImageType(): string
+    {
+        if (!$this->imageID) {
+            return self::IMAGE_NO_IMAGE;
+        }
+
+        if ($this->width == $this->height) {
```
TimWolla, line 139:
```suggestion
        if ($this->width === $this->height) {
```

wcfsetup/install/files/lib/data/unfurl/url/UnfurlUrl.class.php
```diff
@@ -0,0 +1,171 @@
...
+                return $this->imageUrl;
+            }
+        }
+
+        return null;
+    }
+
+    public function hasCoverImage(): bool
+    {
+        return $this->getImageType() == self::IMAGE_COVER && !empty($this->getImageUrl());
```
TimWolla, line 125:
```suggestion
        return $this->getImageType() === self::IMAGE_COVER && !empty($this->getImageUrl());
```

wcfsetup/install/files/lib/data/unfurl/url/UnfurlUrl.class.php
```diff
@@ -0,0 +1,171 @@
...
+    }
+
+    public function hasSquaredImage(): bool
+    {
+        return $this->getImageType() == self::IMAGE_SQUARED && !empty($this->getImageUrl());
```
TimWolla, line 130:
```suggestion
        return $this->getImageType() === self::IMAGE_SQUARED && !empty($this->getImageUrl());
```

wcfsetup/install/files/lib/system/background/job/UnfurlUrlBackgroundJob.class.php
```diff
@@ -0,0 +1,270 @@
...
+
+        return $imageHash;
+    }
+
+    private function getImageExtension(array $imageData): ?string
+    {
+        switch ($imageData[2]) {
+            case \IMAGETYPE_PNG:
+                return 'png';
+                break;
```
TimWolla, line 216: The `break`s are useless here.

TimWolla (changes requested):

wcfsetup/setup/db/install.sql
joshuaruesweg: No, because, if the URL is not fetched yet, it is `0` (and no, we don't use `null` for that at any other place). And every other column has a default value, too.

wcfsetup/install/files/lib/system/background/job/UnfurlUrlBackgroundJob.class.php
```diff
@@ -0,0 +1,270 @@
...
+        array $imageData = []
+    ): void {
+        switch ($status) {
+            case UnfurlUrl::STATUS_PENDING:
+            case UnfurlUrl::STATUS_REJECTED:
+            case UnfurlUrl::STATUS_SUCCESSFUL:
+                break;
+
+            default:
+                throw new BadMethodCallException("Invalid status '{$status}' given.");
```
TimWolla, line 243: InvalidArgumentException?

wcfsetup/install/files/lib/system/background/job/UnfurlUrlBackgroundJob.class.php
```diff
@@ -0,0 +1,270 @@
...
+        $fileLocation = $path . '/' . $imageHash . '.' . $extension;
+
+        \file_put_contents($fileLocation, $image);
+
+        @\touch($fileLocation);
+
+        return $imageHash;
+    }
+
+    private function getImageExtension(array $imageData): ?string
```
TimWolla, line 211: How about [`ImageUtil::getExtensionByMimeType()`](https://github.com/WoltLab/WCF/blob/9e4fe52bb9cf024c0986f6df860aa2780d425b4c/wcfsetup/install/files/lib/util/ImageUtil.class.php#L236-L241)?

wcfsetup/install/files/lib/data/unfurl/url/UnfurlUrl.class.php
```diff
@@ -0,0 +1,171 @@
...
+
+    private const IMAGE_NO_IMAGE = "NOIMAGE";
+
+    public const STATUS_PENDING = "PENDING";
+
+    public const STATUS_SUCCESSFUL = "SUCCESSFUL";
+
+    public const STATUS_REJECTED = "REJECTED";
+
+    public const IMAGE_DIR = "images/unfurlUrl/";
```
TimWolla, line 48:
```suggestion
    private const IMAGE_DIR = "images/unfurlUrl/";
```

wcfsetup/install/files/lib/data/unfurl/url/UnfurlUrlAction.class.php
```diff
@@ -0,0 +1,102 @@
...
+{
+    /**
+     * @inheritDoc
+     */
+    public function create()
+    {
+        if (isset($this->parameters['imageData']) && !empty($this->parameters['imageData'])) {
+            $this->parameters['data']['imageID'] = $this->saveImageData($this->parameters['imageData']);
+        }
+
```
TimWolla, line 32: During creation it is not useful to provide `imageData`.

wcfsetup/install/files/lib/system/background/job/UnfurlUrlBackgroundJob.class.php
```diff
@@ -0,0 +1,270 @@
...
+        try {
+            $unfurlResponse = UnfurlResponse::fetch($unfurlUrl->url);
+
+            if (empty(StringUtil::trim($unfurlResponse->getTitle()))) {
+                $this->save(UnfurlUrl::STATUS_REJECTED);
+
+                return;
+            }
+
+            $title = StringUtil::truncate($unfurlResponse->getTitle(), 255);
```
TimWolla, line 78: The title is not being trimmed which might result in a title that is completely empty after truncation. I recommend to return a trimmed value in `->getTitle()` for consistent behavior.

wcfsetup/install/files/lib/system/background/job/UnfurlUrlBackgroundJob.class.php
```diff
@@ -0,0 +1,270 @@
...
+            $description = "";
+            if ($unfurlResponse->getDescription()) {
```
TimWolla, line 80: Ditto: trim the description.

wcfsetup/install/files/lib/system/background/job/UnfurlUrlBackgroundJob.class.php
```diff
@@ -0,0 +1,270 @@
...
+            $this->save(UnfurlUrl::STATUS_REJECTED);
+        }
+    }
+
+    private function getImageData(UnfurlResponse $unfurlResponse): array
+    {
+        $imageSaveData = [];
+
+        if (empty($unfurlResponse->getImageUrl()) || !Url::is($unfurlResponse->getImageUrl())) {
+            throw new BadMethodCallException("Invalid image given.");
```
TimWolla, line 115:
```suggestion
            throw new \InvalidArgumentException("The given response does not have an image.");
```

wcfsetup/install/files/lib/system/background/job/UnfurlUrlBackgroundJob.class.php
```diff
@@ -0,0 +1,270 @@
...
+        $path = WCF_DIR . 'images/unfurlUrl/' . \substr($imageHash, 0, 2);
+        FileUtil::makePath($path);
+
+        $extension = $this->getImageExtension($imageData);
+
+        $fileLocation = $path . '/' . $imageHash . '.' . $extension;
+
+        \file_put_contents($fileLocation, $image);
+
+        @\touch($fileLocation);
```
TimWolla, line 206: Remove this. This is useless.

TimWolla (changes requested):

wcfsetup/setup/db/install.sql
joshuaruesweg: I won't change it, because it is technically correct. The intention is *not*  to save space, it is to validate incorrect data is inserted there. Yes, theoretically, there are file extensions with more then 4 characters, but we don't use them. We only use `jpg`, `gif`, `png` and, here and there, `webp`. If we want to use a file extension named `.coolExtension`, then we must adjust this, but I don't see any extension like this in the near future. 

AFAIK we use the `VARCHAR(4)` (or `VARCHAR(3)`) for the file extension on many other places, too. Yes, this is not a real argument, but if we want to support other file extensions, we must adjust these places, too.

wcfsetup/install/files/lib/system/background/job/UnfurlUrlBackgroundJob.class.php
joshuaruesweg: If the title is empty (after trimming them), then the url would be rejected. But yes, I should trim the title before I truncate them.

com.woltlab.wcf/templates/unfurlUrl.tpl
joshuaruesweg: Correct me, if I am wrong, but URLs, which contains `'` are not valid. I can't find any valid URL which contains a `'`. 

Invalid URLs are not stored and are rejected (See https://github.com/WoltLab/WCF/pull/4036/files#diff-a83e91d395d63d2a6354a05b163ddce84bb12f5536eacdca1393cc8113538f87R90).

wcfsetup/install/files/lib/system/message/unfurl/UnfurlResponse.class.php
```diff
@@ -0,0 +1,339 @@
...
+    {
+        if (!empty($this->body)) {
+            // og:title
+            $list = $this->domXPath->query("//meta[@property='og:title']");
+            foreach ($list as $node) {
+                foreach ($node->attributes as $attribute) {
+                    if ($attribute->nodeName === 'content') {
+                        return $attribute->nodeValue;
+                    }
+                }
```
TimWolla, line 218: https://www.php.net/manual/en/domelement.getattribute.php?

wcfsetup/install/files/lib/system/background/job/UnfurlUrlBackgroundJob.class.php
```diff
@@ -0,0 +1,275 @@
...
+                        $imageSaveData['imageExtension'] = $this->getImageExtension($imageData);
+                        $imageSaveData['isStored'] = 1;
+                    }
+                }
+            }
+        } catch (UrlInaccessible | DownloadFailed $e) {
+            $imageSaveData = [];
+        }
+
+        return $imageSaveData;
```
TimWolla, line 149:
```suggestion
        $imageSaveData = [];

        if (empty($unfurlResponse->getImageUrl()) || !Url::is($unfurlResponse->getImageUrl())) {
            throw new \InvalidArgumentException("The given response does not have an image.");
        }

        try {
            $imageResponse = $unfurlResponse->getImage();
            $image = $this->downloadImage($imageResponse);
            $imageData = \getimagesizefromstring($image);

            if ($imageData === false) {
                return [];
            }
                if (!$this->validateImage($imageData)) {
                    return [];
                }
                $imageSaveData = [
                 'imageUrl' => $unfurlResponse->getImageUrl(),
                    'imageUrlHash'] => \sha1($unfurlResponse->getImageUrl()),
                    // …
                    ];
                    if (!(MODULE_IMAGE_PROXY || IMAGE_ALLOW_EXTERNAL_SOURCE)) {
                        $this->saveImage($imageData, $image, $imageSaveData['imageUrlHash']);
                        $imageSaveData['imageExtension'] = $this->getImageExtension($imageData);
                        $imageSaveData['isStored'] = 1;
                    }
                    return $imageSaveData;
        } catch (UrlInaccessible | DownloadFailed $e) {
            return [];
        }
```

Ignore the formatting, you get the point: Reduce the nesting level here and use early returns.

joshuaruesweg: I don't think, that this is more readable, but I will apply it…

wcfsetup/install/files/lib/system/background/job/UnfurlUrlBackgroundJob.class.php
```diff
@@ -0,0 +1,275 @@
...
+
+            if (!$this->validateImage($imageData)) {
+                return [];
+            }
+
+            $imageSaveData = [];
+            $imageSaveData['imageUrl'] = $unfurlResponse->getImageUrl();
+            $imageSaveData['imageUrlHash'] = \sha1($unfurlResponse->getImageUrl());
+            $imageSaveData['width'] = $imageData[0];
+            $imageSaveData['height'] = $imageData[1];
```
TimWolla, line 139: I'd merge these as an array literal, this reduces cognitive load, because the variable name only appears once:
```suggestion
            $imageSaveData = [
                'imageUrl' => $unfurlResponse->getImageUrl(),
                'imageUrlHash' => \sha1($unfurlResponse->getImageUrl()),
                'width' => $imageData[0],
                'height' => $imageData[1],
            ];
```

wcfsetup/install/files/lib/system/background/job/UnfurlUrlBackgroundJob.class.php
```diff
@@ -0,0 +1,275 @@
...
+            if (!(MODULE_IMAGE_PROXY || IMAGE_ALLOW_EXTERNAL_SOURCE)) {
+                $this->saveImage($imageData, $image, $imageSaveData['imageUrlHash']);
+                $imageSaveData['imageExtension'] = $this->getImageExtension($imageData);
+                $imageSaveData['isStored'] = 1;
+            }
+        } catch (UrlInaccessible | DownloadFailed $e) {
+            $imageSaveData = [];
+        }
+
+        return $imageSaveData;
```
TimWolla, line 149:
```suggestion
            }

            return $imageSaveData;
        } catch (UrlInaccessible | DownloadFailed $e) {
            return [];
        }
```
Use early return within the catch and do not add a return below the catch, because it is unreachable by definition.

TimWolla (commented):
LGTM if this last remark is applied. I did not yet perform a final functional test.

wcfsetup/install/files/lib/system/background/job/UnfurlUrlBackgroundJob.class.php
```diff
@@ -0,0 +1,280 @@
...
+
+    /**
+     * @inheritDoc
+     */
+    public function perform()
+    {
+        $unfurlUrl = new UnfurlUrl($this->urlID);
+
+        try {
+            $unfurlResponse = UnfurlResponse::fetch($unfurlUrl->url);
```
TimWolla, line 70:
```suggestion
            if (Url::parse($unfurlUrl->url)['port']) {
                throw new UrlInaccessible('Refusing to unfurl non-standard ports');
            }

            $unfurlResponse = UnfurlResponse::fetch($unfurlUrl->url);
```

wcfsetup/install/files/acp/database/update_com.woltlab.wcf_5.4.php
```diff
@@ -1,7 +1,7 @@
 <?php
 
 /**
- * Makes non-critical database adjustments (i.e. everything that is not related
+     * Makes non-critical database adjustments (i.e. everything that is not related
```
joshuaruesweg, line 4: .

TimWolla (approved):

PR #4037, opened 2021-02-26
Improved the UI for the re-authentication prompt

dtdesign:
Overhauled the UI to mirror the appearance of the login dialog. The dysfunctional UI elements were quite distracting and everything behaved in a strange way, since the system does not expect a logged-in user to be _de facto_ unable to do anything.

Keeping the visuals as close as possible to the login dialog also improves safety by providing a UI that the user is familiar with.

image


TimWolla (approved):
Eyeballed the diff, looked at the screenshot, this LGTM.

PR #4048, opened 2021-03-04
Use .scrollableCheckboxList in UserGroupCondition

### LINKED ISSUE #3939: Use scrollableCheckboxList for user group checkbox lists
In installations with many user groups, forms with allow the selection of user groups (like the user search form) can become quite long. To avoid such long user group lists, `scrollableCheckboxList` should be used.

TimWolla:
Resolves #3939


TimWolla:
Bildschirmfoto von 2021-03-04 11-51-07

dtdesign (approved):

PR #4055, opened 2021-03-08
Add ViewableUserOption#setOptionValue event

Sir-Will:
Add ViewableUserOption#setOptionValue event to allow option value manipulation.

While creating spam restrictions I was looking into restricting links in the profile of new users without preventing them from being able to create their profile with links. Unfortunately I couldn't find any way to accomplish that without additional events.

### DIFF
--- a/wcfsetup/install/files/lib/data/user/option/ViewableUserOption.class.php
+++ b/wcfsetup/install/files/lib/data/user/option/ViewableUserOption.class.php
@@ -5,6 +5,7 @@ namespace wcf\data\user\option;
 use wcf\data\DatabaseObjectDecorator;
 use wcf\data\user\User;
 use wcf\system\cache\builder\UserOptionCacheBuilder;
+use wcf\system\event\EventHandler;
 use wcf\system\exception\ImplementationException;
 use wcf\system\exception\SystemException;
 use wcf\system\option\user\IUserOptionOutput;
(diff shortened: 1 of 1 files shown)


TimWolla:
I believe @dtdesign's reply on the sibling issue also applies here: https://github.com/WoltLab/WCF/pull/4056#issuecomment-793092098

dtdesign:
It is a bit odd to enforce the restrictions on runtime rather than preventing the values being set in the first place. This looks like you should reject the rogue data during validation when the user first enters them?

Sir-Will:
The goal is to allow the users to fill out their profile without restrictions but protect other users from malicious links.
Depending on the activity of the user, the user will be promoted and the restrictions will be removed.

Rejecting any links while submitting would mean that the user would need to go back to the profile after the promotion, which creates a bad user experience.
The restrictions are also only visible to other users and not the user itself.

dtdesign:
It is already possible to limit the editing capabilities to the user profile, opening up the customization options for some user groups only. From our experience this is an extremely powerful method to drive spammers away, but have little to no impact on actual users. It is also consistent by only allowing data to be entered if it is actually going to be displayed.

Your approach is quite awkward and contradicts the overall design principles of the UI/UX of the software. Please feel free to provide more arguments that strengthen your point.

Sir-Will:
My approach:
1. User creates account
2. Fills out profile (profile shows link as plain text) 
3. Reads few threads and posts 
4. Gets promoted and the link is no longer being changed to plain text.

Not allowing to fill out the field approach:
1. User creates account
2. Fills out profile (except link related fields, users and staff get less information about the new user, meaning less info to determine if it's a spam user)
3. Reads few threads and posts
4. Gets promoted
5. Needs to be notified about new unlocked profile fields
6. Needs to go back to the profile to fill out the new profile fields

Sir-Will:
I was able to achieve this now by using the OptionHandler#afterReadCache event and overwriting and extending the outputClass.

PR #4056, opened 2021-03-08
Add SignatureCache#getSignature event

Sir-Will:
Add SignatureCache#getSignature event to allow modifying the signature.

Same as https://github.com/WoltLab/WCF/pull/4055

### DIFF
--- a/wcfsetup/install/files/lib/system/user/signature/SignatureCache.class.php
+++ b/wcfsetup/install/files/lib/system/user/signature/SignatureCache.class.php
@@ -3,6 +3,7 @@
 namespace wcf\system\user\signature;
 
 use wcf\data\user\User;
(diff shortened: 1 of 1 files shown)


dtdesign:
This even makes no sense to me. Right above it is the call to the HtmlOutputProcessor which has proper event access including the details on the context (=> user id).

What problems do you aim to solve with this event?

Sir-Will:
Oh, I didn't see that event. It seems to work in this case, though it only provides the userID and not the complete User object which would help to check for a user option value, permissions or groups.

> What problems do you aim to solve with this event?

There are spam users who sign up only to set a link in their signature, profile and so on. If the user is only in the registered group, the links will be turned into plain text until the user gets promoted.

PR #4058, opened 2021-03-09
Show error messages if tainted apps are installed

TimWolla:


wcfsetup/install/files/acp/templates/index.tpl
```diff
@@ -15,6 +15,12 @@
 	{/foreach}
 {/if}
 
+{if !$taintedApplications|empty}
+	{foreach from=$taintedApplications item=$taintedApplication}
+		<div class="error">{lang}wcf.acp.package.application.isTainted{/lang}</div>
+	{/foreach}
+{/if}
```
dtdesign, line 22:
```suggestion
{foreach from=$taintedApplications item=$taintedApplication}
	<div class="error">{lang}wcf.acp.package.application.isTainted{/lang}</div>
{/foreach}
```

Less noise in the template, PHP will happily iterate over an empty array.

wcfsetup/install/files/acp/templates/packageList.tpl
```diff
@@ -58,6 +58,12 @@
 	{/hascontent}
 </header>
 
+{if !$taintedApplications|empty}
+	{foreach from=$taintedApplications item=$taintedApplication}
+		<div class="error">{lang}wcf.acp.package.application.isTainted{/lang}</div>
+	{/foreach}
+{/if}
```
dtdesign, line 65:
```suggestion
{foreach from=$taintedApplications item=$taintedApplication}
	<div class="error">{lang}wcf.acp.package.application.isTainted{/lang}</div>
{/foreach}
```

wcfsetup/install/files/lib/acp/page/IndexPage.class.php
```diff
@@ -170,6 +170,15 @@ public function assignVariables()
             }
         }
 
+        $taintedApplications = [];
+        foreach (ApplicationHandler::getInstance()->getApplications() as $application) {
```
dtdesign, line 174: Maybe use `array_filter()` instead? Although its syntax is a bit awkward, it does improve readability quite a lot. Also, do we even care for the `packageID` as the index?

wcfsetup/install/files/lib/acp/page/PackageListPage.class.php
```diff
@@ -95,9 +96,19 @@ public function assignVariables()
     {
         parent::assignVariables();
 
+        $taintedApplications = [];
+        foreach (ApplicationHandler::getInstance()->getApplications() as $application) {
```
dtdesign, line 100: See the remark on the `wcf\acp\page\IndexPage`.

dtdesign (changes requested):

wcfsetup/install/files/acp/templates/index.tpl
TimWolla: I've stolen this from the surrounding code (above). Will adjust all of the locations.

wcfsetup/install/files/lib/acp/page/IndexPage.class.php
TimWolla: I very much like `array_filter`, but it's also like this for consistency with the code above. The `packageID` in the index is used for consistency with the PackageList, which needs it to show the warning icon besides the package name.

PR #4060, opened 2021-03-09
Fix reading ACL values in non-Ajax form builder forms

Gravatronics:
The wrong data source was used in `AclFormField` (the whole `$_POST` array instead of the dedicated entry) and the data was always stored in `aclValues` instead of a dedicated entry per form field.


TimWolla (approved):
LGTM as far as I am qualified to tell.

PR #4062, opened 2021-03-10
Support filtering the list of user authentication failures

Gravatronics:
Filtering by IP address is not supported due storing IPv4 addresses in IPv6 format but displaying them in as IPv4 so that (partial) IPv4 addresses cannot be (easily) searched for.

See #3395


TimWolla (approved):
LGTM once "Computer says yes".

PR #4064, opened 2021-03-10
Add 'type' to user blocking

### LINKED ISSUE #3981: Split the block feature into two modes
The block feature currently blocks a user, restricts the ability to interact and hides their content. This is fine for most cases, but there are situations where a user should be blocked without removing their content from the view.

* [ ] Stage 1: Block the user, preventing conversations, mentions and follows.
* [ ] Stage 2: Hide the content from the blocked users. (This is the current behavior)

See https://community.woltlab.com/thread/282068-mehr-optionen-f%C3%BCr-die-blockierfunktion/

TimWolla:
Still a draft, because the `type` needs to be integrated into the consumers.
For now only the blocking UI has been adjusted.

-------

Resolves #3981

### DIFF
--- a/com.woltlab.wcf/templates/boxRecentActivitySidebar.tpl
+++ b/com.woltlab.wcf/templates/boxRecentActivitySidebar.tpl
@@ -1,6 +1,6 @@
 <ul class="sidebarItemList">
 	{foreach from=$eventList item=event}
-		<li class="box24{if $__wcf->getUserProfileHandler()->isIgnoredUser($event->getUserProfile()->userID)} ignoredUserContent{/if}">
+		<li class="box24{if $__wcf->getUserProfileHandler()->isIgnoredUser($event->getUserProfile()->userID, 2)} ignoredUserContent{/if}">
 			{user object=$event->getUserProfile() type='avatar24' ariaHidden='true' tabindex='-1'}
 			
 			<div class="sidebarItemTitle">
--- a/com.woltlab.wcf/templates/membersList.tpl
+++ b/com.woltlab.wcf/templates/membersList.tpl
@@ -134,13 +134,10 @@
 	$(function() {
 		WCF.Language.addObject({
 			'wcf.user.button.follow': '{jslang}wcf.user.button.follow{/jslang}',
-			'wcf.user.button.ignore': '{jslang}wcf.user.button.ignore{/jslang}',
 			'wcf.user.button.unfollow': '{jslang}wcf.user.button.unfollow{/jslang}',
-			'wcf.user.button.unignore': '{jslang}wcf.user.button.unignore{/jslang}'
 		});
 		
 		new WCF.User.Action.Follow($('.userList > li'));
-		new WCF.User.Action.Ignore($('.userList > li'));
 	});
 	
 	require(['WoltLabSuite/Core/Ui/User/Search/Input'], (UiUserSearchInput) => {
--- a/ts/WoltLabSuite/Core/Ui/User/Profile/Menu/Item/Ignore.ts
+++ b/ts/WoltLabSuite/Core/Ui/User/Profile/Menu/Item/Ignore.ts
@@ -1,41 +1,47 @@
 import * as Core from "../../../../../Core";
 import * as Language from "../../../../../Language";
-import { AjaxCallbackSetup, ResponseData } from "../../../../../Ajax/Data";
 import * as UiNotification from "../../../../Notification";
 import UiUserProfileMenuItemAbstract from "./Abstract";
+import FormBuilderDialog from "../../../../../Form/Builder/Dialog";
 
-interface AjaxResponse extends ResponseData {
-  returnValues: {
-    isIgnoredUser: 1 | 0;
-  };
+interface AjaxResponse {
+  isIgnoredUser: 1 | 0;
 }
 
 class UiUserProfileMenuItemIgnore extends UiUserProfileMenuItemAbstract {
+  private readonly dialog: FormBuilderDialog;
+
   constructor(userId: number, isActive: boolean) {
     super(userId, isActive);
+
+    this.dialog = new FormBuilderDialog("ignoreDialog", "wcf\\data\\user\\ignore\\UserIgnoreAction", "getDialog", {
+      dialog: {
+        title: Language.get("wcf.user.button.ignore"),
+      },
+      actionParameters: {
+        userID: this._userId,
+      },
+      submitActionName: "submitDialog",
+      successCallback: (r: AjaxResponse) => this._ajaxSuccess(r),
+      destroyOnClose: true,
+    });
   }
 
   _getLabel(): string {
     return Language.get("wcf.user.button." + (this._isActive ? "un" : "") + "ignore");
   }
 
-  _getAjaxActionName(): string {
-    return this._isActive ? "unignore" : "ignore";
-  }
-
   _ajaxSuccess(data: AjaxResponse): void {
-    this._isActive = !!data.returnValues.isIgnoredUser;
+    this._isActive = !!data.isIgnoredUser;
     this._updateButton();
 
     UiNotification.show();
   }
 
-  _ajaxSetup(): ReturnType<AjaxCallbackSetup> {
-    return {
-      data: {
-        className: "wcf\\data\\user\\ignore\\UserIgnoreAction",
-      },
-    };
+  protected _toggle(event: MouseEvent): void {
+    event.preventDefault();
+
+    this.dialog.open();
   }
 }
 
--- a/wcfsetup/install/files/lib/data/user/ignore/UserIgnoreAction.class.php
+++ b/wcfsetup/install/files/lib/data/user/ignore/UserIgnoreAction.class.php
@@ -5,10 +5,17 @@ namespace wcf\data\user\ignore;
 use wcf\data\AbstractDatabaseObjectAction;
 use wcf\data\user\follow\UserFollow;
 use wcf\data\user\follow\UserFollowEditor;
+use wcf\data\user\User;
 use wcf\system\cache\runtime\UserProfileRuntimeCache;
 use wcf\system\exception\IllegalLinkException;
 use wcf\system\exception\PermissionDeniedException;
 use wcf\system\exception\UserInputException;
+use wcf\system\form\builder\data\processor\CustomFormDataProcessor;
+use wcf\system\form\builder\DialogFormDocument;
+use wcf\system\form\builder\field\RadioButtonFormField;
+use wcf\system\form\builder\field\validation\FormFieldValidationError;
+use wcf\system\form\builder\field\validation\FormFieldValidator;
+use wcf\system\form\builder\IFormDocument;
 use wcf\system\user\storage\UserStorageHandler;
 use wcf\system\WCF;
 
@@ -26,6 +33,8 @@ use wcf\system\WCF;
  */
 class UserIgnoreAction extends AbstractDatabaseObjectAction
 {
+    protected $form;
+
     /**
      * Validates the 'ignore' action.
      */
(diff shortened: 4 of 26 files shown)


ts/WoltLabSuite/Core/Ui/User/Profile/Menu/Item/Ignore.ts
```diff
@@ -3,39 +3,48 @@ import * as Language from "../../../../../Language";
...
+      submitActionName: "submitDialog",
+      // TODO: https://github.com/WoltLab/WCF/issues/4063
+      successCallback: (r) => this._ajaxSuccess(r as any),
+      destroyOnClose: true,
+    });
   }
 
   _ajaxSuccess(data: AjaxResponse): void {
-    this._isActive = !!data.returnValues.isIgnoredUser;
+    console.log(data);
```
TimWolla, line 33: Debug Code

wcfsetup/install/files/lib/data/user/ignore/UserIgnoreAction.class.php
```diff
@@ -121,6 +151,100 @@ public function unignore()
...
+                    ->options([
+                        UserIgnore::TYPE_NO_IGNORE => WCF::getLanguage()
+                            ->get('wcf.user.ignore.type.noIgnore'),
+                        UserIgnore::TYPE_BLOCK_DIRECT_CONTACT => WCF::getLanguage()
+                            ->get('wcf.user.ignore.type.blockDirectContact'),
+                        UserIgnore::TYPE_HIDE_MESSAGES => WCF::getLanguage()
+                            ->get('wcf.user.ignore.type.hideMessages'),
+                    ])
+                    ->value($ignore->type ?: 0),
+            ]);
```
TimWolla, line 229: Missing permission check whether the user in question may be ignored.

ts/WoltLabSuite/Core/Ui/User/Profile/Menu/Item/Ignore.ts
```diff
@@ -1,41 +1,47 @@
...
 
-  _ajaxSetup(): ReturnType<AjaxCallbackSetup> {
-    return {
-      data: {
-        className: "wcf\\data\\user\\ignore\\UserIgnoreAction",
-      },
-    };
+  _getLabel(): string {
+    return Language.get("wcf.user.button." + (this._isActive ? "un" : "") + "ignore");
+  }
```
TimWolla, line 39: Move this up again, this uselessly enlarges the diff.

wcfsetup/install/files/lib/data/user/ignore/UserIgnoreAction.class.php
```diff
@@ -42,6 +49,21 @@ public function validateIgnore()
         if ($userProfile->getPermission('user.profile.cannotBeIgnored')) {
             throw new PermissionDeniedException();
         }
+
+        $this->readInteger('type', true, 'data');
+
+        if (!$this->parameters['data']['type']) {
+            $this->parameters['data']['type'] = UserIgnore::TYPE_BLOCK_DIRECT_CONTACT;
```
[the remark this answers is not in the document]
dtdesign: Makes sense, then just keep it as it currently is in the code. This was merely a question.

TimWolla:
I believe I made all the changes required for WoltLab/WCF, so this one is ready to review.

wcfsetup/install/lang/de.xml
```diff
@@ -4809,6 +4809,11 @@ sich{/if} nicht bei uns registriert {if LANGUAGE_USE_INFORMAL_VARIANT}hast{else}
 		<item name="wcf.user.reauthentication.explanation"><![CDATA[{if LANGUAGE_USE_INFORMAL_VARIANT}Du betrittst einen besonders geschützten Bereich. Aus Sicherheitsgründen ist es notwendig, dass du dich durch Eingabe deines Kennworts erneut authentifizierst.{else}Sie betreten einen besonders geschützten Bereich. Aus Sicherheitsgründen ist es notwendig, dass Sie sich durch Eingabe Ihres Kennworts erneut authentifizieren.{/if}]]></item>
 		<item name="wcf.user.reauthentication.loginAs"><![CDATA[Angemeldet als]]></item>
 		<item name="wcf.user.reauthentication.logoutAndChangeUser"><![CDATA[Nicht {$__wcf->user->username}? <a href="{link controller='FullLogout' application='wcf'}t={csrfToken type='url'}{/link}">Abmelden und Benutzer wechseln</a>.]]></item>
+		<item name="wcf.user.ignore.type"><![CDATA[Benutzer blockieren]]></item>
+		<item name="wcf.user.ignore.type.noIgnore"><![CDATA[Nicht Blockieren]]></item>
+		<item name="wcf.user.ignore.type.blockDirectContact"><![CDATA[Direkten Kontakt blockieren]]></item>
+		<item name="wcf.user.ignore.type.hideMessages"><![CDATA[Erstellte Inhalte ausblenden]]></item>
```
Gravatronics, line 4815: This language item does not indicate that direct contact is also blocked with this option (same for English).

wcfsetup/setup/db/install.sql
```diff
@@ -1645,6 +1645,7 @@ CREATE TABLE wcf1_user_ignore (
 	userID INT(10) NOT NULL,
 	ignoreUserID INT(10) NOT NULL,
 	time INT(10) NOT NULL DEFAULT 0,
+	type TINYINT(1) NOT NULL DEFAULT 1,
```
Gravatronics, line 1648: Missing property documentation in `UserIgnore`.

com.woltlab.wcf/templates/membersList.tpl
```diff
@@ -134,13 +134,10 @@
 	$(function() {
 		WCF.Language.addObject({
 			'wcf.user.button.follow': '{jslang}wcf.user.button.follow{/jslang}',
-			'wcf.user.button.ignore': '{jslang}wcf.user.button.ignore{/jslang}',
 			'wcf.user.button.unfollow': '{jslang}wcf.user.button.unfollow{/jslang}',
-			'wcf.user.button.unignore': '{jslang}wcf.user.button.unignore{/jslang}'
 		});
 		
 		new WCF.User.Action.Follow($('.userList > li'));
-		new WCF.User.Action.Ignore($('.userList > li'));
```
Gravatronics, line 143: Why is ignoring no longer possible on list pages?

com.woltlab.wcf/templates/boxRecentActivitySidebar.tpl
```diff
@@ -1,6 +1,6 @@
 <ul class="sidebarItemList">
 	{foreach from=$eventList item=event}
-		<li class="box24{if $__wcf->getUserProfileHandler()->isIgnoredUser($event->getUserProfile()->userID)} ignoredUserContent{/if}">
+		<li class="box24{if $__wcf->getUserProfileHandler()->isIgnoredUser($event->getUserProfile()->userID, 2)} ignoredUserContent{/if}">
```
Gravatronics, line 3: If only class constants could be used in templates…

Gravatronics (dismissed):

com.woltlab.wcf/templates/membersList.tpl
TimWolla: See the commit message here: https://github.com/WoltLab/WCF/pull/4064/commits/27eea3e631312f8111d48550eb342902d00fbe5e

dtdesign:
I do not like the use of the class parameter for the "type" of blocking, it clearly shows its ugly face in templates. Instead, I would like to see this getting split up into two methods:

1. `isIgnoredByUser($userID)` - Behaves as before, full block (equals `TYPE_HIDE_MESSAGES`).
2. `isBlockedByUser($userID)` - Equals `TYPE_BLOCK_DIRECT_CONTACT`.

The naming ain't perfect, but it is close enough to both preserve backwards compatibility and enables cleaner code by avoding magic numbers.

TimWolla:
> The naming ain't perfect, but it is close enough to both preserve backwards compatibility and enables cleaner code by avoding magic numbers.

Should I preserve the `$type` parameter and just make `isBlockedByUser` a small wrapper around `isIgnoredByUser`, passing an  appropriate `$type` for use in templates?

dtdesign:
> > The naming ain't perfect, but it is close enough to both preserve backwards compatibility and enables cleaner code by avoding magic numbers.
> 
> Should I preserve the `$type` parameter and just make `isBlockedByUser` a small wrapper around `isIgnoredByUser`, passing an appropriate `$type` for use in templates?

Can't we just drop the `$type` parameter? It does create some redundancy, but I think that is acceptable.

dtdesign:
Okay, then I would like to suggest to keep the method `isIgnoredUser()` as it is, returning true if the user is blocked, regardless of what setting is enforced. Do we even need to distinguish between these types for the reverse lookup? `isBlockedByUser(int $userID): bool`  should be sufficient to prevent any notifications from being generated and things like a direct contact / follow. The content blocking is not relevant for a reverse lookup.

TimWolla:
> `isBlockedByUser(int $userID): bool` should be sufficient to prevent any notifications from being generated and things like a direct contact / follow. The content blocking is not relevant for a reverse lookup.

Indeed. This might even leak some data that is considered private. I'll fix this and will think about the rest.

TimWolla:
Okay, I looked into this again. I don't think we are able to easily get rid of the magic number in templates:

- `getIgnoredUsers()` / `isIgnoredUser()` _cannot_ be modified for compatibility reasons and thus must return all blocked / ignored users regardless of type.
- Most conditions in the template want to know whether messages should be hidden, thus the users with **type = direct contact** need to be excluded.

Currently this exclusion is performed by explicitly passing the `2` magic number as the filter parameter. A helper method could theoretically be added, but naming is hard and the obvious name (“isIgnoredUser”) is taken. A “isBlockedUser” method would be useless, because this is rarely, if ever needed in templates – we want to get the users for whom messages should be hidden.

Thus the choice is between:

- Magic '2' in templates, and
- Awkwardly named helper method (suggestions?).

Gravatronics:
@TimWolla You can add return types to the new methods in the action class.

wcfsetup/install/lang/de.xml
TimWolla: Any suggestions? Radio buttons unfortunately do not support descriptions and I want to keep the phrases succinct. I'd say that it is pretty much obvious that this includes “direct contact” to be useful.

Maybe simple prefix “Zusätzlich”?

TimWolla:
> @TimWolla You can add return types to the new methods in the action class.

I feel that `array` is pretty much useless there, especially since the methods are never called directly. I'd rather wait for more useful type support.

com.woltlab.wcf/templates/boxRecentActivitySidebar.tpl
TimWolla: I opted to preserve the magic '2', because it is the least ugly solution.

wcfsetup/install/lang/de.xml
Gravatronics: I do not think that repeating the previous phrase is a bad thing.

TimWolla: image
?

I find this hard to scan, because of the common prefix between both options.

Gravatronics: Yes, I prefer it this way because it is clear and unambiguous.

wcfsetup/install/files/lib/data/user/ignore/UserIgnoreAction.class.php
```diff
@@ -121,6 +152,126 @@ public function unignore()
...
+
+        $this->readString('formId');
+
+        $this->getForm()->requestData($this->parameters['data'] ?? []);
+        $this->getForm()->readValues();
+    }
+
+    public function submitDialog()
+    {
+        $this->getForm()->validate();
```
Gravatronics, line 194: Move this line to the end of `validateSubmitDialog()`.

Gravatronics (changes requested):
I think that instead of the delete icon on the page listing the currently blocked users, the dialog should be shown for also being able to change the setting.

wcfsetup/install/files/lib/data/user/ignore/UserIgnoreAction.class.php
Gravatronics: Yes, because the return value inside the condition has to be returned in `submitDialog()`.

TimWolla:
> I think that instead of the delete icon on the page listing the currently blocked users, the dialog should be shown for also being able to change the setting.

I'll create a follow up for this: #4140

dtdesign (approved):

PR #4071, opened 2021-03-12
added missing column wcf1_user_session.spiderID

mutec:
fixes #4067 which want's to use this column on line 1025 in `changeUserVirtual()`

/cc @TimWolla

### DIFF
--- a/wcfsetup/install/files/acp/update_com.woltlab.wcf_5.4_session_2_user_session.php
+++ b/wcfsetup/install/files/acp/update_com.woltlab.wcf_5.4_session_2_user_session.php
@@ -29,6 +29,8 @@ $tables = [
                 ->notNull(),
             IntDatabaseTableColumn::create('userID')
                 ->length(10),
+            IntDatabaseTableColumn::create('spiderID')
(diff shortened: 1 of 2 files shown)


xopez (approved):

TimWolla:
No, this is not the correct fix. The correct fix is reverting `0d262d1080533b952de104f45df7cf5a360d8892` and then applying the change to the `wcf1_session` table.

I could swear I tested this.

mutec:
> No, this is not the correct fix. The correct fix is reverting [0d262d1](https://github.com/WoltLab/WCF/commit/0d262d1080533b952de104f45df7cf5a360d8892) and then applying the change to the `wcf1_session` table.
Damn, it was a 50:50 chance. :D

TimWolla:
> Damn, it was a 50:50 chance. :D

Wouldn't make much sense to introduce a new column just to write a value in it and never read it back out again and *then* forgetting to actually commit it, no? :grin:

mutec:
One thing I learned the last years: I don't have to understand everything. :D

PR #4078, opened 2021-03-14
add missing space during user search

mutec:
Otherwise the query is going to look like: `SELECT user_table.userID FROM wcf1_user user_table LEFT JOIN wcf1_user_option_value option_value ON option_value.userID = user_table.userIDWHERE user_table.username LIKE ? LIMIT 1000`


TimWolla:
Looking at the code it looks like 5.3 should be affected as well, but when testing I am not seeing an error there. I'm confused.

TimWolla:
Fixed differently, thanks.

PR #4079, opened 2021-03-14
include WSC-domain within HTTPRequest useragent

mutec:
this is a big benefit for identifying unwanted sessions in the new security-section of the user account management
It helps the user to identify sessions created by their own instances reading e.g. package information, api-calls or RSS-feeds

### DIFF
--- a/wcfsetup/install/files/lib/system/io/HttpFactory.class.php
+++ b/wcfsetup/install/files/lib/system/io/HttpFactory.class.php
@@ -6,6 +6,7 @@ use GuzzleHttp\Client;
 use GuzzleHttp\ClientInterface;
 use GuzzleHttp\RequestOptions;
 use InvalidArgumentException;
+use wcf\system\application\ApplicationHandler;
 use wcf\system\Regex;
 
 /**
@@ -33,15 +34,16 @@ final class HttpFactory
      */
     public static function getDefaultUserAgent(?string $comment = null): string
     {
+        $domain = ApplicationHandler::getInstance()->getApplicationByID(1)->domainName;
         if ($comment) {
             if (!Regex::compile("^[a-zA-Z0-9_:/\\. -]+$")->match($comment)) {
                 throw new InvalidArgumentException("Invalid comment for user agent given.");
             }
 
-            return \sprintf('WoltLabSuite/%s (%s)', \wcf\getMinorVersion(), $comment);
+            return \sprintf('WoltLabSuite/%s %s (%s)', \wcf\getMinorVersion(), $domain, $comment);
         }
 
-        return \sprintf('WoltLabSuite/%s', \wcf\getMinorVersion());
+        return \sprintf('WoltLabSuite/%s %s', $domain, \wcf\getMinorVersion());
     }
 
     /**
--- a/wcfsetup/install/files/lib/util/HTTPRequest.class.php
+++ b/wcfsetup/install/files/lib/util/HTTPRequest.class.php
@@ -11,6 +11,7 @@ use ParagonIE\ConstantTime\Hex;
 use Psr\Http\Message\RequestInterface;
 use Psr\Http\Message\ResponseInterface;
 use Psr\Http\Message\UriInterface;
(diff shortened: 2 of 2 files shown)


wcfsetup/install/files/lib/system/io/HttpFactory.class.php
```diff
@@ -33,15 +34,16 @@ final class HttpFactory
      */
     public static function getDefaultUserAgent(?string $comment = null): string
     {
+        $domain = ApplicationCacheBuilder::getInstance()->getData()['application'][1]->domainName;
```
Gravatronics, line 37: Why are you using `ApplicationCacheBuilder` instead of `ApplicationHandler` (same below)?

mutec: 503 brain not available. I forgot it was called `ApplicationHandler` and searched for a cache-class; just fixed it.

wcfsetup/install/files/lib/system/io/HttpFactory.class.php
```diff
@@ -33,15 +34,16 @@ final class HttpFactory
...
             if (!Regex::compile("^[a-zA-Z0-9_:/\\. -]+$")->match($comment)) {
                 throw new InvalidArgumentException("Invalid comment for user agent given.");
             }
 
-            return \sprintf('WoltLabSuite/%s (%s)', \wcf\getMinorVersion(), $comment);
+            return \sprintf('WoltLabSuite/%s %s (%s)', \wcf\getMinorVersion(), $domain, $comment);
         }
 
-        return \sprintf('WoltLabSuite/%s', \wcf\getMinorVersion());
+        return \sprintf('WoltLabSuite/%s %s', $domain, \wcf\getMinorVersion());
```
TimWolla, line 46: You mixed up the `\sprintf()` parameter order.

TimWolla, line 46: Even with the correct parameter order this will not result in a valid RFC 7231#5.5.3 User-Agent, thus violating the method contract (neither will the version with the comment).

wcfsetup/install/files/lib/util/HTTPRequest.class.php
```diff
@@ -97,9 +98,10 @@ public function __construct($url, array $options = [], $postParameters = [], arr
 
         // set default headers
         $language = WCF::getLanguage();
+        $domain = ApplicationHandler::getInstance()->getApplicationByID(1)->domainName;
         $this->addHeader(
             'user-agent',
-            "HTTP.PHP (HTTPRequest.class.php; WoltLab Suite/" . WCF_VERSION . "; " . ($language ? $language->languageCode : 'en') . ")"
+            "HTTP.PHP (HTTPRequest.class.php; WoltLab Suite/" . WCF_VERSION . "; " . ($language ? $language->languageCode : 'en') . "; " . $domain . ")"
```
TimWolla, line 104: The User-Agent format in `HTTPRequest` is intentionally unchanged with the introduction of Guzzle for compat reasons, because I know that some services are parsing the User-Agent header for dumb reasons.

TimWolla (commented):
Apart from the technical issues I remarked, I am not interested in this change. In fact we had a short discussion whether an URL should be embedded into the User-Agent used for the URL unfurling feature (#4036).

- The new default User-Agent used with Guzzle is intentionally kept simple, leaving only a simple product name, version (truncated to the first 2 elements) and starting with WoltLab Suite 5.4 a comment to identify the subsystem.
    - This is sufficient to clearly identify the software that makes those requests, blocking it at the web server level.
- Blasting the URL out into the world even for requests to the update servers is questionable with regard to instances used for internal company communications.
- Guzzle is used for almost everything starting with WoltLab Suite 5.4, including internal requests for Elasticsearch.

>  It helps the user to identify sessions created by their own instances reading e.g. package information, api-calls or RSS-feeds

If such a request results in a persistent user session being created then that is a bug within the request target (e.g. the feed generator or the package server). The session list should only contain actual human sessions that will be re-used.

PR #4000, opened 2021-02-17
Reusable Events

TimWolla:

Issue #3609, opened 2020-10-15
Pass the CSRF token for Ajax requests in an X-XSRF-TOKEN request header


TimWolla:
… add this header automatically for same-site requests and pull the value out of the X-XSRF-TOKEN cookie.

Rationale: Passing the CSRF token as an URL parameter is questionable. The described behavior is compatible with common JavaScript frameworks.

Depends on #3593

TimWolla:
Blocked on #3543. It does not make sense to make functional changes while everything is in-flux.

TimWolla:
This probably will become easier with increased PSR-7 use. Moving to 5.6.

TimWolla:
The header will automatically be consumed by the Xsrf middleware that was added in #5059.

TimWolla:
This is effectively resolved with #5170.

PR #4081, opened 2021-03-14
deprecate some old session-related classes and implement new events directly

mutec:
- deprecates some (nearly) unused classes
- implements new events in `\wcf\system\session\SessionHandler` in order to replace the 'old' API in the future completely


TimWolla:
Thanks, but not at this point. We have a few plans regarding the event system (#4000) and I also plan to evaluate the session refactoring after 5.4 is released to further clean this up (cf. #3609, #3647). Your proposed changes might conflict with that, because I don't want to move around these events twice.

PR #4082, opened 2021-03-14
new events for user-online-list and session-management

mutec:
implement some new events to modify user online-appearance and provide more specific information for session-list to present a clear and easy to understand list for the user if some special apps or extensions (like feed readers, external APIs) are interacting with the site

### DIFF
--- a/wcfsetup/install/files/lib/system/session/Session.class.php
+++ b/wcfsetup/install/files/lib/system/session/Session.class.php
@@ -2,6 +2,7 @@
 
 namespace wcf\system\session;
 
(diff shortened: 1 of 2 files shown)


joshuaruesweg:
Hello, 

The security area is an extremely sensitive area. We are very careful here that this area cannot be adapted by plugins if possible. One of the reasons for this is that the user should find identical behaviour in this area on every WoltLab Suite page. Things like feed readers or external APIs should not generate a session. The internal `AbstractAuthedPage`, for example, deliberately does not generate its own session.

PR #4084, opened 2021-03-15
Improve performance of UserImporter

TimWolla:


wcfsetup/install/files/lib/system/importer/UserImporter.class.php
```diff
@@ -206,25 +220,23 @@ public function import($oldID, array $data, array $additionalData = [])
...
-        foreach ($groupIDs as $groupID) {
-            $statement->execute([
-                $user->userID,
-                $groupID,
-            ]);
+        $groupIDs = \array_merge(
+            $groupIDs,
+            $this->defaultGroups[$user->pendingActivation() ? 'notActivated' : 'activated']
+        );
+        if (!empty($groupIDs)) {
```
dtdesign, line 227: `$groupIDs` can never be empty at this point.

wcfsetup/install/files/lib/system/importer/UserImporter.class.php
```diff
@@ -238,19 +250,20 @@ public function import($oldID, array $data, array $additionalData = [])
...
-                                    (userID, eventID)
-                VALUES              (?, ?)";
-        $statement = WCF::getDB()->prepareStatement($sql);
-        WCF::getDB()->beginTransaction();
-        foreach ($this->eventIDs as $eventID) {
-            $statement->execute([
-                $user->userID,
-                $eventID,
-            ]);
+        if (!empty($this->eventIDs)) {
```
dtdesign, line 253: I'm pretty sure that `$this->eventIDs` can never be empty, because even the Core itself provides some.

dtdesign (changes requested):
LGTM

wcfsetup/install/files/lib/system/importer/UserImporter.class.php
TimWolla: They can, because they are filtered for `preset = 1` and the admin might have removed all presets.

PR #4089, opened 2021-03-16
Add wcf1_user_session.creationTime

TimWolla:


joshuaruesweg (approved):
LGTM

jens1o:
Will this be shown on the session management page?

TimWolla:
I added this column because with 4 Bytes per session it's cheap enough to store and “Better to have, and not need, than to need, and not have”. It is not currently planned to expose this in the UI.

PR #4093, opened 2021-03-18
Add RedirectGuard

TimWolla:
- Add RedirectGuard for Guzzle
- Use RedirectGuard during URL unfurling
- Use RedirectGuard in ImageProxy


joshuaruesweg (approved):
LGTM and works fine.

PR #4094, opened 2021-03-19
Update spiderList.xml

tamsch:
I just updated the spiderList.xml.

### DIFF
--- a/spiderList/spiderList.xml
+++ b/spiderList/spiderList.xml
@@ -25,6 +25,10 @@
 		<name>Acoon</name>
 		<url>http://www.acoon.de/robot.asp</url>
 	</spider>
+	<spider ident="Adsbot/3.1">
+		<name>Adsbot</name>
+		<url>https://seostar.co/robot</url>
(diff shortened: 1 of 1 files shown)


spiderList/spiderList.xml
```diff
@@ -703,6 +715,10 @@
 	<spider ident="legs">
 		<name>legs</name>
 	</spider>
+	<spider ident="Let's Encrypt validation server">
```
joshuaruesweg, line 718: I don't get, why this is should be included in this list? This is not a bot, which crawls your site. 
Actually, the bot should not come into contact with your page either.

joshuaruesweg (changes requested):

joshuaruesweg:
Thanks! 

I added the relevant bots: 

https://github.com/WoltLab/WCF/commit/2f4ab6f10ee34601a2931307cc6cc1601943fcd7
https://github.com/WoltLab/WCF/commit/6ca2411ef1f556415d44e5150092e0d64b535954

PR #4099, opened 2021-03-24
Add TemplateScriptingCompiler#beforeCompile event

Sir-Will:
It would be great to have this event to allow the patching of templates. This would add tons of more customization possibilities. 
For example would it allow adding additional buttons in the footer next to the style changer:
```php
if ($parameters['identifier'] === 'pageFooter') {
	$search = '/\s\{if \$__showStyleChanger\}/';
	$replace = '<span class="widthToggle jsOnly">test</span>$0';
	$eventObj->sourceContent = preg_replace($search, $replace, $eventObj->sourceContent);
}
```

The `beforeCompile` event in TemplateEngine doesn't fire for include templates and doesn't allow the modification of the source, which makes it not useable for this case.

### DIFF
--- a/wcfsetup/install/files/lib/system/template/TemplateScriptingCompiler.class.php
+++ b/wcfsetup/install/files/lib/system/template/TemplateScriptingCompiler.class.php
@@ -324,7 +324,10 @@ class TemplateScriptingCompiler
     public function compileString($identifier, $sourceContent, array $metaData = [], $isolated = false)
     {
         $this->sourceContent = $sourceContent;
-        $data = [ 'metaData' => $metaData ]
+        $data = [
+            'identifier' => $identifier,
+            'metaData' => $metaData
+        ];
         EventHandler::getInstance()->fireAction($this, 'beforeCompile', $data);
         $sourceContent = $this->sourceContent;
 
--- a/wcfsetup/install/files/lib/system/template/TemplateScriptingCompiler.class.php
+++ b/wcfsetup/install/files/lib/system/template/TemplateScriptingCompiler.class.php
@@ -7,6 +7,7 @@ use wcf\system\template\plugin\ICompilerTemplatePlugin;
 use wcf\system\template\plugin\IPrefilterTemplatePlugin;
 use wcf\util\StringStack;
 use wcf\util\StringUtil;
+use wcf\system\event\EventHandler;
 
 /**
  * Compiles template sources into valid PHP code.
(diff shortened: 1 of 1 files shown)


dtdesign:
> It would be great to have this event to allow the patching of templates.

No, sorry, we will never do that. We had template patches back in the WCF 1.x era and it was a huge disaster. Pandora's box must remain closed.

Sir-Will:
What issues did it cause?
We have been using it in other softwares for years and never encountered any issues.

dtdesign:
It works for a single consumer that modifies templates, but breaks as soon as multiple consumers try to modify the same template. This even worses when the execution order is unknown or is subject to fluctuations when there is no defined order and the only difference is the installation order.

Template patches are an even more robust solution because they will ensure that a template only matches if the context is correct ("fuzziness"). This will make multiple patches fail if they modify the same parts, but at least this was transparent and had a well-formed behavior. Your suggestion is even weaker by not introducing any safe-guards that prevents incorrect modifications of template code, just because it looks similar to what you were looking for.

We have years or experience with is and history teached us that this is simply not robust enough. You might have been lucky and did not face these challenges yourself so far, but this makes no difference here. We have been through this and we're not looking back on this one, ever. It was a painful time.

Slind14:
What were the issues back then in detail?

XenForo as one example has had this feature forever and it never causes any issue, even with hundreds of extensions on a single instance. The amount of extensions available there is a lot bigger as well.

Sir-Will:
The plan was to add execution order, fail-safe and so on to a plugin, which other plugins are based on.
The proposed event is just the requirement for the planned plugin.

dtdesign:
I can see the temptation of modifying templates at will, including but not limited to the ability to alter the DOM significantly. However, this comes with a lot of drawbacks in terms of interoperability, this is a recurring issue caused by custom templates for styles that alter the DOM in an unexpected way. Also, multiple consumers cannot reliably modify the same part of a template and unless you incorporate some degree sorcery, no "silver bullet" type of code can resolve this.

Please don't get me wrong here, I'm writing this to give you some insight into the experiences that we had in the past, both as the maintainer and as individual developers. There are very few things that we wholeheartedly oppose and this is one of them.

Sir-Will:
I see this as an alternative and optional option to modify templates which doesn't take away anything and doesn't require to be maintained by WoltLab. It's up to the plugin authors if they want to use this option, deal with potential issues and support styles which heavily modify the templates.

This shouldn't be used to heavily modify templates but to do 1-5 lines of changes, which makes it unlikely that multiple installed plugins want to do the same or similar change. If they target the same lines and one fails, then the failed one is an invalid patch anyways, and it's up to the plugin author to make it compatible with the other plugin.

Currently, I have 5+ templates in a plugin where each one only adds 1-3 lines at specific positions. Now I can't modify these templates with any other plugin, although there wouldn't be any conflicts.
Additionally, I have to keep track of any changes to the original templates because of few added lines which makes this not upgrade safe.
This is really daunting and doesn't make me want to modify additional templates.

PR #4114, opened 2021-04-12
Representation of mentions in `text/plain` output

### LINKED ISSUE #4020: text/plain email notifications do not render the `@` in front of mentions
A post like:

```
@test test
```

will render as:

```
test test
```

It works fine for the `text/html` version.

dtdesign:
Fixes #4020


dtdesign:
@TimWolla This PR looks bogus to me, while it does return different values based on the content type, it fails to resolve the underlying issue. Whatever is returned in `getParsedTag()` will be dumped into a `wcfNode`-element that is being discarded before it is evaluated.

PR #4116, opened 2021-04-12
Improve the performance of the HTMLOutputProcessor

dtdesign:
Lessons learned: `DOMElement::getElementsByTagName()` is a slow piece of garbage, `DOMXPath::query()` is extremely faster.


TimWolla (approved):
I trust your measurements and I've eyeballed the diff without seeing any obvious bugs.

PR #4120, opened 2021-04-14
Support script elements in template parameter of `Ui/Confirmation`

Gravatronics:


TimWolla (approved):
If you say that we need to support these then this LGTM.

Gravatronics:
> If you say that we need to support these
Its not necessary anymore in my specific case.

PR #4121, opened 2021-04-14
Add clipboard support for moderation

### LINKED ISSUE #3774: Clipboard in moderation queue list
n/t

Gravatronics:
Close  #3774

### DIFF
--- a/com.woltlab.wcf/clipboardAction.xml
+++ b/com.woltlab.wcf/clipboardAction.xml
@@ -180,5 +180,40 @@
 				<page>wcf\acp\page\ArticleListPage</page>
 			</pages>
 		</action>
+		<action name="assignUserByClipboard">
+			<actionclassname>wcf\system\clipboard\action\ModerationQueueClipboardAction</actionclassname>
+			<showorder>1</showorder>
+			<pages>
+				<page>wcf\page\ModerationListPage</page>
+			</pages>
+		</action>
+		<action name="enableContent">
+			<actionclassname>wcf\system\clipboard\action\ModerationQueueActivationClipboardAction</actionclassname>
+			<showorder>2</showorder>
+			<pages>
+				<page>wcf\page\ModerationListPage</page>
+			</pages>
+		</action>
+		<action name="removeActivationContent">
+			<actionclassname>wcf\system\clipboard\action\ModerationQueueActivationClipboardAction</actionclassname>
+			<showorder>3</showorder>
+			<pages>
+				<page>wcf\page\ModerationListPage</page>
+			</pages>
+		</action>
+		<action name="removeReport">
+			<actionclassname>wcf\system\clipboard\action\ModerationQueueReportClipboardAction</actionclassname>
+			<showorder>4</showorder>
+			<pages>
+				<page>wcf\page\ModerationListPage</page>
+			</pages>
+		</action>
+		<action name="removeReportContent">
+			<actionclassname>wcf\system\clipboard\action\ModerationQueueReportClipboardAction</actionclassname>
+			<showorder>5</showorder>
+			<pages>
+				<page>wcf\page\ModerationListPage</page>
+			</pages>
+		</action>
 	</import>
 </data>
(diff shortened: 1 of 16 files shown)


com.woltlab.wcf/clipboardAction.xml
```diff
@@ -180,5 +180,40 @@
 				<page>wcf\acp\page\ArticleListPage</page>
 			</pages>
 		</action>
+		<action name="assignUserByClipboard">
```
TimWolla, line 183: Why this elaborate name?

ts/WoltLabSuite/Core/Ui/Moderation/Clipboard/AssignUser.ts
```diff
@@ -0,0 +1,187 @@
...
+            assignedUserId.click();
+          });
+
+          content.querySelector("button[data-type=submit]")!.addEventListener("click", submitCallback);
+        },
+        onShow(content: HTMLElement): void {
+          // Reset dialog to initial state.
+          const assignedUsername = content.querySelector("input[name=assignedUsername]") as HTMLInputElement;
+          (content.querySelector("input[name=assignedUserID]:checked") as HTMLInputElement).checked = false;
+          (content.querySelector("input[name=assignedUserID]") as HTMLInputElement).checked = true;
```
TimWolla, line 108: Aren't these equivalent to simply setting `.checked = true` for all inputs with name `assignedUserID`?

wcfsetup/install/files/lib/data/moderation/queue/ModerationQueueActivationAction.class.php
```diff
@@ -69,5 +88,55 @@ public function removeContent()
...
+        foreach ($this->getObjects() as $moderationQueueEditor) {
+            if (
+                !$moderationQueueEditor->canEdit()
+                || !ModerationQueueActivationManager::getInstance()->canRemoveContent($moderationQueueEditor->getDecoratedObject())
+            ) {
+                throw new PermissionDeniedException();
+            }
+        }
+
+        $this->parameters['message'] = isset($this->parameters['message']) ? StringUtil::trim($this->parameters['message']) : '';
```
TimWolla, line 119:
```suggestion
        $this->parameters['message'] = StringUtil::trim($this->parameters['message'] ?? '');
```

wcfsetup/install/files/lib/data/moderation/queue/ModerationQueueReportAction.class.php
```diff
@@ -205,4 +229,52 @@ public function changeJustifiedStatus(): void
...
+        foreach ($this->getObjects() as $moderationQueueEditor) {
+            if (
+                !$moderationQueueEditor->canEdit()
+                || !ModerationQueueReportManager::getInstance()->canRemoveContent($moderationQueueEditor->getDecoratedObject())
+            ) {
+                throw new PermissionDeniedException();
+            }
+        }
+
+        $this->parameters['message'] = isset($this->parameters['message']) ? StringUtil::trim($this->parameters['message']) : '';
```
TimWolla, line 257: ditto

com.woltlab.wcf/clipboardAction.xml
Gravatronics: `assignUser` is already used in `ModerationQueueAction` so that we need to use a different action name here which is mapped to the DBO action name directly.

ts/WoltLabSuite/Core/Ui/Moderation/Clipboard/AssignUser.ts
Gravatronics: `assignedUserID` are radiobuttons.

TimWolla: Ah, yeah. And it's `.querySelector`, so line 108 will match the first element only. I find this logic pretty hard to understand. Are you able to work with `defaultChecked` here? I never tested this, but something like:

```
content.querySelectorAll("input[name=assignedUserID]").forEach((el) => el.checked = el.defaultChecked)
```

could work based off the documentation.

Gravatronics: Yes, it works.

PR #4122, opened 2021-04-15
Include compiled template code in exception logs

### LINKED ISSUE #3943: Include compiled template code in exception logs
If a stacktrace of an exception contains a compiled template, a part of the compiled template (may-be five lines before and after the line with the error) should be included in the exception log.
If it is included base64 coded, the log in the ACP should show the decoded form!

Gravatronics:
Close #3943

### DIFF
--- a/wcfsetup/install/files/lib/core.functions.php
+++ b/wcfsetup/install/files/lib/core.functions.php
@@ -125,6 +125,56 @@ namespace wcf\functions\exception {
 	use wcf\system\exception\SystemException;
 	use wcf\util\FileUtil;
 	use wcf\util\StringUtil;
+
+	/**
+	 * If the stacktrace contains a compiled template, the context of the relevant template line
+	 * is returned, otherwise an empty array is returned.
+	 */
+	function getTemplateContextLines(\Throwable $e): array {
+		try {
+			$contextLineCount = 5;
+			foreach ($e->getTrace() as $traceEntry) {
+				if (isset($traceEntry['file']) && \preg_match('~/templates/compiled/.+\.php$~',
+					$traceEntry['file'])) {
+					$startLine = $traceEntry['line'] - $contextLineCount;
+					if ($startLine < 0) {
+						$startLine = 0;
+					}
+					
+					$file = \fopen($traceEntry['file'], 'r');
+					if (!$file) {
+						break;
+					}
+					
+					for ($line = 0; $line < $startLine; $line++) {
+						if (\substr(\fgets($file, 1024), -1) !== "\n") {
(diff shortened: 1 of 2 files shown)


wcfsetup/install/files/lib/core.functions.php
```diff
@@ -140,6 +140,40 @@ function logThrowable(\Throwable $e, &$logFile = null): string {
...
 		
+		$getExtraInformation = function (\Throwable $e) {
+			$extraInformation = [];
+			$contextLineCount = 5;
+			
+			if ($e instanceof IExtraInformationException) {
+				$extraInformation = $e->getExtraInformation();
+			}
+			
+			foreach ($e->getTrace() as $traceEntry) {
```
TimWolla, line 151: Add a big fat try-catch block around this. We don't want Exceptions being thrown within the Exception handler for some dumb reason.

wcfsetup/install/files/lib/core.functions.php
```diff
@@ -140,6 +140,40 @@ function logThrowable(\Throwable $e, &$logFile = null): string {
...
+				if (isset($traceEntry['file']) && preg_match('~/templates/compiled/.+\.php$~', $traceEntry['file'])) {
+					$lines = file($traceEntry['file']);
```
TimWolla, line 153: This is unsafe for larger templates. Use something like the following to handle the file in a streaming fashion:

```
$f = \fopen($traceEntry['file'], 'r');
for ($line = 0; $line < $startLine; $line++) {
  if (\substr(\fgets($f, 1024), -1) !== "\n") {
    break 2; // We don't want to handle a file where lines exceed 1024 Bytes.
  }
}
```

TimWolla (changes requested):

wcfsetup/install/files/lib/core.functions.php
```diff
@@ -125,6 +125,56 @@ function getMinorVersion(): string {
...
+				if (isset($traceEntry['file']) && \preg_match('~/templates/compiled/.+\.php$~',
+					$traceEntry['file'])) {
+					$startLine = $traceEntry['line'] - $contextLineCount;
+					if ($startLine < 0) {
+						$startLine = 0;
+					}
+					
+					$file = \fopen($traceEntry['file'], 'r');
+					if (!$file) {
+						break;
```
TimWolla, line 146: All these `break`s could simply be a `return [];` after this became a dedicated function.

TimWolla (approved):
Overall this LGTM now and it should be sufficiently safe to not break anything else.

TimWolla (approved):

PR #4123, opened 2021-04-15
Add dialog with anonymous/personalized links for RSS feeds

### LINKED ISSUE #3661: RSS feed selection with and without access token
Currently, the RSS feed is always linked with an access token. To make sharing such links with other people easier, links with and without the access token should be offered (in a dialog?).

Gravatronics:
Close #3661

### DIFF
--- a/ts/WoltLabSuite/Core/Ui/Feed/Dialog.ts
+++ b/ts/WoltLabSuite/Core/Ui/Feed/Dialog.ts
@@ -10,7 +10,6 @@
 import UiDialog from "../Dialog";
 import * as StringUtil from "../../StringUtil";
 import * as Language from "../../Language";
-import * as DomTraverse from "../../Dom/Traverse";
 import * as Clipboard from "../../Clipboard";
 import * as UiNotification from "../Notification";
 
--- a/ts/WoltLabSuite/Core/Ui/Feed/Dialog.ts
+++ b/ts/WoltLabSuite/Core/Ui/Feed/Dialog.ts
@@ -10,7 +10,6 @@
 import UiDialog from "../Dialog";
 import * as StringUtil from "../../StringUtil";
 import * as Language from "../../Language";
-import User from "../../User";
 import * as DomTraverse from "../../Dom/Traverse";
 import * as Clipboard from "../../Clipboard";
 import * as UiNotification from "../Notification";
@@ -22,7 +21,7 @@ async function copy(event: Event): Promise<void> {
   event.preventDefault();
 
   const target = event.currentTarget as HTMLElement;
-  const input = DomTraverse.prevBySel(target, 'input[type="text"]') as HTMLInputElement;
+  const input = target.parentNode!.querySelector('input[type="text"]') as HTMLInputElement;
 
   await Clipboard.copyTextToClipboard(input.value);
 
@@ -38,9 +37,7 @@ function openDialog(event: Event): void {
   const alternative = event.currentTarget as HTMLAnchorElement;
   const linkWithAccessToken = alternative.href;
 
-  const linkWithoutAccessToken = linkWithAccessToken
-    .replace(`at=${User.userId}-${User.accessToken}&`, "&")
-    .replace(new RegExp(`(\\?|&)at=${User.userId}-${User.accessToken}`), "");
+  const linkWithoutAccessToken = linkWithAccessToken.replace(/at=[^&]*&/, "&").replace(/(\\?|&)at=[^&]*/, "");
 
   UiDialog.openStatic(
     "feedLinkDialog",
@@ -81,9 +78,7 @@ function openDialog(event: Event): void {
 }
 
 export function setup(): void {
-  document.querySelectorAll("a[rel=alternate]").forEach((link: HTMLAnchorElement) => {
-    if (new URL(link.href).searchParams.get("at") === `${User.userId}-${User.accessToken}`) {
-      link.addEventListener("click", (ev) => openDialog(ev));
-    }
+  document.querySelectorAll("a.rssFeed").forEach((link: HTMLAnchorElement) => {
+    link.addEventListener("click", (ev) => openDialog(ev));
   });
 }
--- a/ts/WoltLabSuite/Core/Ui/Feed/Dialog.ts
+++ b/ts/WoltLabSuite/Core/Ui/Feed/Dialog.ts
@@ -36,7 +36,7 @@ function openDialog(event: Event): void {
   const alternative = event.currentTarget as HTMLAnchorElement;
   const linkWithAccessToken = alternative.href;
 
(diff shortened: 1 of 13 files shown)


ts/WoltLabSuite/Core/Ui/Feed/Dialog.ts
```diff
@@ -0,0 +1,89 @@
...
+    },
+  );
+}
+
+export function setup(): void {
+  document.querySelectorAll("a[rel=alternate]").forEach((link: HTMLAnchorElement) => {
+    if (new URL(link.href).searchParams.get("at") === `${User.userId}-${User.accessToken}`) {
+      link.addEventListener("click", (ev) => openDialog(ev));
+    }
+  });
```
TimWolla, line 88: This feels too much like magic to me. It's also not necessarily correct. There might be other alternates that are not RSS and also include the `at` (e.g. the iCal export).

ts/WoltLabSuite/Core/Ui/Feed/Dialog.ts
```diff
@@ -0,0 +1,89 @@
...
+ */
+function openDialog(event: Event): void {
+  event.preventDefault();
+
+  const alternative = event.currentTarget as HTMLAnchorElement;
+  const linkWithAccessToken = alternative.href;
+
+  const linkWithoutAccessToken = linkWithAccessToken
+    .replace(`at=${User.userId}-${User.accessToken}&`, "&")
+    .replace(new RegExp(`(\\?|&)at=${User.userId}-${User.accessToken}`), "");
```
TimWolla, line 43: Can't you use the `URL` class for parsing here as well?

wcfsetup/install/lang/de.xml
```diff
@@ -3962,6 +3962,11 @@ Dateianhänge:
 		<item name="wcf.global.button.showSidebarRight"><![CDATA[Rechte Sidebar]]></item>
 		<item name="wcf.global.button.showMenu"><![CDATA[Menü anzeigen]]></item>
 		<item name="wcf.global.button.hideMenu"><![CDATA[Menü verbergen]]></item>
+		<item name="wcf.global.rss.copy"><![CDATA[Link kopieren]]></item>
+		<item name="wcf.global.rss.copy.success"><![CDATA[Der Link wurde erfolgreich kopiert.]]></item>
+		<item name="wcf.global.rss.accessToken.info"><![CDATA[Der Link zum anonymen RSS-Feed enthält nur Inhalte, auf die Gäste Zugriff haben. Der Link zum personalisierten RSS-Feed enthält alle Inhalte, auf die {if LANGUAGE_USE_INFORMAL_VARIANT}Du{else}Sie{/if} Zugriff {if LANGUAGE_USE_INFORMAL_VARIANT}hast{else}haben{/if}.]]></item>
```
TimWolla, line 3967: s/Du/du/

ts/WoltLabSuite/Core/Ui/Feed/Dialog.ts
```diff
@@ -0,0 +1,89 @@
...
+          .forEach((el) => el.addEventListener("click", (ev) => copy(ev)));
+      },
+      title: alternative.title || Language.get("wcf.global.button.rss"),
+    },
+  );
+}
+
+export function setup(): void {
+  document.querySelectorAll("a[rel=alternate]").forEach((link: HTMLAnchorElement) => {
+    if (new URL(link.href).searchParams.get("at") === `${User.userId}-${User.accessToken}`) {
```
TimWolla, line 85: I'm not sure how I feel about specifically matching the user's accessToken vs. matching anything with `at` and is not an external URL.

ts/WoltLabSuite/Core/Ui/Feed/Dialog.ts
Gravatronics: No, because this results in encoding issues with the controller (e.g. `index.php?board-feed/&at=`) because it contains a slash in the parameter name (`board-feed/`) which in encoded when using `.searchParams.delete('at')` requiring additional code to then revert that encoding again.

ts/WoltLabSuite/Core/Ui/Feed/Dialog.ts
Gravatronics: What be the advantage of matching anything else than the expected access token for the active user?

ts/WoltLabSuite/Core/Ui/Feed/Dialog.ts
Gravatronics: The iCal export link does not have a `rel` attribute and this current approach avoids having to change existing template/HTML code resulting in its consistent application in all installed packages.

ts/WoltLabSuite/Core/Ui/Feed/Dialog.ts
TimWolla: I did a cursory check at the URI RFC. URIs don't define a format of the query string (i.e. it does not need to be `application/x-www-form-urlencoded`) and they allow the query string to contain a raw slash. In the URI encoding two URIs that differ in `%2F` and `/` are *not* equivalent, but I'm not sure whether that also applies to `application/x-www-form-urlencoded` encoding. My guess would be that `application/x-www-form-urlencoded` considers these to be equivalent, which would make the result ugly, but not wrong.

I'd like to read up on this in more detail, before considering this to be resolved.

TimWolla: Okay, I looked into this:
- The query generally is to be considered opaque from a URI viewpoint.
- The URL class and PHP use x-www-form-urlencoded.
- In x-www-form-urlencoded all names and values are urldecoded before being interpreted.
- WoltLab Suite specifically handles an encoded controller name: https://github.com/WoltLab/WCF/blob/56e3b7959ec8ccd10c87066cdf075103c5fffc05/wcfsetup/install/files/lib/system/request/RouteHandler.class.php#L357

So: The implicit encoding when using `.searchParams.delete('at')` is ugly at most.

ts/WoltLabSuite/Core/Ui/Feed/Dialog.ts
```diff
@@ -0,0 +1,83 @@
...
+/**
+ * Opens the dialog with an anonymous and personalized link after clicking on the RSS feed link.
+ */
+function openDialog(event: Event): void {
+  event.preventDefault();
+
+  const alternative = event.currentTarget as HTMLAnchorElement;
+  const linkWithAccessToken = alternative.href;
+
+  const linkWithoutAccessToken = linkWithAccessToken.replace(/at=[^&]*&/, "&").replace(/(\\?|&)at=[^&]*/, "");
```
TimWolla, line 39: I believe this will break links that look like the following:

```
https://example.com/forum/rss-feed/?at=123&foo=bar
```
Because the `?` will be stripped. The second replacement probably should replace by `$1` and a third replacement should be added that removes `(\?|&)$`.

PR #4127, opened 2021-04-16
Fix requesting multiple times the UserStorage tables if no cache exists

joshuaruesweg:


TimWolla (approved):

TimWolla (changes requested):
Ah, one thing: Please remove the check in the while-loop. It is no longer required.

TimWolla (approved):

dtdesign (approved):

PR #4131, opened 2021-04-19
Add option to disable FLoC

SoftCreatR:
Google is rolling out Federated Learning of Cohorts (FLoC) for the Chrome browser.

As the Electronic Frontier Foundation explains in their post “[Google’s FLoC is a terrible idea](https://www.eff.org/deeplinks/2021/03/googles-floc-terrible-idea)“, placing people in groups based on their browsing habits is likely to facilitate employment, housing and other types of discrimination, as well as predatory targeting of unsophisticated consumers.

This is in addition to the privacy concerns of tracking people and sharing their data, seemingly without informed consent – and making it more difficult for legislators and regulators to protect people.

TL;DR: FLoC places people in groups based on their browsing habits to target advertising.

Website administrators can disable this new tracking technique by sending a `Permissions-Policy: interest-cohort=()` header. This commit adds a new option to send this header, which is enabled by default. Administrators can disable it, especially in cases, where the underlying webserver is already configured to send the `Permissions-Policy` header.

### DIFF
--- a/wcfsetup/install/files/lib/util/HeaderUtil.class.php
+++ b/wcfsetup/install/files/lib/util/HeaderUtil.class.php
@@ -94,6 +94,11 @@ final class HeaderUtil
             @\header('X-Frame-Options: SAMEORIGIN');
         }
 
+        // send Permissions-Policy
+        if (HTTP_DISABLE_FLOC) {
+            @\header('Permissions-Policy: interest-cohort=()');
+        }
+
         \ob_start([self::class, 'parseOutput']);
     }
 
--- a/wcfsetup/install/files/lib/util/HeaderUtil.class.php
+++ b/wcfsetup/install/files/lib/util/HeaderUtil.class.php
@@ -96,7 +96,7 @@ final class HeaderUtil
 
         // send Permissions-Policy
         if (HTTP_DISABLE_FLOC) {
-            @\header('Permissions-Policy: interest-cohort=()');
+            @\header('permissions-policy: interest-cohort=()');
         }
 
         \ob_start([self::class, 'parseOutput']);
--- a/wcfsetup/install/lang/de.xml
+++ b/wcfsetup/install/lang/de.xml
@@ -1329,6 +1329,8 @@ ACHTUNG: Die oben genannten Meldungen sind stark gekürzt. Sie können Details z
 		<item name="wcf.acp.option.error.tooShort"><![CDATA[Der eingegebene Text ist zu kurz.]]></item>
 		<item name="wcf.acp.option.http_send_x_frame_options"><![CDATA[Einbindung in einem Frame verhindern]]></item>
 		<item name="wcf.acp.option.http_send_x_frame_options.description"><![CDATA[Sendet den <a href="https://de.wikipedia.org/wiki/Clickjacking" class="externalURL">„X-Frame-Options“</a> Header, um die Einbettung dieser Seite in einem Frame zu verhindern (sendet „SAMEORIGIN“).]]></item>
+		<item name="wcf.acp.option.http_disable_floc"><![CDATA[Besucher-Tracking mittels FLoC verhindern]]></item>
+		<item name="wcf.acp.option.http_disable_floc.description"><![CDATA[Verhindert das Tracking mittels des von Google eingeführten <a href="https://de.wikipedia.org/wiki/Federated_Learning_of_Cohorts">„FLoC“</a> durch senden des „Permissions-Policy“ Headers (sendet „interest-cohort=()“).]]></item>
 		<item name="wcf.acp.option.image_adapter_type"><![CDATA[Grafik-Bibliothek]]></item>
 		<item name="wcf.acp.option.image_adapter_type.gd"><![CDATA[GD Graphics Library (Standard)]]></item>
 		<item name="wcf.acp.option.image_adapter_type.imagick"><![CDATA[ImageMagick]]></item>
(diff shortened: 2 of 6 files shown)


wcfsetup/install/files/lib/util/HeaderUtil.class.php
```diff
@@ -94,6 +94,11 @@ public static function sendHeaders()
             @\header('X-Frame-Options: SAMEORIGIN');
         }
 
+        // send Permissions-Policy
+        if (HTTP_DISABLE_FLOC) {
+            @\header('Permissions-Policy: interest-cohort=()');
```
TimWolla, line 99:
```suggestion
            @\header('permissions-policy: interest-cohort=()');
```
We use lowercase names for all newly added headers.

wcfsetup/install/lang/de.xml
```diff
@@ -1329,6 +1329,8 @@ ACHTUNG: Die oben genannten Meldungen sind stark gekürzt. Sie können Details z
 		<item name="wcf.acp.option.error.tooShort"><![CDATA[Der eingegebene Text ist zu kurz.]]></item>
 		<item name="wcf.acp.option.http_send_x_frame_options"><![CDATA[Einbindung in einem Frame verhindern]]></item>
 		<item name="wcf.acp.option.http_send_x_frame_options.description"><![CDATA[Sendet den <a href="https://de.wikipedia.org/wiki/Clickjacking" class="externalURL">„X-Frame-Options“</a> Header, um die Einbettung dieser Seite in einem Frame zu verhindern (sendet „SAMEORIGIN“).]]></item>
+		<item name="wcf.acp.option.http_disable_floc"><![CDATA[Besucher-Tracking mittels FLoC verhindern]]></item>
+		<item name="wcf.acp.option.http_disable_floc.description"><![CDATA[Verhindert das Tracking mittels des von Google eingeführten <a href="https://de.wikipedia.org/wiki/Federated_Learning_of_Cohorts">„FLoC“</a> durch senden des „Permissions-Policy“ Headers (sendet „interest-cohort=()“).]]></item>
```
TimWolla, line 1333:
```suggestion
		<item name="wcf.acp.option.http_disable_floc.description"><![CDATA[Verhindert das Tracking mittels des von Google eingeführten <a href="https://de.wikipedia.org/wiki/Federated_Learning_of_Cohorts">„FLoC“</a> durch Senden des „Permissions-Policy“ Headers (sendet „interest-cohort=()“).]]></item>
```

TimWolla (changes requested):
We are aware of this announcement of course and we did our own research. We believe that this change is not healthy, due to the following reasons:

1. For the vast majority of communities it does exactly nothing, because the FLoC tracking will only be performed if either
    - the website calls the FLoC API (`document.interestCohort()`), or
    - the website embeds something that resembles advertisement (only during the FLoC trial).

Source: https://seirdy.one/2021/04/16/permissions-policy-floc-misinfo.html

In the first case the administrator explicitly opted into this type of tracking. In the latter case contractual obligations might prevent the administrator from disabling this API.

:point_right: The administrator should be required to perform an informed consent.

2. Even if this header did anything useful in the general case, only applying it to HTML responses generated by WoltLab Suite would be wholly insufficient:
    - It does not include web server generated error pages.
    - It does not include any non-HTML output generated by WoltLab Suite.
    - It does not include any other applications running on the same webspace (e.g. a static landing page).

:point_right: This is better solved at the web server's level (e.g. `.htaccess`).

-------

Please also find the inline notes regarding some technical nits I found during review.

SoftCreatR:
While i agree with you from the technical view in most points, even WordPress is thinking about implementing this Header as a security update (so even older versions will get it). I think, implementing it as an option is better than nothing and advanced administrators can simply disable this option and simply implement it by their own (e.g. via .htaccess).

> Even if this header did anything useful in the general case, only applying it to HTML responses generated by WoltLab Suite would be wholly insufficient:
It does not include web server generated error pages.
It does not include any non-HTML output generated by WoltLab Suite.
It does not include any other applications running on the same webspace (e.g. a static landing page).

The same applies for the X-Frame-Options, that are shipped with your software for ages.

TimWolla:
> I think, implementing it as an option is better than nothing

This is not true. The existence of this option might give administrators a false sense of security, making them believe they are protected (or protecting their users) when in fact they might not be. More on this below.

> ... load third-party scripts that might call it.

Sure, but the loading of third party scripts already comes with its own issues with regard to privacy compliant use. Verifying whether FLoC is being used by that script and whether this is lawful or not is just one more bullet point in a list that already needs to exist. If this third-party script is advertisement code then the contract with your advertisement company might or might not have some rules regarding the use of FLoC.

> The same applies for the X-Frame-Options, that are shipped with your software for ages.

This is a straw man argument. X-F-O brings an actual measurable security benefit in all current browsers against click jacking attacks. If the X-F-O is missing on server-generated error pages then no harm can be done (no session being used there), if it is missing on images, no harm done (no clickable areas on images). Third Party contents not being protected is the only bullet point that also applies here. But it is only relevant if the third party contents even make use of a session. Additionally if X-F-O fails it fails rather loudly. It's immediately evident that a page within a frame fails to load.

Disabling the `interest-cohort=()` permission on the other hand protects the user against … their very own web browser. If they can't trust the web browser to not act against them then it's game over anyway. Further it is very hard to verify whether denying the permission actually does anything or breaks anything in a subtle way. Also for this header to be effective it needs to be consistently applied to every single sub-page of your website, because otherwise the browser is just going to be use that one for cohort learning, being able to include e.g. the domain name which by its own should be sufficient data for clustering.

We believe it is more effective to educate administrators how they are able to reliably implement this header on the web server's level and to educate users to use a web browser that does not betray their trust.

Issue #4017, opened 2021-02-23
Use font stack instead of 3rd party fonts by default


SoftCreatR:
A while ago, we switched from default Open Sans to a default font stack:

```css
font-family: -apple-system, BlinkMacSystemFont, Segoe UI, Helvetica, Arial, sans-serif, Apple Color Emoji, Segoe UI Emoji;
```

From my experience, most people can't tell a huge difference in the UI* but avoiding the use of Open Sans or any other 3rd party font greatly improves the performance since they don't require loading any additional font files thus there's no FOIT and no need for additional preload requests, etc.

\* = Of course there are visual differences and of course there are enough people who will notice these differences, but the performance advantage outweighs the disadvantages caused by the visual differences, which are pretty minimal IMHO: 

https://codepen.io/SoftCreatR/full/BaQmYmr

So my idea is to use the proposed font stack, instead of Open Sans by default for the default- and for any new style.

dtdesign:
Duplicate of #3970

SoftCreatR:
@dtdesign U sure, that this is a duplicate? I mean, I can't find any information in your issue, that using a system UI font stack is planned to be the default (instead of Open Sans) in the future, but If that's the case: 👍

dtdesign:
Yes, it absolutely is. I've looked into this for quite some time before creating the issue, in particular making sure that we do use a matching font stack for code-style elements.

SoftCreatR:
Yeah, we use one too:

```css
font-family: SFMono-Regular, Consolas, Liberation Mono, Menlo, 'Courier New', monospace;
```

PR #4136, opened 2021-04-21
Use the OS' native font by default

### LINKED ISSUE #3970: Support for the system UI font
The font selection for styles only permits a fixed set regardless of the user's platform preferences. Offering the selection of the system UI has the benefit of adhering to the user selection without relying on rather expensive web fonts.

dtdesign:
Resolves #3970

/cc @SoftCreatR (#4017)


SoftCreatR:
That's a pretty unusual stack, but better than before 😄 

Maybe use GitHub's?

Default: `-apple-system,BlinkMacSystemFont,Segoe UI,Helvetica,Arial,sans-serif,Apple Color Emoji,Segoe UI Emoji`
Monospace: `SFMono-Regular,Consolas,Liberation Mono,Menlo,monospace`

dtdesign:
The GitHub stack does not provide native fonts for Android and Linux, they would fall back to "Helvetica" or "Arial". I don't understand GitHub's motivation of excluding those, especially Android has _quite some_ market share?

SoftCreatR:
Yeah, GH is still not mobile optimized :/

dtdesign:
For the record, the implementation is designed to allow future modifications of the font stack to address changes in case new fonts emerge. I do not think that GitHub's font stack is better, the font stack proposed in this PR covers their font plus a few more for Android/Linux, but please feel free to suggest changes if you think I missed something here.

SoftCreatR:
It's all good, I like the proposed changes.

wcfsetup/install/files/lib/system/style/StyleCompiler.class.php
```diff
@@ -582,14 +582,26 @@ protected function compileStylesheet(string $scss, array $variables): string
...
             $variables['wcfFontFamilyGoogle'] = '"' . $variables['wcfFontFamilyGoogle'] . '"';
 
             $variables['wcfFontFamily'] = $variables['wcfFontFamilyGoogle'] . ', ' . $variables['wcfFontFamily'];
         }
 
+        // Define the font family set for the OS default fonts. This needs to be happen statically to
+        // allow modifications in the future in case of changes.
+        $variables['wcfFontFamilyMonospace'] = 'ui-monospace, Menlo, Monaco, "Cascadia Mono",
+            "Segoe UI Mono", "Roboto Mono", "Oxygen Mono", "Ubuntu Monospace", "Source Code Pro",
+            "Fira Mono", "Droid Sans Mono", "Courier New", monospace';
```
TimWolla, line 597: Maybe move this up into `private const` class constants. That way one does not have to search the whole file for these strings.

wcfsetup/install/files/lib/system/style/StyleCompiler.class.php
```diff
@@ -582,14 +582,26 @@ protected function compileStylesheet(string $scss, array $variables): string
...
+
+        if ($variables['wcfFontFamily'] === 'system') {
```
TimWolla, line 599: Maybe use a `public const` for the string `system` to allow for re-use in StyleAddForm.

TimWolla (approved):
Overall this LGTM.

TimWolla (approved):

PR #4143, opened 2021-04-23
Fix generating sitemaps with factor 250 objects sitemaps

joshuaruesweg:


wcfsetup/install/files/lib/system/worker/SitemapRebuildWorker.class.php
```diff
@@ -91,8 +91,14 @@ protected function countObjects() {
...
-						// modify count, because we handle only one sitemap object per call
-						$this->count += max(1, ceil($list->countObjects() / $this->limit)) * $this->limit;
+						// We can only handle one sitemap per call. Therefore we must add one to
+						// the count, even if the sitemap has 0 objects. Also we check whether 
+						// the current objects are < 250, if so, we finish the sitemap. If the 
+						// count of the last remaining objects is 250, we need an additional call
+						// to finish the sitemap. 
+						$objectCount = $list->countObjects();
+						$countModifier = ($objectCount % $this->limit === 0) ? $this->limit : 0;
+						$this->count += ceil($objectCount / $this->limit) * $this->limit + $countModifier;
```
TimWolla, line 101: I'd handle this a bit differently to make the code more clear:

```
$iterations = ceil($objectCount / $this->limit);
if (($objectCount % $this->limit) === 0) {
  // We need an additional iteration to finalize the sitemap.
  $iterations++;
}
$this->count += $iterations * $this->limit;
```

Question: Do you even need the check for an exact multiple of 250 or can you always add that extra iteration?

TimWolla (approved):

PR #4145, opened 2021-04-23
Add width to smileys

SoftCreatR:
As of now, the smiley width is unknown, which may lead in unwanted layout shifting. The proposed change will add the width property to smileys' HTML output.

### DIFF
--- a/wcfsetup/install/files/lib/data/smiley/Smiley.class.php
+++ b/wcfsetup/install/files/lib/data/smiley/Smiley.class.php
@@ -19,7 +19,7 @@ use wcf\util\StringUtil;
  * @property-read   int $packageID  id of the package which delivers the smiley
  * @property-read   int|null $categoryID id of the category the smiley belongs to or `null` if it belongs to the default category
  * @property-read   string $smileyPath path to the smiley file relative to wcf's default path
- * @property-read       string $smileyPath2x   path to the smiley file relative to wcf's default path (2x version)
+ * @property-read   string $smileyPath2x   path to the smiley file relative to wcf's default path (2x version)
  * @property-read   string $smileyTitle    title of the smiley or name of language item that contains the title
  * @property-read   string $smileyCode code used for displaying the smiley
  * @property-read   string $aliases    alternative codes used for displaying the smiley
@@ -28,6 +28,8 @@ use wcf\util\StringUtil;
 class Smiley extends DatabaseObject implements ITitledObject
 {
     protected $height;
+
+    protected $width;
 
     public $smileyCodes;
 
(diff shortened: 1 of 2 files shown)


wcfsetup/install/files/lib/data/smiley/Smiley.class.php
```diff
@@ -75,26 +77,54 @@ public function getAliases()
...
+        if ($this->height === null && $this->width === null) {
+            $this->height = $this->width = 0;
 
             $file = WCF_DIR . $this->smileyPath;
             if (\file_exists($file) && \preg_match('~\.(gif|jpe?g|png)$~', $file)) {
                 $data = \getimagesize($file);
                 if ($data !== false) {
-                    // index '1' contains the height of the image
-                    $this->height = $data[1];
+                    [$this->width, $this->height] = $data;
```
dtdesign, line 94:
```suggestion
                    // The first two indices of `getimagesize()` represent the image dimensions.
                    [$this->width, $this->height] = $data;
```
I would consider this to be _clever code_, because one either needs to know what `getimagesize()` returns or has to look it up.

wcfsetup/install/files/lib/data/smiley/Smiley.class.php
```diff
@@ -75,26 +77,54 @@ public function getAliases()
...
-     * @return  int
+     * @since   5.4
+     * @return  int[]
      */
-    public function getHeight()
+    public function getDimensions()
     {
-        if ($this->height === null) {
-            $this->height = 0;
+        if ($this->height === null && $this->width === null) {
```
dtdesign, line 87:
```suggestion
        if ($this->height === null) {
```
This changes makes no sense. If you intended to defend against an unexpected value for `$this->height`, then this should have been a logical "or". Both `$this->height` and `$this->width` equal `null` up until this point, so your extra comparison just makes the code more complex without gaining anything to it.

wcfsetup/install/files/lib/data/smiley/Smiley.class.php
```diff
@@ -75,26 +77,54 @@ public function getAliases()
...
+    /**
+     * Returns the height of the smiley.
+     *
+     * @return  int
+     */
+    public function getHeight()
+    {
+        $dimensions = $this->getDimensions();
+
+        return $dimensions['height'];
```
dtdesign, line 114:
```suggestion
        return $this->getDimensions()['height'];
```
Increased readability, because there is no processing happening and the lines are already very short.

wcfsetup/install/files/lib/data/smiley/Smiley.class.php
```diff
@@ -75,26 +77,54 @@ public function getAliases()
...
+     * Returns the width of the smiley.
+     *
+     * @since   5.4
+     * @return  int
+     */
+    public function getWidth()
+    {
+        $dimensions = $this->getDimensions();
+
+        return $dimensions['width'];
```
dtdesign, line 127:
```suggestion
        return $this->getDimensions()['width'];
```

dtdesign (dismissed):
Otherwise LGTM

TimWolla (approved):

TimWolla:
Thanks!

TimWolla:
Note: This does not currently affect the smileys in the message's content, due to this: https://github.com/WoltLab/WCF/blob/38324ef14ff08c17c004aef3108243ba9749b58f/wcfsetup/install/files/lib/system/html/output/node/HtmlOutputNodeImg.class.php#L42-L68

Ideally that can be delegated to the `->getHtml()` method somehow.

PR #4146, opened 2021-04-23
Don't set stream timeout when fopen() call has failed (guzzle#2265)

DerEnderman:
The connection to the remote crashes with cryptic error:

`stream_set_timeout() expects parameter 1 to be resource, boolean given`

when call to the `fopen()` method has failed for some reason. The root
cause is that code that tries to set timeout on the stream does not
check `fopen()` result for error indication.

Add check that we really do have valid resource before trying to set
timeout on it.

### DIFF
--- a/wcfsetup/install/files/lib/system/api/guzzlehttp/guzzle/src/Handler/StreamHandler.php
+++ b/wcfsetup/install/files/lib/system/api/guzzlehttp/guzzle/src/Handler/StreamHandler.php
@@ -321,7 +321,7 @@ class StreamHandler
                     throw new ConnectException(sprintf('Connection refused for URI %s', $uri), $request, null, $context);
                 }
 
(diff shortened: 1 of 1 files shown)


joshuaruesweg:
The variable can not be `false` in `master`. This is checked in the condition right before your edited condition (see https://github.com/WoltLab/WCF/blob/55a12383b641eb6a9b32954470a1e415ccd05b8a/wcfsetup/install/files/lib/system/api/guzzlehttp/guzzle/src/Handler/StreamHandler.php#L320-L322). 

See also: https://github.com/guzzle/guzzle/pull/2772 

However, the bug is not fixed in WoltLab Suite version 5.3 because we use an older version of Guzzle. 

//cc @TimWolla

TimWolla:
> //cc @TimWolla

I agree with everything you said.

PR #4166, opened 2021-04-29
Use null coalescing operator wherever possible

Gravatronics:
These changes were not done manually but by using IDE actions.

### DIFF
--- a/wcfsetup/install/files/lib/acp/form/AbstractAcpForm.class.php
+++ b/wcfsetup/install/files/lib/acp/form/AbstractAcpForm.class.php
@@ -60,11 +60,7 @@ abstract class AbstractAcpForm extends AbstractForm
      */
     public function getI18nValue($fieldName)
     {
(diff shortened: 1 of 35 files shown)


TimWolla (commented):
While the diff LGTM I'm not sure whether I like this change. It might make it harder to add additional cases (“data sources”) that might or might not be required for future changes, because one cannot simply add an additional `if`, but must refactor the whole method.

dtdesign:
@TimWolla On the other hand, it greatly increases code clarity by removing visual clutter. Refactoring to a full-blown `if` is not nice, but realistically in most cases just one line change among many others. What is the likelihood of us even touching one of these `if` in, like, ever? Even if we may change a few of them, the others will remain, imho making it a worth trade-off?

TimWolla:
Yeah I acknowledge that. I don't particularly care either way and I don't want to decide here. Feel free to proceed with merging if you think this is useful. Close if you don't.

PR #4169, opened 2021-04-30
Lazily load trophy images

TimWolla:
- Use width / height attributes instead of inline CSS in trophyImage.tpl
- Lazily load trophy images


joshuaruesweg (approved):
LGTM

PR #4173, opened 2021-05-01
Fix first custom smiley category not being preloaded, fixes #4172

### LINKED ISSUE #4172: First custom smiley category not being preloaded
When using at least 2 custom smiley categories and removing all smileys from the default category it will not preload the category but instead load it via ajax after the page is loaded.
This can cause the ajax status to popup in the middle of the screen.

Sir-Will:
#4172


Gravatronics:
Thanks. I manually applied it for 5.3 with https://github.com/WoltLab/WCF/commit/edeb3e25323f43e4415cf4c47d282d4addd83b0a.

PR #4174, opened 2021-05-02
Simplify the domain and landing page management

### LINKED ISSUE #3859: Deprecate multi domain setups

### LINKED ISSUE #4142: Overhaul the application management in the admin panel
The current application management is quite technical and error-prone: Editing apps in the wrong order can somewhat brick the admin panel and the ability to change the path is pretty much a foot-gun in disguise. The interface needs to become much more user friendly and preventing user errors as best as possible.

* [x] For existing multi-domain setups, everything remains as it is. Show a warning that these setups are deprecated and will be removed.
* [x] Normal setups will be presented a different page.
  * [x] Dedicated input field to change the domain.
  * [x] List of installed apps with per app drop-down menu for the landing page.

Changes to the path are no longer offered via the UI, especially since users expect this to auto-magically move files around, when it really only updates the `wcfN_application` values. Instead, path changes should be handled by the system recovery mode which is designed to gracefully handle mismatches.

See #3859

dtdesign:
Closes #4142 
Closes #3859


wcfsetup/install/files/lib/acp/page/ApplicationManagementPage.class.php
```diff
@@ -1,62 +0,0 @@
-<?php
```
TimWolla, line 1: You need to explicitly delete this file. Otherwise this is going to collide for upgrades.

see https://github.com/WoltLab/WCF/blob/master/wcfsetup/install/files/acp/update_com.woltlab.wcf_5.4_removeFiles.php

wcfsetup/install/files/lib/acp/form/ApplicationManagementForm.class.php
```diff
@@ -0,0 +1,237 @@
+<?php
+
+namespace wcf\acp\form;
+
+use Stripe\Exception\PermissionException;
```
TimWolla, line 5: You're searching for PermissionDeniedException :grin:

wcfsetup/install/files/lib/acp/form/ApplicationManagementForm.class.php
```diff
@@ -0,0 +1,237 @@
...
+
+/**
+ * Shows the application management form.
+ *
+ * @author Alexander Ebert
+ * @copyright 2001-2021 WoltLab GmbH
+ * @license GNU Lesser General Public License <http://opensource.org/licenses/lgpl-license.php>
+ * @package WoltLabSuite\Core\Acp\Form
+ */
+class ApplicationManagementForm extends AbstractForm
```
TimWolla, line 29:
```suggestion
final class ApplicationManagementForm extends AbstractForm
```
Make everything final by default.

wcfsetup/install/files/acp/templates/applicationManagement.tpl
```diff
@@ -14,42 +14,116 @@
...
+			</dl>
+		</section>
+
+		<script>
+			(() => {
+				const domainName = document.getElementById("domainName");
+				const cookieDomain = document.getElementById("cookieDomain");
+
+				// Keep the cookie domain in sync if it was previously identical.
+				if (domainName.value === cookieDomain.value) {
```
TimWolla, line 70: It's a bit odd that this is checked in JavaScript, instead of only injecting this piece of JavaScript when the values differ using PHP.

wcfsetup/install/files/acp/templates/applicationManagement.tpl
```diff
@@ -14,42 +14,116 @@
...
+<form method="post" action="{link controller='ApplicationManagement'}{/link}">
+	{if !$isMultiDomainSetup}
+		<section class="section">
+			<h2 class="sectionTitle">{lang}wcf.acp.application.management.domain{/lang}</h2>
+
+			<dl{if $errorField == 'domainName'} class="formError"{/if}>
+				<dt><label for="domainName">{lang}wcf.acp.application.management.domainName{/lang}</label></dt>
+				<dd>
+					<div class="inputAddon">
+						<span class="inputPrefix">http(s)://</span>
```
TimWolla, line 30:
```suggestion
						<span class="inputPrefix">https://</span>
```
It's 2021. We should simply assume https:// in the UI.

wcfsetup/install/files/lib/acp/form/ApplicationEditForm.class.php
```diff
@@ -17,6 +17,7 @@
 use wcf\system\Regex;
 use wcf\system\WCF;
 use wcf\util\FileUtil;
+use wcf\util\HeaderUtil;
```
TimWolla, line 20: This import is what GitHub Action complains about.

wcfsetup/install/files/lib/acp/form/ApplicationManagementForm.class.php
```diff
@@ -0,0 +1,237 @@
...
+        }
+
+        $this->saved();
+
+        if (!empty($this->landingPageID[1])) {
+            (new Page($this->landingPageID[1]))->setAsLandingPage();
+        } else {
+            $sql = "UPDATE  wcf" . WCF_N . "_page
+                        SET     isLandingPage = ?
+                        WHERE   isLandingPage = ?";
```
TimWolla, line 196: Broken indentation.

TimWolla (approved):
Diff LGTM, didn't test.

PR #4184, opened 2021-05-04
Add SensitiveArgument attribute

TimWolla:
- Add \wcf\SensitiveArgument attribute
- Consider parameters with password|passphrase|secret in their name to be sensitive
- Make use of \wcf\SensitiveArgument attribute

### DIFF
--- a/wcfsetup/install/files/lib/core.functions.php
+++ b/wcfsetup/install/files/lib/core.functions.php
@@ -717,13 +717,13 @@ EXPLANATION;
 			if (!isset($item['args'])) $item['args'] = [];
 
 			if ($item['class']) {
-				$f = new \ReflectionMethod($item['class'], $item['function']);
+				$function = new \ReflectionMethod($item['class'], $item['function']);
 			}
 			else {
-				$f = new \ReflectionFunction($item['function']);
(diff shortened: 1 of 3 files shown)


Gravatronics (approved):
I explicitly checked the change by calling a undefined function in `loginManually()` using PHP 7.2.

wcfsetup/install/files/lib/core.functions.php
```diff
@@ -710,6 +715,36 @@ function sanitizeStacktrace(\Throwable $e, bool $ignorePaths = false) {
 			if (!isset($item['class'])) $item['class'] = '';
 			if (!isset($item['type'])) $item['type'] = '';
 			if (!isset($item['args'])) $item['args'] = [];
+
+			if ($item['class']) {
+				$f = new \ReflectionMethod($item['class'], $item['function']);
```
dtdesign, line 720: Why `$f`, can it be at least something like `$func` instead?

dtdesign (approved):

wcfsetup/install/files/lib/core.functions.php
TimWolla: Remnants from my initial PoC implementation :smile:

PR #4213, opened 2021-05-14
Fix attachment ids in metacode elements during import

### LINKED ISSUE #4212: `AbstractAttachmentImporter::fixEmbeddedAttachments()` ignores `woltlab-metacode`
https://github.com/WoltLab/WCF/blob/16600646d91df3969087dc8f10d133f78fc6deae/wcfsetup/install/files/lib/system/importer/AbstractAttachmentImporter.class.php#L109-L123

Imports from newer WoltLab Suite versions have messages with attachments represented by `woltlab-metacode` elements instead of the attachment BBCode so that this method does not properly update the ids of these included attachments.

Gravatronics:
Close #4212

### DIFF
--- a/wcfsetup/install/files/lib/system/importer/AbstractAttachmentImporter.class.php
+++ b/wcfsetup/install/files/lib/system/importer/AbstractAttachmentImporter.class.php
@@ -4,6 +4,7 @@ use wcf\data\attachment\Attachment;
 use wcf\data\attachment\AttachmentEditor;
 use wcf\system\exception\SystemException;
 use wcf\util\FileUtil;
+use wcf\util\JSON;
 
 /**
  * Imports attachments.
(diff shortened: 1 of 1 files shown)


wcfsetup/install/files/lib/system/importer/AbstractAttachmentImporter.class.php
```diff
@@ -108,6 +109,32 @@ protected function fixEmbeddedAttachments($message, $oldID, $newID) {
 			
 			return $message;
 		}
+		else {
+			return \preg_replace_callback(
+				'~<woltlab-metacode data-name="attach" data-attributes="(?<attributes>[^"]+)">~',
+				function (array $matches) use ($oldID, $newID): string {
```
TimWolla, line 115:
```suggestion
				static function (array $matches) use ($oldID, $newID): string {
```

Otherwise this will fail after the merge to master.

TimWolla (approved):
Overall this LGTM and it should not be unsafe.

PR #4222, opened 2021-05-18
provide files-property for `package.json`

mutec:
Prevent reading the whole repository when this repository is a dependency. This down-loads `ts/*` and `global.d.ts` only.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,10 @@
 {
   "name": "@woltlab/wcf",
   "version": "5.4.0",
+  "files": [
+    "ts/**/*",
+    "global.d.ts"
+  ],
   "devDependencies": {
     "@types/googlemaps": "^3.43.0",
     "@types/perfect-scrollbar": "^0.6.1",


TimWolla:
Sweet, thank you!

TimWolla:
This appears to shave off ~4 seconds of the `npm install` step for com.woltlab.wcf.conversation in GitHub actions. It's down to 7 seconds from 11 seconds.

mutec:
And it reduces stupid file-duplicates for the PHP-stuff and decreases the size of the project and the time for indexing the project. Since I'm out of RAM it's pretty annoying to index the whole project. ;)
Since the WSC doesn't provide any PHP-"library" in common sense, it shouldn't have any disadvantages.

TimWolla:
>  Since the WSC doesn't provide any PHP-"library" in common sense, it shouldn't have any disadvantages.

Absolutely. Especially since composer is there for PHP. In the future we also plan to just upload the compiled `.d.ts` files to a proper registry to further slim down the size and avoid the dependency on git, but the current solution *works* for now.

PR #4234, opened 2021-05-21
Add 'discarded' status to email log

TimWolla:
- Make EmailDeliveryBackgroundJob::updateStatus() public
- Add EmailLogEntry::STATUS_DISCARDED
- Update email log status to discarded when dropping notification mails

### DIFF
--- a/wcfsetup/install/files/lib/system/background/job/NotificationEmailDeliveryBackgroundJob.class.php
+++ b/wcfsetup/install/files/lib/system/background/job/NotificationEmailDeliveryBackgroundJob.class.php
@@ -2,6 +2,7 @@
 
 namespace wcf\system\background\job;
 
+use wcf\data\email\log\entry\EmailLogEntry;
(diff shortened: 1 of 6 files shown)


wcfsetup/install/files/lib/system/background/job/NotificationEmailDeliveryBackgroundJob.class.php
```diff
@@ -125,6 +131,11 @@ public function perform()
 
             // Drop email if the notification is already confirmed.
             if ($event->isConfirmed()) {
+                $this->job->updateStatus(
+                    EmailLogEntry::STATUS_DISCARDED,
+                    'notification is confirmed'
```
dtdesign, line 136:
```suggestion
                    'obsolete, notification is already confirmed'
```

dtdesign (changes requested):
LGTM

PR #4236, opened 2021-05-21
Fetch UserProfiles live to avoid memory leaks through the RuntimeCache on CLI

### LINKED ISSUE #4233: Creation of WebP cover photos in UserRebuildDataWorker should not use RuntimeCache
https://github.com/WoltLab/WCF/blob/8afa6f697b25dade2e2168d2cd36fd4c968412af/wcfsetup/install/files/lib/system/worker/UserRebuildDataWorker.class.php#L286

RuntimeCaches result in memory "leaks" during CLI rebuilds.

dtdesign:
Fixes #4233


TimWolla (approved):
Diff LGTM, I trust you tested this.

PR #4240, opened 2021-05-26
Do not process the pageLogo in StyleAction::updateVariables() if it is unchanged

TimWolla:
This change ensures that the filename of the logo remains the same, unless the
file actually is modified.

A fix for this issue was initially attempted in
cc17f302902fb62c78b4411dadc3243913f18c74. However that commit was much more
complex and it introduced additional bugs, because it no longer passed an
explicit `null` for removed files, thus breaking the update of the related
database columns. It was subsequently reverted in
dd7c98858d406f8ca9824770ae8a68edde451493.


joshuaruesweg (approved):
Diff LGTM

PR #4243, opened 2021-05-28
Remove X-XSS-Protection response header

TimWolla:
The XSS auditor has been removed in all web browsers by now, making this header
useless.


dtdesign (approved):

TimWolla:
Undrafting, now that master is equal to 5.5.

PR #4255, opened 2021-06-03
Fix displaying `runStandalone` in devtools

Fabii547:
After editing an instruction it is always shown that this instruction is run standalone.


Gravatronics:
Thanks!

PR #4258, opened 2021-06-04
Mark `\wcf\util\UserAgent` as final

TimWolla:
This was found during development of another PR. There is no good reason for
this class to not be final and we're still in Beta. So let's get this done
before it's too late. Opening it back up later is easy enough.


dtdesign (approved):

mutec:
There is no good reason why it should be final. But good reaons against making APIs unchangeable.

TimWolla:
> There is no good reason why it should be final.

OO best practices disagree with you. See also: https://ocramius.github.io/blog/when-to-declare-classes-final/

joshuaruesweg (approved):
LGTM

PR #4259, opened 2021-06-04
Mask IP address in session list

TimWolla:
See https://community.woltlab.com/thread/290548-show-hide-session-ip/

### DIFF
--- a/wcfsetup/install/files/lib/util/IpAddress.class.php
+++ b/wcfsetup/install/files/lib/util/IpAddress.class.php
@@ -109,6 +109,47 @@ final class IpAddress
     }
 
     /**
+     * Returns a masked IP address with the masked parts replaced by bullets.
+     *
+     * @see IpAddress::toMasked()
+     */
+    public function toBulletMasked(int $mask4, int $mask6): string
+    {
+        $masked = $this->toMasked($mask4, $mask6);
+
+        if (($mask4 % 8) !== 0) {
+            throw new \InvalidArgumentException('Given $mask4 is not a multiple of 8.');
+        }
+        if (($mask6 % 4) !== 0) {
+            throw new \InvalidArgumentException('Given $mask6 is not a multiple of 4.');
+        }
+
+        if ($masked->asV4()) {
+            $maskedBlocks = (32 - $mask4) / 8;
+            $replacement = \str_repeat(".\u{2022}\u{2022}\u{2022}", $maskedBlocks);
+
+            return \preg_replace(
+                '/(\.0){' . ($maskedBlocks) . '}$/',
+                $replacement,
+                (string)$masked
+            );
+        } else {
+            $maskedDigits = (128 - $mask6) / 4;
+
+            // Partially masked quadruplet.
+            $replacement = \str_repeat("\u{2022}", ($maskedDigits % 4));
+            // Fully masked quadruplets.
+            $replacement .= \str_repeat(":\u{2022}\u{2022}\u{2022}\u{2022}", ($maskedDigits / 4));
+
+            return \preg_replace(
+                '/.{' . ($maskedDigits % 4) . '}::$/',
+                $replacement,
+                (string)$masked
+            );
+        }
+    }
+
+    /**
      * @see IpAddress::getIpAddress()
      */
     public function __toString(): string
--- a/wcfsetup/install/files/lib/util/IpAddress.class.php
+++ b/wcfsetup/install/files/lib/util/IpAddress.class.php
@@ -80,7 +80,7 @@ final class IpAddress
     /**
      * Returns a new IP address with the last $maskX bits masked off.
      */
-    public function mask(int $mask4, int $mask6): self
+    public function toMasked(int $mask4, int $mask6): self
     {
         if ($mask4 < 0 || $mask4 > 32) {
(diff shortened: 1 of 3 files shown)


wcfsetup/install/files/lib/util/IpAddress.class.php
```diff
@@ -0,0 +1,126 @@
...
+            return new self($ip);
+        }
+
+        return null;
+    }
+
+    /**
+     * Returns a new IP address with the last $maskX bits masked off.
+     */
+    public function mask(int $mask4, int $mask6): self
```
TimWolla, line 83: Should this rather be named `masked()`?

wcfsetup/install/files/lib/util/IpAddress.class.php
```diff
@@ -0,0 +1,126 @@
...
+    /**
+     * @var string
+     */
+    private $ipAddress;
+
+    public function __construct(string $ipAddress)
+    {
+        if (\filter_var($ipAddress, \FILTER_VALIDATE_IP) === false) {
+            throw new \InvalidArgumentException(\sprintf(
+                "The given IP address '%s'is not a valid IP address.",
```
Gravatronics, line 25:
```suggestion
                "The given IP address '%s' is not a valid IP address.",
```

wcfsetup/install/files/lib/util/IpAddress.class.php
Gravatronics: `getMasked()`?

wcfsetup/install/files/lib/util/IpAddress.class.php
```diff
@@ -0,0 +1,126 @@
...
+            $mask .= \chr(0xff << (8 - \min(8, $maskBits)));
+        }
+
+        return new self(\inet_ntop(\inet_pton((string)$ipAddress) & $mask));
+    }
+
+    /**
+     * @see IpAddress::getIpAddress()
+     */
+    public function __toString(): string
```
Gravatronics, line 114: The return type is inconsistent with the return type of `getIpAddress()`.

wcfsetup/install/files/lib/util/IpAddress.class.php
```diff
@@ -0,0 +1,126 @@
...
+    {
+        return $this->getIpAddress();
+    }
+
+    /**
+     * Returns the raw IP address that was used to construct this instance.
+     */
+    public function getIpAddress(): ?string
```
TimWolla, line 122:
```suggestion
    public function getIpAddress(): string
```

wcfsetup/install/files/lib/util/IpAddress.class.php
TimWolla: Indeed. It should not be nullable, of course.

mutec:
Teaching people how to use the internet and not posting every peace of shit they can make a screenshot of would make more sense… ^^

wcfsetup/install/files/lib/util/IpAddress.class.php
dtdesign: `asMasked()` to use a consistent naming schema?

dtdesign (approved):

dtdesign:
> Teaching people how to use the internet and not posting every peace of shit they can make a screenshot of would make more sense… ^^

That is one way to look at this. On the other hand, it does help protecting users from making mistakes in the first place.

I like to think of this like the administrator rights in Windows. You do not have all permissions, but can acquire them if you know how. It's just another a defense-in-depth approach that is designed to prevent accidents. Remember: Mistakes will happen.

wcfsetup/install/files/lib/util/IpAddress.class.php
TimWolla: I prefer not to use the `as` prefix, as the masking operation is lossy and it's not just another presentation of the same information. I opted for `toMasked`, as this logically matches `toString()`.

TimWolla:
Undrafted this PR now, as the planned visual change (show bullets in place of the masked digits) is completed now:

image

TimWolla:
Updated the target branch to 5.4, now that master is equal to 5.5.

wcfsetup/install/files/lib/util/IpAddress.class.php
```diff
@@ -0,0 +1,167 @@
...
+    }
+
+    /**
+     * Returns a masked IP address with the masked parts replaced by bullets.
+     *
+     * @see IpAddress::toMasked()
+     */
+    public function toBulletMasked(int $mask4, int $mask6): string
+    {
+        $masked = $this->toMasked($mask4, $mask6);
```
dtdesign, line 118: I would move this line below the checks for the multiples of 4/8. It makes more sense to validate parameters first that will reject regardless of the output of the call to `toMasked()`.

dtdesign (changes requested):
LGTM

Gravatronics (commented):
I have no further comments.

wcfsetup/install/files/lib/util/IpAddress.class.php
TimWolla: I intentionally selected this order, because I wanted to reuse the parameter validation in `->toMasked()`, validating the range before validating that the mask lines up with a block / quadruplet.

Issue #4178, opened 2021-05-03
Add deletion support for language PIP


Gravatronics:
… instead of having to rely on a PHP script, e.g. https://github.com/WoltLab/WCF/blob/039fe7c90cbb1e7c56c5554a17ada598c52607ce/wcfsetup/install/files/acp/update_com.woltlab.wcf_5.4_deleteLanguageItems.php.

For backwards compatibility, the current language file structure must still be supported, but a new structure with intermediate `<import>` and `<delete>` elements should also be supported.

PR #4261, opened 2021-06-04
Add deletion support for language PIP

Gravatronics:
**Preface:** The current

```xml
<language>
  <category name="…">
    <item name="…">…</item>
  </category>
</language
```

structure remains supported with these changes but is considered deprecated.

---

To support deleting phrases with this PIP, support for a `delete` element is added just like in the other XML-based PIPs:

```xml
<language>
  <delete>
    <item name="…"/>
  </delete>
</language
```

There are no `category` elements between the `delete` element and the `item` elements as for deletion, the category of the phrase is irrelevant as its name is unique globally.

To keep this PIP's structure in sync with other PIPs, in addition to the new support for the `delete` element, the imported phrases can now also be added to a `import` element like in the other XML-based PIPs:

```xml
<language>
  <import>
    <category name="…">
      <item name="…">…</item>
    </category>
  </import>
</language
```

While the old structure without the intermediate `import` element is compatible with the new support for a `delete` element, i.e.

```xml
<language>
  <category name="…">
    <item name="…">…</item>
  </category>
  <delete>
    <item name="…"/>
  </delete>
</language
```

imports and deletes the relevant phrases.

Using `import` and `category` siblings like this

```xml
<language>
  <import>
    <category name="…">
      <item name="…">…</item>
    </category>
  </import>
  <category name="…">
    <item name="…">…</item>
  </category>
</language
```

however, is not supported. Once an `import` element exists, only its phrases are imported and any `category` *siblings*  are ignored.

See #4178

### DIFF
--- a/XSD/language.xsd
+++ b/XSD/language.xsd
@@ -7,14 +7,32 @@
 	<!-- data element -->
 	<xs:element name="language">
 		<xs:complexType>
-			 <xs:sequence>
+			<!-- deprecated since 5.5 -->
+			<xs:sequence>
 				<xs:element name="category" type="category" maxOccurs="unbounded" />
 			</xs:sequence>
+			<!-- /deprecated since 5.5 -->
+			<xs:element name="import" type="import" minOccurs="0" />
+			<xs:element name="delete" type="delete" minOccurs="0" />
 			<xs:attribute name="languagecode" type="woltlab_varchar" use="required" />
 			<xs:attribute name="languagename" type="woltlab_varchar" use="optional" />
 			<xs:attribute name="countrycode" type="woltlab_varchar" use="optional" />
 		</xs:complexType>
 	</xs:element>
+	
+	<!-- import element -->
+	<xs:complexType name="import">
+		<xs:sequence>
+			<xs:element name="category" type="category" maxOccurs="unbounded" />
+		</xs:sequence>
+	</xs:complexType>
+	
+	<!-- delete element -->
+	<xs:complexType name="delete">
+		<xs:sequence>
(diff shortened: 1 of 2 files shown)


XSD/language.xsd
```diff
@@ -7,15 +7,33 @@
 	<!-- data element -->
 	<xs:element name="language">
 		<xs:complexType>
-			 <xs:sequence>
+			<!-- deprecated since 5.5 -->
+			<xs:sequence>
 				<xs:element name="category" type="category" maxOccurs="unbounded" />
 			</xs:sequence>
+			<!-- /deprecated since 5.5 -->
```
TimWolla, line 14: Should this simply be removed, then?

TimWolla (approved):

XSD/language.xsd
Gravatronics: No, otherwise IDEs will treat the current structure as erroneous. We can remove it once the current structure is not supported anymore.

dtdesign:
> Once an `import` element exists, only its phrases are imported and any `category` _siblings_ are ignored.

I would favor a hard rejection, either use the legacy format or the new format, but not both.

XSD/language.xsd
TimWolla: Treating it as erroneous is not wrong. It's easy to adapt this to the new format and I think it's useful to visibly mark the old structure as deprecated before it's actually removed.

Gravatronics:
@dtdesign I implemented the suggested change with https://github.com/WoltLab/WCF/pull/4261/commits/8c9570311cb8eecc6a8a7570c680f652bfc124c6.

PR #4263, opened 2021-06-07
Fix visibility check in `Condition/Page/Dependence`

Fabii547:
In the original js implementation the filter condition was `return pageIds.indexOf(pageId) === -1;`.


TimWolla (approved):

TimWolla:
Thanks!

Issue #4180, opened 2021-05-03
Add PIPs to delete files, templates, and ACP templates


Gravatronics:
… instead of having to rely on a PHP script, e.g. https://github.com/WoltLab/WCF/blob/039fe7c90cbb1e7c56c5554a17ada598c52607ce/wcfsetup/install/files/acp/update_com.woltlab.wcf_5.4_removeFiles.php.

PR #4267, opened 2021-06-07
Add PIP to delete files

Gravatronics:
See #4180

### DIFF
--- a/wcfsetup/install/files/lib/system/package/plugin/FileDeletePackageInstallationPlugin.class.php
+++ b/wcfsetup/install/files/lib/system/package/plugin/FileDeletePackageInstallationPlugin.class.php
@@ -1,0 +1,274 @@
+<?php
+
+namespace wcf\system\package\plugin;
+
+use wcf\data\application\Application;
+use wcf\data\package\Package;
+use wcf\data\package\PackageCache;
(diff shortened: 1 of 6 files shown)


wcfsetup/install/files/lib/system/package/plugin/FileDeletePackageInstallationPlugin.class.php
```diff
@@ -0,0 +1,274 @@
...
+        return $xpath->query('/ns:data/ns:delete/ns:' . $this->tagName);
+    }
+
+    /**
+     * @inheritDoc
+     */
+    protected function getEmptyXml()
+    {
+        $xsdFilename = $this->getXsdFilename();
+        $apiVersion = WSC_API_VERSION;
```
TimWolla, line 265: Is this correct? Probably affects all PIPs.

TimWolla (approved):

wcfsetup/install/files/lib/system/package/plugin/FileDeletePackageInstallationPlugin.class.php
Gravatronics: I moved this into a separate issue https://github.com/WoltLab/WCF/issues/4268.

PR #4271, opened 2021-06-08
Make MysqlSearchIndexManager implement set()

TimWolla:
The use of `add()` is deprecated. `set()` should be the canonical method.


Gravatronics (approved):
LGTM, though I am unsure about the relevance as `SearchIndexManager` only uses `set()` anymore and there should be no reason to use `MysqlSearchIndexManager` directly, or is there?

TimWolla:
> there should be no reason to use `MysqlSearchIndexManager` directly, or is there?

Probably not, but you know: Compability. After this change we can simply remove the methods whenever we like.

PR #4284, opened 2021-06-11
fix possible error while reading i18n data from forms

mutec:
Previous the code caused the following error: `Uncaught TypeError: Cannot set property '1' of undefined` since it tries to set the language's value in an object which doesn't exist (yet).


mutec:
btw. for the future: should I compile the TS and submit the compiled JS within the PR or should I just submit the changed TS?

Gravatronics:
Thanks and yes, please always also commit the compiled JS files.

PR #4285, opened 2021-06-11
force `LF` line endings for generated JavaScript

mutec:
using Windows the generated JavaScript would be formatted using `CRLF` which is pretty much time wasting


TimWolla:
Why would one use platform specific defaults in a modern programming language?!? Thanks.

PR #4287, opened 2021-06-13
Language variable fixed in de.xml & en.xml

felix-d1strict:
References to the developer documentation were no longer up to date.


TimWolla (commented):
The code changes LGTM, but the commit message should be improved. This type of mistake is not a “Typo” and the apostrophe certainly does not belong there either :smile:

Gravatronics (approved):

felix-d1strict:
> The code changes LGTM, but the commit message should be improved. This type of mistake is not a “Typo” and the apostrophe certainly does not belong there either 😄
You're right about that. 😁

Issue #4281, opened 2021-06-11
Usage of deprecated PHP


Gravatronics:
- `Class 'HTTPUnauthorizedException' is deprecated`
https://github.com/WoltLab/WCF/blob/25e1c4184e7048313f040499bc8e413ed6d61b9e/wcfsetup/install/files/lib/system/package/PackageUpdateDispatcher.class.php#L10
- `Class 'HTTPRequest' is deprecated`
https://github.com/WoltLab/WCF/blob/25e1c4184e7048313f040499bc8e413ed6d61b9e/wcfsetup/install/files/lib/system/package/PackageUpdateDispatcher.class.php#L16
- `Class 'HTTPRequest' is deprecated`
https://github.com/WoltLab/WCF/blob/25e1c4184e7048313f040499bc8e413ed6d61b9e/wcfsetup/install/files/lib/system/package/PackageUpdateDispatcher.class.php#L119
- `Class 'HTTPRequest' is deprecated`
https://github.com/WoltLab/WCF/blob/25e1c4184e7048313f040499bc8e413ed6d61b9e/wcfsetup/install/files/lib/system/package/PackageUpdateDispatcher.class.php#L161
- `Class 'HTTPUnauthorizedException' is deprecated`
https://github.com/WoltLab/WCF/blob/25e1c4184e7048313f040499bc8e413ed6d61b9e/wcfsetup/install/files/lib/system/package/PackageUpdateDispatcher.class.php#L183
- `Class 'DatabaseException' is deprecated`
https://github.com/WoltLab/WCF/blob/25e1c4184e7048313f040499bc8e413ed6d61b9e/wcfsetup/install/files/lib/acp/form/DataImportForm.class.php#L7
- `Class 'DatabaseException' is deprecated`
https://github.com/WoltLab/WCF/blob/25e1c4184e7048313f040499bc8e413ed6d61b9e/wcfsetup/install/files/lib/acp/form/DataImportForm.class.php#L242
- `Class 'IPrintableException' is deprecated`
https://github.com/WoltLab/WCF/blob/25e1c4184e7048313f040499bc8e413ed6d61b9e/wcfsetup/install/files/lib/system/WCF.class.php#L23
- `Class 'IPrintableException' is deprecated`
https://github.com/WoltLab/WCF/blob/25e1c4184e7048313f040499bc8e413ed6d61b9e/wcfsetup/install/files/lib/system/WCF.class.php#L292
- `Field 'rankImage' is deprecated`
[...]

PR #4293, opened 2021-06-15
Migrate RecaptchaHandler to Guzzle

TimWolla:
see #4281

### DIFF
--- a/wcfsetup/install/files/lib/system/captcha/RecaptchaHandler.class.php
+++ b/wcfsetup/install/files/lib/system/captcha/RecaptchaHandler.class.php
@@ -2,9 +2,12 @@
 
 namespace wcf\system\captcha;
 
+use GuzzleHttp\ClientInterface;
+use GuzzleHttp\Psr7\Request;
+use Psr\Http\Client\ClientExceptionInterface;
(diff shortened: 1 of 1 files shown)


wcfsetup/install/files/lib/system/captcha/RecaptchaHandler.class.php
```diff
@@ -116,35 +119,37 @@ public function validate()
...
-
+        } catch (ClientExceptionInterface $e) {
             // log error, but accept captcha
             \wcf\functions\exception\logThrowable($e);
         }
 
         WCF::getSession()->register('recaptchaDone', true);
     }
+
+    private function getHttpClient(): ClientInterface
```
Gravatronics, line 151: Is this method really necessary?

Gravatronics (approved):
Diff LGTM

wcfsetup/install/files/lib/system/captcha/RecaptchaHandler.class.php
TimWolla: Necessary is a strong word. I added it for consistency with other classes making HTTP Requests to keep the initialization of the client in an easy-to-find place.

PR #4328, opened 2021-06-21
Add form field dependency for interval of values

Gravatronics:
See https://community.woltlab.com/thread/285533-formbuilder-gr%C3%B6%C3%9Fer-kleiner-als-bedingung/

### DIFF
--- a/ts/WoltLabSuite/Core/Form/Builder/Field/Dependency/ValueInterval.ts
+++ b/ts/WoltLabSuite/Core/Form/Builder/Field/Dependency/ValueInterval.ts
@@ -1,0 +1,62 @@
+/**
+ * Form field dependency implementation that requires the value of a field to be in the interval
+ * [minimum, maximum].
+ *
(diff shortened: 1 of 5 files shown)


ts/WoltLabSuite/Core/Form/Builder/Field/Dependency/ValueInterval.ts
```diff
@@ -0,0 +1,63 @@
...
+      if (this._minimum !== null && this._minimum > value) {
+        return false;
+      } else if (this._maximum !== null && this._maximum < value) {
+        return false;
+      }
+
+      return true;
+    }
+    else {
+      throw new Error("`ValueInterval` is only supported for individual fields.");
```
TimWolla, line 40:
```suggestion
      throw new Error("'ValueInterval' is only supported for individual fields.");
```

TimWolla (approved):

PR #4330, opened 2021-06-22
Start modernizing JavaScript code in templates

Gravatronics:


TimWolla (approved):
These are pretty superficial changes, but not wrong. We should rather aim to remove JavaScript within templates as much as possible, because we can't use TypeScript within a template.

dtdesign:
I agree with @TimWolla, especially the new syntax to exclude files from the bundle make it easy to put those things into separate modules. Bonus points for having true editor support including TypeScript within those files.

PR #4341, opened 2021-06-24
Prevent unfurling DownloadFailed exceptions from ending up in the error log

### LINKED ISSUE #4237: Failures to fetch remote content for url unfurling yields exceptions
The error message "Could not download content." is logged when the attempt to fetch data from a remote host fails. Guzzle reported the message "Connection refused for URI <redacted>". Unless we want to retry the attempt, such failures should be silently discarded.

It was thrown a `DownloadFailed` Exception.

TimWolla:
Resolves #4237


joshuaruesweg (approved):
LGTM

PR #4345, opened 2021-06-25
Update to scssphp/scssphp 1.5

TimWolla:
- Update to scssphp/scssphp 1.5
- Fix scssphp deprecations in StyleCompiler


TimWolla:
Now I also tested WCFSetup. The relevant commit for review is the second.

dtdesign (approved):

PR #4346, opened 2021-06-26
support custom number of decimals in currency modifier

mutec:


wcfsetup/install/files/lib/system/template/plugin/CurrencyModifierTemplatePlugin.class.php
```diff
@@ -23,9 +27,11 @@ class CurrencyModifierTemplatePlugin implements IModifierTemplatePlugin
      */
     public function execute($tagArgs, TemplateEngine $tplObj)
     {
+        $decimals = $tagArgs[1] ?? 2;
```
BurntimeX, line 30:
```suggestion
        $decimals = intval($tagArgs[1] ?? 2);
```

BurntimeX (changes requested):

PR #4289, opened 2021-06-14
Remove deprecated Recaptcha-related classes

Gravatronics:
See #4281

PR #4269, opened 2021-06-08
Add PIPs to delete ACP and frontend templates

Gravatronics:
Close #4180

PR #4348, opened 2021-06-28
Improve speed of PIPs to delete files

Gravatronics:
… by fetching all existing database records at once and deleting obsolete records in one transaction.

See #4269
See #4289


wcfsetup/install/files/lib/system/package/plugin/AbstractFileDeletePackageInstallationPlugin.class.php
```diff
@@ -81,30 +71,67 @@ protected function handleDelete(array $items)
...
+                WHERE       packageID = ?
+                        AND {$this->getFilenameTableColumn()} = ?
+                        AND application = ?";
+        $deleteStatement = WCF::getDB()->prepare($sql);
+
+        WCF::getDB()->beginTransaction();
+        foreach ($logFiles as $application => $files) {
+            foreach ($files as $file => $packageID) {
+                $deleteStatement->execute([
+                    $packageID,
```
TimWolla, line 128: This makes me feel uneasy. While I know that `$packageID === $this->installation->getPackageID()` at this point it is not actually clear. I suggest to either change this to `$this->installation->getPackageID()` or to add an assertion verifying this.

TimWolla (approved):

PR #4352, opened 2021-06-29
Modernize Twitter Auth

TimWolla:
- Use `\hash_equals()` for token validation in TwitterAuthAction
- Use constant time b64 encoder in TwitterAuthAction
- Use `getControllerLink()` to generate TwitterAuth callback URL
- Improve readability of OAuth header generation in TwitterAuthAction
- Migrate Twitter authentication to Guzzle


joshuaruesweg (changes requested):
The email parameter from the callback is optional. You require them now, which leads to an error, if the email is not provided by Twitter.

joshuaruesweg (approved):

PR #4356, opened 2021-07-01
Add UserLoggedIn event and cancel lost password requests upon login

TimWolla:
- Add a UserLoggedIn event
- Implement the UserLoggedIn event
- Cancel lost password requests when the user logs in

### DIFF
--- a/wcfsetup/install/files/lib/data/user/UserAction.class.php
+++ b/wcfsetup/install/files/lib/data/user/UserAction.class.php
@@ -1235,4 +1235,21 @@ class UserAction extends AbstractDatabaseObjectAction implements IClipboardActio
             'languageID' => $this->parameters['languageID'],
         ]);
     }
+
+    /**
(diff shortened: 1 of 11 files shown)


Gravatronics (changes requested):

dtdesign (approved):
LGTM, with exception to the remarks left by @Gravatronics

wcfsetup/install/files/lib/data/user/UserAction.class.php
```diff
@@ -1235,4 +1235,21 @@ public function devtoolsSetLanguage()
             'languageID' => $this->parameters['languageID'],
         ]);
     }
+
+    /**
+     * @since   5.5
+     */
+    public function cancelLostPasswordRequest()
```
Gravatronics, line 1242:
```suggestion
    public function cancelLostPasswordRequest(): void
```
?

PR #4358, opened 2021-07-02
improve TOTP user experience

mutec:


wcfsetup/install/lang/de.xml
```diff
@@ -5015,7 +5015,7 @@ Die E-Mail-Adresse des neuen Benutzers lautet: {@$user->email}.
...
 		<item name="wcf.user.security.multifactor.setupEmail.subject"><![CDATA[Mehrfaktor-Authentifizierung auf {@PAGE_TITLE|language} aktiviert]]></item>
 		<item name="wcf.user.security.multifactor.totp.code"><![CDATA[Einmalcode]]></item>
-		<item name="wcf.user.security.multifactor.totp.code.description"><![CDATA[Der durch die Smartphone-App generierte 6-stellige Einmalcode.]]></item>
+		<item name="wcf.user.security.multifactor.totp.code.description"><![CDATA[Der durch die Smartphone-App generierter 6-stelliger Einmalcode.]]></item>
```
cadeyrn, line 5018: That should be reverted because it's wrong.

wcfsetup/install/files/lib/system/user/multifactor/TotpMultifactorMethod.class.php
```diff
@@ -78,6 +78,7 @@ public function createManagementForm(IFormDocument $form, ?Setup $setup, $return
                     ->label('wcf.user.security.multifactor.totp.code')
                     ->description('wcf.user.security.multifactor.totp.code.description')
                     ->required()
+                    ->autocomplete('one-time-code')
```
TimWolla, line 81: This should be applied within `CodeFormField`.

TimWolla (changes requested):

wcfsetup/install/lang/de.xml
mutec: Oh, i didn't see the "Der". 👀

wcfsetup/install/files/lib/system/user/multifactor/TotpMultifactorMethod.class.php
mutec: Are you sure? I'm asking because it is set to `off` explicitly in the constructor.

TimWolla: Isn't terribly useful to set it to `off`, just to override it in all locations where the field is being used, no?

However: I should be verified whether `one-time-code` should be set in authentication only or in both authentication and registration. The MDN definition of this value is no terribly clear in how it's meant to be used and what the User Agent should do when seeing `one-time-code`.

mutec: That's correct; but I don't have access to the other projects and didn't know why it was explicitly set to `off` before and where it might be used differently.

TimWolla: The field is meant to be used for the use within the TOTP method only. That's why it sits within the TOTP namespace. If anyone uses it for its unrelated purpose then they are deep within warranty void territory.

The `off` value was selected, because the web browser's autocompletion is not useful here. The `one-time-code` value is so underspecified that it was not clear whether it helps, whether it hurts or whether it even does anything at all for TOTP (vs. e.g. SMS codes).

TimWolla (approved):
LGTM from a purely technical PoV. This needs some tests within common configurations to ensure it does not break anything unrelated. I can't test this, as I don't put my TOTP secrets into a password manager.

joshuaruesweg:
The attribute is currently, as Tim mentioned in his review, too underspecified for us to want to use it. It causes more problems than it solves. 

Thanks for your pull request though!

PR #4365, opened 2021-07-04
Don't leak an account's activation status

SoftCreatR:
By entering a username and a random password, it is possible to get information, whether an account is activated, or not. While this isn't much of a security issue, I would authenticate the user first to ensure that the information is not disclosed to everyone.


wcfsetup/install/files/lib/form/RegisterNewActivationCodeForm.class.php
```diff
@@ -71,6 +71,9 @@ public function validate() {
 		// password
 		$this->validatePassword();
 		
+		// activation state
+		validateActivationState();
```
TimWolla, line 75:
```suggestion
		$this->validateActivationState();
```

joshuaruesweg (approved):

joshuaruesweg:
Thanks!

PR #4366, opened 2021-07-04
Fixed small typo

SoftCreatR:
This typo also exists in older versions.


TimWolla:
> This typo also exists in older versions.

Applied the changes manually, because it was impossible to cleanly apply this patch anywhere but 5.4. Thanks.

PR #4369, opened 2021-07-05
Update composer dependencies

TimWolla:


Gravatronics:
Remember to update `fileDelete.xml`.

Issue #4304, opened 2021-06-17
Remove package compatibility/API versions


Gravatronics:
See 0e8af3effdbb51d4b1901c63539ae539adb59135

PR #4371, opened 2021-07-05
Do not store API compatibility information within the database

TimWolla:
These are automatically converted into an exclude of Core 6.0.0 Alpha 1 in
PackageArchive and thus redundant.

### DIFF
--- a/wcfsetup/install/files/acp/templates/package.tpl
+++ b/wcfsetup/install/files/acp/templates/package.tpl
@@ -81,18 +81,6 @@
 					<dd><a href="https://pluginstore.woltlab.com/file/{$pluginStoreFileID}/" class="externalURL">{lang}wcf.acp.pluginStore.file.link{/lang}</a></dd>
 				</dl>
 			{/if}
-			{if $package->packageID != 1}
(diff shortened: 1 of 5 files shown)


Gravatronics (commented):
Should we then also

1. delete the existing entries from this database table during the update to have consistency between different scenarios on when a package was installed (before/after update to 5.5)?
2. drop https://github.com/WoltLab/WCF/blob/be96b55eea61c55177878d7454e396f634169773/wcfsetup/install/files/lib/system/package/PackageInstallationDispatcher.class.php#L554-L558?
3. mark the database as deprecated in the SQL file and the documentation?

Can we do the same for `wcf1_package_update_compatibility`?

TimWolla:
> Should we then also

I guess 1-3 all makes sense.

> Can we do the same for `wcf1_package_update_compatibility`?

I'd do this in a separate PR, because the package servers often behave differently.

Gravatronics:
Also see #4304

dtdesign (approved):

Gravatronics (approved):

PR #4372, opened 2021-07-06
Support sorting users in `UserOnlineListBoxController`

### LINKED ISSUE #4182: The "Users Online" box does not support sort options
The list will always default for an alphabetic sorting by username, but does not allow custom criteria, such as by "Last activity".

Gravatronics:
Close #4182

### DIFF
--- a/wcfsetup/install/files/lib/system/box/UserOnlineListBoxController.class.php
+++ b/wcfsetup/install/files/lib/system/box/UserOnlineListBoxController.class.php
@@ -31,16 +31,36 @@ class UserOnlineListBoxController extends AbstractDatabaseObjectListBoxControlle
     protected $conditionDefinition = 'com.woltlab.wcf.box.userOnlineList.condition';
 
     /**
+     * @inheritDoc
+     */
+    public $defaultSortField = USERS_ONLINE_DEFAULT_SORT_FIELD;
+
+    /**
+     * @inheritDoc
+     */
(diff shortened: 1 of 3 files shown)


wcfsetup/install/files/lib/system/box/UserOnlineListBoxController.class.php
```diff
@@ -65,6 +85,12 @@ protected function readObjects()
     {
         EventHandler::getInstance()->fireAction($this, 'readObjects');
 
+        if ($this->sortOrder && $this->sortField === 'lastActivityTime') {
```
TimWolla, line 88: I think it would be useful to compare the sortfield using `in_array` against an 1-element array or to convert this into a `switch()`, to allow for cleaner extensions in case other fields are of interest.

In fact: Don't you also need to check for `requestURI` here?

Gravatronics: Changes to this code are unlikely as it is basically the same as in the controller, see https://github.com/WoltLab/WCF/blob/be96b55eea61c55177878d7454e396f634169773/wcfsetup/install/files/lib/page/UsersOnlineListPage.class.php#L143.

TimWolla: Oh, the `lastActivityTime` needs special-casing because it exists in both the user as well as the session table. This is dumb.

TimWolla (approved):

PR #4381, opened 2021-07-07
Remove gzopen64() workaround

TimWolla:
PHP bug #53829 (https://bugs.php.net/bug.php?id=53829) is fixed. This should no
longer be required.


Gravatronics (approved):
Diff LGTM.

dtdesign (approved):

PR #4385, opened 2021-07-08
Do not store package compatibility from update servers within the database

TimWolla:
These do not appear to be used anywhere and the official package servers do not
expose them either, making them effectively empty even in current versions.

see #4371

### DIFF
--- a/wcfsetup/install/files/lib/system/package/PackageUpdateDispatcher.class.php
+++ b/wcfsetup/install/files/lib/system/package/PackageUpdateDispatcher.class.php
@@ -524,7 +524,7 @@ class PackageUpdateDispatcher extends SingletonFactory
      */
     protected function savePackageUpdates(array &$allNewPackages, $packageUpdateServerID)
     {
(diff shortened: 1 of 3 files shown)


dtdesign (approved):

wcfsetup/install/files/lib/system/package/PackageUpdateDispatcher.class.php
```diff
@@ -630,45 +630,35 @@ protected function savePackageUpdates(array &$allNewPackages, $packageUpdateServ
...
                 }
 
-                // @deprecated 5.2
-                if (isset($versionData['compatibility'])) {
-                    foreach ($versionData['compatibility'] as $version) {
-                        $compatibilityInserts[] = [
-                            'packageUpdateVersionID' => $packageUpdateVersionID,
-                            'version' => $version,
-                        ];
+                // The API compatibility versions are deprecated, any package that exposes them must
```
Gravatronics, line 633: I would keep a deprecation comment here because at one point in the future, all compatibility code can be removed.

Gravatronics (approved):

PR #4389, opened 2021-07-09
Reduce the size of the Mailbox objects

### LINKED ISSUE #4387: Store the languageID instead of the Language in Mailbox
This reduces the size of the serialized object.

- [ ] Language -> languageID
- [ ] User -> userID (UserMailbox)

TimWolla:
This introduces a small BC break for classes inheriting from Mailbox or
UserMailbox that directly access the `language` or `user` properties instead of
the getter methods. But this is easily fixed.

Resolves #4387

### DIFF
--- a/wcfsetup/install/files/lib/system/email/Mailbox.class.php
+++ b/wcfsetup/install/files/lib/system/email/Mailbox.class.php
@@ -32,9 +32,9 @@ class Mailbox
 
     /**
      * The preferred language of this mailbox.
-     * @var Language
+     * @var int
      */
-    protected $language;
+    protected $languageID;
 
     /**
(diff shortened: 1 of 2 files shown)


Gravatronics:
Can you avoid the BC issue with `__sleep()` and `__wakeup()`?

TimWolla:
That technically still would introduce a BC issue, because an adjusted `Language` in the downstream class will not round-trip the background queue, while it previously did. This is more silent and subtle, so I preferred the “loud” version that will explode due to missing objects.

Gravatronics (approved):
The changes themselves LGTM.

dtdesign (approved):

PR #4393, opened 2021-07-11
add missing phpDoc of `AbstractObjectTypeProcessor::$object`

mutec:


TimWolla (commented):
What does this give you that `getDecoratedObject()` doesn't? I believe the `@property` documentation implies that the property is public, when in fact it is not.

mutec:
I just found some places in a package where `$object` was accessed directly. Since the property should be set during the object setup and is not private, there should be no reason not to use it directly.

PR #4397, opened 2021-07-12
Improve Favicon handling in StyleAction

TimWolla:
- Delete manifest.json and browserconfig.xml when deleting a favicon
- Regenerate Favicon images only when the template changes


joshuaruesweg (approved):
Diff LGTM

PR #4402, opened 2021-07-14
Deprecate `WCF.Search.Message.KeywordList`

TimWolla:
This ideally would have happened a long time ago, as the base class already is
deprecated.


dtdesign (approved):

dtdesign:
I would be fine deprecating this in 5.4 already.

TimWolla:
The migration docs for 5.5 already have the “Deprecation and Removal” list (whereas 5.4 does not), so I'd rather put it there: https://docs.woltlab.com/5.5/migration/wsc54/deprecations_removals/#javascript

PR #4407, opened 2021-07-19
fix missing brackets and make sure languageID is valid in `I18nDatabaseObjectList`

mutec:
Missing brackets might lead to strange behavior.
I experienced some issues with an empty `languageID` which lead to an syntax error.

### DIFF
--- a/wcfsetup/install/files/lib/data/I18nDatabaseObjectList.class.php
+++ b/wcfsetup/install/files/lib/data/I18nDatabaseObjectList.class.php
@@ -39,7 +39,7 @@ abstract class I18nDatabaseObjectList extends DatabaseObjectList
         parent::__construct();
 
         if ($languageID === null) {
-            $languageID = WCF::getUser()->languageID;
+            $languageID = WCF::getLanguage()->languageID;
         }
 
         if (!empty($this->i18nFields)) {
--- a/wcfsetup/install/files/lib/data/I18nDatabaseObjectList.class.php
+++ b/wcfsetup/install/files/lib/data/I18nDatabaseObjectList.class.php
@@ -57,8 +57,8 @@ abstract class I18nDatabaseObjectList extends DatabaseObjectList
                 $this->sqlSelects .= (!empty($this->sqlSelects) ? ', ' : '') . "COALESCE(" . $matchTable . ".languageItemValue, " . $this->getDatabaseTableAlias() . "." . $key . ") AS " . $value;
                 $this->sqlJoins .= "
                     LEFT JOIN   wcf" . WCF_N . "_language_item " . $matchTable . "
-                    ON          " . $matchTable . ".languageItem = " . $this->getDatabaseTableAlias() . "." . $key . "
-                            AND " . $matchTable . ".languageID = " . $languageID;
+                    ON (        " . $matchTable . ".languageItem = " . $this->getDatabaseTableAlias() . "." . $key . "
+                            AND " . $matchTable . ".languageID = " . $languageID . ")";
             }
         }
     }


wcfsetup/install/files/lib/data/I18nDatabaseObjectList.class.php
```diff
@@ -55,10 +58,7 @@ public function __construct($languageID = null)
                 $matchTable = 'i18n_' . StringUtil::getHash($key);
 
                 $this->sqlSelects .= (!empty($this->sqlSelects) ? ', ' : '') . "COALESCE(" . $matchTable . ".languageItemValue, " . $this->getDatabaseTableAlias() . "." . $key . ") AS " . $value;
-                $this->sqlJoins .= "
-                    LEFT JOIN   wcf" . WCF_N . "_language_item " . $matchTable . "
-                    ON          " . $matchTable . ".languageItem = " . $this->getDatabaseTableAlias() . "." . $key . "
-                            AND " . $matchTable . ".languageID = " . $languageID;
+                $this->sqlJoins .= "LEFT JOIN wcf" . WCF_N . "_language_item " . $matchTable . " ON (" . $matchTable . ".languageItem = " . $this->getDatabaseTableAlias() . "." . $key . " AND " . $matchTable . ".languageID = " . $languageID . ")";
```
joshuaruesweg, line 61: Please keep the original indentation.

mutec: Sure, if you prefer that chaos…

wcfsetup/install/files/lib/data/I18nDatabaseObjectList.class.php
```diff
@@ -41,6 +41,9 @@ public function __construct($languageID = null)
         if ($languageID === null) {
             $languageID = WCF::getUser()->languageID;
         }
```
TimWolla, line 43: I wonder if this fall back should be removed. Either one wants an explicit language selection or one wants the effective current language. I can't imagine a case where the users configured language might be preferred over the language that is currently being in use.

TimWolla (changes requested):
It's also not clear to me why the parentheses are required here. Can you explain?

TimWolla:
Oh and please don't bundle two unrelated changes into a single commit.

wcfsetup/install/files/lib/data/I18nDatabaseObjectList.class.php
mutec: The problem was that the user has no languageID set which resulted in invalid SQL. Using this fallback it was fixed.
And I don't want to use a language explicitly every time. Why should I do that by default? By default I want to use the user's or session's language.

TimWolla: > The problem was that the user has no languageID set which resulted in invalid SQL.

Yes, I understand that.

> Using this fallback it was fixed.

I specifically commented on lines 41 to 43 (i.e. the `WCF::getUser()->languageID`) part. My understanding is that this value is never the correct value. Either I want an explicit language or the language that's currently in use by the language subsystem (i.e. the lines you added).

TimWolla:
@mutec Do you plan to make the requested changes or should we carry this across the finish line?

wcfsetup/install/files/lib/data/I18nDatabaseObjectList.class.php
```diff
@@ -57,8 +60,8 @@ public function __construct($languageID = null)
                 $this->sqlSelects .= (!empty($this->sqlSelects) ? ', ' : '') . "COALESCE(" . $matchTable . ".languageItemValue, " . $this->getDatabaseTableAlias() . "." . $key . ") AS " . $value;
                 $this->sqlJoins .= "
                     LEFT JOIN   wcf" . WCF_N . "_language_item " . $matchTable . "
-                    ON          " . $matchTable . ".languageItem = " . $this->getDatabaseTableAlias() . "." . $key . "
-                            AND " . $matchTable . ".languageID = " . $languageID;
+                    (ON         " . $matchTable . ".languageItem = " . $this->getDatabaseTableAlias() . "." . $key . "
```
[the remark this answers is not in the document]
mutec: Do you want it on the first query-line?

mutec:
@TimWolla I just came home yesterday. ;)
I tried to split things up and pushed it again. You might want to have a look.

I don't have the specific exception anymore, but it was an error during the overhaul of the shop package. I don't remember where and under which circumstances since I cleaned up some weeks ago.

wcfsetup/install/files/lib/data/I18nDatabaseObjectList.class.php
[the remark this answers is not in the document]
TimWolla: I'd like it behind the `ON`, because it's a syntax error otherwise. I'll put a suggestion.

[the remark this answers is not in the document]
mutec: Ah, of course; just had a look on my local fix. :D

TimWolla:
> I don't have the specific exception anymore, but it was an error during the overhaul of the shop package. I don't remember where and under which circumstances since I cleaned up some weeks ago.

I just scanned the full codebase for `ON\s*\(` and was not able to find a single occurrence. I assume that your custom code extended `->sqlJoins` and did not add whitespace at the start, thus generating invalid SQL which was then fixed by placing the parentheses.

Unless you can demonstrate that the missing parentheses cause actual issues I'm not taking that one, because that would result in inconsistent code.

I just picked the other commit manually into 5.4 to safe you from another rebase + force push. Thanks!

mutec:
I'll keep my eyes open. :)

PR #4411, opened 2021-07-20
Fall-back to GD if ImageMagick does not satisfy the requirements

dtdesign:
n/t

### DIFF
--- a/wcfsetup/test.php
+++ b/wcfsetup/test.php
@@ -347,18 +347,16 @@ function checkResult() {
 	foreach ($requiredExtensions as $extension) {
 		if (!extension_loaded($extension)) return false;
 	}
-	
-	if (!extension_loaded('gd') && !extension_loaded('imagick')) return false;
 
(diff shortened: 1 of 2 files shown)


dtdesign:
I recommend using the "Split" view to compare the changes, because some of the logic was inversed.

wcfsetup/test.php
```diff
@@ -347,18 +347,16 @@ function checkResult() {
...
 	}
 
-	if (extension_loaded('gd') && empty(\gd_info()['WebP Support'])) {
-		return false;
+	if (extension_loaded('gd') && !empty(\gd_info()['WebP Support'])) {
+		return true;
 	}
 	
-	return true;
+	return false;
```
TimWolla, line 359: I don't like the logic inversion here, because one can't easily add additional conditions at the bottom, as a successful check for the image processing library will return true.

I guess adding a `$imageProcessingWorks = true;` in place of the `return true` does the job most cleanly.

TimWolla (changes requested):
Other than that one remark this change LGTM.

TimWolla (approved):

PR #4415, opened 2021-07-22
Help TSC understand the data flow in Dropdown/Simple#toggle

TimWolla:
Newer TSC versions determine that `firstListItem` must always be `null`, thus
resulting in `firstListItem` being of type `never` within the `if
(firstListItem !== null)` condition, ultimately complaining about calling
`.focus()` on a `never` type:

> Property 'focus' does not exist on type 'never'.

The issue lies in the TypeScript compiler not properly understanding the
semantics of the `.forEach()`, assuming it might not synchronously call the
callback function.

Rewrite this logic to support TSC in understanding the code and to also make
the intent clearer to the human reader by adding additional helper variables.

### DIFF
--- a/ts/WoltLabSuite/Core/Ui/Dropdown/Simple.ts
+++ b/ts/WoltLabSuite/Core/Ui/Dropdown/Simple.ts
@@ -176,7 +176,6 @@ function toggle(
   _activeTargetId = "";
   _dropdowns.forEach((dropdown, containerId) => {
     const menu = _menus.get(containerId) as HTMLElement;
-    let firstListItem: HTMLLIElement | null = null;
 
     if (dropdown.classList.contains("dropdownOpen")) {
       if (!preventToggle) {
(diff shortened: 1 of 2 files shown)


ts/WoltLabSuite/Core/Ui/Dropdown/Simple.ts
```diff
@@ -222,16 +221,19 @@ function toggle(
...
         menu.removeEventListener("keydown", dropdownMenuKeyDown);
         menu.addEventListener("keydown", dropdownMenuKeyDown);
-        menu.querySelectorAll("li").forEach((listItem) => {
-          if (!listItem.clientHeight) return;
-          if (firstListItem === null) firstListItem = listItem;
-          else if (listItem.classList.contains("active")) firstListItem = listItem;
 
+        const nonEmptyListItems = Array.from(menu.querySelectorAll("li")).filter(
+          (listItem) => listItem.clientHeight > 0,
+        );
```
dtdesign, line 233:
```suggestion
        const nonEmptyListItems = Array.from(menu.querySelectorAll("li"))
          .filter((listItem) => listItem.clientHeight > 0);
```

I find this much easier to read, because the `filter()` in the front of the line adds context. Also I'm sort of being used to have chained calls being put into a single line every time.

dtdesign (commented):
LGTM, although I would format it slightly differently. Since this is just an opinion, I leave it up to you to decide.

ts/WoltLabSuite/Core/Ui/Dropdown/Simple.ts
TimWolla: This is what came out of Prettier. I'll see if I can convince it to accept your suggestion, otherwise I'm going to merge.

TimWolla: Nope, it does not. After I manually adjust this to your suggestion re-running prettier will move this back to how it was.

dtdesign: Okay, I'm fine with that.

PR #4425, opened 2021-07-28
Foreach tag unset variables that have the value NULL

BurntimeX:
ref https://community.woltlab.com/thread/291560-5-4-kalender-fehlermeldung/

### DIFF
--- a/wcfsetup/install/files/lib/system/template/TemplateScriptingCompiler.class.php
+++ b/wcfsetup/install/files/lib/system/template/TemplateScriptingCompiler.class.php
@@ -819,17 +819,13 @@ class TemplateScriptingCompiler {
 		$foreachData['itemVar'] = $itemVar;
 		
 		$phpCode .= "\$this->foreachVars['{$hash}'] = [];\n";
-		$phpCode .= "if (isset({$itemVar})) {\n";
-		$phpCode .= "\$this->foreachVars['{$hash}']['item'] = {$itemVar};\n";
-		$phpCode .= "}\n";
+		$phpCode .= "\$this->foreachVars['{$hash}']['item'] = {$itemVar} ?? null;\n";
 		
 		if (isset($args['key'])) {
 			$keyVar = mb_substr($args['key'], 0, 1) != '$' ? "\$this->v[".$args['key']."]" : $args['key'];
 			$foreachData['keyVar'] = $keyVar;
 			
-			$phpCode .= "if (isset({$keyVar})) {\n";
-			$phpCode .= "\$this->foreachVars['{$hash}']['key'] = {$keyVar};\n";
-			$phpCode .= "}\n";
+			$phpCode .= "\$this->foreachVars['{$hash}']['key'] = {$keyVar} ?? null;\n";
 			
 			$phpCode .= "foreach (".$foreachHash." as {$keyVar} => {$itemVar}) {\n";
 		}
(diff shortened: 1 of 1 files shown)


dtdesign:
This is quite a tricky change with some back-story. The `{foreach}` loop had the side effect of possibly overwriting variables that have existed before, because key and item are not scoped. For example, consider a page controller that sets `$category` and somewhere a random box uses a loop with the same index variable.

This was originally addressed by explicitly unsetting the loop variables to prevent them from leak their values. Also we had to check if the loop variables existed before in order to preserve and later restore their value.

However, PHP's `isset` has the annoying behavior that existing variables/array keys that evaluate to `null` are considered `false`/"not set". This is an issue when a variable was defined, but set to `null`, since it will then neither be captured nor restored.

Our solution was to use `array_key_exists()`, but since this requires different parameters, we ran into issues when the key or item was not a plain string, but a variable itself, e.g. `item=foo` vs `item=$foo`.

The most reliable fix was to use the null coalescing operator, because it allows us to safely attempt to read from an uninitialized variable by falling back to `null`. This has the side effect that the key and item loop (to be precise, they key is only set if it was provided because it is optional) variables will now always be defined as `null` if they did not exist before. This is a tradeoff that we feel comfortable with to finally resolve this issues.

wcfsetup/install/files/lib/system/template/TemplateScriptingCompiler.class.php
```diff
@@ -819,17 +819,13 @@ protected function compileForeachTag($foreachTag) {
 		$foreachData['itemVar'] = $itemVar;
 		
 		$phpCode .= "\$this->foreachVars['{$hash}'] = [];\n";
-		$phpCode .= "if (isset({$itemVar})) {\n";
-		$phpCode .= "\$this->foreachVars['{$hash}']['item'] = {$itemVar};\n";
-		$phpCode .= "}\n";
+		$phpCode .= "\$this->foreachVars['{$hash}']['item'] = {$itemVar} ?? null;\n";
```
TimWolla, line 822: This will introduce a behavioral change, because `item` will now always be filled. Thus this will "leak" variables containing `null` after the loop exits.

I suggest to simply replace the `isset()` with `array_key_exists` in all the modified locations.

BurntimeX: `array_key_exists` doesn't work here, because `$args['item']` can be a variable (e.g. `{foreach item=$foo}`).

dtdesign: > Our solution was to use `array_key_exists()`, but since this requires different parameters, we ran into issues when the key or item was not a plain string, but a variable itself, e.g. `item=foo` vs `item=$foo`.

I have also explained that the "leaking" as `null` is an acceptable trade-off.

TimWolla: Ah, sorry. I did not read the explanation once more after looking at it this morning and forgot about those paragraphs :sweat_smile:

TimWolla (approved):

wcfsetup/install/files/lib/system/template/TemplateScriptingCompiler.class.php
```diff
@@ -862,13 +858,11 @@ protected function compileForeachEndTag() {
 		// unset `item` and `key` variables and restore their sandboxed values
 		$phpCode = "<?php }\n";
 		$phpCode .= "unset({$foreachData['itemVar']});";
```
TimWolla, line 860: This `unset()` technically is useless now.

PR #4426, opened 2021-07-30
Properly show the label form field's selection if it has a value upon load

TimWolla:
This got broken during the TypeScript migration, because the constructor's
`labelId` parameter is typed as string, but in reality it was passed an
integer. This broke a `===` comparison with the `data-label-id` property of the
list items, thus failing to properly select the label.

### DIFF
--- a/com.woltlab.wcf/templates/__labelFormField.tpl
+++ b/com.woltlab.wcf/templates/__labelFormField.tpl
@@ -32,7 +32,7 @@
 		
 		new FormBuilderFieldLabel(
 			'{@$field->getPrefixedId()}',
-			{if $field->getValue()}{@$field->getValue()}{else}null{/if},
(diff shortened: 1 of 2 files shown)


dtdesign (commented):
Are you missing `|encodeJS` here?

TimWolla:
This is guaranteed to be an integer, but I guess an `|encodeJS` does not hurt.

dtdesign:
Oh, I see. In that case, the PR is okay the way it is, I missed the point about the int.

TimWolla:
> Oh, I see. In that case, the PR is okay the way it is, I missed the point about the `int`.

I adjusted it nonetheless. The fact that you needed to ask makes this non-obvious and in this case it's nothing performance critical, making the tiny performance overhead acceptable.

PR #4427, opened 2021-07-30
Add missing check for dependencies in LabelFormField processor

TimWolla:


dtdesign (approved):
Why did you remove the local `$value` variable, does `$this->checkDependencies()` cause any side-effects that alter the value?

TimWolla:
> Why did you remove the local `$value` variable, does `$this->checkDependencies()` cause any side-effects that alter the value?

It does not, but I wanted to match the other field types with a custom processor as closely as possible:

https://github.com/WoltLab/WCF/blob/ba04396a4059989da14c7edec402038b53f27ed0/wcfsetup/install/files/lib/system/form/builder/field/tag/TagFormField.class.php#L123-L125

https://github.com/WoltLab/WCF/blob/c29b75e0e43c5680f7d4182325b5958902997661/wcfsetup/install/files/lib/system/form/builder/field/acl/simple/SimpleAclFormField.class.php#L64-L66

PR #4432, opened 2021-08-03
Do not store optional information from update servers within the database

TimWolla:
These do not appear to be used anywhere and the official package servers do not
expose them either, making them effectively empty even in current versions.

see fd3511448663b86f4428e848bbefa07b93bf4bad
see #4371
see #4385


TimWolla:
Build failures are unrelated to the changes in this PR, but instead caused by the new PHP CS Fixer release: https://github.com/FriendsOfPHP/PHP-CS-Fixer/releases/tag/v3.0.1

I'll check whether I can find a 3.0.1 to run locally.

dtdesign (approved):

PR #4438, opened 2021-08-05
Fix having an incorrect parameter if a hash does not contains a salt

### LINKED ISSUE #4416: Correctly handle missing colons in password hashes
Some very old imported password hashes might not contain the colon that separated the hash from the salt, leading to exceptions upon login. This should be handle correctly.

joshuaruesweg:
Fixes #4416


TimWolla (approved):

TimWolla:
Adjusted the target branch to 5.4.

PR #4440, opened 2021-08-06
Add proxy_sourcemap.php

TimWolla:


dtdesign (approved):

wcfsetup/install/files/proxy_sourcemap.php
```diff
@@ -0,0 +1,230 @@
...
+    $mapOrErrorResponse = getMapFromQuery($request->getUri()->getQuery());
+
+    if ($mapOrErrorResponse instanceof ResponseInterface) {
+        return $mapOrErrorResponse;
+    }
+
+    $map = $mapOrErrorResponse;
+
+    $cacheFilename = \sprintf(
+        '%s/tmp/sourcemap_%s.map',
```
jens1o, line 124: Why did you opted to use `tmp/` and do not use `FileUtil::getTempFolder()` (which would probably make this more error resilient)? :)

TimWolla: Because this script very intentionally does not use anything from within WoltLab Suite and `tmp/` within WoltLab Suite Core's directory effectively always exists.

PR #4442, opened 2021-08-06
Prefixed CSS First

SoftCreatR:
Ensure vendor-prefixed versions of a CSS property are listed before the standardized, unprefixed version.

When multiple versions of the same CSS property are specified, the last supported one will be used due to how browsers handle fallback values. This means the order matters when using both vendor-prefixed and unprefixed versions of the same property. Specifically, the unprefixed version must be listed last to ensure standardized behavior takes precedence.


dtdesign:
Thanks!

PR #4447, opened 2021-08-09
Use PSR-7 responses in additional actions

TimWolla:
Blocked on #4436

### DIFF
--- a/wcfsetup/install/files/lib/action/AJAXUploadAction.class.php
+++ b/wcfsetup/install/files/lib/action/AJAXUploadAction.class.php
@@ -2,6 +2,7 @@
 
 namespace wcf\action;
 
+use Laminas\Diactoros\Response\JsonResponse;
 use wcf\system\upload\UploadHandler;
 use wcf\util\JSON;
 
@@ -34,10 +35,10 @@ class AJAXUploadAction extends AJAXProxyAction
             parent::sendResponse();
         }
 
+        $response = new JsonResponse($this->response);
         // IE9 is mad if iframe response is application/json
-        \header('Content-type: text/plain');
-        echo JSON::encode($this->response);
+        $response = $response->withHeader('content-type', 'text/plain');
 
-        exit;
+        return $response;
     }
 }
--- a/wcfsetup/install/files/lib/action/FacebookAuthAction.class.php
+++ b/wcfsetup/install/files/lib/action/FacebookAuthAction.class.php
@@ -5,6 +5,7 @@ namespace wcf\action;
 use GuzzleHttp\Psr7\Request;
 use Laminas\Diactoros\Response\RedirectResponse;
 use wcf\data\user\User;
+use wcf\form\AccountManagementForm;
 use wcf\form\RegisterForm;
 use wcf\system\event\EventHandler;
 use wcf\system\exception\NamedUserException;
@@ -155,7 +156,11 @@ final class FacebookAuthAction extends AbstractOauth2Action
                 WCF::getSession()->register('__oauthUser', $oauthUser);
 
                 return new RedirectResponse(
-                    LinkHandler::getInstance()->getLink('AccountManagement') . '#3rdParty'
+                    LinkHandler::getInstance()->getControllerLink(
+                        AccountManagementForm::class,
+                        [],
+                        '#3rdParty'
+                    )
                 );
             } else {
                 // This account does not belong to anyone and we are not logged in.
--- a/wcfsetup/install/files/lib/action/FacebookAuthAction.class.php
+++ b/wcfsetup/install/files/lib/action/FacebookAuthAction.class.php
@@ -3,6 +3,7 @@
 namespace wcf\action;
 
 use GuzzleHttp\Psr7\Request;
+use Laminas\Diactoros\Response\RedirectResponse;
 use wcf\data\user\User;
 use wcf\form\RegisterForm;
 use wcf\system\event\EventHandler;
@@ -11,7 +12,6 @@ use wcf\system\request\LinkHandler;
 use wcf\system\user\authentication\event\UserLoggedIn;
 use wcf\system\user\authentication\oauth\User as OauthUser;
 use wcf\system\WCF;
-use wcf\util\HeaderUtil;
 use wcf\util\JSON;
 use wcf\util\StringUtil;
 
@@ -141,9 +141,9 @@ final class FacebookAuthAction extends AbstractOauth2Action
                     new UserLoggedIn($user)
                 );
 
-                HeaderUtil::redirect(LinkHandler::getInstance()->getLink());
-
-                exit;
+                return new RedirectResponse(
+                    LinkHandler::getInstance()->getLink()
+                );
             }
         } else {
             WCF::getSession()->register('__3rdPartyProvider', 'facebook');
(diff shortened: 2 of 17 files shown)


wcfsetup/install/files/lib/util/HeaderUtil.class.php
```diff
@@ -106,6 +107,28 @@ public static function sendNoCacheHeaders()
         @\header('Cache-Control: max-age=0, no-cache, no-store, must-revalidate');
     }
 
+    /**
+     * Returns an response matching the given response with headers preventing in-browser
+     * caching attached.
+     */
+    public static function withNoCacheHeaders(ResponseInterface $response): ResponseInterface
```
TimWolla, line 114: I wonder if this should be moved somewhere to `lib/system/io/http/`?

wcfsetup/install/files/lib/util/HeaderUtil.class.php
```diff
@@ -106,6 +107,28 @@ public static function sendNoCacheHeaders()
         @\header('Cache-Control: max-age=0, no-cache, no-store, must-revalidate');
     }
 
+    /**
+     * Returns an response matching the given response with headers preventing in-browser
+     * caching attached.
```
TimWolla, line 112: `@since`

wcfsetup/install/files/lib/action/AJAXUploadAction.class.php
```diff
@@ -34,10 +35,10 @@ protected function sendResponse()
             parent::sendResponse();
         }
 
+        $response = new JsonResponse($this->response);
         // IE9 is mad if iframe response is application/json
-        \header('Content-type: text/plain');
-        echo JSON::encode($this->response);
+        $response = $response->withHeader('content-type', 'text/plain');
```
dtdesign, line 40: Please just remove this non-sense, it is an old legacy hack for IE9 :D

wcfsetup/install/files/lib/action/FacebookAuthAction.class.php
```diff
@@ -154,9 +154,9 @@ protected function processUser(OauthUser $oauthUser)
 
                 WCF::getSession()->register('__oauthUser', $oauthUser);
 
-                HeaderUtil::redirect(LinkHandler::getInstance()->getLink('AccountManagement') . '#3rdParty');
-
-                exit;
+                return new RedirectResponse(
+                    LinkHandler::getInstance()->getLink('AccountManagement') . '#3rdParty'
```
dtdesign, line 158: Can you replace this one with `getControllerLink()`?

wcfsetup/install/files/lib/action/GithubAuthAction.class.php
```diff
@@ -144,9 +144,9 @@ protected function processUser(OauthUser $oauthUser)
 
                 WCF::getSession()->register('__oauthUser', $oauthUser);
 
-                HeaderUtil::redirect(LinkHandler::getInstance()->getLink('AccountManagement') . '#3rdParty');
-
-                exit;
+                return new RedirectResponse(
+                    LinkHandler::getInstance()->getLink('AccountManagement') . '#3rdParty'
```
dtdesign, line 148: Can you replace this one with `getControllerLink()`?

wcfsetup/install/files/lib/action/GoogleAuthAction.class.php
```diff
@@ -166,9 +166,9 @@ protected function processUser(OauthUser $oauthUser)
 
                 WCF::getSession()->register('__oauthUser', $oauthUser);
 
-                HeaderUtil::redirect(LinkHandler::getInstance()->getLink('AccountManagement') . '#3rdParty');
-
-                exit;
+                return new RedirectResponse(
+                    LinkHandler::getInstance()->getLink('AccountManagement') . '#3rdParty'
```
dtdesign, line 170: Can you replace this one with `getControllerLink()`?

wcfsetup/install/files/lib/action/MessageQuoteAction.class.php
```diff
@@ -131,12 +132,10 @@ public function execute()
         WCF::getSession()->disableUpdate();
 
         if ($returnValues !== null) {
-            // send JSON-encoded response
-            \header('Content-type: application/json; charset=UTF-8');
-            echo JSON::encode($returnValues);
+            return new JsonResponse($returnValues);
+        } else {
+            return new EmptyResponse(200);
```
dtdesign, line 137: You can get rid of the `else` here.

wcfsetup/install/files/lib/util/HeaderUtil.class.php
dtdesign: This would only make sense if you're looking forward to phase out this class?

dtdesign (changes requested):
LGTM

wcfsetup/install/files/lib/action/AJAXUploadAction.class.php
TimWolla: As explained in [the commit message](https://github.com/WoltLab/WCF/pull/4447/commits/4dfe137efc1fb1dfea1ae435335749218d13c2cb):

> The IE 9 fallback technically should no longer be required, but the (legacy)
> JavaScript still references the `isFallback` parameter.

I'd rather clean this up once the JavaScript is cleaned up as well to not mix 2 unrelated things.

wcfsetup/install/files/lib/util/HeaderUtil.class.php
TimWolla: Looking at the methods in `HeaderUtil`:

- `setCookie`: This is one of the issues that should be solved with PSR-7. Will need to be adapted, might be replaced by: https://github.com/dflydev/dflydev-fig-cookies
- `getCookieDomain`: Might be simplified once multi-domain setups are gone?
- `sendHeaders`: Will become obsolete with PSR-7 and is then better solved using a middleware.
- `sendNoCacheHeaders`: This is replaced by the method we are currently discussing.
- `parseOutput`: Is intimately tied to `sendHeaders`.
- `redirect`: Will be replaced by `RedirectResponse`.
- `delayedRedirect`: Should probably be replaced by a specific `DelayedRedirectResponse`.

Basically all of the methods in `HeaderUtil` will become obsolete with the migration to PSR-7. Thus why I want to discuss whether re-architecting this might make sense. For Guzzle we already have the RedirectGuard as a single-purpose class:

https://github.com/WoltLab/WCF/blob/master/wcfsetup/install/files/lib/system/io/http/RedirectGuard.class.php

wcfsetup/install/files/lib/action/MessageQuoteAction.class.php
TimWolla: In this case it's intentional to make it more clear that these are the only 2 cases. An early return IMO would not be as fitting.

The `EmptyResponse` case can also be removed once the `markForRemoval` action is removed (I deprecated it in a previous PR).

wcfsetup/install/files/lib/util/HeaderUtil.class.php
dtdesign: I agree, spreading things out while still having a mix of everything certainly does not improve the situation.

wcfsetup/install/files/lib/action/AJAXUploadAction.class.php
dtdesign: OK, I was under the false impression that the fallback was removed a while ago, because it is effectively unsupported.

dtdesign (approved):

PR #4450, opened 2021-08-09
Rename `Ui/Object/Action/Toogle` to `Toggle`

joshuaruesweg:


wcfsetup/install/files/lib/acp/form/DevtoolsProjectEditForm.class.php
```diff
@@ -340,7 +340,7 @@ protected function setFormObjectData()
         $instructionsField->value($instructions);
     }
 
-    protected function tooglePackageXmlFieldAvailabilty($available)
```
joshuaruesweg, line 343: This bug is introduced with WoltLab Suite 3.1. Should we fix them only in `5.4` or should we backport them into `3.1` or should we ignore the typo here?

Issue #4273, opened 2021-06-09
Set `cache-control: private` for *all* responses that use a session


TimWolla:

TimWolla:
This is non-trivial, because PHP defaults to completely replacing headers whenever `header()` is being called. Thus we can't simply set `cache-control: private` in a central location, as it might be replaced at will. For now I've committed a change to 5.4 to at the very least set it on redirects which are already handled with a helper function.

TimWolla:
Linking to #4437, as that one kicks off the necessary changes for this to happen.

PR #4454, opened 2021-08-11
Add `cache-control: private` to PSR-7 responses

TimWolla:
While we cannot rely on every controller returning a PSR-7 response, we are
also unable to reliably set `cache-control: private` without these PSR-7
responses.

For this reason this change still provides a net benefit, as it improves the
situation in some cases, while not making it worse for other cases.

-----------------

see #4273

### DIFF
--- a/wcfsetup/install/files/lib/system/request/RequestHandler.class.php
+++ b/wcfsetup/install/files/lib/system/request/RequestHandler.class.php
@@ -158,7 +158,9 @@ class RequestHandler extends SingletonFactory
     {
         // Storing responses in a shared cache is unsafe, because they all contain session specific information.
         // Add the 'private' value to the cache-control header and remove any 'public' value.
-        $cacheControl = [];
+        $cacheControl = [
+            'private',
+        ];
         foreach ($this->splitCacheControl($response->getHeader('cache-control')) as $value) {
             [$field] = \explode('=', $value, 2);
 
@@ -174,7 +176,6 @@ class RequestHandler extends SingletonFactory
 
             $cacheControl[] = $value;
         }
-        $cacheControl[] = 'private';
 
         $response = $response->withHeader(
             'cache-control',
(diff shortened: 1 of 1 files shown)


wcfsetup/install/files/lib/system/request/RequestHandler.class.php
```diff
@@ -106,6 +105,88 @@ public function handle($application = 'wcf', $isACPRequest = false)
...
+    }
+
+    /**
+     * @since 5.5
+     */
+    private function sendPsr7Response(ResponseInterface $response)
+    {
+        // Storing responses in a shared cache is unsafe, because they all contain session specific information.
+        // Add the 'private' value to the cache-control header and remove any 'public' value.
+        $cacheControl = [];
```
dtdesign, line 161:
```suggestion
        $cacheControl = ['private'];
```

The `'private'` value is always set, moving it into the initialization makes more sense. It also provides a better explanation for line 165 which talks about the "duplication" of this value.

wcfsetup/install/files/lib/system/request/RequestHandler.class.php
```diff
@@ -106,6 +105,88 @@ public function handle($application = 'wcf', $isACPRequest = false)
...
+            }
+
+            // Drop the 'public' field.
+            if ($field === 'public') {
+                continue;
+            }
+
+            $cacheControl[] = $value;
+        }
+        $cacheControl[] = 'private';
```
dtdesign, line 177:
```suggestion
```

See line 161 above

dtdesign (changes requested):
LGTM

wcfsetup/install/files/lib/system/request/RequestHandler.class.php
TimWolla: Ah, indeed. The code initially used `array_filter`, until I added `splitCacheControl` (which is just a generator / iterator).

PR #4459, opened 2021-08-12
Unify error handling in several templates

SoftCreatR:
As of now, not every template allows setting custom error types for input fields, which is pretty inflexible. Trying it (without dirty JS hacks) results in a broken output:

image

Within WSC, this affects multiple templates.

/cc @TimWolla


TimWolla (approved):
Thanks!

PR #4466, opened 2021-08-19
fix missing property name for Popover's selector

mutec:


TimWolla (changes requested):
I don't see what this is fixing. This is legal code: https://alligator.io/js/object-property-shorthand-es6/

What am I missing?

TimWolla:
@mutec PTAL at my review. Can this PR be closed?

mutec:
@TimWolla might have been something else. It works at the moment. 👀

PR #4471, opened 2021-08-20
Fix validation of `UserFormField` if field is not required

Fabii547:


TimWolla:
Can you explain which case is being fixed by this PR? Is it submitting the form with an empty `UserFormField` or something else?

Fabii547:
I have a form where the field is not required and if nothing is entered the value is `null` when submitting the form.

TimWolla:
I am unable to reproduce this (using a current master). If I leave the field empty then I either get `null`, `0` or an empty array depending of the combination of `->multiple()` and `->nullable()` without any exceptions during validation.

Please provide an example form that reproduces the issue.

Fabii547:
The error is reproducible if the field is not included in the request (e.g. by removing the HTML of the field before submitting the form). In this case `$this->getValue()` returns `null` inside `UserFormField::validate()`.

In my case the error occurs because I use the form builder to build a filter which supports reading the form data also from the query string.

TimWolla:
> The error is reproducible if the field is not included in the request

Thanks for the clarification. I'll have a look.

TimWolla:
Fixed differently for 5.2 and completely cleaned up the validation for 5.4. Thank you!

PR #4477, opened 2021-08-25
Deprecate HTTP_SEND_X_FRAME_OPTIONS

### LINKED ISSUE #4474: Deprecate HTTP_SEND_X_FRAME_OPTIONS
Embedding in frames:

- Reduces the security (by forcing SameSite=none on cookies and allowing clickjacking).
- Imposes bad UX (by not showing the actual URL in the URL bar).
- Interoperates especially badly with Firefox' scoped cookie jars, because WoltLab Suite will see differing cookies depending on whether the site is embedded in a frame or not.

TimWolla:
Shall I also add a warning message on IndexPage? Just adjusting the field
description might not be sufficiently visible and removal of this option might
bring signifant breakage unless anticipated.

------------

Resolves #4474


dtdesign (approved):

dtdesign:
> Shall I also add a warning message on IndexPage?

Yes. I would like to see deprecated options that are subject to removal in the next version to show a warning on the index page. This should give operators enough time to plan ahead and adjust their setup.

PR #4486, opened 2021-08-30
use default captcha type when creating `CaptchaFormField`

mutec:
I assume by default one would want to use the captcha type selected by the administrator, so this might require one line less custom code than before.
When one needs a specific captcha type, he/she still can set it manually as done before.

### DIFF
--- a/wcfsetup/install/files/lib/system/form/builder/field/CaptchaFormField.class.php
+++ b/wcfsetup/install/files/lib/system/form/builder/field/CaptchaFormField.class.php
@@ -49,6 +49,14 @@ class CaptchaFormField extends AbstractFormField implements IObjectTypeFormNode
     /**
      * @inheritDoc
      */
+    public function __construct()
+    {
+        $this->objectType(CAPTCHA_TYPE);
+    }
+
+    /**
+     * @inheritDoc
+     */
     public function cleanup()
     {
         try {


TimWolla:
Did you test this? `TObjectTypeFormNode` uses write-once semantics: 

https://github.com/WoltLab/WCF/blob/2cec802c8b4c7ba3309431ca0af56ff2c806dd6f/wcfsetup/install/files/lib/system/form/builder/TObjectTypeFormNode.class.php#L54-L56

mutec:
> Did you test this? `TObjectTypeFormNode` uses write-once semantics:

Thanks for that catch! Would it be better to set the property `null` right before `defaultObjectType()` or override the whole method? I'd prefer the first one for future compatibility when `TObjectTypeFormNode::validate()` changes (in theory).

TimWolla:
I'm not sure if I want to lift this requirement. Changing the `->objectType()` while parts of the logic already acted on the previous object type can lead to broken/mismatched state. I assume that this is the reason behind that property being write-once. Special-casing the Captcha field doesn't feel right and only saves one from writing a single line of code.

`defaultObjectType()` is already in use, but perhaps that `use` could be made private and renamed and `defaultObjectType()` then be made a wrapper that at least avoids having to put `CAPTCHA_TYPE` there?

TimWolla:
> `defaultObjectType()` is already in use, but perhaps that `use` could be made private and renamed and `defaultObjectType()` then be made a wrapper that at least avoids having to put `CAPTCHA_TYPE` there?

@mutec Opinion?

Issue #2683, opened 2018-05-29
Toggle the list of items for display of notices


dtdesign:
The current implementation is a strict whitelist, where each new item must be explicitly checked in order to have a notice being displayed there. It would be much easier to implement a switch that negates this behavior (similar to the boxes) which turns it into a list of items where the notice should be _hidden_.

ref https://community.woltlab.com/thread/263630-usability-der-hinweis-seiten-kategorien-auswahl/
ref https://community.woltlab.com/thread/263631-wildcard-f%C3%BCr-seiten-kategorien-auswahl-bei-hinweisen/

PR #4489, opened 2021-08-31
Box Visibility Conditions

joshuaruesweg:
* [x] Add update script

### DIFF
--- a/wcfsetup/install/files/acp/update_com.woltlab.wcf_5.5_boxCondtions.php
+++ b/wcfsetup/install/files/acp/update_com.woltlab.wcf_5.5_boxCondtions.php
@@ -1,0 +1,52 @@
+<?php
+
+/**
+ * Creates for all boxes own conditions.
+ *
+ * @author Joshua Ruesweg
+ * @copyright 2001-2021 WoltLab GmbH
+ * @license GNU Lesser General Public License <http://opensource.org/licenses/lgpl-license.php>
+ * @package WoltLabSuite\Core
+ */
+
+use wcf\data\box\Box;
+use wcf\data\box\BoxList;
+use wcf\data\condition\Condition;
+use wcf\data\object\type\ObjectTypeCache;
+use wcf\system\condition\ConditionHandler;
+use wcf\system\condition\page\MultiPageCondition;
+
+$boxList = new BoxList();
+$boxList->readObjects();
+
+foreach ($boxList as $box) {
+    $conditions = ConditionHandler::getInstance()->getConditions(
+        Box::VISIBILITY_CONDITIONS_OBJECT_TYPE_NAME,
+        $box->boxID
+    );
+
+    if (!empty($conditions)) {
+        // The box already has conditions (maybe from a previous upgrade attempt).
+        // Skip this box to ensure, that the condition will not attached twice.
+        continue;
+    }
+
+    $pageCondition = ObjectTypeCache::getInstance()->getObjectTypeByName(
+        Box::VISIBILITY_CONDITIONS_OBJECT_TYPE_NAME,
+        'com.woltlab.wcf.page'
+    );
+
+    \assert($pageCondition->getProcessor() instanceof MultiPageCondition);
+
+    $pageCondition->getProcessor()->setData(new Condition(null, [
+        'conditionData' => \serialize([
+            'pageIDs' => $box->getPageIDs(),
+            'pageIDs_reverseLogic' => $box->visibleEverywhere,
+        ]),
+    ]));
+
+    ConditionHandler::getInstance()->createConditions(
+        $box->boxID,
+        [$pageCondition]
+    );
+}
--- a/wcfsetup/install/files/acp/update_com.woltlab.wcf_5.5_boxCondtions.php
+++ b/wcfsetup/install/files/acp/update_com.woltlab.wcf_5.5_boxCondtions.php
@@ -1,7 +1,7 @@
 <?php
 
 /**
- * Creates for all boxes own conditions.
+ * Create conditions for all boxes that use the legacy page filter.
  *
  * @author Joshua Ruesweg
  * @copyright 2001-2021 WoltLab GmbH
--- a/wcfsetup/install/files/lib/acp/form/BoxAddForm.class.php
+++ b/wcfsetup/install/files/lib/acp/form/BoxAddForm.class.php
@@ -405,38 +405,18 @@ class BoxAddForm extends AbstractForm
         );
 
         $this->pageIDs = $this->visibleEverywhere = null;
-        foreach ($this->groupedConditionObjectTypes as $groupedObjectTypes) {
-            foreach ($groupedObjectTypes as $objectTypes) {
-                if (\is_array($objectTypes)) {
-                    foreach ($objectTypes as $objectType) {
-                        $objectType->getProcessor()->readFormParameters();
-                        if ($objectType->objectTypeID == $pageConditionObjectTypeID) {
-                            \assert($objectTypes->getProcessor() instanceof MultiPageCondition);
+        foreach ($this->toFlatList($this->groupedConditionObjectTypes) as $objectType) {
+            $objectType->getProcessor()->readFormParameters();
+            if ($objectType->objectTypeID == $pageConditionObjectTypeID) {
+                \assert($objectType->getProcessor() instanceof MultiPageCondition);
 
-                            $data = $objectType->getProcessor()->getData();
-                            if ($data !== null) {
-                                $this->pageIDs = $data['pageIDs'];
-                                $this->visibleEverywhere = $data['pageIDs_reverseLogic'] ? 1 : 0;
-                            } else {
-                                $this->pageIDs = [];
-                                $this->visibleEverywhere = 1;
-                            }
-                        }
-                    }
+                $data = $objectType->getProcessor()->getData();
+                if ($data !== null) {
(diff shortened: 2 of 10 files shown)


wcfsetup/install/files/lib/acp/form/BoxAddForm.class.php
```diff
@@ -393,6 +393,37 @@ public function readFormParameters()
...
         }
+
+        $this->readConditions();
+    }
+
+    private function readConditions(): void
+    {
+        foreach ($this->groupedConditionObjectTypes as $groupedObjectTypes) {
+            foreach ($groupedObjectTypes as $objectTypes) {
+                if (\is_array($objectTypes)) {
```
TimWolla, line 404: In which cases is this not an array? Can you avoid the duplication by doing the following:

```
if (!\is_array($objectTypes)) {
  $objectTypes = [$objectTypes];
}
```
?

wcfsetup/install/files/lib/acp/form/BoxAddForm.class.php
```diff
@@ -532,6 +563,18 @@ public function validate()
             $statement->execute($conditionBuilder->getParameters());
             $this->pageIDs = $statement->fetchAll(\PDO::FETCH_COLUMN);
         }
+
+        foreach ($this->groupedConditionObjectTypes as $groupedObjectTypes) {
+            foreach ($groupedObjectTypes as $objectTypes) {
+                if (\is_array($objectTypes)) {
```
TimWolla, line 569: ditto, possibly elsewhere as well.

wcfsetup/install/files/lib/acp/form/BoxAddForm.class.php
```diff
@@ -393,6 +393,37 @@ public function readFormParameters()
...
+    }
+
+    private function readConditions(): void
+    {
+        foreach ($this->groupedConditionObjectTypes as $groupedObjectTypes) {
+            foreach ($groupedObjectTypes as $objectTypes) {
+                if (\is_array($objectTypes)) {
+                    foreach ($objectTypes as $objectType) {
+                        $objectType->getProcessor()->readFormParameters();
+                        if ($objectType->getProcessor() instanceof MultiPageCondition) {
```
TimWolla, line 407: It should not be possible that this occurs multiple times, but if it does for some reason, then the `pageIDs` will be overriden by an incorrect value, possibly resulting in an hard to find bug.

Can you add a safety check that ensures `pageIDs` will only ever be set once in this loop, throwing an exception if this is not true? You can use `\assert()`.

TimWolla, line 407: Should you compare against objectTypeID here, like you do in BoxHandler? You would not need the assertion then, because an ID obviously is unique.

wcfsetup/install/files/lib/data/box/Box.class.php
```diff
@@ -548,6 +561,32 @@ public function getConditions()
...
+    /**
+     * Returns the conditions for the visibility of the box.
+     *
+     * @return  Condition[]
+     * @since   5.5
+     */
+    public function getVisibilityConditions(): array
+    {
+        return ConditionHandler::getInstance()->getConditions(
+            'com.woltlab.wcf.condition.box', // @TODO make const
```
TimWolla, line 573: TODO

wcfsetup/install/files/lib/system/box/BoxHandler.class.php
```diff
@@ -185,6 +189,56 @@ public function addBoxToPageAssignments($boxIdentifier, array $pageIdentifiers,
                 $statement->execute([$box->boxID, $page->pageID, $visible ? 1 : 0]);
             }
         }
+
+        $conditionObjectTypeID = ObjectTypeCache::getInstance()->getObjectTypeIDByName(
+            'com.woltlab.wcf.condition.box',
+            'com.woltlab.wcf.page'
+        );
+        $oldCondition = [];
```
TimWolla, line 197: Why is this an array?

wcfsetup/install/files/lib/system/box/BoxHandler.class.php
```diff
@@ -185,6 +189,56 @@ public function addBoxToPageAssignments($boxIdentifier, array $pageIdentifiers,
...
+        foreach ($box->getVisibilityConditions() as $condition) {
+            if ($condition->objectTypeID === $conditionObjectTypeID) {
+                $oldCondition[] = $condition;
+                break;
+            }
+        }
+
```
TimWolla, line 204: Add a safety check that you indeed found a matching condition.

wcfsetup/install/files/lib/system/box/BoxHandler.class.php
```diff
@@ -185,6 +189,56 @@ public function addBoxToPageAssignments($boxIdentifier, array $pageIdentifiers,
...
+    private function createPageConditions(array $pages, bool $reverseLogic, Box $box): array
+    {
+        $pageCondition = ObjectTypeCache::getInstance()->getObjectTypeByName(
+            'com.woltlab.wcf.condition.box',
+            'com.woltlab.wcf.page'
+        );
+
+        $pageIDs = \array_merge($box->getPageIDs(), \array_map(static function ($page) {
+            return $page->pageID;
+        }, $pages));
```
TimWolla, line 228: You can use `\array_column($pages, 'pageID')` here.

wcfsetup/install/files/lib/system/box/BoxHandler.class.php
joshuaruesweg: Why? It is totally fine, if the array is empty.

wcfsetup/install/files/lib/system/box/BoxHandler.class.php
joshuaruesweg: The value can be empty, so that is the easiest solution if the default value is an array.

wcfsetup/install/files/lib/data/box/Box.class.php
```diff
@@ -530,11 +535,24 @@ public function getPageIDs()
     }
 
     /**
-     * Returns the conditions of the notice.
+     * Returns the conditions for the box controller.
      *
      * @return  Condition[]
+     * @deprecated since 5.5 - use self::getControllerConditions() instead
```
dtdesign, line 541:
```suggestion
     * @deprecated 5.5 - use self::getControllerConditions() instead
```

wcfsetup/install/files/lib/data/box/Box.class.php
```diff
@@ -530,11 +535,24 @@ public function getPageIDs()
...
+    }
+
+    /**
+     * Returns the conditions for the box controller.
+     * The conditions are intended for the contents of the box.
+     *
+     * @return  Condition[]
+     * @since   5.5
+     */
+    public function getControllerConditions()
```
dtdesign, line 555:
```suggestion
    public function getControllerConditions(): array
```

wcfsetup/install/files/lib/acp/form/BoxEditForm.class.php
```diff
@@ -239,6 +257,26 @@ public function readData()
...
 
             $this->readBoxImages();
+
+            $conditions = $this->box->getVisibilityConditions();
+            $conditionsByObjectTypeID = [];
+            foreach ($conditions as $condition) {
+                $conditionsByObjectTypeID[$condition->objectTypeID] = $condition;
+            }
+
+            foreach ($this->groupedConditionObjectTypes as $objectTypes1) {
```
dtdesign, line 267: This overlaps with the logic in line 170+ above. How about introducing a helper method to do this, it will also save you from the awkward names like `$objectTypes2`:
```php
private function toFlatList(array $groupedObjects): array { /* … */ }
```

dtdesign (changes requested):

wcfsetup/install/files/acp/update_com.woltlab.wcf_5.5_boxCondtions.php
```diff
@@ -0,0 +1,52 @@
...
+        Box::VISIBILITY_CONDITIONS_OBJECT_TYPE_NAME,
+        'com.woltlab.wcf.page'
+    );
+
+    \assert($pageCondition->getProcessor() instanceof MultiPageCondition);
+
+    $pageCondition->getProcessor()->setData(new Condition(null, [
+        'conditionData' => \serialize([
+            'pageIDs' => $box->getPageIDs(),
+            'pageIDs_reverseLogic' => $box->visibleEverywhere,
```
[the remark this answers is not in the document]
joshuaruesweg: Yes, but that terminus `pageIDs_reverseLogic`  is used since years for this condition (it is not introduced by me). I think it is not very helpful, to create breaking changes to change this, even if this may not be accessed by plugins. 

See https://github.com/WoltLab/WCF/issues/2683
See https://github.com/WoltLab/WCF/commit/988fc118d188eab587b5c3d13d026aa0c1e8b02f

wcfsetup/install/files/lib/acp/form/BoxEditForm.class.php
joshuaruesweg: As a note, this is fully c&p code. However, since you are the second to note this and find the code hard to read, I have adjusted this and introduced the appropriate helper method. It may make sense to make the method globally available in the future and to use it everywhere where the code is used.

dtdesign (approved):

PR #4490, opened 2021-09-02
Remove informal salutation in acp & correct texts

Bantor:
[Forum discussion](https://community.woltlab.com/thread/292299-acp-grunds%C3%A4tzlich-formelle-anrede/)


felix-d1strict:
I don't understand what that's about.  There is a distinction between informal and formal. Why do all informal language variables have to be removed?

PR #4491, opened 2021-09-04
Fix miswording in language variable of devtools

Fighter456:
The german language variable `wcf.acp.pip.page.name.description` uses the word `Boxname` but is used in the `PagePackageInstallationPlugin`.


joshuaruesweg:
Thanks!

PR #4492, opened 2021-09-06
5.5 package exclude specification

joshuaruesweg:


wcfsetup/install/files/lib/system/package/PackageArchive.class.php
```diff
@@ -329,6 +329,19 @@ protected function readPackageInfo()
                 $data[$attribute->name] = $attribute->value;
             }
 
+            if (
+                !isset($data['version'])
+                || (!Package::isValidVersion($data['version']) && $data['version'] !== '*')
```
dtdesign, line 334:
```suggestion
                || ($data['version'] !== '*' && !Package::isValidVersion($data['version']))
```
Besides this being a short-circuit, this also makes it clear that we only validate the version if it does not equal `*`.

dtdesign (changes requested):
LGTM

PR #4506, opened 2021-09-21
Deprecate the escapeString() helper

TimWolla:
Developers are strongly encouraged to use prepared statements. If this is not
possible for compatibility reasons, they should use the `->escapeString()`
method directly.

Deprecating the helper ultimately allows cleaning up core.functions.php which
has become a dumping ground for all type of stuff over time.

### DIFF
--- a/wcfsetup/install/files/lib/core.functions.php
+++ b/wcfsetup/install/files/lib/core.functions.php
@@ -18,7 +18,7 @@ namespace {
 	spl_autoload_register([WCF::class, 'autoload']);
 	
 	/**
-	 * @deprecated 5.5 Use prepared statements if possible. Directly WCF::getDB()->escapeString() if prepared statements cannot be used.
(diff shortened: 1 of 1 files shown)


wcfsetup/install/files/lib/core.functions.php
```diff
@@ -18,11 +18,7 @@
 	spl_autoload_register([WCF::class, 'autoload']);
 	
 	/**
-	 * Escapes a string for use in sql query.
-	 * 
-	 * @see	\wcf\system\database\Database::escapeString()
-	 * @param	string		$string
-	 * @return	string
+	 * @deprecated 5.5 Use prepared statements if possible. Directly WCF::getDB()->escapeString() if prepared statements cannot be used.
```
dtdesign, line 21:
```suggestion
	 * @deprecated 5.5 Use prepared statements if possible. Directly call WCF::getDB()->escapeString() if prepared statements cannot be used.
```

dtdesign (changes requested):
LGTM

PR #4514, opened 2021-09-23
Recommend 64-bit PHP during WCFSetup

### LINKED ISSUE #4512: Recommend 64 Bit PHP during WCFSetup
Shall we add a 64 Bit recommendation to WCFSetup with 5.5?

_Originally posted by @TimWolla in https://github.com/WoltLab/WCF/issues/3023#issuecomment-925025160_

TimWolla:
Resolves #4512


dtdesign (changes requested):
LGTM

PR #4518, opened 2021-09-23
Notifications for subscribers of parent objects

BurntimeX:
See https://github.com/WoltLab/com.woltlab.wbb/issues/386


wcfsetup/install/files/lib/system/user/object/watch/UserObjectWatchHandler.class.php
```diff
@@ -171,4 +171,67 @@ public function getSubscribers(string $objectType, int $objectID): array
...
+                    $parentUserIDs,
+                    $filterRecipients,
+                    \ARRAY_FILTER_USE_BOTH
+                )
+            );
+            $recipientIDs = \array_unique(\array_merge($recipientIDs, $parentRecipientIDs));
+        }
+
+        if (!empty($recipientIDs)) {
+            if (!empty($recipientIDs)) {
```
TimWolla, line 226: This condition is duplicated with the line above.

PR #4530, opened 2021-09-28
Add `WoltLabSuite/Core/Ui/ScrollSpy`

joshuaruesweg:


ts/WoltLabSuite/Core/Ui/ScrollSpy.ts
```diff
@@ -0,0 +1,89 @@
...
+    this.callback = callback;
+    this.reference = reference;
+    this.target = target;
+
+    this.start();
+
+    this.scroll();
+  }
+
+  private scroll(): void {
```
TimWolla, line 40:
```suggestion
  private onScroll(): void {
```

ts/WoltLabSuite/Core/Ui/ScrollSpy.ts
```diff
@@ -0,0 +1,89 @@
...
+
+type Callback = () => void;
+
+export class ScrollSpy {
+  private threshold: number;
+  private callback: Callback;
+  private reference?: HTMLElement;
+  private target?: HTMLElement;
+
+  public constructor(threshold: number, callback: Callback, target?: HTMLElement, reference?: HTMLElement) {
```
TimWolla, line 25: Does it make sense to only provide one of `target` and `reference`?

If not:
```suggestion
  public constructor(threshold: number, callback: Callback);
  public constructor(threshold: number, callback: Callback, target: HTMLElement, reference: HTMLElement);
  public constructor(threshold: number, callback: Callback, target?: HTMLElement, reference?: HTMLElement) {
```
(untested)

ts/WoltLabSuite/Core/Ui/ScrollSpy.ts
```diff
@@ -0,0 +1,89 @@
...
+      this.callback();
+    }
+  }
+
+  private getTargetHeight(): number {
+    if (this.target) {
+      return this.target.clientHeight;
+    }
+
+    return document.querySelector("body")!.scrollHeight;
```
TimWolla, line 55:
```suggestion
    return document.body.scrollHeight;
```

ts/WoltLabSuite/Core/Ui/ScrollSpy.ts
joshuaruesweg: Yes. The common usage is to apply a `target` but not a `reference`.

ts/WoltLabSuite/Core/Ui/ScrollSpy.ts
```diff
@@ -0,0 +1,89 @@
+/**
+ * Executes provided callback if scroll threshold is reached. Usable to determine
+ * if user reached the bottom of an element to load new elements on the fly.
+ *
+ * If you do not provide a value for `reference` and `target` it will assume you're
+ * monitoring page scrolls, otherwise a valid HTMLElement must be provided.
+ *
+ * If reference is `null` the entire window will be the reference.
```
TimWolla, line 8: `null` is not an allowed value.

PR #4538, opened 2021-10-04
Disable unfurled urls in signatures

joshuaruesweg:


TimWolla (approved):
Not particularly happy with the hardcoded context check, but LGTM. With the property being `private` it can be refactored if the need arises.

PR #4543, opened 2021-10-08
Replace charset directly after head

xopez:
The `charset` meta element should be the first thing in `<head>`.
Further reading: https://webhint.io/docs/user-guide/hints/hint-meta-charset-utf-8/


TimWolla:
Duplicate of #3998

PR #4570, opened 2021-10-28
Add support for inverted permissions in simple acl form field

joshuaruesweg:


wcfsetup/install/files/lib/system/form/builder/field/acl/simple/SimpleAclFormField.class.php
```diff
@@ -82,9 +89,37 @@ public function readValue()
...
 
         return $this;
     }
+
+    /**
+     * Enables or disables the support for inverted permissions.
+     *
+     * @return  static      this field
+     */
+    public function supportInvertedPermissions(bool $supportInvertedPermissions = true)
```
joshuaruesweg, line 109: Missing `@since`

TimWolla (approved):

PR #4572, opened 2021-11-01
Suggest performing a system check when key system information change

### LINKED ISSUE #4562: Suggest performing a system check when key system information change
Running a system check should be recommended when key system information change:

- [x] The document root (indicating a switch of web host)
- [x] The PHP release branch (possibly missing an important extension, because the host forgot to install it)
- [x] The MySQL release branch

These values should be stored when a successful system check was run and checked on the IndexPage, suggesting to run a system check when the values differ.

TimWolla:
Resolves #4562


wcfsetup/install/files/lib/system/Environment.class.php
```diff
@@ -0,0 +1,57 @@
...
+     * the PHP minor version and the MySQL fork and minor version.
+     */
+    public static function getSystemId(): string
+    {
+        $ctx = \hash_init('sha256');
+
+        $sqlVersion = WCF::getDB()->getVersion();
+
+        foreach (
+            [
```
dtdesign, line 31: I would move this into a separate variable in front of the `foreach()`, makes it easier to read.

wcfsetup/install/files/lib/system/Environment.class.php
```diff
@@ -0,0 +1,57 @@
...
+                // 1) The path to this class as a proxy for the document root.
+                \sprintf("__FILE__ %s", __FILE__),
+                // 2) The PHP minor version.
+                \sprintf('PHP %s', self::getMinorVersion(\PHP_VERSION)),
```
dtdesign, line 35:
```suggestion
                \sprintf('PHP %s.%s', \PHP_MAJOR_VERSION, \PHP_MINOR_VERSION),
```
See https://www.php.net/manual/en/reserved.constants.php#reserved.constants.core

dtdesign (changes requested):

wcfsetup/install/files/lib/system/Environment.class.php
TimWolla: I initially considered that, but this would be inconsistent with `SystemCheck` itself:

https://github.com/WoltLab/WCF/blob/master/wcfsetup/install/files/lib/acp/page/SystemCheckPage.class.php#L351

dtdesign: I see, then disregard my comment please. I'm not sure why we went the extra mile in the system check, but you are right about being consistent.

dtdesign (approved):

PR #4577, opened 2021-11-04
fix dependency issues with ItemList-based fields

mutec:


TimWolla:
Thanks, I've already picked the commit fixing the typo in the comment.

As for the fix for the dependencies: I'm not sure I'm happy with exposing this new function in Ui/ItemList. Any piece of public API will need to be preserved going forward. Your solution also tightly ties the dependencies and ItemLists together.

I'll have a look at this, maybe your suggestion is the best solution after all. Maybe not.

mutec:
Another way would be using
```TS
          try {
            return UiItemList.getValues(field.id).length >/=== 0;
          }
          catch (e) {
            // bypass
          }
```
since `getValues()` throws an `Error` in case no `ItemList` is registered matching the given id. However you have to access the `ItemList`-API in you don't want to rewrite `ItemList` completely in a minor bugfix-update.

PR #4582, opened 2021-11-08
support `UploadFile` in `UploadFormField::updatedObject()`

mutec:


wcfsetup/install/files/lib/system/form/builder/field/UploadFormField.class.php
```diff
@@ -362,7 +362,15 @@ public function updatedObject(array $data, IStorableObject $object, $loadValues
...
+                        if (!\file_exists($v->getLocation())) {
+                            throw new \InvalidArgumentException(
+                                "The {$method} must return an array of strings or object of class '" . UploadFile::class . "' with the valid file locations for field '{$this->getId()}'."
+                            );
+                        }
+
+                        return $v;
+                    } elseif (!\is_string($v) || !\file_exists($v)) {
                         throw new \InvalidArgumentException(
                             "The {$method} must return an array of strings with the file locations for field '{$this->getId()}'."
```
TimWolla, line 375: You need to update this error message. Maybe it would be easier to just `if ($v instanceof Uploadfile) $v = $v->getLocation();`?

wcfsetup/install/files/lib/system/file/upload/UploadFile.class.php
```diff
@@ -219,9 +219,11 @@ public function setProcessed($newLocation)
      *
      * @param string $link
      */
-    public function setImageLink($link)
+    public function setImageLink(string $link): UploadFile
```
TimWolla, line 222: Adding a new return value/return type is a breaking change, as this class is not `final`. Additionally this is inconsistent with e.g. `setProcessed()` not returning anything at all.

TimWolla (changes requested):

wcfsetup/install/files/lib/system/form/builder/field/UploadFormField.class.php
mutec: This would make it impossible to use the image link-feature of `UploadFile` and setting a filename differing from the used path (e.g. attachment's original filename instead of id and hash).

wcfsetup/install/files/lib/system/file/upload/UploadFile.class.php
mutec: The return type definition doesn't matter for me; I was used to do so since we're doing it strictly everywhere for the (internal code of the) new shop. ^^
Important for me is that it returns itself for a usage like that:
```php
	public function getFileUploadFileLocations() : array {
		$path = $this->getLocation();
		
		if (\file_exists($path)) {
			return [
				(new UploadFile($path, $this->getFilename()))
					->setImageLink($this->getLink()),
			];
		}
		
		return [$path];
	}
```

However the phpDoc is missing the return type; totally forgot that.

TimWolla: >  Important for me is that it returns itself for a usage like that:

This would be pretty inconsistent across the setters in the `UploadFile` class, because not all of them `return $this;`

> Important for me is that it returns itself for a usage like that:

In any case I don't really see the benefit of such an interface here: It's not like PHP DDL/form builder where you need to define _large amounts_ of “static data” (i.e. a hardcoded list of 10+ fields).

You are actually creating a dynamic object here – just use a temporary variable.

joshuaruesweg (changes requested):
Basically I agree with Tim in all points. 

Other than that, passing an `UploadFile` is a good idea. But I think about an own method located in the DBO object. What about `get{$fieldName}UploadFiles`, which only can return this objects? Returning this files in `get{$fieldName}UploadFiles` is quite confusing for new developers.

mutec:
I'm not sure about the own function, I remember about an issue when using `UploadFile` in case the file does not exist (yet). I'm going to have a look whether it was an experiment of mine or a serious issue. If possible it's more intuitive, you're right.

In general, what's your opinion on adding an own function @TimWolla?

joshuaruesweg:
> I'm not sure about the own function, I remember about an issue when using UploadFile in case the file does not exist (yet). I'm going to have a look whether it was an experiment of mine or a serious issue. If possible it's more intuitive, you're right.

I don't get the point. You can return an empty array, if the file does not exists, as in the other methods. You have exactly the same problem, if you pass a file, which not exists, with any other method, because it is converted directly into an `UploadFile`.

TimWolla:
> In general, what's your opinion on adding an own function @TimWolla?

I did not realize that there were getter methods, because they did not show up in the Diff and this is not the part of the code where I'm experienced. But I agree that it would be useful to have a properly named getter as well.

mutec:
> > In general, what's your opinion on adding an own function @TimWolla?
> 
> I did not realize that there were getter methods, because they did not show up in the Diff and this is not the part of the code where I'm experienced. But I agree that it would be useful to have a properly named getter as well.

I changed the code to use a own function if available - preferred before the old one to make sure the features of `UploadFile` are used when possible. Would you and @joshuaruesweg please have a look on the new changes?

wcfsetup/install/files/lib/system/form/builder/field/UploadFormField.class.php
```diff
@@ -360,29 +369,36 @@ public function updatedObject(array $data, IStorableObject $object, $loadValues
...
-            if (\is_array($value)) {
-                $value = \array_map(function ($v) use ($method) {
-                    if (!\is_string($v) || !\file_exists($v)) {
-                        throw new \InvalidArgumentException(
-                            "The {$method} must return an array of strings with the file locations for field '{$this->getId()}'."
-                        );
+            // in any case the given value has to be a valid array
+            if (!\is_array($value)) {
+                throw new \InvalidArgumentException(
+                    "The {$method} must return an array of " . ($expectUploadFile ? "objects of class '" . UploadFile::class . "'" : "strings") . " with the valid file locations for field '{$this->getId()}'."
```
joshuaruesweg, line 375: This, and the other exceptions are wrong worded, if the method does return the objects: 

`The {$method} must return an array of objects of class '{UploadFile::class}' with the valid file locations for field '{$this->getId()}'.`

wcfsetup/install/files/lib/system/file/upload/UploadFile.class.php
```diff
@@ -212,16 +213,21 @@ public function setProcessed($newLocation)
 
         $this->location = $newLocation;
         $this->processed = true;
+
+        return $this;
```
joshuaruesweg, line 217: This is pretty inconsistent. I dislike the return of `$this`.

wcfsetup/install/files/lib/system/form/builder/field/UploadFormField.class.php
```diff
@@ -360,29 +369,36 @@ public function updatedObject(array $data, IStorableObject $object, $loadValues
...
+                );
+            }
+
+            $value = \array_map(function ($v) use ($method, $expectUploadFile) {
+                if ($expectUploadFile && $v instanceof UploadFile) {
+                    if (!\file_exists($v->getLocation())) {
```
joshuaruesweg, line 381: That is unnecessary, the upload file checks themselves, whether the file location exists or not within the constructor.

wcfsetup/install/files/lib/system/form/builder/field/UploadFormField.class.php
```diff
@@ -345,13 +345,22 @@ public function getFieldHtml()
...
+                ]);
+                $method = "method '" . \get_class($object) . "::get" . \ucfirst($this->getObjectProperty()) . "UploadFiles()'";
+                $expectUploadFile = true;
+            } elseif (\method_exists($object, 'get' . \ucfirst($this->getObjectProperty()) . 'UploadFileLocations')) {
                 $value = \call_user_func([
                     $object,
                     'get' . \ucfirst($this->getObjectProperty()) . 'UploadFileLocations',
                 ]);
-                $method = "method '" . \get_class($object) . "::get" . \ucfirst($this->getObjectProperty()) . "UploadFileLocations()'";
+                $method = "method '" . \get_class($object) . '::get' . \ucfirst($this->getObjectProperty()) . "UploadFileLocations()'";
```
joshuaruesweg, line 363: Why this change here?

wcfsetup/install/files/lib/system/form/builder/field/UploadFormField.class.php
```diff
@@ -360,29 +369,36 @@ public function updatedObject(array $data, IStorableObject $object, $loadValues
...
-                            "The {$method} must return an array of strings with the file locations for field '{$this->getId()}'."
-                        );
+            // in any case the given value has to be a valid array
+            if (!\is_array($value)) {
+                throw new \InvalidArgumentException(
+                    "The {$method} must return an array of " . ($expectUploadFile ? "objects of class '" . UploadFile::class . "'" : "strings") . " with the valid file locations for field '{$this->getId()}'."
+                );
+            }
+
+            $value = \array_map(function ($v) use ($method, $expectUploadFile) {
```
joshuaruesweg, line 379: Overall this is very hard to read and understand. I dislike, that both value types are handled here. Maybe we can get two methods, which are handling the different inputs? That would be much easier to understand and read. And thus methods could be called within the if-else structure, which determines the source of the values.

joshuaruesweg (changes requested):

wcfsetup/install/files/lib/system/form/builder/field/UploadFormField.class.php
mutec: I assume I copied or moved the part and the IDE did some "improvements" since it's configured to prefer `'`.

wcfsetup/install/files/lib/system/form/builder/field/UploadFormField.class.php
```diff
@@ -345,13 +345,22 @@ public function getFieldHtml()
     public function updatedObject(array $data, IStorableObject $object, $loadValues = true)
     {
         if ($loadValues) {
-            // first check, whether an getter for the field exists
-            if (\method_exists($object, 'get' . \ucfirst($this->getObjectProperty()) . 'UploadFileLocations')) {
+            $expectUploadFile = false;
+
+            // first check, whether a getter for the field exists; preferring get{$fieldName}UploadFiles() before get{$fieldName}UploadFileLocations()
+            if (\method_exists($object, 'get' . \ucfirst($this->getObjectProperty()) . 'UploadFiles')) {
```
TimWolla, line 351: This code is in order to deserve a dedicated temporary variable for `\ucfirst($this->getObjectProperty())` to then be able to:
```suggestion
            if (\method_exists($object, "get{$fieldName}UploadFiles")) {
```
making this much more readable.

TimWolla (commented):
Please apologize the delay in the review. Overall I agree with Joshua's remarks, so I'm just going to add one comment, as I expect the remainder of the code will change anyway.

joshuaruesweg (changes requested):

wcfsetup/install/files/lib/system/form/builder/field/UploadFormField.class.php
```diff
@@ -340,54 +340,113 @@ public function getFieldHtml()
...
         }
 
         return $this;
     }
 
+    /**
+     * Reads an array of `UploadFile` objects and validates it. Returns `false` if validation failed.
+     *
+     * @param array $data
+     * @return array|false
```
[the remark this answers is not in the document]
mutec: Because I'd have to pass the error message to the function which would be a bit strange.

wcfsetup/install/files/lib/system/form/builder/field/UploadFormField.class.php
```diff
@@ -340,54 +340,113 @@ public function getFieldHtml()
...
+     */
```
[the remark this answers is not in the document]
mutec: Since the class is not final and someone might extend the class and override `updatedObject()` this person would have to re-implement the function. So I'd prefer using the tagging only.

wcfsetup/install/files/lib/system/form/builder/field/UploadFormField.class.php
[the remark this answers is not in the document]
joshuaruesweg: > or capture the Exception and re-throw it.

Rethrowing is preferred, make sure to use the `$previous` to properly chain the exceptions.

wcfsetup/install/files/lib/system/form/builder/field/UploadFormField.class.php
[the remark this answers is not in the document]
joshuaruesweg: I don't see the use point for this specific method within parent classes (keep in mind, you can use `value()` to add values). And if, re-implementing this method is easy and better for us, because we can adjust the method in the future, without maintain backwards compatibility.

wcfsetup/install/files/lib/system/form/builder/field/UploadFormField.class.php
```diff
@@ -340,52 +340,117 @@ public function getFieldHtml()
...
+     * @param string[] $data
+     * @return UploadFile[]
+     * @throws \InvalidArgumentException
+     * @since 5.5
+     */
+    private function readUploadLocations($data): array
+    {
+        if (!\is_array($data)) {
+            throw new \InvalidArgumentException('Given data is not an array.');
+        }
```
TimWolla, line 435: This is better solved with an actual parameter type hint. It might be required to extend the catch-block with `| \TypeError` or `| \ValueError` (or whatever exception PHP throws).

wcfsetup/install/files/lib/system/form/builder/field/UploadFormField.class.php
```diff
@@ -340,52 +340,117 @@ public function getFieldHtml()
...
+     * @param UploadFile[] $data
+     * @return UploadFile[]
+     * @throws \InvalidArgumentException
+     * @since 5.5
+     */
+    private function readUploadFiles($data): array
+    {
+        if (!\is_array($data)) {
+            throw new \InvalidArgumentException('Given data is not an array.');
+        }
```
TimWolla, line 412: ditto

wcfsetup/install/files/lib/system/form/builder/field/UploadFormField.class.php
```diff
@@ -340,52 +340,117 @@ public function getFieldHtml()
...
+                    throw new \InvalidArgumentException("The method {$className}::get{$propertyName}() must return an array of objects of strings with the valid file locations for field '{$this->getId()}'.", 0, $e);
+                }
             } else {
-                $value = $data[$this->getObjectProperty()];
-                $method = "variable '" . \get_class($object) . "::$" . $this->getObjectProperty() . "'";
+                $variableName = $this->getObjectProperty();
+                try {
+                    $value = $this->readUploadLocations($data[$variableName] ?? null);
+                } catch (\InvalidArgumentException $e) {
+                    throw new \InvalidArgumentException("The variable {$className}::\${$variableName}UploadFiles() must return an array of strings with the valid file locations for field '{$this->getId()}'.", 0, $e);
```
TimWolla, line 390:
```suggestion
                    throw new \InvalidArgumentException("The property {$className}::\${$variableName}UploadFiles must contain an array of strings with the valid file locations for field '{$this->getId()}'.", 0, $e);
```

TimWolla (commented):
Some small nits, other than that this LGTM based on a review of the diff.

wcfsetup/install/files/lib/system/form/builder/field/UploadFormField.class.php
```diff
@@ -340,52 +340,117 @@ public function getFieldHtml()
...
-                $value = \call_user_func([
-                    $object,
-                    'get' . \ucfirst($this->getObjectProperty()) . 'UploadFileLocations',
-                ]);
-                $method = "method '" . \get_class($object) . "::get" . \ucfirst($this->getObjectProperty()) . "UploadFileLocations()'";
-            } elseif (\method_exists($object, 'get' . \ucfirst($this->getObjectProperty()))) {
-                $value = \call_user_func([$object, 'get' . \ucfirst($this->getObjectProperty())]);
-                $method = "method '" . \get_class($object) . "::get" . \ucfirst($this->getObjectProperty()) . "()'";
+            $propertyName = \ucfirst($this->getObjectProperty());
+            $className = static::class;
```
joshuaruesweg, line 349: Previously, the Variable `className` contains the classname of the given object, not the class of the `UploadFormField`.

joshuaruesweg (approved):
Other than the small issue, the PR LGTM! :)

wcfsetup/install/files/lib/system/form/builder/field/UploadFormField.class.php
mutec: That's still wrong. :D

wcfsetup/install/files/lib/system/form/builder/field/UploadFormField.class.php
```diff
@@ -14,7 +14,7 @@
 /**
  * Implementation of a form field for to uploads.
  *
- * @author  Joshua Ruesweg
+ * @author  Joshua Ruesweg, Florian Gail
```
TimWolla, line 17: Names should be ordered alphabetically by last name.

wcfsetup/install/files/lib/system/form/builder/field/UploadFormField.class.php
```diff
@@ -340,52 +340,109 @@ public function getFieldHtml()
...
+     * @since 5.5
+     */
+    private function readUploadLocations(array $data): array
+    {
+        foreach ($data as $key => $value) {
+            if (!\is_string($value)) {
+                throw new \InvalidArgumentException('At least one value is not a string.');
+            } elseif (!\file_exists($value)) {
+                throw new \InvalidArgumentException("File '{$value}' could not be found.");
+            }
```
TimWolla, line 434:
```suggestion
            }
            if (!\file_exists($value)) {
                throw new \InvalidArgumentException("File '{$value}' could not be found.");
            }
```
I'd avoid `else` if the `if` forces the function to return.

TimWolla (approved):
LGTM other than those two nits. I've checked with Joshua, he'll make those changes himself and apply the PR manually to save you another back-and-forth cycle.

joshuaruesweg:
Thanks! 

In addition to the comments Tim has made, I have made the following additional changes to your commit: 

* Reword the commit message. 
* I have made the error messages more readable by using `sprintf`

PR #4585, opened 2021-11-14
Draft for a new ajax API built on top of `fetch()`

dtdesign:
The design of the existing AJAX API is quite old and dates back to the behavior of `WCF.Action.Proxy`. It lacks any support for asynchronous actions (Promises) and has a subpar control flow caused by the different methods to prepare (`_ajaxSetup()`), invoke and evaluate the result (`_ajaxSuccess()` and `_ajaxFailure()`). Especially modules serving multiple endpoints suffer from `_ajaxSuccess()` being a glorified `switch` to keep the code readable. The lack of proper types for the request and response make it hard to work with it.

At some point I have evaluated [Axios](https://github.com/axios/axios/) as an alternative, but at the end of the day decided against it. It adds a lot of bells and whistles, but its major selling point are the compatibility with ancient browsers as well as some nice helper features. We already maintain a working implementation on top of `XMLHttpRequest`, so we are familiar with the requirements and have proven to be able to maintain our own wrapper around it. From my perspective, Axios is a nice library, but adds another dependency that we do not strictly need.

Usage of the library in this draft:
```ts
type ResponseMakeSnafucated = {
  foo: {
    bar: number;
  };
  message: string;
};

class MyModule {
  constructor(buttonId: string) {
    const button = document.getElementById(buttonId);
    button?.addEventListener("click", (event) => void this.onClick(event));
  }

  async onClick(event: MouseEvent) {
    event.preventDefault();

    const data: ResponseMakeSnafucated = await Api.dboAction("makeSnafucated", "wcf\\data\\foo\\FooAction").dispatch();
    if (data.foo.bar > 0) {
      console.log(data.message);
    }
  }
}
```


ts/WoltLabSuite/Core/Api.ts
```diff
@@ -0,0 +1,294 @@
+import * as AjaxStatus from "./Ajax/Status";
```
TimWolla, line 1: Missing module level comment.

ts/WoltLabSuite/Core/Api.ts
```diff
@@ -0,0 +1,294 @@
...
+    return true;
+  }
+
+  private async genericError(result: ErrorResult): Promise<void> {
+    const html = await this.getErrorHtml(result);
+
+    if (html !== "") {
+      const uiDialog = await import("./Ui/Dialog");
+      const domUtil = await import("./Dom/Util");
+      const language = await import("./Language");
```
TimWolla, line 237: One can assume that all of these are immediately available, but maybe use `Promise.all()` to load them in parallel?

ts/WoltLabSuite/Core/Api.ts
```diff
@@ -0,0 +1,294 @@
...
+    }
+
+    const init: RequestInit = {
+      method: "POST",
+      headers: {
+        "Content-Type": "application/x-www-form-urlencoded; charset=UTF-8",
+        "X-Requested-With": "XMLHttpRequest",
+        "X-XSRF-TOKEN": Core.getXsrfToken(),
+      },
+      body: Core.serialize(body),
```
TimWolla, line 129: [URLSearchParams](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams) will automatically be correctly serialized by `fetch()`. Maybe it would make sense relying on that and requiring the API user to pass a valid `URLSearchParams` object as the payload?

ts/WoltLabSuite/Core/Api.ts
```diff
@@ -0,0 +1,294 @@
...
+
+      // This is an explicit wait for the event loop to execute the
+      // callee function before we execute additional tasks.
+      return Promise.resolve(json.returnValues).then(async (result) => {
+        if (json.forceBackgroundQueuePerform) {
+          const backgroundQueue = await import("./BackgroundQueue");
+          backgroundQueue.invoke();
+        }
+        return result;
+      });
```
TimWolla, line 180: Not sure whether I would do this with a `.then()`, because that will require the loading of the background queue to succeed for the AJAX request to succeed. Instead:

```
if (json.forceBackgroundQueuePerform) {
 // Fire and forget.
 void import("./BackgroundQueue").then((q) => q.invoke());
}

// an explicit Promise.resolve() is no longer required.
return json.returnValues;
```

ts/WoltLabSuite/Core/Api.ts
```diff
@@ -0,0 +1,294 @@
...
+      try {
+        json = await response.json();
+      } catch (e) {
+        const result = this.handleError(ErrorCode.INVALID_JSON, response);
+
+        return Promise.reject(result);
+      }
+
+      // This is an explicit wait for the event loop to execute the
+      // callee function before we execute additional tasks.
```
TimWolla, line 173: I don't understand what this is attempting to say.

dtdesign: You will often times see calls to `resolve()` without awaiting for them, but technically `void resolve()` would make it more clear that you don't care for its result.

Here we want to execute code after the action has run to mimic the existing behavior, so we need to chain it to `Promise.resolve()` which is a promise itself.

ts/WoltLabSuite/Core/Api.ts
dtdesign: I don't see much value there, at the end of the day it is the same logic as `Core.serialize()`, plus its ergonomics for creating nested structures do not look appealing to me.

I would even go so far to say that the user should not even care what happens, because this is merely an implementation detail. We could transparently change the API to accept a `POST` request with a JSON body instead, but this gets messy when we start requiring non vanilla objects. The current implementation hides the fact _how_ the data gets to the server, it just asks for _what_ should be transmitted.

ts/WoltLabSuite/Core/Api.ts
dtdesign: That makes sense, that code was refactored a few times but I missed this.

ts/WoltLabSuite/Core/Api.ts
```diff
@@ -0,0 +1,303 @@
...
+        throw new InvalidJson(response);
+      }
+
+      if (json.forceBackgroundQueuePerform) {
+        void import("./BackgroundQueue").then((BackgroundQueue) => BackgroundQueue.invoke());
+      }
+
+      return json.returnValues;
+    } catch (error) {
+      if (error instanceof ExpectedJson || error instanceof InvalidJson || error instanceof StatusNotOk) {
```
TimWolla, line 144: Should this also check for `ApiError`? You generally don't want to re-pack an `ApiError` into a `ConnectionError`.

ts/WoltLabSuite/Core/Api.ts
```diff
@@ -0,0 +1,303 @@
...
+  }
+}
+
+export class ApiError extends Error {
+  constructor(message: string) {
+    super(message);
+
+    this.name = "ApiError";
+  }
+}
```
TimWolla, line 175: Same applies for the other classes:

```suggestion
export class ApiError extends Error {
  name = "ApiError"
}
```

TimWolla (approved):
The public API of this module LGTM now. The remaining comments are just about “code style” within this module (and missing doc comments).

dtdesign:
@TimWolla Please verify the latest changes and compare it against the documentation in case there are any discrepancies.

TimWolla:
> @TimWolla Please verify the latest changes and compare it against the documentation in case there are any discrepancies.

LGTM. I tested the first example in the documentation (against UserProfileAction::getUserProfile()) with the following HTML:

```
<button id="testbutton">Test</button>
<div id="latestFoo"></div>
<script>
require(['WoltLabSuite/Core/MyModule'], (MyModule) => {
  new MyModule.default(1, "bar", "testbutton");
});
</script>
```

TimWolla (approved):

PR #4586, opened 2021-11-19
Reaction update event for plugin usage

Cyperghost:
Insert event handler to update reaction from outside.


TimWolla:
What's the use case behind this?

Cyperghost:
> What's the use case behind this?

Currently it's impossible to say the system that the data has change. And I would do that think with a plugin, without reloading the page.

TimWolla:
I'd rather not have a method that allows injecting arbitrary data into the internal state, because that will break whenever changes to the internal data handling need to be made.

Instead a `refresh()` method that performs an AJAX request fetching the updated data could be added.

PR #4587, opened 2021-11-19
Fix Image width of the unfurl card images

Morik:
Looks like it was missed while uniufying the width of the card itself, see:
https://github.com/WoltLab/WCF/commit/533e1ed2861322b3b598797492f2a99f8d06841a


TimWolla:
Careful here when merging: The target branch is incorrect and I assume that I can't easily change it without GitHub suggesting that I merge the whole of 5.5 into 5.4.

Morik:
Sorry for that mishap

dtdesign:
Thanks, I've manually added a commit in 5.4 that fixes the issue.

PR #4594, opened 2021-11-28
force int|null type for `TLengthDatabaseTableColumn::getLength()`

mutec:


TimWolla (changes requested):
The casting should be performed in `->length()` so that the internal state of the object is correct.

mutec:
> The casting should be performed in `->length()` so that the internal state of the object is correct.

I explicitly didn't do that since I hope you are going to use typed arguments in the future which makes it senseless to cast after the type check. So I thought it might make more sense to ensure the provided value is an `int` when the function is called.
I'd rather use `function length(int $length)` additionally to that change.

TimWolla:
> I explicitly didn't do that since I hope you are going to use typed arguments in the future which makes it senseless to cast after the type check.

Such a cast can easily be removed once the parameter is properly typed. It is important however, that the internal state of the object is correct, because everything else results in a mess. The property is already defined to be `int|null` in the PHP doc.

> I'd rather use `function length(int $length)` additionally to that change.

It should be `?int`. I'm wondering whether that would be a “safe” change: Passing a purely numeric string to a `int` parameter should result in PHP casting this automatically and not error. If you don't pass a purely numeric string, then your code already is broken.

--------

In any case: Performing the cast in the getter is not the correct change.

TimWolla:
@mutec Do you plan on making (and testing) the suggested changes, or should I handle this?

mutec:
@TimWolla Sure, but I have been busy the whole week.

mutec:
@TimWolla seems to work. Should I add the type for the function's parameter too or would you like to wait for WSC 6.0?

TimWolla:
> Should I add the type for the function's parameter too or would you like to wait for WSC 6.0?

I'd not type the parameter for now. With 5.6 the PHP requirements will be increased once more to PHP 7.4, allowing the use of typed properties. It makes sense to then adjust all the PHP DDL DTOs in a swooping fashion to ensure proper consistency.

Your change LGTM now. I'll probably take it manually to slightly adjust the commit message. Thanks!

mutec:
You know I'm very creative in writing commit messages. ;)

mutec:
Btw. maybe it's worth it to check `MySQLDatabaseEditor::getColumns()` in WSC 6.0 since the parameter is set as a string (coming from a regex match) over there.

TimWolla:
> You know I'm very creative in writing commit messages. ;)

It wasn't bad per se. However I found it important to give the full reasoning behind this change in the commit message itself. This makes it *much* easier to figure out what happened, later on in git blame and friends. In the end I did a full rewrite of the message, but you did the hard work in creating (and testing) the change, so that's fine!

TimWolla:
> Btw. maybe it's worth it to check `MySQLDatabaseEditor::getColumns()` in WSC 6.0 since the parameter is set as a string (coming from a regex match) over there.

We'll probably come across this naturally when adding proper type support. But as I said above: As long as `strict_types` is not enabled PHP will automatically cast the parameter to an int, if the given string is completely numeric. So this will automatically ensure clean types in the "modern" part of the code, without requiring changes to these very very old parts (MySQLDatabaseEditor).

At a cursory glance the `getColumns()` method should probably be migrated to the well-defined INFORMATION_SCHEMA instead of using `SHOW COLUMNS` + a regular expression. But that would be busy-work for no really measurable benefit.

PR #4596, opened 2021-11-30
Fix suffix declaration in XSD

SoftCreatR:
As of now, a suffix is only allowed for option categories, which is wrong, because a suffix can only be applied to an option itself.


TimWolla:
Thanks, applied manually to 5.3+ while including the PR's description within the commit message itself.

PR #4603, opened 2021-12-08
Overhaul of the user panel menu

dtdesign:
The menu has been rebuilt from the ground up with simplicity, consistent styling and a11y in mind. It introduces a set of new style variables to avoid the pitfall of deriving colors from different elements just because the happen to fit in nicely in the default style.

This change is the first step towards the new mobile menu which will be built on top of this. For now the conversation menu has been left unchanged to serve as a blueprint for the backwards compatibility layer.

image

image


wcfsetup/install/files/lib/data/moderation/queue/ModerationQueueAction.class.php
```diff
@@ -110,8 +112,66 @@ public function validateGetOutstandingQueues()
      * Returns a list of outstanding queues.
      *
      * @return  string[]
+     * @deprecated 5.5
```
TimWolla, line 115: Please add a short explanation blurb: `5.5 - Use foo instead.` or `5.5 - This was tightly bound to the old and removed userpanel implementation.`.

wcfsetup/install/files/lib/data/user/notification/UserNotificationAction.class.php
```diff
@@ -190,6 +196,7 @@ public function validateGetOutstandingNotifications()
      * Loads user notifications.
      *
      * @return  mixed[]
+     * @deprecated 5.5
```
TimWolla, line 199: ditto

TimWolla (commented):
Eyeballed everything, except for pageHeaderUser.tpl. The changes are just too large in there.

ts/WoltLabSuite/Core/Ui/User/Menu/Data/ModerationQueue.ts
```diff
@@ -0,0 +1,165 @@
...
+  private updateCounter(counter: number): void {
+    let badge = this.button.querySelector<HTMLElement>(".badge");
+    if (badge === null && counter > 0) {
+      badge = document.createElement("span");
+      badge.classList.add("badge badgeUpdate");
+
+      this.button.querySelector("a")!.append(badge);
+    }
+
+    if (badge) {
```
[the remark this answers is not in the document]
dtdesign: No, the badge could not exist *and* the `counter` equals `0`.

dtdesign:
The commit c9e8db8 might not be immediately obvious, but this is pretty much what the previous implementation does by slapping `top: 50px !important` into the CSS. I'm taking the extra steps here to make it possible for style creators to overwrite the position with `absolute` and having the computed value match. This is useful when the header is not fixed and was horribly awkward to work around.

TimWolla (approved):
As far as I'm qualified to tell this LGTM now. I did not perform a test, this was a review purely based on the diff.

PR #4604, opened 2021-12-09
Use the "focus-trap" library for existing components

dtdesign:
All changes have already been manually tested.


TimWolla (changes requested):
Diff LGTM, but please check the commit messages. The second "reaction picker" should probably be called "dialog".

PR #4605, opened 2021-12-09
Search form overhaul

BurntimeX:
This overhaul of the search form has two goals:
1. By combining SearchForm and SearchResultPage, a better UI is provided to the user. The user can make changes to his search directly above the search results and does not have to call a separate form for this. In addition, the URL now always contains all search parameters and the user can easily share them with other users.

Overall, the UI of the search function should now behaves in a similar way to search engines like Google.

2. The program logic of the search function was previously very strongly coupled to the SearchForm and the SearchResultPage. With the changes, the code is now independent and can be accessed much more easily from other sources (e.g. a REST endpoint).


wcfsetup/install/files/lib/system/search/ISearchProvider.class.php
```diff
@@ -0,0 +1,110 @@
...
+     * This method is only called, if this provider is the only provider that is searched.
+     */
+    public function getCustomSortField(string $sortField): string;
+
+    /**
+     * Replaces the outer SQL query with a custom version. Querying the search index requires the
+     * placeholder {WCF_SEARCH_INNER_JOIN} within an empty INNER JOIN() statement.
+     */
+    public function getOuterSQLQuery(
+        string $q,
```
TimWolla, line 106: I'd drop the `$q` parameter from here. The search provider can't really rely on the search query, because it does not know what search engine will actually execute the search and what type of preprocessing was performed on the query.

wcfsetup/install/files/lib/page/SearchResultPage.class.php
```diff
@@ -247,4 +217,12 @@ protected function initObjectList()
     protected function readObjects()
     {
     }
+
+    /**
+     * @inheritDoc
+     */
+    public function __run()
+    {
+        throw new IllegalLinkException();
```
TimWolla, line 226: As this controller is effectively non-functional you could remove it outright. Compatibility with code relying on this is broken anyway.

wcfsetup/install/files/lib/page/SearchPage.class.php
```diff
@@ -0,0 +1,58 @@
...
+/**
+ * Shows the search form.
+ *
+ * @author  Marcel Werk
+ * @copyright   2001-2021 WoltLab GmbH
+ * @license GNU Lesser General Public License <http://opensource.org/licenses/lgpl-license.php>
+ * @package WoltLabSuite\Core\Page
+ * @since 5.5
+ */
+class SearchPage extends AbstractPage
```
TimWolla, line 18:
```suggestion
final class SearchPage extends AbstractPage
```

wcfsetup/install/files/lib/form/SearchForm.class.php
```diff
@@ -564,27 +565,14 @@ protected function getConditions()
...
+        return $this->userIDs;
+    }
 
-        return $userIDs;
+    /**
+     * @inheritDoc
+     */
+    public function __run()
+    {
+        throw new IllegalLinkException();
```
TimWolla, line 576: Same as SearchResultPage. Especially since this Form conflicts with the new SearchPage based on the name.

wcfsetup/install/files/lib/data/search/SearchAction.class.php
```diff
@@ -22,4 +28,108 @@ class SearchAction extends AbstractDatabaseObjectAction
...
+            if (SearchEngine::getInstance()->getObjectType($this->parameters['type']) === null) {
+                throw new IllegalLinkException();
+            }
+        }
+
+        if (\in_array($this->parameters['sortField'], ['subject', 'time', 'username', 'relevance'])) {
+            $this->parameters['sortField'] = SEARCH_DEFAULT_SORT_FIELD;
+        }
+
+        if ($this->parameters['sortOrder'] !== 'ASC' && $this->parameters['sortOrder'] !== 'DESC') {
```
TimWolla, line 63: For consistency I'd validate with with `\in_array()`, like the `sortField`.

wcfsetup/install/files/lib/data/search/SearchAction.class.php
```diff
@@ -22,4 +28,108 @@ class SearchAction extends AbstractDatabaseObjectAction
...
+            $this->parameters['sortOrder'] = SEARCH_DEFAULT_SORT_ORDER;
+        }
+    }
+
+    public function search(): array
```
TimWolla, line 68: `@since` (applies to all new methods)

wcfsetup/install/files/lib/data/search/SearchAction.class.php
```diff
@@ -22,4 +28,108 @@ class SearchAction extends AbstractDatabaseObjectAction
...
+            'query' => $resultHandler->getQuery(),
+        ]);
+
+        return [
+            'count' => $resultHandler->countSearchResults(),
+            'title' => WCF::getLanguage()->getDynamicVariable('wcf.search.results.title', [
+                'count' => $resultHandler->countSearchResults(),
+                'query' => $resultHandler->getQuery(),
+            ]),
+            'pages' => ceil($resultHandler->countSearchResults() / SEARCH_RESULTS_PER_PAGE),
```
TimWolla, line 98:
```suggestion
            'pages' => \ceil($resultHandler->countSearchResults() / SEARCH_RESULTS_PER_PAGE),
```

ts/WoltLabSuite/Core/Ui/Search/Page.ts
```diff
@@ -106,4 +98,21 @@ export function init(objectType: string): void {
...
+}
+
+function submit(form: HTMLFormElement, input: HTMLInputElement): void {
+  const url = new URL(form.action);
+  url.search += url.search !== "" ? "&" : "?";
+  url.search += "q=" + encodeURIComponent(input.value.trim());
+
+  parameters.forEach((value, key) => {
+    url.search += "&" + encodeURIComponent(key) + "=" + encodeURIComponent(value);
+  });
```
TimWolla, line 115: This could be much simpler as https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams.

ts/WoltLabSuite/Core/Ui/Search/Extended.ts
```diff
@@ -0,0 +1,197 @@
+import { dboAction } from "../../Ajax";
```
TimWolla, line 1: Module comment missing here.

ts/WoltLabSuite/Core/Ui/Search/Extended.ts
```diff
@@ -0,0 +1,197 @@
...
+    }
+  }
+
+  private updateQueryString(): void {
+    const url = new URL(this.form.action);
+
+    new FormData(this.form).forEach((value, key) => {
+      if (value.toString()) {
+        url.search += url.search !== "" ? "&" : "?";
+        url.search += encodeURIComponent(key) + "=" + encodeURIComponent(value.toString());
```
TimWolla, line 95: ditto URLSearchParams

ts/WoltLabSuite/Core/Ui/Search/Extended.ts
```diff
@@ -0,0 +1,197 @@
...
+      callbackSwitch: (pageNo) => {
+        this.changePage(pageNo);
+      },
+    });
+  }
+
+  private async changePage(pageNo: number): Promise<void> {
+    if (this.lastSearchResultRequest) {
+      this.lastSearchResultRequest.abort();
+    }
```
TimWolla, line 164: I'd either explicitly clear the value after the abort:

```suggestion
    if (this.lastSearchResultRequest) {
      this.lastSearchResultRequest.abort();
      this.lastSearchRequest = undefined;
    }
```

 or just do:
```suggestion
    this.lastSearchResultRequest?.abort();
```

wcfsetup/install/files/lib/page/SearchResultPage.class.php
BurntimeX: It's still used here: https://github.com/WoltLab/WCF/blob/3cfed0e1c3883117ff3d44e65655429a1d3b4b54/wcfsetup/install/files/lib/system/search/SearchResultHandler.class.php#L164

ts/WoltLabSuite/Core/Ui/Search/Page.ts
BurntimeX: Unfortunately, using URLSearchParams leads to an undefined order of the parameters and to an unwanted encoding of the first parameter (`index.php?search/` -> `index.php?search%2F`).

TimWolla: It should be fine if you just use it to append the additional parameters like this:

```
const url = new URL(form.action);
url.search += url.search !== "" ? "&" : "?";
url.search += (new URLSearchParams({q: input.value.trim(), ...parameters})).toString();
```

wcfsetup/install/files/lib/system/search/SearchHandler.class.php
```diff
@@ -0,0 +1,348 @@
...
+    }
+
+    private function buildTypeBasedConditionBuilders(): void
+    {
+        $form = $this->getSearchFormEmulation();
+
+        foreach ($this->objectTypeNames as $key => $objectTypeName) {
+            $objectType = SearchEngine::getInstance()->getObjectType($objectTypeName);
+            if ($objectType === null) {
+                throw new SystemException('unknown search object type ' . $objectTypeName);
```
TimWolla, line 163: No SystemException in new code, please. `$this->objectTypeNames` should probably be validated at the point where untrusted information may be inserted (i.e. initObjectTypeNames).

wcfsetup/install/files/lib/system/search/SearchResultHandler.class.php
```diff
@@ -0,0 +1,172 @@
...
+                    throw new ImplementationException(\get_class($message), ISearchResultObject::class);
+                }
+
+                $customIcon = '';
+                if ($message instanceof ICustomIconSearchResultObject) {
+                    $customIcon = $message->getCustomSearchResultIcon();
+                }
+
+                $this->messages[] = $message;
+                $this->customIcons[\spl_object_hash($message)] = $customIcon;
```
TimWolla, line 131: https://www.php.net/manual/en/class.splobjectstorage.php might be a bit cleaner, avoiding the need for the manual call to `spl_object_hash()`.

wcfsetup/install/files/lib/system/search/mysql/MysqlSearchEngine.class.php
```diff
@@ -59,7 +60,11 @@ public function search(
             }
             $additionalConditionsConditionBuilder = ($additionalConditions[$objectTypeName] ?? null);
 
-            $query = $objectType->getOuterSQLQuery($q, $searchIndexCondition, $additionalConditionsConditionBuilder);
+            if ($objectType instanceof ISearchProvider) {
+                $query = $objectType->getOuterSqlQuery($additionalConditionsConditionBuilder);
+            } else {
+                $query = $objectType->getOuterSQLQuery($q, $searchIndexCondition, $additionalConditionsConditionBuilder);
+            }
```
TimWolla, line 67: This is not ideal, because it breaks existing search engines (e.g. Elasticsearch). Perhaps the `getOuterSqlQuery()` method in the ISearchProvider could be renamed to something more fitting [1] and then a `->getOuterSQLQuery()` method be added to the abstract parent class that provides the backwards compatibility with old search engines?

[1] getFetchObjectsQuery() perhaps?

com.woltlab.wcf/option.xml
```diff
@@ -719,11 +719,6 @@ block:wcf.acp.option.blacklist_sfs_action.block]]></selectoptions>
 				<optiontype>boolean</optiontype>
 				<defaultvalue>1</defaultvalue>
 			</option>
-			<option name="search_use_captcha">
-				<categoryname>security.antispam.captcha</categoryname>
-				<optiontype>boolean</optiontype>
-				<defaultvalue>0</defaultvalue>
-			</option>
```
TimWolla, line 726: With the removal of the captcha (which I consider a good thing), we probably should add support for the flood control with a reasonable limit to prevent the most obvious cases of abuse / DoS attacks. Something like the following should affect any legitimate uses:

- at most 20 searches within the last 60 seconds, but not more than
- 300 searches within the last 12 hours.

wcfsetup/install/files/lib/system/search/SearchHandler.class.php
```diff
@@ -0,0 +1,347 @@
...
+                $parameters[] = WCF::getUser()->userID;
+            }
+
+            $sql = "SELECT  searchID
+                    FROM    wcf" . WCF_N . "_search
+                    WHERE   searchHash = ?
+                        AND searchType = ?
+                        AND searchTime > ?
+                        " . (WCF::getUser()->userID ? 'AND userID = ?' : 'AND userID IS NULL');
+            $statement = WCF::getDB()->prepareStatement($sql);
```
TimWolla, line 280: The condition handling could be cleaned up with a PreparedStatementConditionBuilder.

TimWolla (approved):
I don't have any further concerns, but I purely reviewed this on a diff basis, without testing and without checking the general concept.

PR #4607, opened 2021-12-15
Upgrade to sabberworm/php-css-parser 8.4

TimWolla:
This comes with native PHP 8.1, avoiding the need to install from our fork.


TimWolla:
Unfortunately this is not really reviewable, because neither Git, nor GitHub detect the renamed files as the contents also drastically changed due to an update in code style.

PR #4608, opened 2021-12-19
Overhaul of the mobile menu

dtdesign:
**This PR is currently in work and used to keep track of outstanding tasks.**

* [x] General
  * [x] Missing phrases for the a11y labels (search for `TODO`).
  * [x] Overflow behavior is not yet tested on tablets.
  * [x] The mobile header needs to be updated.
  * [x] Indicator for `.badge.badgeOutstanding` is missing from the icons.
  * [x] Padding at the bottom to account for the browser UI.
  * [x] Missing `aria-label` for the menu buttons.
  * [x] Breaks the admin panel during the login, possibly affects the setup too.
* [x] User
  * [x] DOM might be rebuilt too often.
  * [x] Currently bricked for guests.
  * [x] Check if the tap targets are large enough (should be 44px).
  * [x] There is no indicator for the number of unread items on the tabs.
  * [x] Suppress hover states for touch devices.
* [x] Main
  * [x] Breaks the admin panel because of the different menu DOM.
  * [x] Enforce a minimum gap between the main menu and the footer menu.
  * [x] Missing visual indication for items with unread content.
  * [x] Incomplete support for non-link items (`a[href="#"]`).
* [x] Clean up
  * [x] Remove the template for the mobile menu.
  * [x] Remove the CSS for the legacy Android swipe menu (conflicts with the browser navigation).
* [x] WCFSetup
  * [x] ~Rebuild the CSS for the setup.~
  * [x] Verify if the setup still works because of the differences in the UI.


com.woltlab.wcf/templates/headIncludeJavaScript.tpl
```diff
@@ -16,6 +16,7 @@
 	var ENABLE_PRODUCTION_DEBUG_MODE = {if ENABLE_PRODUCTION_DEBUG_MODE}true{else}false{/if};
 	var ENABLE_DEVELOPER_TOOLS = {if ENABLE_DEVELOPER_TOOLS}true{else}false{/if};
 	var WSC_API_VERSION = {@WSC_API_VERSION};
+	var PAGE_TITLE = '{PAGE_TITLE|language|encodeJS}';
```
TimWolla, line 19: I really dislike all these global JavaScript variables, because they are not really ergonomic. Especially in TypeScript where they need to be defined in a `.d.ts`. Maybe this could be stored as a `data-` attribute on `<html>` in `documentHeader.tpl`? This one is also easy to access as `document.documentElement`.

ts/WoltLabSuite/Core/Acp/Ui/Page/Menu/Main/Backend.ts
```diff
@@ -0,0 +1,57 @@
+import { MenuItem, PageMenuMainProvider } from "../../../../../Ui/Page/Menu/Main/Provider";
```
TimWolla, line 1: Missing module level comment.

ts/WoltLabSuite/Core/Bootstrap.ts
```diff
@@ -30,9 +30,11 @@ import * as UiEmpty from "./Ui/Empty";
 import * as UiObjectAction from "./Ui/Object/Action";
 import * as UiObjectActionDelete from "./Ui/Object/Action/Delete";
 import * as UiObjectActionToggle from "./Ui/Object/Action/Toggle";
+import { init as initSearch } from "./Ui/Search";
 
 // perfectScrollbar does not need to be bound anywhere, it just has to be loaded for WCF.js
 import "perfect-scrollbar";
+import { PageMenuMainProvider } from "./Ui/Page/Menu/Main/Provider";
```
TimWolla, line 37: This is an odd placement for this `import`. I'd move this up above perfect scrollbar.

ts/WoltLabSuite/Core/Ui/Page/Menu/Container.ts
```diff
@@ -0,0 +1,110 @@
+import { PageMenuProvider } from "./Provider";
```
TimWolla, line 1: Missing module level comment.

ts/WoltLabSuite/Core/Ui/Page/Menu/Main/Frontend.ts
```diff
@@ -0,0 +1,53 @@
+import { MenuItem, MenuItemDepth, PageMenuMainProvider } from "./Provider";
```
TimWolla, line 1: Missing module level comment.

ts/WoltLabSuite/Core/Ui/Page/Menu/Main/Frontend.ts
```diff
@@ -0,0 +1,53 @@
...
+
+function normalizeMenuItem(menuItem: HTMLElement, depth: MenuItemDepth): MenuItem {
+  const anchor = menuItem.querySelector(".boxMenuLink") as HTMLAnchorElement;
+  const title = anchor.querySelector(".boxMenuLinkTitle")!.textContent as string;
+
+  let counter = 0;
+  const outstandingItems = anchor.querySelector(".boxMenuLinkOutstandingItems");
+  if (outstandingItems) {
+    counter = +outstandingItems.textContent!.replace(/[^0-9]/, "");
```
TimWolla, line 10: Use `parseInt(…, 10)`, please. This is much more visible when reading the code.

ts/WoltLabSuite/Core/Ui/Page/Menu/Main/Provider.ts
```diff
@@ -0,0 +1,14 @@
+export type MenuItemDepth = 0 | 1 | 2;
```
TimWolla, line 1: Missing module level comment.

ts/WoltLabSuite/Core/Ui/Page/Menu/Provider.ts
```diff
@@ -0,0 +1,13 @@
+export interface PageMenuProvider {
```
TimWolla, line 1: Missing module level comment. This file, being an interface, would probably also benefit from method-level comments.

ts/WoltLabSuite/Core/Ui/User/Menu/Manager.ts
```diff
@@ -94,6 +94,10 @@ function getView(provider: UserMenuProvider): UserMenuView {
   return views.get(provider)!;
 }
 
+export function getUserMenuProviders(): Set<UserMenuProvider> {
```
TimWolla, line 97: I wonder whether the return value can|should be a `readonly Set`?

ts/WoltLabSuite/Core/Ui/Search.ts
```diff
@@ -0,0 +1,192 @@
+/**
+ * Manages the sticky page header.
```
TimWolla, line 2: Is this accurate?

ts/WoltLabSuite/Core/Ui/Search.ts
```diff
@@ -0,0 +1,192 @@
...
+ *
+ * @author  Alexander Ebert
+ * @copyright  2001-2019 WoltLab GmbH
+ * @license  GNU Lesser General Public License <http://opensource.org/licenses/lgpl-license.php>
+ * @module  WoltLabSuite/Core/Ui/Page/Header/Fixed
```
TimWolla, line 7: This certainly isn't.

ts/WoltLabSuite/Core/Ui/Search.ts
```diff
@@ -0,0 +1,192 @@
...
+
+    if (searchButton.getAttribute("aria-expanded") === "true") {
+      closeSearch();
+
+      searchButton.setAttribute("aria-expanded", "false");
+    } else {
+      if (Environment.platform() === "ios") {
+        _scrollTop = document.body.scrollTop;
+        UiScreen.scrollDisable();
+      }
```
TimWolla, line 87: I assume this was copied from the old implementation and I believe this was because iOS tends to jump around the page. However this warrants a comment.

TimWolla (commented):
A purely diff-based review. Will install this branch and test around later.

com.woltlab.wcf/templates/headIncludeJavaScript.tpl
dtdesign: I thought about it, but placing it on `<html>` is a bit odd, because it may contain UTF-8. There is already the issue of the `<title>` being placed before the meta charset definition, adding another possible UTF-8 sequence before makes it even worse?

ts/WoltLabSuite/Core/Bootstrap.ts
dtdesign: Thanks, it was implicitly added by the IDE.

com.woltlab.wcf/templates/headIncludeJavaScript.tpl
TimWolla: Ah, indeed. Good point. Disregard my suggestion then. I don't have a suggestion for a better location then.

ts/WoltLabSuite/Core/Ui/Page/Menu/Provider.ts
```diff
@@ -1,13 +1,46 @@
...
 
+  /**
+   * Provides the button that is used to change the visibility of the overlay container.
+   */
   getMenuButton(): HTMLElement;
 
+  /**
+   * Suspends the activity of the container and returns any borrowed HTML elements
+   * to their previous position. It is primarily used to restore the UI in case the
+   * mobile view bcomes inactive.
```
jens1o, line 36: `bcomes` should read `becomes` :)

dtdesign: Good catch!

TimWolla (approved):

PR #4611, opened 2021-12-21
Add UploadHandler session cleanup

joshuaruesweg:
Since the sessions stick way longer than with the introduction of the system, it is necessary, to clean up old file uploads in the session storage.

With this patch, a field in the session saves the creation time of the StorageField in the session and delete the field after 24 hours automatically.

### DIFF
--- a/wcfsetup/install/files/lib/system/file/upload/UploadHandler.class.php
+++ b/wcfsetup/install/files/lib/system/file/upload/UploadHandler.class.php
@@ -33,11 +33,24 @@ class UploadHandler extends SingletonFactory
     const VALID_IMAGE_EXTENSIONS = ['jpeg', 'jpg', 'png', 'gif', 'webp'];
 
     /**
+     * @since 5.5
+     */
+    private const UPLOAD_FIELD_SESSION_TIMEOUT = 3600 * 24;
+
+    /**
      * Contains the registered upload fields.
      *
      * @var UploadField[]
      */
     protected $fields = [];
+
+    /**
+     * @inheritDoc
+     */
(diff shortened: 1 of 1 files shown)


wcfsetup/install/files/lib/system/file/upload/UploadHandler.class.php
```diff
@@ -430,6 +444,26 @@ private function getStorage()
...
+    /**
+     * @since 5.5
+     */
+    private function cleanupStorage(): void
+    {
+        $hasChanges = false;
+        $storage = $this->getStorage();
+
+        foreach ($storage as $internalID => $storageData) {
+            if (!isset($storageData['registered']) || $storageData['registered'] < \TIME_NOW - self::UPLOAD_FIELD_SESSION_TIMEOUT) {
```
TimWolla, line 456:
```suggestion
            if (!isset($storageData['registered']) || $storageData['registered'] < (\TIME_NOW - self::UPLOAD_FIELD_SESSION_TIMEOUT)) {
```

wcfsetup/install/files/lib/system/file/upload/UploadHandler.class.php
```diff
@@ -32,13 +32,27 @@ class UploadHandler extends SingletonFactory
      */
     const VALID_IMAGE_EXTENSIONS = ['jpeg', 'jpg', 'png', 'gif', 'webp'];
 
+    /**
+     * @var int
```
TimWolla, line 36:
```suggestion
```

TimWolla (approved):

PR #4620, opened 2022-01-04
Update `htmlspecialchars` flags for PHP 8.1 default

TimWolla:
This change:

- Encodes `'` as `&#039;`, whereas it previously was not touched.
- Inserts the Unicode replacement character instead of returning an empty
  string when an invalid UTF-8 sequence is passed.

The first change might slightly improve security, whereas the second change
might improve debugging.

see also: https://php.watch/versions/8.1/html-entity-default-value-changes


TimWolla:
This might break `StringUtil::encodeJSON()`, will have to check.

dtdesign (approved):

PR #4621, opened 2022-01-04
Updates the UI/UX for the list of outstanding updates

dtdesign:
The new UI is much cleaner and focused on the relevant actions. It is also much more condensed to allow more updates to be visible at once, providing a better experience than the previous UI that spaces out updates greatly.

image

(The version numbers had been modified for testing purposes.)


wcfsetup/install/files/acp/templates/packageUpdate.tpl
```diff
@@ -1,56 +1,51 @@
...
-	<button class="buttonPrimary">{lang}wcf.global.button.submit{/lang}</button>
+	<button class="buttonPrimary" id="packageUpdateSubmitButton">{lang}wcf.global.button.submit{/lang}</button>
 </div>
 
+<script data-relocate="true">
+	require(["WoltLabSuite/Core/Language", "WoltLabSuite/Core/Acp/Ui/Package/Update/Manager"], (Language, { setup }) => {
+		Language.addObject({
+			'wcf.acp.package.update.excludedPackages': '{jslang}wcf.acp.package.update.excludedPackages{/jslang}',
+			'wcf.acp.package.update.title': '{jslang}wcf.acp.package.update.title{/jslang}',
+			'wcf.acp.package.update.unauthorized': '{jslang}wcf.acp.package.update.unauthorized{/jslang}'
```
joshuaruesweg, line 44:
```suggestion
			'wcf.acp.package.update.unauthorized': '{jslang}wcf.acp.package.update.unauthorized{/jslang}',
```

ts/WoltLabSuite/Core/Acp/Ui/Package/Update/Manager.ts
```diff
@@ -0,0 +1,200 @@
...
+    });
+  }
+
+  private updateSubmitButtonState(): void {
+    let disableButton = true;
+    this.packages.forEach((checkbox) => {
+      if (!checkbox || checkbox.checked) {
+        disableButton = false;
+      }
+    });
```
TimWolla, line 80: This technically could be `.every(checkbox => checkbox?.checked)`, however that would require an additional `Array.from()`. I leave this up to you.

ts/WoltLabSuite/Core/Acp/Ui/Package/Update/Manager.ts
```diff
@@ -0,0 +1,200 @@
...
+
+    this.submitButton.disabled = disableButton;
+  }
+
+  private getSelectedPackages(): Record<string, string> {
+    const packages = {};
+    this.packages.forEach((checkbox, element) => {
+      if (checkbox === null || checkbox.checked) {
```
TimWolla, line 88: This condition is inconsistent with line 77 (`=== null` vs `!`).

ts/WoltLabSuite/Core/Acp/Ui/Package/Update/Manager.ts
```diff
@@ -0,0 +1,200 @@
...
+      this.submitButton.disabled = false;
+    }
+
+    if (response.type === ResponseType.AuthorizationRequired) {
+      this.promptCredentials(response.template);
+    } else if (response.type === ResponseType.Conflict) {
+      this.showConflict(response.template);
+    } else {
+      this.startInstallation(response.queueID);
+    }
```
TimWolla, line 127: As `response.type` is an enum it would be desirable if this was a `switch()` without a `default:`. This might allow the TypeScript compiler to catch unhandled cases.

wcfsetup/install/files/acp/templates/packageUpdate.tpl
```diff
@@ -1,56 +1,51 @@
...
+		{foreach from=$availableUpdates item=update}
+		<div class="contentItem packageUpdate" data-package="{$update[package]}" data-version="{$update[newVersion][packageVersion]}">
+			<div class="contentItemContent">
+				<div class="contentItemTitle">{$update[packageName]|language}</div>
+				<div class="contentItemDescription">
+					<small{if $update[newVersion][servers][0][packageUpdateServerID] === $woltlabUpdateServer->packageUpdateServerID} class="packageSearchAuthorWoltlab"{/if}>{$update[author]}</small>
+					<p class="packageUpdateAvailable">
+						{$update[packageVersion]}
+						→
+						</span>
```
TimWolla, line 18: This does not look like correct markup to me.

wcfsetup/install/files/acp/templates/packageUpdate.tpl
```diff
@@ -1,56 +1,51 @@
...
+	<div class="contentItemList packageUpdateList">
+		{foreach from=$availableUpdates item=update}
+		<div class="contentItem packageUpdate" data-package="{$update[package]}" data-version="{$update[newVersion][packageVersion]}">
+			<div class="contentItemContent">
+				<div class="contentItemTitle">{$update[packageName]|language}</div>
+				<div class="contentItemDescription">
+					<small{if $update[newVersion][servers][0][packageUpdateServerID] === $woltlabUpdateServer->packageUpdateServerID} class="packageSearchAuthorWoltlab"{/if}>{$update[author]}</small>
+					<p class="packageUpdateAvailable">
+						{$update[packageVersion]}
+						→
```
TimWolla, line 17: It would be desirable if the full `{$from} → {$to}` construct would be part of a language item to allow proper localization into languages where such an arrow notation does not exist.

ts/WoltLabSuite/Core/Acp/Ui/Package/Update/Manager.ts
dtdesign: This isn't equal, because `checkbox` might be `null` and thus it is implicitly checked. It would be much cleaner if I could do a return inside the condition. I don't think that using an array makes it much cleaner.

TimWolla: > This isn't equal, because checkbox might be null and thus it is implicitly checked.

Ah, indeed. I've messed up the condition when doing De Morgan's rule in my head. The suggestion primarily was about the `.every()` instead of `.forEach()`, but it's not a clear-cut case here. So leaving this up to you.

dtdesign: I'm leaving it as-is, because I wouldn't favor one over the other here.

TimWolla (approved):

PR #4622, opened 2022-01-04
Allow manual setting of dbName

SoftCreatR:
As of now, the database name depends on the name of the installation directory, which is pretty inflexible and somewhat unforeseen behavior, if you don't look at the source code.

In addition, the current behavior could also lead to error messages because, as far as I know, hyphens are not allowed in database names, while a directory name may contain hyphens.

This PR allows to optionally set `dbName` in the JSON configuration.

### DIFF
--- a/wcfsetup/install/files/lib/system/devtools/DevtoolsSetup.class.php
+++ b/wcfsetup/install/files/lib/system/devtools/DevtoolsSetup.class.php
@@ -61,15 +61,17 @@ class DevtoolsSetup extends SingletonFactory
             return null;
         }
 
-        // dirname return a single backslash on Windows if there are no parent directories
-        $dir = \dirname($_SERVER['SCRIPT_NAME']);
-        $dir = ($dir === '\\') ? '/' : FileUtil::addTrailingSlash($dir);
-        if ($dir === '/') {
-            throw new \RuntimeException("Refusing to install in the document root.");
-        }
+        if (empty($dbConfig['dbName'])) {
+            // dirname return a single backslash on Windows if there are no parent directories
+            $dir = \dirname($_SERVER['SCRIPT_NAME']);
+            $dir = ($dir === '\\') ? '/' : FileUtil::addTrailingSlash($dir);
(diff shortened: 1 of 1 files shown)


dtdesign:
This change makes no sense to me, providing the name via the config means that the database name is fixed. This goes straight against the intention of the dev installation (which, for the record, is mostly designed around _our_ needs) that is built around the promise of repeatable installations based on a pattern.

For example, the use of the directory names is anything but inflexible. I have a habit of grouping installations per generation, so you would end up with paths like `/55/6/` which would be the 6th installation based on the WoltLab Suite 5.5. This also plays nicely with phpMyAdmin which can automatically group those database by interpreting `55` as a prefix.

SoftCreatR:
The proposed change doesn't change the way, how you use it. You can still omit the database name from the config, but with this change, you can optionally set any database name, you like. So, no breaking changes, just slightly more flexibility.

dtdesign:
I perfectly understand what your proposed changed does, it just doesn't make sense in the context of the dev installation. It is laser-focused on a single specific purpose: Install the software based on patterns for development purposes.

Your suggested change does not fit into this and I'm hesitant to introduce changes that go against this goal. We're maintaining this specific bit of the software for our own needs. It is not documented for a reason, we will make changes to it whenever _we_ see fit. Please take your time and explain your reasons behind this as it helps us to better understand how you are going to benefit from this. There is little point in having this change introduced just for us to make a different changer later that break your workflow again.

mutec:
Is there any real argument against implementing this which might help running installations more easy in some cases?

As already said the current behavior will not be changed, but extended by the possibility to use a generated config-file using a database-name that might match another pattern? E.g. using `test_*` since the user `test` has access to the databases `test` and `test_*` which means no new user has to be setup or modified before.

SoftCreatR:
Needles to say, that even if it's designed to fit your needs, it might also be used by others. However, i personally don't care much since I already perform a completely headless installations without this whole thing.

PR #4623, opened 2022-01-06
Fix PHP 8.1 compatibility when saving I18n options

TimWolla:
Really not sure whether this is the best way to fix this, or if a special
`null` check should be added to each option type. It's really evident here that
the option handling shows its age when looking at control and data flow.

------

As I18n options are special-cased, they will not be provided in `rawValues`,
thus passing `null` to `->getData()`, which the option types are not prepared
to handle. Before PHP 8.1 this was implicitly treated as an empty string, with
the types introduced to native functions, e.g. `explode()` or `preg_replace()`
this will result in an error.

### DIFF
--- a/wcfsetup/install/files/lib/system/option/OptionHandler.class.php
+++ b/wcfsetup/install/files/lib/system/option/OptionHandler.class.php
@@ -393,18 +393,22 @@ class OptionHandler implements IOptionHandler
         // get new value
         $newValue = $this->rawValues[$option->optionName] ?? null;
 
-        // get save value
-        $this->optionValues[$option->optionName] = $typeObj->getData($option, $newValue);
+        // Skip validation for I18n options. They are special cased and are not provided
+        // within the `rawValues`, thus causing validation to be effectively useless.
+        if (!$this->supportI18n || !$option->supportI18n) {
+            // get save value
+            $this->optionValues[$option->optionName] = $typeObj->getData($option, $newValue);
 
-        // validate with pattern
-        if ($option->validationPattern) {
-            if (
-                !\preg_match(
-                    '~' . \str_replace('~', '\~', $option->validationPattern) . '~',
-                    $this->optionValues[$option->optionName]
-                )
-            ) {
-                throw new UserInputException($option->optionName, 'validationFailed');
+            // validate with pattern
+            if ($option->validationPattern) {
+                if (
+                    !\preg_match(
+                        '~' . \str_replace('~', '\~', $option->validationPattern) . '~',
+                        $this->optionValues[$option->optionName]
+                    )
+                ) {
+                    throw new UserInputException($option->optionName, 'validationFailed');
+                }
             }
         }
 


wcfsetup/install/files/lib/system/option/OptionHandler.class.php
```diff
@@ -393,18 +393,22 @@ protected function validateOption(Option $option)
...
-            if (
-                !\preg_match(
-                    '~' . \str_replace('~', '\~', $option->validationPattern) . '~',
-                    $this->optionValues[$option->optionName]
-                )
-            ) {
-                throw new UserInputException($option->optionName, 'validationFailed');
+        // Skip validation for I18n options. They are special cased and are not provided
+        // within the `rawValues`, thus causing validation to be effectively useless.
+        if (!$this->supportI18n || !$option->supportI18n) {
```
dtdesign, line 398:
```suggestion
        if (!$this->supportI18n && !$option->supportI18n) {
```
The comment implies that this is an invalid code path for i18n options.

dtdesign (changes requested):

wcfsetup/install/files/lib/system/option/OptionHandler.class.php
TimWolla: The condition is correct:

- I18n is supported when `$this->supportI18n && $option->supportI18n` (this is the condition that is used everywhere else).
- We need to invert the full condition here.
- According to De Morgan's rules we need to invert each condition and replace the `&&` by `||`: `!$this->supportI18n || !$option->supportI18n`

dtdesign: I know, it is just that `$this->supportI18n && $option->supportI18n` looks more like a defense in depth. I'm not sure what happens when `$option->supportI18n` is actually `true` while `$this->supportI18n` is falsy.

TimWolla: I have no idea either, this code is not easy to follow. But now I understand what you're hinting at.

Maybe instead of defaulting `$newValue` to `null` it should be defaulted to `''` as a less invasive fix? I can also add explicit string casts where necessary to each option type. Then the behavior should be fully identical to PHP < 8.1.

dtdesign: > Maybe instead of defaulting `$newValue` to `null` it should be defaulted to `''` as a less invasive fix?

Yes, that sounds reasonable. It is still a bit strange, but more matches the behavior of older PHP versions. I'm not aware of any code explicitly relying on checks against `null`.

TimWolla: Okay, I've done this. I've intentionally kept the previous attempt and reverted it with an appropriate message, so that it will be visible in the git history.

dtdesign (approved):

PR #4627, opened 2022-01-08
Add size detection for WebP smileys

SoftCreatR:
As of now, `Smiley::getDimensions()` only returns sizes for gif, jpg & png files. It would be good, if webp files are also supported.


dtdesign (approved):

dtdesign:
Thanks!

PR #4631, opened 2022-01-11
Uninstalling from the package details

felix-d1strict:
Just a QOL-Feature: Adds a button to uninstall a package on the details page. 
1:1 like the following plugin: https://www.woltlab.com/pluginstore/file/7106-uninstall-button/

Maybe a feature for the 5.5? Otherwise please move to the correct branch/tag.

### DIFF
--- a/wcfsetup/install/files/acp/templates/package.tpl
+++ b/wcfsetup/install/files/acp/templates/package.tpl
@@ -18,7 +18,15 @@
 	{hascontent}
 		<nav class="contentHeaderNavigation">
 			<ul>
-				{content}{event name='contentHeaderNavigation'}{/content}
+				{content}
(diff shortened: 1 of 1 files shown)


xopez:
I like the idea to integrate it directly.

dtdesign:
Thanks for you work! I have decided to create an own commit instead, for three reasons:
* The button is not available when the package cannot be uninstalled, this is more in line with the behavior of these kind of buttons.
* Removed the tooltip which was part of the icon-style button in the package list.
* Uses a `<button>` element for better accessibility.

felix-d1strict:
All right, thanks for the feedback. I have incorporated your change in the above plugin and excluded the WSC 5.5. Thanks for the quick processing.

Issue #4616, opened 2022-01-01
Secure Installer from unauthenticated access


GodMod:
The current Installer of WCF allows the usage of an external Database Server, without proofing if the installing user is the owner of this Database Server. This can result in overtaking the webspace (install the WCF with external database > login as admin > install plugin with webshell > access to OS), if the user just uploads the Installer, but does not install the WCF or is just doing it.

Other CMS like Joomla (see https://nvd.nist.gov/vuln/detail/CVE-2017-11364) have already reacted and implemented additional checks, like writing a file that a user must delete in order to continue the installation, if he uses an external database server (see https://docs.joomla.org/J3.x:Secured_procedure_for_installing_Joomla_with_a_remote_database).

More information about this issue can be found here: https://media.defcon.org/DEF%20CON%2025/DEF%20CON%2025%20presentations/DEF%20CON%2025%20-%20Hanno-Boeck-Abusing-Certificate-Transparency-Logs-UPDATED.pdf

GodMod:
Another reference, which describes how this issue is automatically exploited for Wordpress instances: https://smitka.me/2022/07/01/wordpress-installer-attack-race/

PR #4636, opened 2022-01-15
Introducing the UltimateWoltlabMaster

Justman10000:


syntafin:
Please... no!

UltimateWoltlabMaster.sh
```diff
@@ -0,0 +1,123 @@
...
+echo "Zurück [3]"
+read WhatDo
+
+clear
+
+FileName=WoltlabSuite.zip
+
+case $WhatDo in
+    "1")
+        wget https://assets.woltlab.com/release/woltlab-suite-5.4.10.zip
```
felix-d1strict, line 18: The version is hardcoded. It would be better to implement a check to always download the latest version of the WSC. Or alternatively, that you can choose a version yourself.

UltimateWoltlabMaster.sh
```diff
@@ -0,0 +1,123 @@
+reset
+
+echo "UltimateWoltlabMaster von Justman10000"
+echo "Dies ist unter der GNU General Public License v3.0 lizenziert"
+
+echo "Was soll getan werden?"
+echo "Installieren [1]"
+echo "Deinstallieren [2]"
+echo "Zurück [3]"
```
felix-d1strict, line 9: The texts are only in German. However, since the WSC is multilingual (at least German & English by default), English should also be available.

UltimateWoltlabMaster.sh
```diff
@@ -0,0 +1,123 @@
...
+        mv woltlab-suite-5.4.10.zip $FileName
+        sleep 1
+        unzip $FileName
+        sleep 1
+        rm -r *.txt
+        sleep 1
+
+        clear
+
+        echo "In welchem Pfad soll WoltLab installiert werden?"
```
felix-d1strict, line 29: The texts are only in German. However, since the WSC is multilingual (at least German & English by default), English should also be available.

UltimateWoltlabMaster.sh
```diff
@@ -0,0 +1,123 @@
...
+
+        echo "Soll die test.php Datei beim Aufruf der URL aufgerufen werden? Dadurch kann Woltlab installiert werden,"
+        echo "ohne die install.php oder test.php direkt aufrufen zu müssen"
+        sleep 2
+        echo "-------------------------------------------------------------------------------------------------------"
+        echo "Warnung, sollte die Installation nicht direkt durchgeführt werden, wird die Antwort Nein empfohlen"
+        echo "Siehe: https://www.woltlab.com/community/thread/293784-test-php-index-php"
+        sleep 2
+        echo "-------------------------------------------------------------------------------------------------------"
+        echo "Ja   [1]"
```
felix-d1strict, line 62: The texts are only in German. However, since the WSC is multilingual (at least German & English by default), English should also be available.

UltimateWoltlabMaster.sh
```diff
@@ -0,0 +1,123 @@
...
+        case $Rename in
+            "1")
+                mv $WhereInstall/test.php $WhereInstall/index.php
+            ;;
+
+            "2")
+            ;;
+
+            *)
+                echo "Ungültige Antwort"
```
felix-d1strict, line 77: The texts are only in German. However, since the WSC is multilingual (at least German & English by default), English should also be available.

UltimateWoltlabMaster.sh
```diff
@@ -0,0 +1,123 @@
...
+
+        bash UltimateWoltlabMaster.sh
+    ;;
+
+    "3")
+        exit
+    ;;
+
+    *)
+        echo "Ungültige Antwort"
```
felix-d1strict, line 116: The texts are only in German. However, since the WSC is multilingual (at least German & English by default), English should also be available.

UltimateWoltlabMaster.sh
```diff
@@ -0,0 +1,123 @@
...
+        read WhereInstalled
+
+        clear
+
+        rm -r $WhereInstalled/*
+        sleep 1
+        rm -r $WhereInstalled
+
+        echo "Bedenke, dass dieses Script nicht die Datenbank lerren oder entfernen kann"
+        echo "Tue dies daher manuell..."
```
felix-d1strict, line 105: The texts are only in German. However, since the WSC is multilingual (at least German & English by default), English should also be available.

UltimateWoltlabMaster.sh
```diff
@@ -0,0 +1,123 @@
...
+        rm -r $FileName
+        sleep 1
+        chmod -R 777 $WhereInstall/*
+        sleep 2
+
+        bash UltimateWoltlabMaster.sh
+    ;;
+
+    "2")
+        echo "In welchem Verzeichnis ist Woltlab Installiert?"
```
felix-d1strict, line 95: The texts are only in German. However, since the WSC is multilingual (at least German & English by default), English should also be available.

UltimateWoltlabMaster.sh
felix-d1strict, line 9: Also, you could try keeping your script a little smaller by using `echo -e "text"` instead of creating a new `echo` for each new line.

UltimateWoltlabMaster.sh
```diff
@@ -0,0 +1,123 @@
...
+        echo "ohne die install.php oder test.php direkt aufrufen zu müssen"
+        sleep 2
+        echo "-------------------------------------------------------------------------------------------------------"
+        echo "Warnung, sollte die Installation nicht direkt durchgeführt werden, wird die Antwort Nein empfohlen"
+        echo "Siehe: https://www.woltlab.com/community/thread/293784-test-php-index-php"
+        sleep 2
+        echo "-------------------------------------------------------------------------------------------------------"
+        echo "Ja   [1]"
+        echo "Nein [2]"
+        read Rename
```
felix-d1strict, line 64: Also, you could try keeping your script a little smaller by using `echo -e "text"` instead of creating a new `echo` for each new line.

UltimateWoltlabMaster.sh
felix-d1strict, line 105: Also, you could try keeping your script a little smaller by using `echo -e "text"` instead of creating a new `echo` for each new line.

felix-d1strict (commented):
An installation does not take place here. It only downloads an archive & moves its contents. I find that this script is rather unnecessary, because you can download & unpack the WSC yourself with a few clicks. The actual installation I have to do with this script manually anyway.
I have now only superficially looked over it. However, a shell validator spits out a few errors for me.

syntafin:
I think the script does nothing you can do faster from hand, the only thing it does is renaming the ``test.php`` to ``index.php``.
As long as there is no prevention to use an external database, I think this is a security problem.

felix-d1strict:
> As long as there is no prevention to use an external database, I think this is a security problem.

Well, that is a [problem of the WSC](https://github.com/WoltLab/WCF/issues/4616) itself. After all, someone who knows the WSC also knows that you can "call" the installation routine via /install.php. ^^

syntafin:
> someone who knows the WSC also knows that you can "call" the installation routine via /install.php. ^^

correct, but the most attacks are automated and a bot will mostly call the ``index.php`` or things like ``wp-install.php`` (idk if this file even exists, I see this so many times in my Logs).

Justman10000:
> I think the script does nothing you can do faster from hand, the only thing it does is renaming the `test.php` to `index.php`. As long as there is no prevention to use an external database, I think this is a security problem.

Die Sache ist, dass man ja selbst enscheiden kann, ob das Script die Datei umbenennt, wählt man bei der Frage 2, tut das Script in diesem Punkt nichts, lässt also die Datei bei test.php

Justman10000:
> An installation does not take place here. It only downloads an archive & moves its contents. I find that this script is rather unnecessary, because you can download & unpack the WSC yourself with a few clicks. The actual installation I have to do with this script manually anyway. I have now only superficially looked over it.

Maybe I could see if I could implement a full installation process in the script... However, the point is that the script takes the work out of it, even if it's not a lot!

> However, a shell validator spits out a few errors for me.

And which would be?

felix-d1strict:
> The thing is that you can decide for yourself whether the script renames the file, if you choose question 2, the script does nothing at this point, so leave the file at test.php

I would leave that part out, just for the reasons mentioned above. This more or less prevents a security problem from arising in the first place.

> Maybe I could see if I could implement a full installation process in the script... However, the point is that the script takes the work out of it, even if it's not a lot!

You name it, it's not really much work. In my opinion, an administrator should know how to download the WSC and unpack it anyway. From there, I still have the same opinion as above, that this script is unnecessary. 
What difference does it make if I download and run the shell – or if I download, unzip and run the WSC. The latter is probably even faster because I don't have to go through 3 prompts.

> And which would be?

You're the developer. ;-)
You should run your code through a validator first. Here is an example: https://www.shellcheck.net/ 
In that case, it's an error, a warning. Of course, you can implement the "improvement information". :)

dtdesign:
Thank you for your efforts and taking your time to share this script with others.

Unfortunately, this script is currently out of scope for what we want to achieve with the framework. It does not provide immediate value for the end customers, because while you deem some parts of the archive to be irrelevant (such as the readme files), they are placed there for a reason. Your script offers _some_ benefit to developers, but at the end of the day does not really accelerate the part that really consumes the most time: The installation itself.

Please feel free to share any ideas you have with us and we can see if we can work out something in the future. However, such a script as proposed here does not fit into our general goals.

PR #4638, opened 2022-01-16
Add `var_dump` to allowed enterprise functions

Krymonota:
This function is used in ACP templates of some of my VieCode Shop plugins that are labelled as "Compatible with WoltLab Cloud" to display API responses from payment service providers.


TimWolla:
Thanks, merged. FWIW: `print_r` is also whitelisted and might be a little more user-friendly for scalar values.

PR #4639, opened 2022-01-16
implement optional forcing decimals in `StringUtil.formatNumeric()`

mutec:
With this change the current behavior of `StringUtil.formatNumeric()` is kept als default behavior, but it is possible to force the given number of decimals instead of treating the parameter as `$maxDecimals`.
This change makes it possible to get the same result as using `CurrencyModifierTemplatePlugin` in order to display things in a more consistent way.

see https://www.woltlab.com/community/thread/294033/

### DIFF
--- a/ts/WoltLabSuite/Core/StringUtil.ts
+++ b/ts/WoltLabSuite/Core/StringUtil.ts
@@ -41,13 +41,14 @@ export function escapeRegExp(string: string): string {
 /**
  * Rounds number to given count of floating point digits, localizes decimal-point and inserts thousands separators.
  */
-export function formatNumeric(number: number, decimalPlaces?: number): string {
+export function formatNumeric(number: number, decimalPlaces?: number, forceDecimalPlaces = false): string {
   let tmp = NumberUtil.round(number, decimalPlaces || -2).toString();
   const numberParts = tmp.split(".");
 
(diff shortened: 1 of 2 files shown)


TimWolla:
Please apologize the delay in reviewing this. I wasn't super happy with the proposed API and needed some time to think about this.

We're not merging this and likely will completely deprecate `StringUtil.formatNumeric()` in favor of using JavaScript's [`Intl.NumberFormat`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat) API that is available in all modern browsers and exposes a nice and powerful API.

Please use something similar to the following:

```
const formatter = new Intl.NumberFormat(document.documentElement.lang, {minimumFractionDigits: 2}):
formatter.format(1.00);
```

The Intl API even is capable of properly formatting currencies:

```
const formatter = new Intl.NumberFormat(document.documentElement.lang, { style: 'currency', currency: 'EUR' });
```

mutec:
@TimWolla thanks. I'm going to implement this in the 5.4 version already so it doesn't get deprecated in a few months already. :)

TimWolla:
> I'm going to implement this in the 5.4 version already

Yeah, that's fine. The Intl API is stable and clearly solves your needs.

PR #4642, opened 2022-01-18
Fix the replacing of WCF_N in PHP DDL during app installation

TimWolla:
During app installation the newly installed app might not yet be stored within
the application cache, thus failing to replace the `1` within the table
structure definition.

Fix this by setting the `skipCache` parameter to `true`. This will increase the
number of database queries, because applications will be checked once for each
defined table and for each defined FOREIGN KEY, but I don't see a simple fix
for this issue that avoids this increase in query count. Specifically we cannot
simply reset the application cache after inserting the application into
wcf1_application.

### DIFF
--- a/wcfsetup/install/files/lib/system/database/table/DatabaseTable.class.php
+++ b/wcfsetup/install/files/lib/system/database/table/DatabaseTable.class.php
@@ -49,7 +49,7 @@ class DatabaseTable {
 	 * @param	string		$name	name of the database table
 	 */
 	protected function __construct($name) {
-		$this->name = ApplicationHandler::insertRealDatabaseTableNames($name);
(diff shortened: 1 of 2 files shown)


TimWolla:
> The method `ApplicationHandler::insertRealDatabaseTableNames()` is designed for use during the installation/update

Yes, however I already create an issue to merge the logic with the `->prepare()` method: #4641 

> Wouldn't it make sense to change it into a runtime cache based on a local `static` instead? This makes the behavior consistent for the lifetime of the request and gets rid of the persistent cache.

Not sure whether this would fix the issue: As all the PIPs run a single request, it might (or might not?) be possible that the runtime cache already is filled with outdated data, then the application is inserted, then the method is used again. It's likely not going to be worse than that static cache, though. The `$skipCache` parameter should be preserved, though.

dtdesign (approved):

PR #4656, opened 2022-02-02
Fix rendering of alert boxes at the top of a dialog

TimWolla:


dtdesign (approved):

joshuaruesweg (approved):
Overall LGTM. 

Please take a look in the current templates / stylesheets an adjust extra CSS which avoids the problem (e.g. `removeUserContentDialog.tpl`).

PR #4658, opened 2022-02-03
Overhauled handling of dynamic phrases in templates

### LINKED ISSUE #4657: Handling of dynamic phrases in templates
The `|language` modifier was added to allow the piping of the phrase through other functions. This has some unwanted side effects when used with plain strings that should not support variable interpolation. Another difference to `{lang}` is the evaluation on runtime rather than at compile time, allowing the phrase to be taken from a variable instead.

In general it is often used in places where a proper method exists, such as `{$obj->title|language}` instead of `{$obj->getTitle()}` and is more of a legacy leftover rather than an intentional use. However, we cannot change the semantics of the `|language` modifier without breaking compatibility, therefore we will introduce a new modifier and phase out the old one.

* [x] Introduce the new modifier `|phrase` as a thin wrapper around `\wcf\system::WCF::getLanguage()->get()`.
* [x] Deprecate the `|language` modifier and remove it in 6.0.
* [x] Replace existing uses of `|language` with the proper method calls instead, for example, the aforementioned `{$obj->getTitle()}`.

Related to #2976, #4107

BurntimeX:
Closes #4657

### DIFF
--- a/com.woltlab.wcf/templates/email_html.tpl
+++ b/com.woltlab.wcf/templates/email_html.tpl
@@ -144,7 +144,7 @@
 	</head>
 	<body>
 	{capture assign='header'}
-	<h1>{@PAGE_TITLE|phrase}</h1>
+	<h1>{PAGE_TITLE|phrase}</h1>
 	{/capture}
 	{include file='email_paddingHelper' block=true class='header' content=$header sandbox=true}
 
--- a/com.woltlab.wcf/templates/email_html.tpl
+++ b/com.woltlab.wcf/templates/email_html.tpl
@@ -144,7 +144,7 @@
 	</head>
 	<body>
 	{capture assign='header'}
-	<h1>{@PAGE_TITLE|language}</h1>
+	<h1>{@PAGE_TITLE|phrase}</h1>
 	{/capture}
 	{include file='email_paddingHelper' block=true class='header' content=$header sandbox=true}
 
@@ -158,10 +158,10 @@
 	{hascontent}
 	<span style="font-size: 0;">-- <br></span>
 	{content}
-	{if MAIL_SIGNATURE_HTML|language}
-	{@MAIL_SIGNATURE_HTML|language}
+	{if MAIL_SIGNATURE_HTML|phrase}
+	{@MAIL_SIGNATURE_HTML|phrase}
 	{else}
-	{@MAIL_SIGNATURE|language|newlineToBreak}
+	{@MAIL_SIGNATURE|phrase|newlineToBreak}
 	{/if}
 	{/content}{/hascontent}{/capture}
 	{include file='email_paddingHelper' block=true class='footer' content=$footer sandbox=true}
--- a/wcfsetup/install/files/acp/templates/header.tpl
+++ b/wcfsetup/install/files/acp/templates/header.tpl
@@ -4,7 +4,7 @@
 	<meta charset="utf-8">
 	<meta name="viewport" content="width=device-width, initial-scale=1">
 	<meta name="robots" content="noindex">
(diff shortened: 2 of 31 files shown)


com.woltlab.wcf/templates/email_html.tpl
```diff
@@ -144,7 +144,7 @@
 	</head>
 	<body>
 	{capture assign='header'}
-	<h1>{@PAGE_TITLE|language}</h1>
+	<h1>{@PAGE_TITLE|phrase}</h1>
```
TimWolla, line 147: The `@` should be removed here.

wcfsetup/install/files/acp/templates/header.tpl
```diff
@@ -4,7 +4,7 @@
 	<meta charset="utf-8">
 	<meta name="viewport" content="width=device-width, initial-scale=1">
 	<meta name="robots" content="noindex">
-	<title>{if $pageTitle|isset}{@$pageTitle|language} - {/if}{jslang}wcf.global.acp{/jslang}{if PACKAGE_ID} - {PAGE_TITLE|language}{/if}</title>
+	<title>{if $pageTitle|isset}{@$pageTitle|phrase} - {/if}{jslang}wcf.global.acp{/jslang}{if PACKAGE_ID} - {PAGE_TITLE|phrase}{/if}</title>
```
TimWolla, line 7: The `@` in front of the `$pageTitle` should be removed.

wcfsetup/install/lang/de.xml
```diff
@@ -3235,7 +3235,7 @@ aus Sicherheitsgründen erforderlich, dass {if LANGUAGE_USE_INFORMAL_VARIANT}du{
 		<item name="wcf.acp.user.sendNewPassword.mail.html.intro"><![CDATA[
 <p>ein Administrator hat {if LANGUAGE_USE_INFORMAL_VARIANT}dein{else}Ihr{/if} Kennwort zurückgesetzt. Es ist nun
 erforderlich, dass {if LANGUAGE_USE_INFORMAL_VARIANT}du{else}Sie{/if} ein neues Kennwort {if LANGUAGE_USE_INFORMAL_VARIANT}setzt{else}setzen{/if}, damit {if LANGUAGE_USE_INFORMAL_VARIANT}du dein {else}Sie Ihr {/if}
-Benutzerkonto {$mailbox->getUser()->username} auf der Seite <a href="{link isHtmlEmail=true}{/link}">{@PAGE_TITLE|language}</a>
+Benutzerkonto {$mailbox->getUser()->username} auf der Seite <a href="{link isHtmlEmail=true}{/link}">{@PAGE_TITLE|phrase}</a>
```
TimWolla, line 3238: The `@` should be removed.

wcfsetup/install/lang/en.xml
```diff
@@ -3163,7 +3163,7 @@ the lost password form [URL:{link controller='LostPassword' isEmail=true}{/link}
 		<item name="wcf.acp.user.sendNewPassword.mail.html.headline"><![CDATA[Dear {$mailbox->getUser()->username},]]></item>
 		<item name="wcf.acp.user.sendNewPassword.mail.html.intro"><![CDATA[
 <p>An administrator has reset your password. You are now required to set a new password to be able to use your
-user account {$mailbox->getUser()->username} on the website <a href="{link isHtmlEmail=true}{/link}">{@PAGE_TITLE|language}</a> again:</p>]]></item>
+user account {$mailbox->getUser()->username} on the website <a href="{link isHtmlEmail=true}{/link}">{@PAGE_TITLE|phrase}</a> again:</p>]]></item>
```
TimWolla, line 3166: `@`

TimWolla (commented):
I'll leave it up to you whether or not to remove the `@` in 5.4 or lower. All the values are trustworthy.

TimWolla (approved):

PR #4662, opened 2022-02-07
Rework Poll JS to TypeScript

joshuaruesweg:


ts/WoltLabSuite/Core/Ui/Poll/Manager/Vote.ts
```diff
@@ -0,0 +1,131 @@
...
+    }
+  }
+
+  private getSelectedOptions(): number[] {
+    const numbers = new Array<number>();
+    this.inputs.forEach((input) => {
+      if (input.checked) {
+        numbers.push(parseInt(input.value, 10));
+      }
+    });
```
TimWolla, line 89:
```
Array.from(this.inputs).filter(input => input.checked).map(input => parseInt(input.value, 10));
```

TimWolla (commented):
Can't comment to the overall changes, just some remarks regarding localized changes.

ts/WoltLabSuite/Core/Ui/Poll/Manager/Manager.ts
```diff
@@ -0,0 +1,125 @@
...
+    const innerContainer = this.poll.querySelector<HTMLElement>(".pollInnerContainer");
+
+    if (!innerContainer) {
+      throw new Error(`Could not find inner container for poll "${this.pollID}"`);
+    }
+
+    return innerContainer;
+  }
+
+  protected setInnerContainer(html: string): void {
```
dtdesign, line 109: The choice between `private` and `protected` appears arbitrary, better just stick to `private`?

ts/WoltLabSuite/Core/Ui/Poll/Manager/Manager.ts
```diff
@@ -0,0 +1,125 @@
...
+      this.participants = new Participants(this);
+      this.participants.showButton();
+    }
+  }
+
+  private canViewParticipants(): boolean {
+    return this.canViewResults && this.isPublic;
+  }
+
+  private getInnerContainer(): HTMLElement {
```
dtdesign, line 99: This is not a good solution, there is no reason to inject arbitrary HTML every time the view is changed. Either set the inactive containers to `hidden` or at the very least recycle the DOM elements by appending and removing them at will.

Also, the check doesn't make much sense, the `setInnerContainer()` method should handle it and reject invalid contents immediately.

ts/WoltLabSuite/Core/Ui/Poll/Manager/View/Abstract.ts
```diff
@@ -0,0 +1,89 @@
...
+  protected button: HTMLButtonElement;
+
+  public constructor(manager: Manager) {
+    this.pollManager = manager;
+
+    this.initButton();
+  }
+
+  private apiCall(actionName: string, data?: RequestOptions): void {
+    const request = new AjaxRequest({
```
dtdesign, line 27: You are rewriting the JS API from the ground up, why not take your time and throw away the old garbage `wcf\action\PollAction`? These manual invocations are a huge mess.

ts/WoltLabSuite/Core/Ui/Poll/Manager/View/Abstract.ts
```diff
@@ -0,0 +1,89 @@
...
+ * @module  WoltLabSuite/Core/Ui/Poll/Manager/View/Abstract
+ * @since   5.5
+ */
+
+import * as Core from "../../../../Core";
+import AjaxRequest from "../../../../Ajax/Request";
+import { RequestOptions, ResponseData } from "../../../../Ajax/Data";
+import Manager, { PollViews } from "../Manager";
+
+export abstract class Abstract {
```
dtdesign, line 16: I don't think that this module is required. If you just need 2 or 3 helper functions, just use a "util" style module with exports and go for composition for the views.

ts/WoltLabSuite/Core/Ui/Poll/Manager/View/Participants.ts
```diff
@@ -0,0 +1,64 @@
...
+ * @since   5.5
+ */
+
+import Manager from "../Manager";
+import UiUserList from "../../../User/List";
+
+export class Participants {
+  protected readonly pollManager: Manager;
+  private button: HTMLButtonElement;
+  private userList?: UiUserList;
```
dtdesign, line 17:
```suggestion
  private userList?: UiUserList = undefined;
```

Defining properties on runtime is somewhat error-prone and has the drawback of nuking the prototype inside the JS engine. Please feel free to compare the difference it makes in the JS output to understand what is happening with this change.

ts/WoltLabSuite/Core/Ui/Poll/Manager/Vote.ts
```diff
@@ -0,0 +1,122 @@
...
+import Manager, { PollViews } from "./Manager";
+import { ResponseData } from "../../../Ajax/Data";
+import * as Core from "../../../Core";
+
+export class Vote {
+  private pollManager: Manager;
+
+  private button: HTMLButtonElement;
+
+  private inputs: NodeListOf<HTMLInputElement>;
```
dtdesign, line 21: I would use an array instead, it isn't a live list anyway so there is no value in not converting it to an array right away.

ts/WoltLabSuite/Core/Ui/Poll/Manager/Vote.ts
```diff
@@ -0,0 +1,122 @@
...
+ * @since   5.5
+ */
+
+import AjaxRequest from "../../../Ajax/Request";
+import Manager, { PollViews } from "./Manager";
+import { ResponseData } from "../../../Ajax/Data";
+import * as Core from "../../../Core";
+
+export class Vote {
+  private pollManager: Manager;
```
dtdesign, line 17:
```suggestion
  private readonly pollManager: Manager;
```

ts/WoltLabSuite/Core/Ui/Poll/Manager/Vote.ts
```diff
@@ -0,0 +1,122 @@
...
+  private checkInputs(): void {
+    let selectedInputCount = 0;
+    this.inputs.forEach((input) => {
+      if (input.checked) {
+        selectedInputCount++;
+      }
+
+      if (this.pollManager.maxVotes > 1) {
+        input.disabled = false;
+      }
```
dtdesign, line 65: This should be done exactly once during the initialization instead of with every invocation.

dtdesign (changes requested):

com.woltlab.wcf/templates/poll.tpl
```diff
@@ -1,16 +1,22 @@
...
+<script data-relocate="true">
+	require(['WoltLabSuite/Core/Ui/Poll/Manager/Manager'], function({ Manager }) {
+		new Manager(
+			{@$poll->pollID},
+			{if $poll->canSeeResult()}true{else}false{/if},
+			{if $poll->canVote()}true{else}false{/if},
+			{if $poll->isPublic}true{else}false{/if},
+			{@$poll->maxVotes},
+			"{$poll->question}"
+		);
```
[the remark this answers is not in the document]
joshuaruesweg: I don't like the mapping of the parameters into the DOM. What is the benefit from this? The dataset attributes are not currently used with the new implementation either way, so I might remove them.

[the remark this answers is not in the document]
dtdesign: We're already doing this and it has its upsides:
* **You can style elements based on the attribute values.**
  We're already doing this with the new user menu that relies on the aria attributes to toggle states. Apart from this, you could easily style non public polls in a different fashion simply by targeting the `[data-is-public="true"]` selector.
* **Lazy initialization of modules.**
   The current approaches around the `DomNodeInsertedHandler` are not only cumbersome, but they also require us to load modules ahead of time. Relying on simple attributes allow us to migrate this into a lazy initialization that does not require a `<script>` tag to supply the values.
* **Similar to the behavior of custom elements.**
  This overlaps a bit with the previous bullet point, but custom elements are something I can imagine using in the future (including the shadow DOM). Relying on attributes to initialize the elements and tracking their state via these attributes is considered to be a good practice.

ts/WoltLabSuite/Core/Ui/Poll/Manager/Poll.ts
```diff
@@ -0,0 +1,205 @@
+/**
+ * Handles the poll UI.
+ *
+ * @author  Joshua Ruesweg
+ * @copyright  2001-2022 WoltLab GmbH
+ * @license  GNU Lesser General Public License <http://opensource.org/licenses/lgpl-license.php>
+ * @module  WoltLabSuite/Core/Ui/Poll/Manager/Poll
```
dtdesign, line 7: Why isn't this just `WoltLabSuite/Core/Ui/Poll.ts`? What is the point of nesting this into `/Manager/`, it doesn't add any value.

ts/WoltLabSuite/Core/Ui/Poll/Manager/Poll.ts
```diff
@@ -0,0 +1,205 @@
...
+import VoteHandler from "./Vote";
+
+export enum PollViews {
+  vote = "vote",
+  results = "results",
+}
+
+export class Poll {
+  public readonly pollID: number;
+  protected poll: HTMLElement;
```
dtdesign, line 26:
```suggestion
  private readonly poll: HTMLElement;
```
The class isn't being derived, therefore we can simply flag everything as `private` instead of `protected`. Same goes for other fields/methods.

ts/WoltLabSuite/Core/Ui/Poll/Manager/Poll.ts
```diff
@@ -0,0 +1,205 @@
...
+
+  private voteView?: VoteView;
+  private resultsView?: Results;
+  private participants?: Participants;
+
+  private voteHandler?: VoteHandler;
```
dtdesign, line 32:
```suggestion
  private readonly voteView?: VoteView = undefined;
  private readonly resultsView?: Results = undefined;
  private readonly participants?: Participants = undefined;

  private readonly voteHandler?: VoteHandler = undefined;
```
First of all, all these members are only written inside the constructor, it is safe to mark them as `readonly`.

Second, and I really can't stress this enough, you must always initialize members, otherwise these do not exist on runtime and you will face all sort of strange prototype behaviors. Since their initialization is conditional, you must explicitly define them as `undefined` in their definition!

ts/WoltLabSuite/Core/Ui/Poll/Manager/Poll.ts
```diff
@@ -0,0 +1,205 @@
...
+
+  private views: Map<PollViews, HTMLElement> = new Map();
```
dtdesign, line 34:
```suggestion
  private readonly views: Map<PollViews, HTMLElement> = new Map();
```

ts/WoltLabSuite/Core/Ui/Poll/Manager/Poll.ts
```diff
@@ -0,0 +1,205 @@
...
+  get canViewResults(): boolean {
+    return this.poll.dataset.canViewResult === "true";
+  }
+
+  set canViewResults(canViewResults: boolean) {
+    this.poll.dataset.canViewResult = canViewResults ? "true" : "false";
+  }
+}
+
+export class PollSetup {
```
dtdesign, line 179: What is the intention behind this “one-off” class instead of using a plain function together with module level const map?

ts/WoltLabSuite/Core/Ui/Poll/Manager/Poll.ts
dtdesign: Please rename it to `element`, that makes more sense. You could then also ditch the `getElement()` method and just expose it through a `get element(): HTMLElement`.

ts/WoltLabSuite/Core/Ui/Poll/Manager/Poll.ts
```diff
@@ -0,0 +1,205 @@
...
+import VoteView from "./View/Vote";
+import VoteHandler from "./Vote";
+
+export enum PollViews {
+  vote = "vote",
+  results = "results",
+}
+
+export class Poll {
+  public readonly pollID: number;
```
dtdesign, line 25:
```suggestion
  public readonly pollId: number;
```
This is already annoying enough in PHP, please try to stick with strict camel case in TypeScript to keep it consistent.

ts/WoltLabSuite/Core/Ui/Poll/Manager/View/Results.ts
```diff
@@ -0,0 +1,65 @@
...
+      }
+
+      this.button.disabled = false;
+    });
+  }
+
+  protected async loadView(): Promise<void> {
+    const request = Ajax.dboAction("getResultTemplate", "wcf\\data\\poll\\PollAction");
+    request.objectIds([this.pollManager.pollID]);
+    const results = (await request.dispatch()) as ResponseData;
```
dtdesign, line 50: Please do not use `ResponseData`, it is part of the internal API and is basically just an "this is an object with arbitrary values". Use a module local `type` definition instead to model the response.

ts/WoltLabSuite/Core/Ui/Poll/Manager/View/Vote.ts
```diff
@@ -0,0 +1,67 @@
...
+      }
+
+      this.button.disabled = false;
+    });
+  }
+
+  protected async loadView(): Promise<void> {
+    const request = Ajax.dboAction("getVoteTemplate", "wcf\\data\\poll\\PollAction");
+    request.objectIds([this.pollManager.pollID]);
+    const results = (await request.dispatch()) as ResponseData;
```
dtdesign, line 52: See my previous remarks about the use of `ResponseData`.

ts/WoltLabSuite/Core/Ui/Poll/Manager/Vote.ts
```diff
@@ -0,0 +1,115 @@
...
+
+        this.pollManager.canVote = data.changeableVote ? true : false;
+        this.pollManager.canViewResults = true;
+
+        this.pollManager.addView(PollViews.results, data.template);
+        this.pollManager.displayView(PollViews.results);
+        this.pollManager.changeTotalVotes(data.totalVotes, data.totalVotesTooltip);
+      },
+    });
+    request.sendRequest();
```
dtdesign, line 107: This is still using the legacy PHP API and not a proper DBOAction?!

dtdesign (changes requested):

dtdesign (commented):
LGTM

PR #4666, opened 2022-02-09
Improve UserAvatarImporter

TimWolla:
**THIS IS UNTESTED**

### DIFF
--- a/wcfsetup/install/files/lib/system/importer/UserAvatarImporter.class.php
+++ b/wcfsetup/install/files/lib/system/importer/UserAvatarImporter.class.php
@@ -29,7 +29,7 @@ class UserAvatarImporter extends AbstractImporter
     public function import($oldID, array $data, array $additionalData = [])
     {
         // check file location
-        if (!@\file_exists($additionalData['fileLocation'])) {
+        if (!\is_readable($additionalData['fileLocation'])) {
             return 0;
         }
 
--- a/wcfsetup/install/files/lib/system/importer/UserAvatarImporter.class.php
+++ b/wcfsetup/install/files/lib/system/importer/UserAvatarImporter.class.php
@@ -7,6 +7,7 @@ use wcf\data\user\avatar\UserAvatarEditor;
 use wcf\system\exception\SystemException;
 use wcf\system\WCF;
 use wcf\util\FileUtil;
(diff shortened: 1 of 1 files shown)


wcfsetup/install/files/lib/system/importer/UserAvatarImporter.class.php
```diff
@@ -29,7 +30,7 @@ class UserAvatarImporter extends AbstractImporter
     public function import($oldID, array $data, array $additionalData = [])
     {
         // check file location
-        if (!@\file_exists($additionalData['fileLocation'])) {
+        if (!\is_readable($additionalData['fileLocation'])) {
```
BurntimeX, line 33: According to the PHP documentation, `is_readable` can emit an E_WARNING error message. You should suppress that error message by using the `@` operator. Otherwise, the complete import will abort if an error occurs while importing an avatar.

BurntimeX (changes requested):

wcfsetup/install/files/lib/system/importer/UserAvatarImporter.class.php
TimWolla: a) The exact same check already exists within AbstractAttachmentImporter: https://github.com/WoltLab/WCF/blob/f466b0922af60c5d9436fc4f9ad67a25e92ccb4f/wcfsetup/install/files/lib/system/importer/AbstractAttachmentImporter.class.php#L38

b) The warning will not be emitted for non-existent nor for non-readable files. I just checked PHP's source code: The warning basically can only happen if the file system misbehaves. This is a case where an abort is desirable, as there's likely some major issue lurking somewhere.

Example: Even passing a single null-byte as the filename (invalid in any file system out there) will not trigger a warning.
```
php > var_dump(is_readable('/etc/passwd'));
php shell code:1:
bool(true)
php > var_dump(is_readable('/etc/shadow'));
php shell code:1:
bool(false)
php > var_dump(is_readable('/unknown-file'));
php shell code:1:
bool(false)
php > var_dump(is_readable("\0"));
php shell code:1:
bool(false)
```

BurntimeX: All right. Then we can leave it like that.

PR #4670, opened 2022-02-14
fix anchor link in i18n article add template

mutec:
caused by c8abd7df4a2c0e434964c66ef86d236613e11fe4


TimWolla:
Good catch, thanks.

TimWolla:
Applied manually to include the commit reference within the commit message itself (it's less useful if it's just within the PR message).

PR #4676, opened 2022-02-18
handle available api of form builder buttons (correctly)

mutec:
Before buttons did support the addition of dependencies, validation-methods and adding a description without any usage of those features.

Setting descriptions for `FormButtons` will be disabled with this PR and are going to throw a `BadMethodCallException`. Child-classes might override this and implement a handling for descriptions.
In favor of the current layout of the `formSubmit`-part of forms I decided to not implement descriptions of buttons by default since I didn't find an appropriate way to present them (besides using `jsTooltip`).

`FormButton`s require a label now. Creating a button of `FormButton` without a label would't make sense, but was (theoretically) possible.

Dependencies of button are now handled correctly as well as the execution of buttons' `validate()` method has been added.

### DIFF
--- a/com.woltlab.wcf/templates/__emptyFormFieldDependency.tpl
+++ b/com.woltlab.wcf/templates/__emptyFormFieldDependency.tpl
@@ -1,7 +1,7 @@
 require(['WoltLabSuite/Core/Form/Builder/Field/Dependency/Empty'], function(EmptyFieldDependency) {
 	// dependency '{@$dependency->getId()}'
 	new EmptyFieldDependency(
-		'{@$dependency->getDependentNode()->getPrefixedId()}Container',
+		'{@$dependency->getDependentPrefixedId()}',
 		'{@$dependency->getField()->getPrefixedId()}'
 	);
 });
--- a/com.woltlab.wcf/templates/__formButton.tpl
+++ b/com.woltlab.wcf/templates/__formButton.tpl
@@ -13,3 +13,5 @@
 		*}{if $button->getAccessKey()} accesskey="{$button->getAccessKey()}"{/if}{*
 	*}>{$button->getLabel()}</button>
 {/if}
+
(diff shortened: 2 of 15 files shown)


TimWolla:
Thanks for the contribution, but ugh. This is much more involved than I expected and I'd rather go the other route and explicitly remove support for dependencies and all the other stuff from buttons.

IIRC `FormButton`s are not really meant for this, as also evidenced by you needing to block of `->description()`. You might've missed it, but there's also a [`ButtonFormField`](https://github.com/WoltLab/WCF/blob/master/wcfsetup/install/files/lib/system/form/builder/field/ButtonFormField.class.php). That one is a proper form field. You can find reference usage in [`TotpMultifactorMethod`](https://github.com/WoltLab/WCF/blob/master/wcfsetup/install/files/lib/system/user/multifactor/TotpMultifactorMethod.class.php).

mutec:
`IFormFieldDependency` is basically correct since it's a dependency on a field, not a button. Only the dependent node is a `FormButton`-object. So you could also use it's called `getDependentNode()` and not `getDependentField()` as an argument. ;)
Btw. all those functions are features of Form-Nodes, not especially Form-Fields.

I know it looks pretty big, but it should be non-breaking. I don#t think anyone modified the changed templates and the PHP-API is explicitly kept as-is (except the `description()` which could also be postponed to WSC 5.5 or WSC 6).

However I don't see any reason why form-buttons should not have dependencies on fields.

TimWolla:
> However I don't see any reason why form-buttons should not have dependencies on fields.

Okay, let's tackle this from this side first: What's your specific use-case here and why doesn't `ButtonFormField` work for you?

“Why not” is not a good explanation for API design, because this causes one to miss better solution, as one is focused on a a small piece into of the whole. Very similar to [the XY problem](https://xyproblem.info/).

PR #4678, opened 2022-02-23
Randomize the times of the data update cronjobs

TimWolla:
- Randomize the times of the data update cronjobs
- Randomize the times of the data update cronjobs when upgrading to 5.5

### DIFF
--- a/com.woltlab.wcf/cronjob.xml
+++ b/com.woltlab.wcf/cronjob.xml
@@ -7,7 +7,7 @@
 			<description language="de">Aktualisiert Paket-Informationen</description>
 			<startminute>0</startminute>
 			<starthour>2</starthour>
-			<startdom>*/2</startdom>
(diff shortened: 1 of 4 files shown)


dtdesign (approved):
LGTM. Wouldn't it make sense to just set the refresh `refreshSearchRobots` cronjobs to the same hour and minute, but with the time shifted by 12 hours? That will make it sufficiently random, but prevents both cronjobs running at the same time. I'm thinking of this:
```php
$hour = ($hourOfRefreshPackages + 12) % 24;
```

TimWolla:
> Wouldn't it make sense to just set the refresh `refreshSearchRobots` cronjobs to the same hour and minute, but with the time shifted by 12 hours?

The refreshSearchRobots cronjob only runs once per month anway, but your suggestion certainly makes sense to make this a little more predictable. So, done.

PR #4692, opened 2022-03-04
Move the reset of the import mapping into a dedicated JavaScript module

TimWolla:
This removes a non-trivial amount of JavaScript and specifically:

- Fixes the use of the legacy `index.php/AJAXInvoke` syntax.
- Avoids the use of the deprecated SECURITY_TOKEN pseudo-constant.

### DIFF
--- a/ts/WoltLabSuite/Core/Acp/Ui/DataImport/MappingReset.ts
+++ b/ts/WoltLabSuite/Core/Acp/Ui/DataImport/MappingReset.ts
@@ -1,0 +1,36 @@
+/**
+ * Provides the program logic for the import mapping reset.
+ *
+ * @author  Tim Duesterhus
+ * @copyright  2001-2022 WoltLab GmbH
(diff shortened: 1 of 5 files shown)


ts/WoltLabSuite/Core/Acp/Ui/DataImport/MappingReset.ts
```diff
@@ -0,0 +1,37 @@
...
+          success() {
+            window.location.reload();
+          },
+          url: "index.php?ajax-invoke&t=" + Core.getXsrfToken(),
+        });
+      },
+      message: link.dataset.confirmMessage!,
+    });
+
+    return false;
```
dtdesign, line 35: This has no effect and is most likely a remnant from the jQuery implementation, instead the event's `preventDefault()` should be invoked.

dtdesign (changes requested):

dtdesign (approved):

PR #4696, opened 2022-03-08
Protect the XSRF token against BREACH attacks

TimWolla:
- Read the `SECURITY_TOKEN` from the cookie to protect it against BREACH
- Implement WoltLabSuite/Core/Core#getXsrfToken without relying on SECURITY_TOKEN
- Add WoltLabSuite/Core/Form/XsrfToken.ts to automatically insert the XSRF-TOKEN into matching inputs
- Add the `xsrfTokenInput` class to the XSRF token form fields

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -30,7 +30,8 @@ module.exports = {
     "@typescript-eslint/no-unsafe-return": 0,
     "@typescript-eslint/no-unused-vars": [
       "error", {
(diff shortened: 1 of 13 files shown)


dtdesign (approved):
Looks good so far, although I strongly want the value of the static `<input>` to be blanked out now. We need to test this as early as possible to identify any edge cases, although I do not expect any issues.

PR #4700, opened 2022-03-10
Disallow default values for BLOB and TEXT columns

dtdesign:
See https://www.woltlab.com/community/thread/294121-databasepackageinstallationplugin-hat-probleme-mit-bestimmtem-defaultvalue/

### DIFF
--- a/wcfsetup/install/files/lib/system/database/table/column/BlobDatabaseTableColumn.class.php
+++ b/wcfsetup/install/files/lib/system/database/table/column/BlobDatabaseTableColumn.class.php
@@ -13,6 +13,8 @@ namespace wcf\system\database\table\column;
  */
 class BlobDatabaseTableColumn extends AbstractDatabaseTableColumn
 {
+    use TUnsupportedDefaultValue;
+
     /**
      * @inheritDoc
      */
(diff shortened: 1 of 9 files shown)


TimWolla (changes requested):
This won't work due to: https://github.com/WoltLab/WCF/blob/43f27fc5d769a0f0d0503841cc58a765205770b4/wcfsetup/install/files/lib/system/database/table/column/AbstractDatabaseTableColumn.class.php#L222

We'll likely need a separate interface, similar to `ILengthDatabaseTableColumn`.

dtdesign:
Introducing a new interface sounds like a horrible idea, because it would break backwards compatibility. We could change the logic of the trait that it will reject non-empty values only, which would also solve the problem but without breaking compatibility.

TimWolla:
> Introducing a new interface sounds like a horrible idea, because it would break backwards compatibility.

I'll have a look into how bad that is going to be. I can imagine that the interface is implemented by default on the abstract base class for now, then developers implementing their own types can implement it explicitly and then in 5.6 the interface is remove from the abstract base class.

dtdesign:
Are we able to change the trait to only throw an exception when a non-empty value is provided? So far it appears that an empty string does not cause any issues in regards to the existing implementation?

TimWolla:
> So far it appears that an empty string does not cause any issues in regards to the existing implementation?

Likely `null` instead of an empty string, I'll test this.

- 5.4: Allow null.
- 5.5: Add the explicit interface to the abstract base class and all matching data types.
- 5.6 (or later): Remove the interface from the base class.

PR #4738, opened 2022-04-14
Add yarn workaround for the files property, fixes #4711

### LINKED ISSUE #4711: TypeScript types without PHP files
Hey,

is there any way to install the TypeScript types without having the whole repo with all the PHP files in every package?

Sir-Will:
See #4711

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,9 @@
 {
   "name": "@woltlab/wcf",
   "version": "5.4.0",
+  "scripts": {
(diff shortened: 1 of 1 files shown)


TimWolla:
I thought I already replied to the issue, but apparently not:

Sorry, we're not interested in adding workarounds for non-standard implementations of npm clients, especially since yarn 2 is intentionally moving away from being a drop-in replacement for npm.

PR #4746, opened 2022-04-20
Cleaned up the user menu, extra profile link

dtdesign:
image


Alex-Fotoecke:
I don't think the new decision to no longer display links or the text at all is expedient.

dtdesign:
The drop-down menu has an overflow handling, when there are to many items the list containing the option categories becomes scrollable. Effectively everything above and including the link to the admin panel remains visible as well as the logout link. Additionally, the link(s) immediately following the username/avatar have a different semantic compared to the option categories below.

PR #4752, opened 2022-04-22
 Fix calculation of next cronjob execution for list-separated hour declarations

IceQ1337:
See https://www.woltlab.com/community/thread/294564-fehler-bei-der-berechnung-von-cronjobs/

Fixes the calculation for cronjobs such as `0 4,16 * * *`. Cronjobs like `0 1 * * *`,  `*/10 * * * *` or `0 4-16 * * *` should not be affected, but further testing may be required.

If the hour (for the next execution) is greater than the current hour, the hour must not be increased when the minute declaration has elapsed. This may only happen if the next execution is to take place at the current hour, but this being no longer possible.

### DIFF
--- a/wcfsetup/install/files/lib/util/CronjobUtil.class.php
+++ b/wcfsetup/install/files/lib/util/CronjobUtil.class.php
@@ -332,8 +332,8 @@ final class CronjobUtil
     }
 
     /**
-     * Calculates hour of next execution. Returns number of days which
-     * had been added in order to match expression for hour and minutes.
+     * Calculates hour of next execution. Returns true if hour-declaration
+     * has already elapsed, thus requiring at least one day to be added.
      *
      * @param array $values
      * @param int $timeBase
--- a/wcfsetup/install/files/lib/util/CronjobUtil.class.php
+++ b/wcfsetup/install/files/lib/util/CronjobUtil.class.php
@@ -199,10 +199,10 @@ final class CronjobUtil
      */
     protected static function calculateDay(array &$values)
     {
-        $addAnDay = self::calculateHour($values, self::$timeBase);
+        $addADay = self::calculateHour($values, self::$timeBase);
         $timeBase = self::$timeBase;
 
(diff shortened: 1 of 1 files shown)


TimWolla:
@IceQ1337 I'm seeing that you pushed another commit. Your PR is not forgotten, it's just that it's a low priority issue within delicate code and I could not yet allocate the time to look into this.

IceQ1337:
> @IceQ1337 I'm seeing that you pushed another commit. Your PR is not forgotten, it's just that it's a low priority issue within delicate code and I could not yet allocate the time to look into this.

No problem at all! I was just installing a test instance of WSC 5.5 and when transferring these changes I noticed something was off here. I just added it for the sake of completeness :)

TimWolla:
Status update: I've already picked two of the commits into 5.5, improving the PHPDoc commentary.

TimWolla:
Thanks. An alternative solution likely would've been to calculate the minutes first, but your patch appears reasonable to me.

I've left out the commit fixing the a/an typo, as two consecutive upper characters in a variable name don't look good either.

IceQ1337:
Thanks for looking into this! As for the typo, that's probably a non-issue, but you could always just use a different wording for the variable name, like renaming both 'addAnDay' and 'addAnHour' to 'addDay' and 'addHour', which would not affect readability or anything like that. I don't think you use articles in variable names anywhere else either.

TimWolla:
> but you could always just use a different wording for the variable name, like renaming both 'addAnDay' and 'addAnHour' to 'addDay' and 'addHour', which would not affect readability or anything like that. I don't think you use articles in variable names anywhere else either.

I considered that, but I was also strongly contemplating completely rewriting the logic from scratch. The current logic is super old (WCF 1 times) and has a very complicated control and data flow making it hard to understand what's going on (and hard to verify the correctness of the fix). I still might do the latter.

TimWolla:
@IceQ1337 Just FYI: #4923.

PR #4753, opened 2022-04-23
Ensures that the date of a published article is not in the future.

BurntimeX:
A future date messes up the "mark as read" function.

### DIFF
--- a/wcfsetup/install/files/lib/acp/form/ArticleAddForm.class.php
+++ b/wcfsetup/install/files/lib/acp/form/ArticleAddForm.class.php
@@ -488,7 +488,7 @@ class ArticleAddForm extends AbstractForm
         }
 
         $data = [
-            'time' => $this->timeObj->getTimestamp(),
+            'time' => $this->getArticleTimestamp(),
             'categoryID' => $this->categoryID,
             'publicationStatus' => $this->publicationStatus,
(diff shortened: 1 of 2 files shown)


wcfsetup/install/files/lib/acp/form/ArticleAddForm.class.php
```diff
@@ -617,4 +617,19 @@ public function assignVariables()
...
+
+    /**
+     * Ensures that the date of a published article is not in the future.
+     *
+     * @since 5.5
+     */
+    protected function getArticleTimestamp(): int
+    {
+        $time = $this->timeObj->getTimestamp();
+        if ($time > \TIME_NOW && $this->publicationStatus == Article::DELAYED_PUBLICATION) {
```
TimWolla, line 629: Is this condition correct? Should it be `!= DELAYED`?

TimWolla (commented):
Okay, with the condition fixed it's clearer what should be happening here.

However I believe the date should not be silently changed, but rather rejected in `validate()`. There already is validation of the `$timeObj` and there's also validation of `publicationDateObj` gated to the selected `publicationStatus`. So the necessary "infrastructure" for validation already exists.

BurntimeX:
This would require an individual error message (language variable).

TimWolla:
> This would require an individual error message (language variable).

It could be merged with `blog.entry.date.error.invalid` into a reusable phrase.

TimWolla (approved):

PR #4757, opened 2022-04-26
Improve VisitTracker behavior

TimWolla:
** THIS IS UNTESTED **


dtdesign (approved):
Do we want to deprecate the per-object lifetimes? That feature is currently unused and increases the complexity without adding any value. I would even argue that arbitrary lifetime values create an inconsistent UX.

dtdesign (approved):

PR #4761, opened 2022-04-28
Add and update several web spiders

IceQ1337:
Uptime-Kuma is a self-hosted monitoring tool like "Uptime Robot".
See: https://github.com/louislam/uptime-kuma

User-Agent: `Uptime-Kuma/{VERSION}` Example: `Uptime-Kuma/1.18.0`
See: https://github.com/louislam/uptime-kuma/blob/1.18.0/server/model/monitor.js#L258

### DIFF
--- a/spiderList/spiderList.xml
+++ b/spiderList/spiderList.xml
@@ -1315,6 +1315,10 @@
 		<name>Unpartisan</name>
 		<url>http://www.unpartisan.com</url>
 	</spider>
+	<spider ident="Uptime-Kuma">
+		<name>Uptime-Kuma</name>
+		<url>https://uptime.kuma.pet/</url>
+	</spider>
 	<spider ident="UptimeRobot/2.0">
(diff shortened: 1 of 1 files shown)


dtdesign (approved):

IceQ1337:
Sorry for adding so much to this PR after you've already reviewed the changes. I noticed that most of the current entries no longer exist, some even for many years. I didn't remove those entries because it doesn't really matter and who knows, maybe some pre-2000 bots will come back at some point :) But I thought it would be useful to at least add a whole bunch of new entries to the list where it is actually likely that they will hit a website.

Rest assured I'm done now :)

dtdesign (approved):

dtdesign:
Thank you for taking your time to check the items on the list. I can imagine that a significant portion of the spiders are no longer active, but this is sometimes quite hard to verify.

PR #4764, opened 2022-04-29
Use icons in content interaction buttons

BurntimeX:


com.woltlab.wcf/templates/articleList.tpl
```diff
@@ -63,7 +63,7 @@
 
 {capture assign='contentInteractionButtons'}
 	<div class="contentInteractionButton dropdown jsOnly">
-		<a href="#" class="button small dropdownToggle">{lang}wcf.article.button.sort{/lang}</a>
+		<a href="#" class="button small dropdownToggle"><span class="icon icon16 fa-sort-amount-desc"></span> <span>{lang}wcf.article.button.sort{/lang}</span></a>
```
TimWolla, line 66: I wonder if it would be useful if the icon dynamically adapted to the actual sort order.

dtdesign (approved):

PR #4766, opened 2022-05-01
fix application management in multi domain setups

mutec:
Changing the landing pages of apps was failing since the domain name for single-domain-setups was validated for any case, but is not set when using a multi-domain-setup.
This lead to an un-meaningful error-message saying something is incorrect.


TimWolla (approved):
Thanks.

PR #4767, opened 2022-05-02
fix support of inverted simple acls for page-sitemap

mutec:


TimWolla:
Can you share what exactly was broken previously, so that it:
1. can be incorporated in the commit message, and it
2. allows me to verify that the fix works properly.

mutec:
@TimWolla I didn't get any issues since I'm not using inverted permissions on pages at the moment, but this looks badly wrong:

current check in sitemap:
```PHP
$canAccess  = SimpleAclResolver::getInstance()->canAccess('com.woltlab.wcf.page', $object->pageID);
```

current check in page itself:
```PHP
        $canAccess = SimpleAclResolver::getInstance()->canAccess('com.woltlab.wcf.page', $this->pageID);

        if ($this->invertPermissions) {
            $canAccess = !$canAccess;
        }
```

PR #4787, opened 2022-05-11
User online statistics were invisible when all active users were invisible

BurntimeX:


wcfsetup/install/files/lib/system/box/UserOnlineListBoxController.class.php
```diff
@@ -127,7 +127,9 @@ public function hasContent()
             return false;
         }
 
-        return parent::hasContent();
+        parent::hasContent();
```
TimWolla, line 130: This line looks incorrect to me. Either:
- Remove the call.
- Add a comment why the call is necessary (e.g. if it modifies state).
- Use the return value.

TimWolla (changes requested):

TimWolla (approved):

PR #4791, opened 2022-05-12
Remove the WCF_N input in WCFSetup

TimWolla:
Users should install separate instances into separate databases for security
reasons. This also avoids issues with users running a non-standard number and
any existing guides / queries not working, because the database tables have a
different name.


BurntimeX (approved):
LGTM

PR #4798, opened 2022-05-12
Update URL to WoltLab Community and Plugin-Store in acp/pageHeaderUser.tpl

pehbeh:


TimWolla:
Thanks! Can't believe I updated the one URL and missed the big fat subdomains in front of the remaining ones :man_facepalming:

PR #4805, opened 2022-05-14
Added documentation of the void element

BurntimeX:
ref https://www.woltlab.com/community/thread/295602-void-in-package-xml-f%C3%BChrt-in-ide-zu-fehler/


TimWolla:
Note: I adjusted the target branch, `master` is 5.6.

TimWolla (approved):
I don't know anything about XSD, but this seems reasonable to me.

PR #4883, opened 2022-06-29
Partial Data Binding for the Main Menu

dtdesign:
Synchronizes the mobile menu with the changes to the main menu, in particular keeps the unread counter in sync.

See https://www.woltlab.com/community/thread/296147-alle-als-gelesen-markieren-entfernt-counter-badges-nicht-mehr/

### DIFF
--- a/ts/WoltLabSuite/Core/Ui/Page/Menu/Main.ts
+++ b/ts/WoltLabSuite/Core/Ui/Page/Menu/Main.ts
@@ -19,6 +19,7 @@ export class PageMenuMain implements PageMenuProvider {
   private readonly callbackOpen: CallbackOpen;
   private readonly container: PageMenuContainer;
   private readonly mainMenu: HTMLElement;
+  private readonly menuItemBadges = new Map<string, HTMLElement>();
   private readonly menuItemProvider: PageMenuMainProvider;
   private readonly observer: MutationObserver;
 
@@ -220,6 +221,10 @@ export class PageMenuMain implements PageMenuProvider {
         counter.setAttribute("aria-hidden", "true");
         counter.textContent = menuItem.counter.toString();
 
+        if (menuItem.identifier !== false) {
+          this.menuItemBadges.set(menuItem.identifier, counter);
+        }
+
         link.append(counter);
       }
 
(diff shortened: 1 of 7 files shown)


ts/WoltLabSuite/Core/Ui/Page/Menu/Main.ts
```diff
@@ -314,6 +319,28 @@ export class PageMenuMain implements PageMenuProvider {
...
+  private refreshUnreadBage(menuItem: MenuItem): void {
+    if (menuItem.identifier !== false) {
+      const counter = this.menuItemBadges.get(menuItem.identifier);
+      if (counter) {
+        if (menuItem.counter === 0) {
+          counter.remove();
+          this.menuItemBadges.delete(menuItem.identifier);
+        } else {
+          const value = parseInt(counter.textContent!, 10);
+          if (value !== menuItem.counter) {
```
TimWolla, line 336: Is this check necessary (e.g. for performance)? Otherwise I'd remove it (and the line above) to keep the indentation level lower.

ts/WoltLabSuite/Core/Ui/Page/Menu/Main/Provider.ts
```diff
@@ -16,6 +16,7 @@ export type MenuItem = {
   children: MenuItem[];
   counter: number;
   depth: MenuItemDepth;
+  identifier: string | false;
```
TimWolla, line 19: `false` is an odd choice for "no identifier set". I'd rather suggest `null`.

ts/WoltLabSuite/Core/Ui/Page/Menu/Main.ts
dtdesign: DOM operations are always somewhat expensive, updating the `textContent` is effectively this operation:
```ts
counter.childNodes[0].remove();
counter.append(document.createTextNode(menuItem.counter.toString()));
```

We can simply this into a one-liner, but it doesn‘t really make much of a difference, does it?

TimWolla: > DOM operations are always somewhat expensive,

Understood. Then leave it as it is. A one-liner doesn't bring any benefit here if a conditional still is required.

PR #4912, opened 2022-07-27
Allow @-sign in username

SoftCreatR:
Currently, the check for email addresses as username is too strict, as an email address in user@host format is valid. However, this results in usernames no longer being allowed to contain an @ character, otherwise they will be recognized as a valid email address.

See https://www.woltlab.com/community/thread/296653-benutzernamen-mit-im-namen-bei-5-5-nicht-mehr-m%C3%B6glich


TimWolla:
Manually applied to:

1. Also add an explanatory comment.
2. Use `str_contains()` as per: https://docs.woltlab.com/5.5/migration/wsc54/libraries/#symfony-php-polyfills

Thanks!

PR #4921, opened 2022-08-04
Fix Facebook media provider

SoftCreatR:
In Facebook video urls, usernames CAN contain periods, but they're currently not supported. There's also no need for a trailing slash (after `watch` and/or after the whole url).

Example: https://www.facebook.com/RaiPlay.it/videos/1059491774481091


TimWolla (approved):

TimWolla:
Thanks!

PR #4922, opened 2022-08-04
Add support for PSR-15 controllers

TimWolla:


TimWolla:
PSR-15-style controllers are already usable simply by implementing the `RequestHandlerInterface` interface. The use of a separate namespace will be deferred to a later version, because it does not bring a real benefit and even more possible naming conflicts. Instead PSR-15-style controllers should be placed in one of the existing namespaces as appropriate.

PR #4935, opened 2022-08-08
Move ban checking from WCF boot to middleware

TimWolla:
The previous location in WCF made sense back when `initBlacklist()` also
checked the IP address, User-Agent, or hostname blocklist to prevent processing
the request as much as possible. As all these checks are gone now, the only
thing that remains is the inexpensive ban check.

Move it into a middleware, it does not really belong into WCF which should just
be responsible for booting the framework independently of the request in
question.

### DIFF
--- a/wcfsetup/install/files/lib/http/middleware/CheckUserBan.class.php
+++ b/wcfsetup/install/files/lib/http/middleware/CheckUserBan.class.php
@@ -1,0 +1,67 @@
+<?php
+
+namespace wcf\http\middleware;
+
+use Psr\Http\Message\ResponseInterface;
+use Psr\Http\Message\ServerRequestInterface;
+use Psr\Http\Server\MiddlewareInterface;
+use Psr\Http\Server\RequestHandlerInterface;
+use wcf\data\user\User;
+use wcf\system\exception\AJAXException;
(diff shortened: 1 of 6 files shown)


wcfsetup/install/files/lib/http/middleware/CheckUserBan.class.php
```diff
@@ -0,0 +1,63 @@
...
+                WCF::getSession()->deleteUserSessionsExcept($user);
+
+                throw new NamedUserException(WCF::getLanguage()->getDynamicVariable('wcf.user.error.isBanned'));
+            }
+        }
+
+        return $handler->handle($request);
+    }
+
+    private function isBanned(User $user): bool
```
BurntimeX, line 51: Wouldn't it be better to move this code to `User.class.php`?

TimWolla: Yes, it would. Looking through the code, this should be done in a follow-up PR, because various locations already check for `->banned` without taking owner access into account.

TimWolla: see #4938

BurntimeX (approved):
LGTM

PR #4936, opened 2022-08-08
Font Awesome 6: `{icon}` template helper and `<fa-icon>` element

dtdesign:
n/t


dtdesign:
The `<fa-icon>` currently has a few limitations and conceptual issues:
* It is possible to change the attributes without affecting the item, even invalid combinations are possible.
* The other icon styles that are part of the Pro license cannot be used.
* Specifying the icon style at all times is verbose.

The components needs to be overhauled with these goals in mind:
* The `solid` style is the only special case, because unlike the other (Pro) styles, it is the only style that alters the icon. A good example is the `bell` icon, where the `solid` style is the only icon that is entirely filled.
* `solid` must become an explicit opt-in attribute.
* If `solid` is not specified:
  *  If the font is `Font Awesome 6 Free`:
    * Check if a `regular` version is available and use it. Otherwise fall back to `solid`.
  * If the font is `Font Awesome 6 Pro`:
    * Use whatever font-weight is specified through `--fa-icon-weight`.
* Watch for changes to the updates and update accordingly.

wcfsetup/install/files/lib/system/template/plugin/IconFunctionTemplatePlugin.class.php
```diff
@@ -0,0 +1,69 @@
...
+        if (!\in_array($size, self::SIZES)) {
+            throw new \InvalidArgumentException("An unsupported size `{$size}` was requested.");
+        }
+
+        if ($name === '') {
+            throw new \InvalidArgumentException("The `name` attribute must be present and non-empty.");
+        }
+
+        if ($type !== '' && !\in_array($type, self::TYPES)) {
+            throw new \InvalidArgumentException("An unsupported type `${type}` was specified.");
```
TimWolla, line 44:
```suggestion
            throw new \InvalidArgumentException("An unsupported type `{$type}` was specified.");
```
Your previous syntax is wonky and deprecated with 8.2: https://wiki.php.net/rfc/deprecate_dollar_brace_string_interpolation. Apparently the Syntax checker workflow does not support deprecations as of right now, otherwise it would have caught this. The same is used below in line 50.

wcfsetup/install/files/lib/system/template/plugin/IconFunctionTemplatePlugin.class.php
```diff
@@ -0,0 +1,69 @@
...
+     * @inheritDoc
+     */
+    public function execute($tagArgs, TemplateEngine $tplObj)
+    {
+        $size = \intval($tagArgs['size'] ?? 0);
+        $name = $tagArgs['name'] ?? '';
+        $type = $tagArgs['type'] ?? '';
+
+        if (!\in_array($size, self::SIZES)) {
+            throw new \InvalidArgumentException("An unsupported size `{$size}` was requested.");
```
TimWolla, line 36:
```suggestion
            throw new \InvalidArgumentException("An unsupported size '{$size}' was requested.");
```
We usually use single quotes instead of backticks in error messages. Same applies elsewhere.

ts/WoltLabSuite/WebComponent/fa-brand.ts
```diff
@@ -0,0 +1,60 @@
+(() => {
+  const HeightMap = new Map<number, number>([
```
TimWolla, line 2: This might benefit from type aliases for the two `number`s.

ts/WoltLabSuite/WebComponent/fa-icon.ts
```diff
@@ -0,0 +1,138 @@
...
+      return name !== null && window.getFontAwesome6IconMetadata(name) !== undefined;
+    }
+
+    private isValidIconStyle(name: string, isSolid: boolean): boolean {
+      if (!isSolid && isFontAwesome6Free()) {
+        const [, styles] = window.getFontAwesome6IconMetadata(name)!;
+        if (!styles.includes("regular")) {
+          return false;
+        }
+      }
```
TimWolla, line 72: This part would benefit from a short (!) explanatory comment:

> We must reject non-solid icons that are not part of FA 6 Free.

or something like this.

TimWolla (changes requested):

wcfsetup/install/files/lib/system/template/plugin/IconFunctionTemplatePlugin.class.php
dtdesign: Good catch, looks like I mixed up the variable interpolation of JavaScript by accident.

PR #4951, opened 2022-08-11
Migration Tools for Font Awesome 6

dtdesign:
n/t


extra/migrate-fa-v4.php
```diff
@@ -0,0 +1,153 @@
...
+if (\PHP_SAPI !== 'cli') {
+    exit;
+}
+
+// JSON Structure:
+// [key: string]: {
+//   name: string;
+//   type: "brand" | "regular" | "solid";
+// }
+$json = <<<JSON
```
TimWolla, line 23:
```suggestion
$json = <<<'JSON'
```
To be extra sure, that `$` won't be interpreted in there.

extra/migrate-fa-v4.php
```diff
@@ -0,0 +1,153 @@
...
+            $fileExtension = pathinfo($filename, PATHINFO_EXTENSION);
+            return in_array($fileExtension, $targetFilenames);
+        }
+    );
+
+    $updatedFiles = 0;
+
+    $iterator = new RecursiveIteratorIterator($filter);
+    /** @var SplFileInfo $fileInfo */
+    foreach ($iterator as $fileInfo) {
```
TimWolla, line 56:
```suggestion
    foreach ($iterator as $fileInfo) {
        \assert($fileInfo instanceof SplFileInfo);
```

PR #4955, opened 2022-08-12
Permit the use of the native `<button>` element

dtdesign:
Strip all styling and allow it to be used in places where `a[role="button"]` and similar workarounds had to be used.

### DIFF
--- a/com.woltlab.wcf/templates/commentAddGuestDialog.tpl
+++ b/com.woltlab.wcf/templates/commentAddGuestDialog.tpl
@@ -22,5 +22,5 @@
 
 <div class="formSubmit">
 	<input type="submit" value="{lang}wcf.global.button.submit{/lang}" accesskey="s">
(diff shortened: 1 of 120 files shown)


com.woltlab.wcf/templates/commentAddGuestDialog.tpl
```diff
@@ -22,5 +22,5 @@
 
 <div class="formSubmit">
 	<input type="submit" value="{lang}wcf.global.button.submit{/lang}" accesskey="s">
-	<button data-type="cancel">{lang}wcf.global.button.cancel{/lang}</button>
+	<button clas="button" data-type="cancel">{lang}wcf.global.button.cancel{/lang}</button>
```
cadeyrn, line 25: @dtdesign `clas` should be `class`.

dtdesign: Good catch, thanks!

PR #4958, opened 2022-08-13
Explicitly convert float to int

SoftCreatR:
Implicit conversion from float to int is deprecated in PHP 8.


TimWolla:
Applied to 5.5 with a slightly adjusted commit message (adding the necessary context of where the change happened). Thanks!

PR #4959, opened 2022-08-13
Explicitly cast param to int

SoftCreatR:
round(): Passing null to parameter #1 ($num) of type int|float is deprecated in PHP 8.

### DIFF
--- a/wcfsetup/install/files/lib/util/FileUtil.class.php
+++ b/wcfsetup/install/files/lib/util/FileUtil.class.php
@@ -365,13 +365,17 @@ final class FileUtil
     /**
      * Formats the given filesize.
      *
-     * @param int $byte
+     * @param ?int $byte
      * @param int $precision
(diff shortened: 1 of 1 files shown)


TimWolla (changes requested):
No, this is not the correct fix. The PHPDoc is defined to only accept `int`, not `?int`. Thus the callers pass invalid values and they should be the ones to be adjusted. Actually using proper types here, will make this even more visible: Then passing `null` will throw and the `if()` will never be reached.

TimWolla:
Leaving this open as a reminder to actually add the types. Feel free to propose a PR if you'd like to tackle this yourself.

SoftCreatR:
What about this?

TimWolla:
Nope. This just waters up the types. This function really is not intended to accept `null` as a valid input. It's a bug in the caller and should be fixed there.

PR #4960, opened 2022-08-15
Update reactions & smilies (Fluent Emoji)

SoftCreatR:
Replace EmojiOne assets with Fluent Emoji (https://github.com/microsoft/fluentui-emoji)

For compatibility reasons, paths and file names shouldn't change.

image


SoftCreatR:
As an alternative, I've also prepared a PR to use Twemoji instead.

cadeyrn:
Why? Maybe first open an issue to discuss that change?

IMO Fluent Emoji are one of the most ugliest emojis that are available.

SoftCreatR:
Because the emojione icons used are no longer maintained for years now, and newer versions can't be used due to license restrictions. If you want to use matching reactions for example, you are very limited.

Both Twemoji and Fluent emoji are under active development and the licenses allow them to be used within the WSC without any problems.

PR #4963, opened 2022-08-21
Reaction summary web component

BurntimeX:
In addition, the selected reaction is now visually highlighted in the summary.


com.woltlab.wcf/templates/reactionSummaryList.tpl
```diff
@@ -1,16 +1,13 @@
 {if $__wcf->session->getPermission('user.like.canViewLike')}
-	{assign var='_reactionSummaryListReactions' value=null}
-        {if $reactionData[$objectID]|isset}
-		{assign var='_reactionSummaryListReactions' value=$reactionData[$objectID]->getReactions()}
+	{assign var='__reactionSummaryJson' value='[]'}
+    {if $reactionData[$objectID]|isset}
+		{assign var='__reactionSummaryJson' value=$reactionData[$objectID]->getReactionsJson()}
```
TimWolla, line 4: This could use the `|json` modifier, saving the new method.

dtdesign: That isn’t correct, the new method returns a different data structure by omitting superfluous data included in `getReactions()`.

PR #4972, opened 2022-08-31
Web component for google maps

BurntimeX:


ts/WoltLabSuite/WebComponent/woltlab-core-google-maps.ts
```diff
@@ -0,0 +1,80 @@
+{
+  let initCalled = false;
+  const callbackPromise = new Promise<void>((resolve) => {
+    window.__initGoogleMaps = resolve;
```
TimWolla, line 4: This is only used internally, instead of adding it to `global.d.ts`, I recommend to rather:

```suggestion
    (window as any).__initGoogleMaps = resolve;
```

And perhaps even call the property `woltlab_core_google_maps_callback` to namespace it like the component itself.

TimWolla (approved):
I don't have any further remarks based on a review of the diff.

PR #4987, opened 2022-09-08
Add support for embedded audio/video media files

BurntimeX:


wcfsetup/install/files/lib/data/media/Media.class.php
```diff
@@ -274,6 +274,22 @@ public function hasThumbnail($size)
         return false;
     }
 
+    /**
+     * @since 6.0
+     */
+    public function isVideo(): bool
+    {
+        return \substr($this->fileType, 0, 6) === 'video/';
```
TimWolla, line 282: `str_starts_with()` and ditto below.

TimWolla (approved):
The remainder looks reasonable to me.

PR #4988, opened 2022-09-08
Remove obsolete option "profile_mail_use_captcha"

SoftCreatR:
For some reason, the option that was already removed in #4399 is fully back in the code. This PR aims to remove it again.


TimWolla:
Thanks for pointing this out, I'll have an investigation where the (presumed) bad merge happened first.

TimWolla:
I've looked into this now: #4399 was merged into 5.5 and I am not seeing the option in either 5.5 and master.

The removal was a belated compared to the removal of the form itself, but we didn't want to remove the option in a patch release for compatibility concerns.

So unless I've missed something, this PR is not necessary and can be closed.

PR #5011, opened 2022-09-20
Lazy loader for modules

dtdesign:
A lot of times entire modules have to be loaded ahead of time, because certain elements might be present on the page. This can be solved by introducing a lazy loader that will take a selector to look for and notifies a callback whenever this selector matches for the first time.

For performance reasons the lazy loader will wait for the page to fully load before notifying any callbacks. This should not be used for components that alter the visual representation of the page, eventually causing layout shifts.

```ts
whenFirstSeen(".myFancyClass", () => {
  void import("./Foo").then(({ bar }) => bar());
});
```

Another helpful feature is the ability to register a callback that is invoked whenever a matching element is added to the document. Removing and adding the identical element to the document will trigger the callback again.

```ts
wheneverSeen(".myFancyClass", (element: HTMLElement) => {
  /* … */
});
```

The additional function `findUniqueElement()` tracks all previously seen elements and will report them only once. This is especially useful for functions that want to perform certain actions, such as binding event listeners, only once.

```ts
findUniqueElements(".myFancyClass", (element: HTMLElement) => {
  /* … */
});
```

### DIFF
--- a/ts/WoltLabSuite/Core/LazyLoader.ts
+++ b/ts/WoltLabSuite/Core/LazyLoader.ts
@@ -1,0 +1,77 @@
+/**
+ * Efficient lazy loader that executes a callback once a selector matches
+ * for the first time and the document has finished loading.
+ *
+ * Designed for actions that do not require eager execution, such as
+ * binding specific event listeners on runtime. It should not be used for
+ * components that alter the visible UI to prevent layout shifts.
+ *
+ * Based on the work of GitHub‘s Catalyst library (MIT license).
+ * See https://github.com/github/catalyst/blob/c7983581adffd88f059e3c70674350b4fea4ee47/src/lazy-define.ts
+ *
+ * @author Alexander Ebert
+ * @copyright 2001-2022 WoltLab GmbH
+ * @license GNU Lesser General Public License <http://opensource.org/licenses/lgpl-license.php>
+ * @module WoltLabSuite/Core/LazyLoader
+ */
+
+type CallbackWhenSeen = () => void;
+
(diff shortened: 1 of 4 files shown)


ts/WoltLabSuite/Core/LazyLoader.ts
```diff
@@ -0,0 +1,78 @@
...
+    element,
+    window.requestAnimationFrame(() => {
+      for (const selector of selectors.keys()) {
+        if (element.matches(selector) || element.querySelector(selector) !== null) {
+          for (const callback of selectors.get(selector)!) {
+            void documentReady.then(() => callback());
+          }
+
+          selectors.delete(selector);
+          timers.delete(element);
```
TimWolla, line 47: This should likely not be part of the loop / part of the if. `requestAnimationFrame` is a once-off action. So if the element fails to match, the `timers` registration will still be active, even though the callback executed, ultimately canceling an animation frame that already executed. While this might not cause issues, it's confusing.

dtdesign:
I‘m not quite sure what to document here. To be precise, I know what to document, but not how a few comments would actually be of any help. The implementation is heavily geared towards a sub-ms execution time in order to not have a negative impact on the UX. 

Understanding the logic requires a somewhat proficient knowledge in how browsers work. The use of `requestAnimationFrame()` takes advantage of the scheduling inside the rendering engine when it comes to layout recalculation and microtasks.

On the surface this is nothing but a fancy debounce that takes advantage of built-in features of the browser rather than using arbitrary timeouts.

PR #5032, opened 2022-10-06
add missing return types in form builder classes

mutec:
fixes 28e142216081467c4d40fa32d730e40b0b101e8a

### DIFF
--- a/wcfsetup/install/files/lib/system/form/builder/field/AbstractFormFieldDecorator.class.php
+++ b/wcfsetup/install/files/lib/system/form/builder/field/AbstractFormFieldDecorator.class.php
@@ -211,7 +211,7 @@ abstract class AbstractFormFieldDecorator implements IFormField
     /**
      * @inheritDoc
      */
-    public function addClass($class)
+    public function addClass($class): static
     {
         $this->field->addClass($class);
 
@@ -221,7 +221,7 @@ abstract class AbstractFormFieldDecorator implements IFormField
     /**
      * @inheritDoc
      */
-    public function addClasses(array $classes)
+    public function addClasses(array $classes): static
     {
         $this->field->addClasses($classes);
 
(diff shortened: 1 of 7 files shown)


wcfsetup/install/files/lib/system/form/builder/field/AbstractFormFieldDecorator.class.php
```diff
@@ -271,31 +271,31 @@ public function cleanup()
     /**
      * @inheritDoc
      */
-    public function checkDependencies()
+    public function checkDependencies(): static
```
TimWolla, line 274: This one is not correct.

wcfsetup/install/files/lib/system/form/builder/field/acl/AclFormField.class.php
```diff
@@ -192,8 +192,10 @@ public function readValue()
     /**
      * @inheritDoc
      */
-    public function cleanup()
+    public function cleanup(): static
     {
         ACLHandler::getInstance()->resetValues($this->getObjectType()->objectTypeID);
+
+        return $this;
```
TimWolla, line 199: As this is an actual bug: Can you split this in a separate commit that I can apply independently to 5.5?

wcfsetup/install/files/lib/system/form/builder/field/wysiwyg/WysiwygFormField.class.php
```diff
@@ -113,9 +113,11 @@ public function autosaveId($autosaveId)
     /**
      * @inheritDoc
      */
-    public function cleanup()
+    public function cleanup(): static
     {
         MessageQuoteManager::getInstance()->saved();
+
+        return $this;
```
TimWolla, line 120: ditto.

TimWolla (commented):
Thanks, looking good apart from the few annotations.

mutec:
@TimWolla I just updated it.

TimWolla:
After a rebase on the latest master, only one change remained that I applied with an updated commit message. Thanks.

PR #5033, opened 2022-10-06
remove unused imports

mutec:


TimWolla:
Thank you, but we won't accept PRs for this: They are impossible to verify. Checking for unused imports is part of the (currently disabled) code style workflow. It will be reenabled, once the bugs in the CS tools are fixed.

Regarding the second commit: This file is a (horribly outdated) third party dependency that must not be manually modified.

PR #5039, opened 2022-10-13
User list component

BurntimeX:


ts/WoltLabSuite/Core/Component/User/List.ts
```diff
@@ -0,0 +1,103 @@
...
+
+import { dboAction } from "../../Ajax";
+import WoltlabCoreDialogElement from "../../Element/woltlab-core-dialog";
+import UiPagination from "../../Ui/Pagination";
+import { dialogFactory } from "../Dialog";
+
+export class UserList {
+  readonly #options: AjaxRequestOptions;
+  readonly #dialogTitle: string;
+  readonly #cache = new Map<number, string>();
```
TimWolla, line 18: I've never been a fan of these caches in the frontend. Are they actually worth it in this case where the dialog is only opened after explicit interaction?

ts/WoltLabSuite/Core/Controller/User/Profile.ts
```diff
@@ -0,0 +1,50 @@
...
+    button.addEventListener("click", () => {
+      if (userList === undefined) {
+        userList = new UserList(
+          {
+            className: className,
+            parameters: {
+              userID: userId,
+            },
+          },
+          button.closest("section")!.querySelector(".boxTitle")!.textContent!.replace(/ \d+$/, ""),
```
TimWolla, line 26: The replacement regex makes quite a few assumptions about the HTML layout. Should a clearly identifiable `<span>` or something like that be added?

dtdesign: This is not possible, because the title is taken from the box title which is unknown in JavaScript. The only source is the generated HTML which at least is directly based off the `__box` template so that we can at least assume the HTML to be matching. However, it might be better to use `.closest(".box")` instead of `.closest("section")`.

TimWolla: This module is specific to the user profile, can't `userSidebar.tpl` be modified from:

https://github.com/WoltLab/WCF/blob/09aab20f27805dca11b564d613d5c30bdb17ccc9/com.woltlab.wcf/templates/userSidebar.tpl#L4

to:

```
<h2 class="boxTitle"><span class="this_is_the_real_title">{lang}wcf.user.profile.following{/lang}</span> <span class="badge">{#$followingCount}</span></h2>
```

or similar? I'm mostly concerned about the number replacement, which is super ugly.

TimWolla (approved):
Looks reasonable to me, but please wait for @dtdesign's review, he's the expert with regard to this area of the code.

dtdesign (approved):

PR #5048, opened 2022-10-15
Use the Intl extension for locale dependent components

dtdesign:
* [x] Require the `intl` extension in PHP. #5047
* [x] Add support for the locale selection of languages.
* [x] Convert the relative timestamps to use `Intl.*`.
  * [x] Account for the drift between the server and the client clock.
* [x] Migrate server-side relative timestamps to the `intl` extension _or_ migrate the relative timestamps to a web component.
* [x] Use intl plural rules.
  * [x] PHP
  * [x] TypeScript
* [x] Format numbers using intl.
  * [x] PHP
  * [x] TypeScript


wcfsetup/install/files/lib/util/StringUtil.class.php
```diff
@@ -841,6 +802,18 @@ public static function normalizeCsv($string): string
         return \implode(',', ArrayUtil::trim(\explode(',', $string)));
     }
 
+    private static function getNumberFormatter(): \NumberFormatter
+    {
+        static $formatters = [];
+
+        $locale = WCF::getLanguage()->getLocale();
+        if (!isset($formatters[$locale])) {
+            $formatters[$locale] = new \NumberFormatter($locale, \NumberFormatter::TYPE_DEFAULT);
```
TimWolla, line 811:
```suggestion
            $formatters[$locale] = new \NumberFormatter($locale, \NumberFormatter::DEFAULT_STYLE);
```

The type constants are used as the second parameter of `->format()`. `TYPE_DEFAULT == PATTERN_DECIMAL` which breaks the output format, because the default pattern is `null`.

wcfsetup/install/files/lib/util/StringUtil.class.php
```diff
@@ -200,51 +186,26 @@ public static function formatInteger($integer): string
...
+        return $double;
     }
 
     /**
      * Adds thousands separators to a given number.
-     *
-     * @param mixed $number
      */
-    public static function addThousandsSeparator($number): string
+    public static function addThousandsSeparator(int|float $number): string
```
TimWolla, line 206: Should the majority of the functions be deprecated and perhaps just `formatNumeric` remain?

wcfsetup/install/files/lib/system/template/plugin/TimeModifierTemplatePlugin.class.php
```diff
@@ -33,6 +33,11 @@ public function execute($tagArgs, TemplateEngine $tplObj)
         $isFutureDate = ($timestamp > TIME_NOW);
         $dateTime = DateUtil::getRelativeTime($dateTimeObject, $timestamp, $date, $time, $isFutureDate);
 
+        return \sprintf(
+            '<woltlab-core-time date="%s"></woltlab-core-time>',
```
TimWolla, line 37: Can we render *some* placeholder as the `textContent` here, so that the times at least show up without JavaScript? Or will this cause FOUC?

dtdesign: I’ve added a placeholder date. FOUC isn’t a concern, because this is an eager web component and thus is being executed in a micro task while the DOM is being parsed.

ts/WoltLabSuite/WebComponent/woltlab-core-time.ts
```diff
@@ -0,0 +1,155 @@
...
+        value = datePartsWithoutDayOfWeek.join("");
+      }
+
+      return value;
+    }
+  }
+
+  window.customElements.define("woltlab-core-time", WoltlabCoreTimeElement);
+
+  const refreshAllTimeElements = function () {
```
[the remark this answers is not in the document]
dtdesign: I did, but https://eslint.org/docs/latest/rules/no-inner-declarations appears to not play nicely with my use of `{ … }` instead of an IIFE. So I need to either use an IIFE or use a lambda function.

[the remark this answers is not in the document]
TimWolla: Okay, fair enough.

ts/WoltLabSuite/WebComponent/woltlab-core-time.ts
```diff
@@ -0,0 +1,191 @@
...
+      this.refresh(true);
+    }
+
+    refresh(updateTitle: boolean) {
+      const date = this.date;
+      const difference = Math.trunc((Date.now() - date.getTime() - drift) / 1_000);
+
+      if (this.#timeElement === undefined) {
+        this.#timeElement = document.createElement("time");
+        this.#timeElement.dateTime = date.toISOString();
```
TimWolla, line 105: I believe this will go out of sync if the `.date` setter is used.

ts/WoltLabSuite/WebComponent/woltlab-core-time.ts
```diff
@@ -0,0 +1,191 @@
...
+        }
+      }
+
+      value = value.charAt(0).toUpperCase() + value.slice(1);
+
+      this.#timeElement.textContent = value;
+    }
+
+    /**
+     * The date formatter was not provide a reliable way to generate
```
TimWolla, line 141:
```suggestion
     * The date formatter does not provide a reliable way to generate
```

TimWolla (approved):

PR #5052, opened 2022-10-18
Use explicit casts instead of function invocation

SoftCreatR:
Since there's no base used for conversion, we can safely switch from function invocation to explicit casts, which can increase the performance greatly. This also saves us 4 bytes per replacement.

### DIFF
--- a/wcfsetup/install/files/lib/util/DateUtil.class.php
+++ b/wcfsetup/install/files/lib/util/DateUtil.class.php
@@ -443,13 +443,13 @@ final class DateUtil
         $year = $month = $day = 0;
         $value = \explode('-', $date);
         if (isset($value[0])) {
-            $year = \intval($value[0]);
+            $year = (int)$value[0];
         }
         if (isset($value[1])) {
-            $month = \intval($value[1]);
+            $month = (int)$value[1];
         }
         if (isset($value[2])) {
-            $day = \intval($value[2]);
+            $day = (int)$value[2];
         }
 
         // calc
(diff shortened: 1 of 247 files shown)


TimWolla:
> which can increase the performance greatly

Do you have any numbers for common requests (e.g. accessing a Thread)?

> This also saves us 4 bytes per replacement.

4 bytes of what? Memory Usage? Disk usage?

TimWolla:
In any case such a large-scale replacement is not acceptable for stable branches (5.5) and needs to target master.

wcfsetup/install/files/lib/util/DateUtil.class.php
```diff
@@ -492,7 +492,7 @@ public static function validateDate($date)
     public static function getFirstDayOfTheWeek()
     {
         if (self::$firstDayOfTheWeek === null) {
-            self::$firstDayOfTheWeek = \intval(WCF::getLanguage()->get('wcf.date.firstDayOfTheWeek'));
+            self::$firstDayOfTheWeek = (int)WCF::getLanguage(->get('wcf.date.firstDayOfTheWeek'));
```
TimWolla, line 495: Did you perform the replacement using a naïve regular expression or with a proper tool (e.g. Rector)? If this is a naïve regular expression, then there is a great chance that an incorrect replacement introduces bugs, not just obvious syntax errors.

SoftCreatR: Did most of it by hand. Just missed that one.

SoftCreatR:
> In any case such a large-scale replacement is not acceptable for stable branches (5.5) and needs to target master.

Found a bug anyways. I'll re-create this for master.

> Do you have any numbers for common requests (e.g. accessing a Thread)?

Not really, just pretty simple benchmarks.

> 4 bytes of what? Memory Usage? Disk usage?

Disk. Of course, this is not really a significant improvement. At least the conversion would now be consistent.

wcfsetup/install/files/lib/util/DateUtil.class.php
TimWolla: Before you spend more (manual) time on this: I'm not really sold on this being an overall improvement. As an example the lack of parentheses makes precedence much less clear and `\intval()` is not really in super hot paths anyway. When I last checked with XDebug's profiler, `intval` was not anywhere near the expensive locations.

PR #5053, opened 2022-10-18
Harden the JavaScript relocator with per-request nonces

TimWolla:


wcfsetup/install/files/lib/system/WCF.class.php
```diff
@@ -1127,6 +1127,23 @@ public function useDesktopNotifications()
...
+     */
+    final public static function getRequestNonce(string $scope): string
+    {
+        static $key = null;
+
+        if ($key === null) {
+            $key = \random_bytes(16);
+        }
+
+        return $scope . '_' .\hash_hmac('md5', WCF_UUID . ':' . self::class . ':' . $scope, $key);
```
dtdesign, line 1144:
```suggestion
        return $scope . '_' .\hash_hmac('md5', \WCF_UUID . ':' . self::class . ':' . $scope, $key);
```

dtdesign (changes requested):
Otherwise LGTM

PR #5054, opened 2022-10-19
Add logic to sync a user's TZ with their browser timezone

TimWolla:


ts/WoltLabSuite/Core/Controller/User/Profile/TimezoneSync.ts
```diff
@@ -0,0 +1,36 @@
+/**
+ * Updates the user's timezone option on the server if the browser's
+ * timezone differs.
+ *
+ * @author  Tim Duesterhus
+ * @copyright   2001-2022 WoltLab GmbH
+ * @license GNU Lesser General Public License <http://opensource.org/licenses/lgpl-license.php>
+ * @package WoltLabSuite\Core\Data\User
```
dtdesign, line 8:
```suggestion
 * @package WoltLabSuite\Core\Data\User
 * @since 6.0
```

ts/WoltLabSuite/Core/Controller/User/Profile/TimezoneSync.ts
```diff
@@ -0,0 +1,36 @@
...
+
+function getClientTimezone(): string
+{
+  return Intl.DateTimeFormat().resolvedOptions().timeZone
+}
+
+export async function maybeSyncServerTimezone(): Promise<void>
+{
+  if (getServerTimezone() !== getClientTimezone()) {
+    await fetch(`${window.WSC_API_URL}index.php?user-timezone-sync/`, {
```
dtdesign, line 27: Fetch can fail loud. This should be enclosed in `try … catch`, because we do not care about failed requests (outside of the debug mode).

wcfsetup/install/files/lib/action/UserTimezoneSyncAction.class.php
```diff
@@ -0,0 +1,91 @@
...
+            $errors = [];
+
+            foreach ($form as $field) {
+                if ($field instanceof IFormField && $field->getValidationErrors() !== []) {
+                    $errors[$field->getId()] = \array_map(
+                        static fn (FormFieldValidationError $e) => $e->getMessage(),
+                        $field->getValidationErrors()
+                    );
+                }
+            }
```
dtdesign, line 58: This smells like a good opportunity for a ready-made helper method in `$form`.

ts/WoltLabSuite/Core/Controller/User/Profile/TimezoneSync.ts
TimWolla, line 8: fix

wcfsetup/install/files/lib/action/UserTimezoneSyncAction.class.php
```diff
@@ -0,0 +1,91 @@
...
+use wcf\system\WCF;
+use wcf\util\DateUtil;
+
+/**
+ * Updates the user's timezone option.
+ *
+ * @author  Tim Duesterhus
+ * @copyright   2001-2022 WoltLab GmbH
+ * @license GNU Lesser General Public License <http://opensource.org/licenses/lgpl-license.php>
+ * @package WoltLabSuite\Core\Data\User
```
TimWolla, line 27: fix

dtdesign (changes requested):

wcfsetup/install/files/lib/action/UserTimezoneSyncAction.class.php
```diff
@@ -0,0 +1,98 @@
...
+    }
+
+    /**
+     * @inheritDoc
+     */
+    public function handle(ServerRequestInterface $request): ResponseInterface
+    {
+        if (!WCF::getSession()->checkSecurityToken($request->getHeaderLine('x-xsrf-token'))) {
+            throw new InvalidSecurityTokenException();
+        }
```
TimWolla, line 65: With #5059 this (c|sh)ould be transformed to:

```suggestion
        if ($request->getAttribute(Xsrf::HAS_VALID_HEADER_ATTRIBUTE) !== true) {
            throw new InvalidSecurityTokenException();
        }
```

Issue #4885, opened 2022-06-30
Modernize the favicon handling


TimWolla:
- [x] Do not generate a .ico favicon: Browsers support `.png` well.
    - [x] This avoids the need for an external library that is no longer maintained.
- [ ] Increase the expected size of the favicon template to 512x512 (#4449)
- [x] https://www.woltlab.com/community/thread/296627-favicon-definition-in-head-veraltet/

TimWolla:
> * [ ]  This avoids the need for an external library that is no longer maintained.

see also: #4911 4911

dtdesign:
The entire manifset fike is a bit dated and should be refreshed, but this makes most sense in the context of the introduction of a lightweight PWA.

PR #5063, opened 2022-10-21
Generate favicons in bigger sizes

SoftCreatR:
As [recommended by Google](https://developers.google.com/search/docs/appearance/favicon-in-search#guidelines), favicons should be at least 48x48 pixels. Currently, a favicon will be generated with a maximum dimension of 32x32 pixels.

> Your favicon must be a multiple of 48px square, for example: 48x48px, 96x96px, 144x144px and so on.

However, I don't think there's any need for anything else but 48px. Following the Google recommendation (generating a favicon in 48, 96 & 144px), the generated favicon file would exceed 100 kB, while the single-size favicon is reasonably tiny in file size.


TimWolla:
see #4885

PR #5065, opened 2022-10-22
Fix double implements in TimeDatabaseTableColumn

Hanashi:
In class `TimeDatabaseTableColumn` was a double `implements`. The first `implements` was appended to the abstract class name. Because of this the `extends` was wrong.


dtdesign:
Thanks!

PR #5067, opened 2022-10-24
Add missing return types in FormBuilder

Hanashi:
In some classes the return values were missing although they were defined in the parent classes or interfaces. This caused an error in the IDE and possibly in the PHP parser.


TimWolla (approved):

TimWolla:
Thanks!

PR #5070, opened 2022-10-25
Clean up JavaScript Templates

TimWolla:


wcfsetup/install/files/acp/templates/bbcodeAdd.tpl
```diff
@@ -43,13 +43,13 @@
 {/capture}
 
 <script data-relocate="true">
-	$(function() {
+	require(['WoltLabSuite/Core/Template'], (Template) => {
```
dtdesign, line 46: This is an unsafe transformation, because jQuery might not be initialized at this point. You need to preserve `$(function () {` inside the `require()` function body.

dtdesign (changes requested):
Otherwise LGTM

PR #5102, opened 2022-10-31
Fixed multiple typos

SoftCreatR:
I know, it's a lot. But not my fault ;)

No code has been modified, just the documentation.


wcfsetup/install/files/lib/data/user/User.class.php
```diff
@@ -720,7 +720,7 @@ public function canEmailConfirm()
     }
 
     /**
-     * Returns true, if the user must confirm his email by themself.
+     * Returns true, if the user must confirm his email by himself.
```
TimWolla, line 723: This change is incorrect, but I believe it should be “themselves” instead of “themself”.

wcfsetup/install/files/lib/data/user/User.class.php
```diff
@@ -732,7 +732,7 @@ public function mustSelfEmailConfirm()
 
     /**
      * Returns true if the user is a member of a user group that requires
-     * multi-factor authentication to be enabled.
+     * multifactor authentication to be enabled.
```
TimWolla, line 735: This change is incorrect: https://en.wikipedia.org/wiki/Multi-factor_authentication

wcfsetup/install/files/lib/data/user/activity/event/ViewableUserActivityEventList.class.php
```diff
@@ -94,7 +94,7 @@ public function readObjects()
     }
 
     /**
-     * Returns timestamp of oldest entry fetched.
+     * Returns timestamp of the oldest entry fetched.
```
TimWolla, line 97: Not incorrect, but incomplete. It should be “Returns the timestamp”.

TimWolla (commented):
How did you create this PR? As indicated by the inline comments, several of the changes are incorrect and with the amount of changes I'm not really comfortable with reviewing them all, especially since the value-add for a/an fixes, hyphenation fixes, and similar is very small.

wcfsetup/install/files/lib/data/user/User.class.php
SoftCreatR: Yeah, I wasn't 100% sure about that one.

wcfsetup/install/files/lib/data/user/User.class.php
SoftCreatR: Different websites, different opinions ;

https://learn.microsoft.com/en-us/style-guide/a-z-word-list-term-collections/m/multifactor-authentication

wcfsetup/install/files/lib/data/user/activity/event/ViewableUserActivityEventList.class.php
SoftCreatR: Whoops.

SoftCreatR:
No automation was used. I did all corrections by hand.

wcfsetup/install/files/lib/system/database/Database.class.php
```diff
@@ -484,7 +484,7 @@ public function getDatabaseName()
     /**
      * Returns the name of the database user.
      *
-     * @return  string      user name
+     * @return  string      username
```
TimWolla, line 487: As an example: This can be dropped and should likely be replaced by a proper return type.

TimWolla (commented):
> However, I was thinking about recreating the whole PR based on WSC 6, due to possible merge conflicts.

I'll be happy to merge well-written spelling and grammar fixes if they are somewhat split by “module” (e.g. fix all controllers, fix all form builder stuff, fix all *Util classes). Splitting by commit is sufficient, they do not need to be separate PRs (but may be, if that works better for you) This allows us to review them in small chunks and give early feedback if some fixes are incorrect.

As you're fixing them by hand anyway: It would be useful if you would also take a second and think about if the current comment even makes any sense. If it is tautological, the entire comment can just be dropped. If could possibly be useful (but currently is not), then it should be rewritten.

Some general guidelines:
- Gendered pronouns are always incorrect (if there are any, then it's a bug). Either use either the singular “they” or “it” (when not referring to humans).
- For “multi-factor” as mentioned previously: It’s multi-factor with a hyphen for us. That naming is also used in the user-visible phrases. Generally: If both variants are correct, no fix is required.

PR #5106, opened 2022-11-03
Migrate moderation queue user assignment to PSR-15 controller

TimWolla:


wcfsetup/install/files/lib/action/ModerationQueueAssignUserAction.class.php
```diff
@@ -0,0 +1,148 @@
...
+use wcf\system\exception\InvalidSecurityTokenException;
+use wcf\system\exception\PermissionDeniedException;
+use wcf\system\form\builder\field\SingleSelectionFormField;
+use wcf\system\form\builder\FormDocument;
+use wcf\system\moderation\queue\command\AssignUser;
+use wcf\system\WCF;
+use wcf\util\JSON;
+
+/**
+ * Updates the user's timezone option.
```
TimWolla, line 25: TODO

ts/WoltLabSuite/Core/Controller/Moderation/AssignUser.ts
```diff
@@ -0,0 +1,59 @@
+import { dialogFactory } from "../../Component/Dialog";
```
TimWolla, line 1: module comment

ts/WoltLabSuite/Core/Ajax/Backend.ts
```diff
@@ -0,0 +1,167 @@
...
+  }
+
+  disableLoadingIndicator(): this {
+    this.#showLoadingIndicator = false;
+
+    return this;
+  }
+
+  async fetchAsJson(): Promise<unknown> {
+    const response = await this.fetchAsResponse();
```
TimWolla, line 67: This should ideally add a `accept: application/json` header to the request.

wcfsetup/install/files/lib/action/ModerationQueueAssignUserAction.class.php
```diff
@@ -0,0 +1,160 @@
...
+    }
+
+    /**
+     * @inheritDoc
+     */
+    public function handle(ServerRequestInterface $request): ResponseInterface
+    {
+        if (
+            $request->getMethod() === 'POST'
+            && !$request->getAttribute(Xsrf::HAS_VALID_HEADER_ATTRIBUTE, false)
```
TimWolla, line 60: Maybe add `Xsrf::assertHasXsrfToken($request)` that checks the request for the attribute and throws the exception otherwise?

dtdesign: Actually, I would rather opt for a “secure by default” approach.

The actual request should be wrapped in a middleware that checks if it implements `RequestHandlerInterface`. If this true, the XSRF check should be enforced.

An opt-out should be possible through an attribute like `#[DisableXsrfCheck]`.

This not only saves time for most implementations, but also makes it harder to “forget” the check.

dtdesign: On second thought, we might need a parameter to opt-out for some request methods only.

```php
#[DisableXsrfCheck]
final class NeverCheckForXsrf implements RequestHandlerInterface
```

```php
#[DisableXsrfCheck('GET')]
final class CheckForXsrfUnlessGET implements RequestHandlerInterface
```

```php
#[DisableXsrfCheck('GET', 'HEAD')]
final class CheckForXsrfUnlessGETorHEAD implements RequestHandlerInterface
```

If present without parameters the attribute will disable the XSRF validation for all methods. A list of HTTP verbs can be provided as a list of string parameters (the attribute consumes `array ...$verbs`), disabling the check for these verbs only. Providing an unknown verb causes an exception.

TimWolla: Now resolved with #5124

PR #5112, opened 2022-11-07
Add new template event

SoftCreatR:
In 6.0, the way how share-button providers are initialized has been changed. Unfortunately, with this change, it became somewhat impossible to add new providers to the sharing dialog, because there's no longer a way to extend the list.

With this PR, a new template event is introduced, that allows adding individual providers easily.


dtdesign (approved):

dtdesign:
Thanks!

PR #5114, opened 2022-11-07
Add 4k support for avatar images

BurntimeX:


wcfsetup/install/files/lib/system/upload/AvatarUploadFileSaveStrategy.class.php
```diff
@@ -78,10 +78,18 @@ public function save(UploadFile $uploadFile)
 
             // shrink avatar if necessary
             try {
+                $newWidth = $newHeight = UserAvatar::AVATAR_SIZE;
+                
+                // Save 4k version if possible.
```
cadeyrn, line 83: Maybe "HiDPI" is a more appropriate name. It's not really related to the term "4K".

wcfsetup/install/files/lib/data/user/avatar/UserAvatar.class.php
```diff
@@ -32,6 +32,12 @@ class UserAvatar extends DatabaseObject implements IUserAvatar, ISafeFormatAvata
      */
     const AVATAR_SIZE = 128;
 
+    /**
+     * minimum height and width of an uploaded avatar (4k version)
```
TimWolla, line 36: 4k also referenced here.

dtdesign (approved):

PR #5117, opened 2022-11-09
Preload phrases for use in TypeScript

dtdesign:
Phrases are registered in JavaScript/TypeScript through the templates, with `Language.addObject()` being the standard. This works pretty well and with then introduction of `{jslang}` is also a pretty robust solution.

However, over the time the need for more and more phrases emerged, especially for those that _may_ be used on runtime, but cannot be statically determined. Examples include phrases for dialogs or date pickers, where we cannot predict if they are needed or not. This led to a large amount of phrases being registered ahead of time in `headIncludeJavaScript` with the list growing with every release.

This approach has two significant downsides:

1. Phrases are registered through the `Language` module, effectively restricting their availability to `require.js` based modules. Eager web components cannot make use of those phrases, because they need to run much earlier.
2. These phrases are completely static and are only evaluated on runtime, which means that their value does not depend on server-side conditions. This means that we have a large set of phrases (~100 at time of this PR) that are always exported to the template and transmitted with every request.

The first issue is solved with #5107 which effectively exports the `Language` module as a vanilla JS and bundles it along the other eager web components. This allows those components to rely on phrases and makes it possible to register phrases before the `require.js` modules have been initialized.

This PR aims to resolve the second issue by allowing phrases to be flagged for preloading ahead of time. Those phrases are then exported into a static JS file that is referenced immediately after the web component bundle. This static export will register all listed phrases before they are needed.

Since this is a static JS file the browser can aggressively cache it, reducing the server response size by ~3 kB (after gzip!). This reduces the overall response size after gzip by 10-15% on average.


wcfsetup/install/files/lib/system/language/preload/RegisterPreloadPhrases.class.php
```diff
@@ -0,0 +1,38 @@
+<?php
+
+namespace wcf\system\language\preload;
```
TimWolla, line 3:
```suggestion
namespace wcf\system\language\preload\event;
```

Events are grouped into the `event` namespace of an appropriate “subsystem”.

wcfsetup/install/files/lib/system/language/preload/RegisterPreloadPhrases.class.php
```diff
@@ -0,0 +1,38 @@
...
+
+use wcf\data\language\Language;
+use wcf\system\event\IEvent;
+
+final class RegisterPreloadPhrases implements IEvent
```
TimWolla, line 8: The event namespace is especially important here, because the name does not make it clear that this is an event. It could also be a *command*.

In this case it's hard to choose an appropriate name, because this event is not about something that *happened* (“User Logged In”, “Cache was cleared”, “Application booted successfully”), but that *will happen*.

The Microsoft style guide for C# appears to use the Gerundium for events like these: https://stackoverflow.com/a/724375/782822

How about: `CollectingPreloadedPhrases`.

wcfsetup/install/files/lib/system/language/preload/LanguagePreloader.class.php
```diff
@@ -0,0 +1,62 @@
...
+     * @param PreloadPhrase[] $preloadPhrases
+     */
+    private  function generatePreloadPayload(array $preloadPhrases): string
+    {
+        return '';
+    }
+
+    public static function reset(): void
+    {
+        DirectoryUtil::getInstance(self::STORAGE_PATH)->removePattern(new Regex('.*\.preload\.js$'));
```
TimWolla, line 60: Avoid `DirectoryUtil` in new code, the API is not great and this is better replaced by an appropriate SPL Iterator directly, similar to DiskCacheSource:

https://github.com/WoltLab/WCF/blob/bb7156fa65109349f5a955fd50ee7af11995f127/wcfsetup/install/files/lib/system/cache/source/DiskCacheSource.class.php#L25

wcfsetup/install/files/lib/system/language/preload/LanguagePreloader.class.php
```diff
@@ -0,0 +1,62 @@
...
+{
+    private readonly EventHandler $eventHandler;
+    private readonly Language $language;
+
+    private const STORAGE_PATH = \WCF_DIR . 'js/preload/';
+
+    public function __construct(Language $language)
+    {
+        $this->eventHandler = EventHandler::getInstance();
+        $this->language = $language;
```
TimWolla, line 21: `LanguagePreloader` is effectively a service and thus should avoid state. This is important for a possible future migration to dependency injection.

The `Language` parameter should be moved into the `rebuild()` function and then passed along.

Alternatively this could be transformed into a command pattern where the dependency injection issue is solved by the introduction of a command bus. In that case there should only be a single public `__invoke()` method.

TimWolla: > LanguagePreloader is effectively a service and thus should avoid state. This is important for a possible future migration to dependency injection.

More specifically: Yes, it should avoid state as much as possible. However it absolutely must not be parameterized, treat it as if it was a singleton.

wcfsetup/install/files/lib/system/language/preload/LanguagePreloader.class.php
```diff
@@ -0,0 +1,62 @@
...
+            '%s%s.preload.js',
+            self::STORAGE_PATH,
+            $this->language->getLocale(),
+        );
+    }
+
+    /**
+     * @param PreloadPhrase[] $preloadPhrases
+     */
+    private  function generatePreloadPayload(array $preloadPhrases): string
```
TimWolla, line 53:
```suggestion
    /**
     * @param iterable<PreloadPhrase> $preloadPhrases
     */
    private  function generatePreloadPayload(iterable $preloadPhrases): string
```

For this having an iterable is likely sufficient. By taking an iterable instead of an array you are restricting yourself a little, but this often makes the API better (e.g. by allowing you to also take a generator).

wcfsetup/install/files/lib/system/language/event/PhraseChanged.class.php
```diff
@@ -0,0 +1,27 @@
...
+final class PhraseChanged implements IEvent
+{
+    public readonly Language $language;
+    public readonly string $name;
+
+    public function __construct(Language $language, string $name)
+    {
+        $this->language = $language;
+        $this->name = $name;
+    }
```
TimWolla, line 26: Something fundamental that needs to be discussed:

I've generally avoid storing the original DBO in Value Objects (and thus events) allow for a cleaner serialization.

see:
https://github.com/WoltLab/WCF/blob/87c040a366b227f321c3fc9e747391a9e489806b/wcfsetup/install/files/lib/system/user/authentication/event/UserLoggedIn.class.php#L29

and also:

https://github.com/WoltLab/WCF/blob/87c040a366b227f321c3fc9e747391a9e489806b/wcfsetup/install/files/lib/system/email/Mailbox.class.php#L51-L55

wcfsetup/install/files/lib/system/language/preload/PreloadPhrase.class.php
```diff
@@ -0,0 +1,25 @@
...
+final class PreloadPhrase
+{
+    public readonly string $name;
+    public readonly bool $literal;
+
+    public function __construct(string $name, bool $literal)
+    {
+        $this->name = $name;
+        $this->literal = $literal;
+    }
```
TimWolla, line 24: This could cleanly use constructor property promotion.

wcfsetup/install/files/lib/system/language/preload/event/PreloadPhrasesCollecting.php
```diff
@@ -0,0 +1,48 @@
...
+     * @var PreloadPhrase[]
+     */
+    private array $phrases = [];
+
+    public function __construct(Language $language)
+    {
+        $this->language = $language;
+    }
+
+    public function preload(string|PreloadPhrase $phrase): void
```
TimWolla, line 32: I'm not a fan of union types of this kind. In any case this method needs a doc comment.

wcfsetup/install/files/lib/system/language/preload/event/PreloadPhrasesCollecting.php
```diff
@@ -0,0 +1,48 @@
...
+            $phrase = new PreloadPhrase($phrase, false);
+        }
+
+        $this->phrases[] = $phrase;
+    }
+
+    /**
+     * @return PreloadPhrase[]
+     */
+    public function getPhrases(): array
```
TimWolla, line 44:
```suggestion
    /**
     * @return iterable<PreloadPhrase>
     */
    public function getPhrases(): iterable
```

Iterable is preferred here.

wcfsetup/install/files/lib/system/language/preload/command/CachePreloadPhrases.class.php
```diff
@@ -0,0 +1,61 @@
...
+final class CachePreloadPhrases
+{
+    private readonly EventHandler $eventHandler;
+    private readonly Language $language;
+
+    public function __construct(Language $language)
+    {
+        $this->eventHandler = EventHandler::getInstance();
+        $this->language = $language;
+    }
```
TimWolla, line 31:
```suggestion
    public function __construct(Language $language)
    {
        $this->language = $language;

        $this->eventHandler = EventHandler::getInstance();
    }
```

Visually split the actual parameters from the dependencies. Move the dependencies to the bottom, because this is a more natural order.

wcfsetup/install/files/lib/system/WCF.class.php
```diff
@@ -813,6 +814,10 @@ protected function assignDefaultTemplateVariables()
             }
         }
 
+        self::getTPL()->assign([
+            '__phrasePreloadCacheUrl' => PhrasePreloadCache::getUrl(self::getLanguage()),
+        ]);
```
TimWolla, line 819: Please do not add new logic to `wcf\system\WCF`. WCF should perform bootstrapping only. I've previously moved the registration of the cronjob template variable to the bootstrap file to keep WCF clean.

In this case the registration in WCF is especially bad, because it relies on the language and thus a request-specific value. In the long run this might even need to be moved into a middleware.

wcfsetup/install/files/lib/bootstrap/com.woltlab.wcf.php
```diff
@@ -13,4 +19,15 @@
     );
 
     EventHandler::getInstance()->register(UserLoggedIn::class, UserLoginCancelLostPasswordListener::class);
+    EventHandler::getInstance()->register(PackageListChanged::class, static function () {
```
TimWolla, line 22: Maybe add empty lines (and perhaps a comment) to visually group related events.

wcfsetup/install/files/lib/system/event/listener/PhraseChangedPreloadListener.class.php
```diff
@@ -0,0 +1,59 @@
...
+        if ($resetCache) {
+            $command = new ResetPreloadCache($event->language);
+            $command();
+        }
+    }
+
+    /**
+     * @return PreloadPhrase[]
+     */
+    private function getPhrasesForPreloading(Language $language): array
```
TimWolla, line 52:
```suggestion
    /**
     * @return iterable<PreloadPhrase>
     */
    private function getPhrasesForPreloading(Language $language): iterable
```

TimWolla (commented):
Overall I like the result very much. Great work!

wcfsetup/install/files/lib/system/WCF.class.php
```diff
@@ -812,6 +812,7 @@ protected function assignDefaultTemplateVariables()
             }
         }
 
+
```
TimWolla, line 815: This is a left-over change that should be removed.

wcfsetup/install/files/lib/system/language/preload/command/CachePreloadPhrases.class.php
```diff
@@ -0,0 +1,68 @@
...
+    {
+        $event = new PreloadPhrasesCollecting($this->language);
+        $this->eventHandler->fire($event);
+
+        $file = new Stream(\WCF_DIR . $this->language->getPreloadCacheFilename(), 'wb');
+        $file->write(
+            \sprintf(
+                "/* cache for '%s' (generated at %s) -- DO NOT EDIT */\n",
+                $this->language->getLocale(),
+                \gmdate('r'),
```
TimWolla, line 43: Prefer creating a fresh `DateTimeImmutable` and the `c` format string, the functional API for date-time is not great.

com.woltlab.wcf/templates/headIncludeJavaScript.tpl
```diff
@@ -24,7 +24,8 @@
 	{/if}
 </script>
 
-<script src="{$__wcf->getPath()}js/WoltLabSuite/WebComponent.js"></script>
+<script src="{$phrasePreloader->getUrl($__wcf->language)}"></script>
+<script src="{$__wcf->getPath()}js/WoltLabSuite/WebComponent.js?v={@LAST_UPDATE_TIME}"></script>
```
TimWolla, line 28: This is likely the wrong order: The WebComponent bundle needs to be loaded first for the called functions to exist.

wcfsetup/install/files/lib/system/language/event/LanguageImported.class.php
```diff
@@ -0,0 +1,24 @@
...
+ * @author Alexander Ebert
+ * @copyright 2001-2022 WoltLab GmbH
+ * @license GNU Lesser General Public License <http://opensource.org/licenses/lgpl-license.php>
+ * @package WoltLabSuite\Core\System\Language\Event
+ * @since 6.0
+ */
+final class LanguageImported implements IEvent
+{
+    public function __construct(
+        public readonly Language $language
```
TimWolla, line 21:
```suggestion
        private readonly Language $language
```

and add a getter. This allows us to more easily refactor this to only store the ID for serialization in the future.

Ditto for the other events.

wcfsetup/install/files/lib/system/language/preload/PhrasePreloader.class.php
```diff
@@ -0,0 +1,49 @@
...
+ */
+final class PhrasePreloader
+{
+    /**
+     * Returns the URL to the preload cache. Will implicitly
+     * create the cache if it does not exist.
+     */
+    public function getUrl(Language $language): string
+    {
+        if (!$this->needsRebuild($language)) {
```
TimWolla, line 27:
```suggestion
        if ($this->needsRebuild($language)) {
```

Instead the return value should be inverted within the method. Otherwise the code reads like nonsense.

wcfsetup/install/files/lib/system/language/preload/command/CachePreloadPhrases.class.php
```diff
@@ -0,0 +1,68 @@
...
+
+        $this->eventHandler = EventHandler::getInstance();
+    }
+
+    public function __invoke(): void
+    {
+        $event = new PreloadPhrasesCollecting($this->language);
+        $this->eventHandler->fire($event);
+
+        $file = new Stream(\WCF_DIR . $this->language->getPreloadCacheFilename(), 'wb');
```
TimWolla, line 38: We likely need the `AtomicWriter`, instead of a Laminas Stream here. Otherwise multiple requests might write into the file at the same time. The method calls should likely be identical, it might require an additional `->flush()` before the `->close()`, though.

wcfsetup/install/files/lib/system/language/preload/command/ResetPreloadCache.class.php
```diff
@@ -0,0 +1,30 @@
...
+
+    public function __construct(Language $language)
+    {
+        $this->language = $language;
+    }
+
+    public function __invoke(): void
+    {
+        // Try to remove the file if it exists.
+        @\unlink(\WCF_DIR . $this->language->getPreloadCacheFilename());
```
TimWolla, line 28:
```suggestion
        if (\WCF_DIR . $this->language->getPreloadCacheFilename()) {
            \unlink(\WCF_DIR . $this->language->getPreloadCacheFilename());
        }
```
Avoid placing an `@` by default. In this case I don't expect to see a TOCTOU problem, because this command is not called in some hot path.

wcfsetup/install/files/lib/system/language/preload/event/PreloadPhrasesCollecting.class.php
```diff
@@ -0,0 +1,48 @@
...
+ *
+ * @author Alexander Ebert
+ * @copyright 2001-2022 WoltLab GmbH
+ * @license GNU Lesser General Public License <http://opensource.org/licenses/lgpl-license.php>
+ * @package WoltLabSuite\Core\System\Language\Preload\Event
+ * @since 6.0
+ */
+final class PreloadPhrasesCollecting implements IEvent
+{
+    public readonly Language $language;
```
TimWolla, line 20: Likewise this should be `private`.

PR #5118, opened 2022-11-10
Remove legacy cookie banner

BurntimeX:
From a legal point of view, this type of cookie banner is useless.


TimWolla (commented):
Diff LGTM, but should the CookiePolicy page also be removed: https://github.com/WoltLab/WCF/blob/6087c1ebd9905b2dc10e4cd212307029eef22627/com.woltlab.wcf/page.xml#L655?

TimWolla:
> I would keep the page in case someone linked to it elsewhere or added their own content.

It could be removed for new installations only + a comment in the `<delete>` section why it's not actively deleted.

PR #5123, opened 2022-11-14
Migrate the forced check of the background queue to a HTTP response header

dtdesign:
Some AJAX actions triggered background jobs that should be attempted to be executed immediately. This was communicated to the client by injecting a property into the JSON response which is both cumbersome and does not play nice with non-JSON-responses.

Using the custom header `woltlab-background-queue-check: yes` allows us to provide the hint regardless of the actual response.

### DIFF
--- a/wcfsetup/install/files/lib/action/AJAXProxyAction.class.php
+++ b/wcfsetup/install/files/lib/action/AJAXProxyAction.class.php
@@ -3,13 +3,12 @@
 namespace wcf\action;
 
 use wcf\data\IDatabaseObjectAction;
-use wcf\http\middleware\TriggerBackgroundQueue;
+use wcf\system\background\BackgroundQueueHandler;
 use wcf\system\exception\ImplementationException;
 use wcf\system\exception\ParentClassException;
 use wcf\system\exception\UserInputException;
 use wcf\system\request\RequestHandler;
 use wcf\system\WCF;
-use wcf\system\WCFACP;
 use wcf\util\ArrayUtil;
 use wcf\util\StringUtil;
 
@@ -120,8 +119,8 @@ class AJAXProxyAction extends AJAXInvokeAction
             @\header(
                 \sprintf(
                     '%s: %s',
-                    TriggerBackgroundQueue::HEADER_NAME,
-                    TriggerBackgroundQueue::HEADER_VALUE,
+                    BackgroundQueueHandler::FORCE_CHECK_HTTP_HEADER_NAME,
+                    BackgroundQueueHandler::FORCE_CHECK_HTTP_HEADER_VALUE,
                 )
             );
         }
--- a/wcfsetup/install/files/lib/action/AJAXProxyAction.class.php
+++ b/wcfsetup/install/files/lib/action/AJAXProxyAction.class.php
@@ -114,11 +114,6 @@ class AJAXProxyAction extends AJAXInvokeAction
             }
         }
 
-        // force background queue invocation
-        if (!\class_exists(WCFACP::class, false) && WCF::getSession()->getVar('forceBackgroundQueuePerform')) {
-            $this->response['forceBackgroundQueuePerform'] = true;
-        }
-
         parent::sendResponse();
     }
 }
--- a/wcfsetup/install/files/lib/action/AJAXProxyAction.class.php
+++ b/wcfsetup/install/files/lib/action/AJAXProxyAction.class.php
@@ -3,9 +3,11 @@
 namespace wcf\action;
 
 use wcf\data\IDatabaseObjectAction;
+use wcf\http\middleware\TriggerBackgroundQueue;
 use wcf\system\exception\ImplementationException;
 use wcf\system\exception\ParentClassException;
 use wcf\system\exception\UserInputException;
+use wcf\system\request\RequestHandler;
 use wcf\system\WCF;
 use wcf\system\WCFACP;
 use wcf\util\ArrayUtil;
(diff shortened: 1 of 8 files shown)


wcfsetup/install/files/lib/http/middleware/TriggerBackgroundQueue.class.php
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    /**
+     * @inheritDoc
+     */
+    public function process(ServerRequestInterface $request, RequestHandlerInterface $handler): ResponseInterface
+    {
+        $response = $handler->handle($request);
+        if (
+            \class_exists(WCFACP::class, false)
```
TimWolla, line 42:
```suggestion
            RequestHandler::getInstance()->isACPRequest()
```

wcfsetup/install/files/lib/system/background/BackgroundQueueHandler.class.php
```diff
@@ -20,6 +20,12 @@
  */
 final class BackgroundQueueHandler extends SingletonFactory
 {
+    /**
+     * Indicates that the client should trigger a check for
+     * pending jobs in the background queue.
```
TimWolla, line 25: This explanation should rather be moved into the getter’s PHPDoc, because the latter is part of the public API.

wcfsetup/install/files/lib/http/middleware/TriggerBackgroundQueue.class.php
```diff
@@ -0,0 +1,58 @@
...
+            || !$this->backgroundQueueHandler->hasPendingCheck()
+        ) {
+            return $response;
+        }
+
+        if ($response instanceof LegacyPlaceholderResponse) {
+            \header(
+                \sprintf('%s: %s', self::HEADER_NAME, self::HEADER_VALUE)
+            );
```
TimWolla, line 51: This might be unsafe for two reasons:

- The non-PSR-15 controller might’ve already emitted some output and thus headers are already sent, causing an error.
- The non-PSR-15 controller might call `exit;`, preventing any middlewares from running in the response path (this is what happens in AJAXInvokeAction and thus AJAXProxyAction).

For `LegacyPlaceholderResponse` the rule is: Don’t touch, here be dragons.

-----------

Instead for compatibility the following might be required:

- `BackgroundQueueHandler::forceCheck()` adds the header on a best-effort basis (with `@\header()`, including the `@`).
- This middleware removes the header again with `\header_remove()` and then adds it to the PSR-7 response object to be emitted by the PSR-7 pipeline, similarly to `PreventMimeSniffing`.

-------------------------

Alternatively the legacy AJAXProxyAction continues to add the metadata to the JSON itself and the header is only used for new-style controllers. Technically `AJAXInvokeAction` could also be adjusted to emit a PSR-7 response in `AJAXInvokeAction::sendResponse()` but that might be considered a break.

wcfsetup/install/files/lib/action/AJAXProxyAction.class.php
```diff
@@ -114,9 +115,14 @@ protected function sendResponse()
             }
         }
 
-        // force background queue invocation
-        if (!\class_exists(WCFACP::class, false) && WCF::getSession()->getVar('forceBackgroundQueuePerform')) {
-            $this->response['forceBackgroundQueuePerform'] = true;
+        if (!RequestHandler::getInstance()->isACPRequest() && WCF::getSession()->getVar('forceBackgroundQueuePerform')) {
```
TimWolla, line 118: As this is modified anyway: This could also check `BackgroundQueueHandler::getInstance()->hasPendingCheck()`.

wcfsetup/install/files/lib/http/middleware/TriggerBackgroundQueue.class.php
```diff
@@ -0,0 +1,56 @@
...
+
+        if ($response instanceof LegacyPlaceholderResponse) {
+            return $response;
+        }
+
+        return $response->withHeader(
+            BackgroundQueueHandler::FORCE_CHECK_HTTP_HEADER_NAME,
+            BackgroundQueueHandler::FORCE_CHECK_HTTP_HEADER_VALUE,
+        );
+    }
```
TimWolla, line 55:
```suggestion
        if ($this->requestHandler->isACPRequest()) {
            return $handler->handle($request);
        }

        $response = $handler->handle($request);

        if ($response instanceof LegacyPlaceholderResponse) {
            return $response;
        }

        if (!$this->backgroundQueueHandler->hasPendingCheck()) {
            return $response;
        }

        return $response->withHeader(
            BackgroundQueueHandler::FORCE_CHECK_HTTP_HEADER_NAME,
            BackgroundQueueHandler::FORCE_CHECK_HTTP_HEADER_VALUE,
        );
    }
```

I've slightly reordered the logic for the follow reasons:

1. Completely bypass the middleware logic for ACP requests. I consider this a little cleaner. If the “is ACP request” information is moved into an attribute on the `ServerRequestInterface`, then acting on information from the request in the response path is generally a little questionable: The middleware can’t know what happened to the request (and its attributes) in the following middlewares. By moving the ACP check into a dedicated check, this is nicely separated.

2. Handle the `LegacyPlaceholderResponse` first in the response path to ensure a “poisoned” response is not accidentally touched.

TimWolla (commented):
Overall this PR LGTM now.

wcfsetup/install/files/lib/action/AJAXProxyAction.class.php
dtdesign: I have considered this, but opted out. First of all, this is as close to the existing implementation and second I consider the session variable to be a superset of `hasPendingCheck()`.

At worst we’re invoking the background queue “twice” in case there is a concurrent request that enabled that flag. And there is no such thing as “invoking the background queue _too often_”.

TimWolla: Okay, fair enough.

PR #5124, opened 2022-11-14
Implicit XSRF checks for PSR-15 controllers with opt-out

dtdesign:
Calls to PSR-15 controllers will implicitly validate the XSRF-Token for requests using any HTTP verb other than `GET` or `HEAD`. This “security by default” approach makes it easy to use controllers without worrying too much about security.

Controllers can explicitly opt-out of the XSRF validation through the attribute `#[DisableXsrfCheck]`.

### DIFF
--- a/wcfsetup/install/files/lib/http/middleware/Xsrf.class.php
+++ b/wcfsetup/install/files/lib/http/middleware/Xsrf.class.php
@@ -54,14 +54,19 @@ final class Xsrf implements MiddlewareInterface
         );
 
         if (
-            $request->getMethod() !== 'GET'
-            && $request->getMethod() !== 'HEAD'
+            $this->isSafeHttpMethod($request->getMethod())
             && $this->requestHandler->getActiveRequest()
         ) {
             $this->assertHasValidXsrfToken($this->requestHandler->getActiveRequest(), $hasValidXsrfToken);
         }
 
         return $handler->handle($request);
+    }
+
+    private function isSafeHttpMethod(string $verb): bool {
+        // HTTP requests using the 'GET' or 'HEAD' verb are safe
+        // by design, because those should not alter the state.
+        return $verb === 'GET' || $verb === 'HEAD';
     }
 
     private function assertHasValidXsrfToken(Request $request, $hasValidXsrfToken): void
--- a/wcfsetup/install/files/lib/http/middleware/Xsrf.class.php
+++ b/wcfsetup/install/files/lib/http/middleware/Xsrf.class.php
@@ -56,18 +56,33 @@ final class Xsrf implements MiddlewareInterface
             $request->getMethod() !== 'GET'
             && $request->getMethod() !== 'HEAD'
             && $this->requestHandler->getActiveRequest()
-            && \is_subclass_of($this->requestHandler->getActiveRequest()->getClassName(), RequestHandlerInterface::class)
         ) {
-            if (!$this->validateXsrfToken($this->requestHandler->getActiveRequest(), $hasValidXsrfToken)) {
-                throw new InvalidSecurityTokenException();
-            }
+            $this->assertHasValidXsrfToken($this->requestHandler->getActiveRequest(), $hasValidXsrfToken);
         }
 
         return $handler->handle($request);
     }
 
-    private function validateXsrfToken(Request $request, $hasValidXsrfToken): bool
+    private function assertHasValidXsrfToken(Request $request, $hasValidXsrfToken): void
     {
-        return $hasValidXsrfToken;
+        if (!\is_subclass_of($request->getClassName(), RequestHandlerInterface::class)) {
+            // Skip the XSRF check for legacy controllers.
+            return;
+        }
+
+        $reflectionClass = new \ReflectionClass($request->getClassName());
+        if ($reflectionClass->getAttributes('DisableXsrfCheck') !== []) {
+            // Controller has opted out of the XSRF check.
+            return;
+        }
+
+        if (!$hasValidXsrfToken) {
+            throw new InvalidSecurityTokenException();
+        }
     }
+}
+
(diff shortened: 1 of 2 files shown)


wcfsetup/install/files/lib/http/middleware/Xsrf.class.php
```diff
@@ -34,11 +44,45 @@ public function process(ServerRequestInterface $request, RequestHandlerInterface
...
+
+    private function assertHasValidXsrfToken(Request $request, $hasValidXsrfToken): void
+    {
+        if (!\is_subclass_of($request->getClassName(), RequestHandlerInterface::class)) {
+            // Skip the XSRF check for legacy controllers.
+            return;
+        }
+
+        $reflectionClass = new \ReflectionClass($request->getClassName());
+        if ($reflectionClass->getAttributes('DisableXsrfCheck') !== []) {
```
TimWolla, line 74:
```suggestion
        if ($reflectionClass->getAttributes(DisableXsrfCheck::class) !== []) {
```

wcfsetup/install/files/lib/http/middleware/Xsrf.class.php
```diff
@@ -34,11 +44,45 @@ public function process(ServerRequestInterface $request, RequestHandlerInterface
...
+        if (!$hasValidXsrfToken) {
+            throw new InvalidSecurityTokenException();
+        }
+    }
+}
+
+#[\Attribute(\Attribute::TARGET_CLASS)]
+final class DisableXsrfCheck
+{
 }
```
TimWolla, line 88: This should be placed in a dedicated file/namespace for autoloading / autocompletion to work properly.

wcfsetup/install/files/lib/http/middleware/Xsrf.class.php
```diff
@@ -34,11 +44,45 @@ public function process(ServerRequestInterface $request, RequestHandlerInterface
...
+
         $request = $request->withAttribute(
             self::HAS_VALID_HEADER_ATTRIBUTE,
-            \hash_equals($xsrfToken, $request->getHeaderLine('x-xsrf-token'))
+            $hasValidXsrfToken
         );
 
+        if (
+            $request->getMethod() !== 'GET'
+            && $request->getMethod() !== 'HEAD'
```
TimWolla, line 57: Perhaps add a helper method `private function isSafeMethod(string $method): bool` that returns true for GET/HEAD and false otherwise.

wcfsetup/install/files/lib/http/middleware/Xsrf.class.php
```diff
@@ -34,11 +45,45 @@ public function process(ServerRequestInterface $request, RequestHandlerInterface
...
+        $hasValidXsrfToken = \hash_equals($xsrfToken, $request->getHeaderLine('x-xsrf-token'));
+
         $request = $request->withAttribute(
             self::HAS_VALID_HEADER_ATTRIBUTE,
-            \hash_equals($xsrfToken, $request->getHeaderLine('x-xsrf-token'))
+            $hasValidXsrfToken
         );
 
+        if (
+            $this->isSafeHttpMethod($request->getMethod())
```
TimWolla, line 57:
```suggestion
            !$this->isSafeHttpMethod($request->getMethod())
```

wcfsetup/install/files/lib/http/middleware/Xsrf.class.php
```diff
@@ -34,11 +45,45 @@ public function process(ServerRequestInterface $request, RequestHandlerInterface
...
         return $handler->handle($request);
     }
+
+    private function isSafeHttpMethod(string $verb): bool {
+        // HTTP requests using the 'GET' or 'HEAD' verb are safe
+        // by design, because those should not alter the state.
+        return $verb === 'GET' || $verb === 'HEAD';
+    }
+
+    private function assertHasValidXsrfToken(Request $request, $hasValidXsrfToken): void
```
TimWolla, line 72:
```suggestion
    private function assertHasValidXsrfToken(Request $request, bool $hasValidXsrfToken): void
```

TimWolla (commented):
The overall structure LGTM if the those annotations are fixed. I did not test this, though.

dtdesign:
I haven’t actually tested it myself, because there is no proper test case at this time. I intend to put it to the test with #5106 though, which was the primary motivation for this PR anyway.

wcfsetup/install/files/lib/http/middleware/Xsrf.class.php
```diff
@@ -34,11 +45,45 @@ public function process(ServerRequestInterface $request, RequestHandlerInterface
...
+
+        $reflectionClass = new \ReflectionClass($request->getClassName());
+        if ($reflectionClass->getAttributes(DisableXsrfCheck::class) !== []) {
+            // Controller has opted out of the XSRF check.
+            return;
+        }
+
+        if (!$hasValidXsrfToken) {
+            throw new InvalidSecurityTokenException();
+        }
```
TimWolla, line 87:
```suggestion
        if ($hasValidXsrfToken) {
            // No need to do anything for a valid token.
            return;
        }

        if (!\is_subclass_of($request->getClassName(), RequestHandlerInterface::class)) {
            // Skip the XSRF check for legacy controllers.
            return;
        }

        $reflectionClass = new \ReflectionClass($request->getClassName());
        if ($reflectionClass->getAttributes(DisableXsrfCheck::class) !== []) {
            // Controller has opted out of the XSRF check.
            return;
        }

        // The controller requires a valid XSRF Token and no valid
        // token was provided, abort the processing.
        throw new InvalidSecurityTokenException();
```

One more optimization: Move the test for `$hasValidXsrfToken` to the top of this function and return early otherwise. This ensures that the reflection logic does not need to run for the happy-path of a valid token (which should be the majority of requests).

TimWolla (approved):

PR #5127, opened 2022-11-16
Further clean up the package system

TimWolla:


TimWolla:
Dropping the remaining changes, there is no real benefit in them, they were just done while building the mental model for the actual changes.

PR #5130, opened 2022-11-18
Improve registration of rebuild workers

TimWolla:


wcfsetup/install/files/lib/system/worker/event/RebuildWorkerCollecting.class.php
```diff
@@ -0,0 +1,53 @@
...
+            $this->queue->insert(
+                new RegisteredWorker($objectType->className, $objectType),
+                -$objectType->nicevalue ?: 0
+            );
+        }
+    }
+
+    /**
+     * Registers the name of a phrases that should
+     * be added to the list of preloaded values.
```
TimWolla, line 39: Fix

wcfsetup/install/files/lib/system/worker/RegisteredWorker.class.php
```diff
@@ -0,0 +1,54 @@
...
+ */
+final class RegisteredWorker
+{
+    public function __construct(
+        public readonly string $classname,
+        private readonly ?ObjectType $legacyObjectType = null,
+    ) {
+    }
+
+    public function getName(): string
```
dtdesign, line 25: I would like to see a `private getIdentifier(): string` that returns the `objectType` or the sanitized `$className`. This allows both `getName()` and `getDescription()` to stop caring about the type of worker and simplify their code to a plain `return`.

dtdesign (changes requested):
Otherwise LGTM

dtdesign (approved):

PR #5147, opened 2022-11-25
Add experimental context filter for search backends

dtdesign:
Every search consists of an outer and inner query, with the inner query directly targeting the search backend. For performance reasons the results from the inner query are capped (although at a much greater limit). This limitation can cause issue if the filters on the outer query remove too many results, because the inner query was too broad.

### DIFF
--- a/wcfsetup/install/files/lib/system/search/IContextAwareSearchEngine.class.php
+++ b/wcfsetup/install/files/lib/system/search/IContextAwareSearchEngine.class.php
@@ -1,0 +1,59 @@
+<?php
+
+namespace wcf\system\search;
+
+use wcf\system\database\util\PreparedStatementConditionBuilder;
+
+/**
+ * Default interface for search engines that support
+ * filtering by context.
+ *
+ * @author Alexander Ebert
+ * @copyright 2001-2022 WoltLab GmbH
+ * @license GNU Lesser General Public License <http://opensource.org/licenses/lgpl-license.php>
+ * @package WoltLabSuite\Core\System\Search
+ * @since 6.0
+ */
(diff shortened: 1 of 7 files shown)


wcfsetup/install/files/lib/system/search/IContextAwareSearchEngine.class.php
```diff
@@ -0,0 +1,59 @@
...
+
+    /**
+     * Returns the inner join query and the condition parameters. This method is allowed to return NULL for both the
+     * 'fulltextCondition' and 'searchIndexCondition' index instead of a PreparedStatementConditionBuilder instance:
+     *
+     * array(
+     *  'fulltextCondition' => $fulltextCondition || null,
+     *  'searchIndexCondition' => $searchIndexCondition || null,
+     *  'sql' => $sql
+     * );
```
TimWolla, line 34:
```suggestion
     * @return array{
     *             fulltextCondition: ?whateverthetypeshouldbe
     *             searchIndexCondition: ?whateverthetypeshouldbe
     *             sql: string
     *         }
```

This uses the array shape syntax as supported by Psalm: https://psalm.dev/docs/annotating_code/type_syntax/array_types/#array-shapes

dtdesign: This was copied from `getInnerJoin()` and the type is `?PreparedStatementConditionBuilder`. However, my IDE doesn’t support this syntax and I have no other tool at hand to validate it. Please feel free to add the proper typing to this interface and the base interface.

TimWolla (commented):
I lack the necessary background info to review this appropriately, but I don't see anything obviously wrong.

PR #5155, opened 2022-11-30
Migrate to native CSS variables

dtdesign:
Switching to native CSS variables makes it much easier to modify the values on runtime.

### DIFF
--- a/wcfsetup/install/files/acp/style/layout.scss
+++ b/wcfsetup/install/files/acp/style/layout.scss
@@ -1,5 +1,7 @@
-var(--wcfAcpMenuWidth): 150px;
-var(--wcfAcpSubMenuWidth): 300px;
+html {
+	--wcfAcpMenuWidth: 150px;
+	--wcfAcpSubMenuWidth: 300px;
+}
 
 .layoutBoundary {
 	margin: 0;
--- a/wcfsetup/install/files/acp/style/layout.scss
+++ b/wcfsetup/install/files/acp/style/layout.scss
@@ -278,7 +278,7 @@ html {
 	.pageContainer.acpPageSubMenuActive {
 		.acpPageContentContainer,
 		.pageFooter {
-			padding-left: var(--wcfAcpMenuWidth) + var(--wcfAcpSubMenuWidth);
+			padding-left: calc(var(--wcfAcpMenuWidth) + var(--wcfAcpSubMenuWidth));
 		}
 	}
 }
--- a/wcfsetup/install/files/acp/style/layout.scss
+++ b/wcfsetup/install/files/acp/style/layout.scss
@@ -1,5 +1,5 @@
-$wcfAcpMenuWidth: 150px;
-$wcfAcpSubMenuWidth: 300px;
+var(--wcfAcpMenuWidth): 150px;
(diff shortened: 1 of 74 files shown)


wcfsetup/install/files/style/bbcode/code.scss
```diff
@@ -168,7 +168,7 @@
 	}
 
 	> .toggleButton {
-		background-color: opacify(var(--wcfContentBackground), 0.9);
+		background-color: rgba(var(--wcfContentBackground-rgb) / 0.9);
```
TimWolla, line 171: I just compared this with scssphp, this change results in a visual change if `$wcfContentBackground` is a color with existing alpha channel:

```
$foo: rgba(255, 0, 27, 0.3);

body {
    color: opacify($foo, 0.5);
}
```
results in:
```
body {
  color: rgba(255, 0, 27, 0.8);
}
```

So any color with more than 10% opacity would no longer have any alpha channel, whereas other colors would have something between 90% and 100% opacity. Now everything has a 90% alpha channel.

Same applies to the other changes here.

wcfsetup/install/files/acp/style/layout.scss
```diff
@@ -1,5 +1,7 @@
-var(--wcfAcpMenuWidth): 150px;
-var(--wcfAcpSubMenuWidth): 300px;
+html {
```
TimWolla, line 1: Is there a difference between `:root` and `html`?

dtdesign: Yes: `:root` is more specific than `html`:

```css
:root {
  --color: red;
}

html {
  --color: blue;
}

body: {
  color: var(--color); // red
}
```

cadeyrn: Yes, :root has a higher specifity than html (and it's the de facto standard for defining HTML variables). But I can't judge if there are reasons to define them in html here.

wcfsetup/install/files/style/bbcode/code.scss
dtdesign: I’m 99.9% confident that the use of `opacity()` was incorrect before. Both before and after looks the same for the default colors, so it’s very likely that this was an unintentional misuse.

wcfsetup/install/files/acp/style/layout.scss
dtdesign: My intention was to offer customization through `:root {}` without having to care about the selector order. In the long run this could be used to offer an easy way to overwrite CSS variables through the style editor without actually changing the generated values for `html {}`.

From a technical standpoint it doesn’t really matter which of these two selectors is actually used, because they will both target the same element. My approach has the upside of being able to make use of the difference in specificity, for better or for worse.

wcfsetup/install/files/style/bbcode/code.scss
TimWolla: > I’m 99.9% confident that the use of opacity() was incorrect before.

(Note: opaciFy, not opacity)

Yeah, I agree with that conclusion, the described behavior of "something between 90 and 100 with the vast majority 100" doesn't appear to be super useful. I just wanted to point out the difference so this can the double-checked.

PR #5156, opened 2022-12-02
Several fixes in phpdoc

mutec:


wcfsetup/install/files/lib/util/ArrayUtil.class.php
```diff
@@ -81,7 +81,7 @@ public static function encodeHTML($array)
      * Applies stripslashes on all elements of the given array.
      *
      * @param array $array
-     * @return  array
+     * @return  array|string
```
TimWolla, line 84: This one is not more correct than the existing one (because string may only be returned if `$array` is not an array) and much less useful.

wcfsetup/install/files/lib/system/package/PackageInstallationSQLParser.class.php
```diff
@@ -260,8 +260,6 @@ protected function getIndexOwnerID($tableName, $indexName)
             return $row['packageID'];
         } elseif (isset($this->knownTables[$tableName])) {
             return $this->knownTables[$tableName];
-        } else {
-            return;
```
TimWolla, line 264: This should probably rather be an explicit `return null;` to make this a total function, but one would need to carefully check all users and add real types first.

wcfsetup/install/files/lib/system/form/builder/field/IconFormField.class.php
```diff
@@ -18,7 +18,7 @@ class IconFormField extends AbstractFormField implements IImmutableFormField
 {
     use TImmutableFormField;
 
-    private FontAwesomeIcon $icon;
```
TimWolla, line 21: This one appears to be correct, but does not belong in this PR, because this is not PHPDoc.

There might also be some bug lurking here, because the `$icon` is not invalidated when the value changes. Perhaps the property should be removed entirely.

wcfsetup/install/files/lib/data/user/group/UserGroup.class.php
```diff
@@ -69,7 +69,7 @@ class UserGroup extends DatabaseObject implements ITitledObject
 
     /**
      * group cache
-     * @var UserGroup[]
+     * @var UserGroup[][]|int[][]
```
TimWolla, line 72: This should rather be an array shape or completely cleaned up to not require this type in the first place.

TimWolla (changes requested):

TimWolla:
Picked the correct ones, thanks.

PR #5158, opened 2022-12-02
Add web component for pagination

BurntimeX:


ts/WoltLabSuite/WebComponent/woltlab-core-pagination.ts
```diff
@@ -0,0 +1,194 @@
...
+      });
+      li.append(button);
+
+      return li;
+    }
+
+    getLinkUrl(page: number): string {
+      const url = new URL(this.url);
+      url.search += url.search !== "" ? "&" : "?";
+      url.search += new URLSearchParams([["pageNo", page.toString()]]);
```
TimWolla, line 171:
```suggestion
      const url = new URL(this.url);
      url.searchParams.set('pageNo', page.toString());
```

ts/WoltLabSuite/WebComponent/woltlab-core-pagination.ts
```diff
@@ -0,0 +1,194 @@
...
+    get count(): number {
+      return this.hasAttribute("count") ? parseInt(this.getAttribute("count")!) : 0;
+    }
+
+    get page(): number {
+      return this.hasAttribute("page") ? parseInt(this.getAttribute("page")!) : 1;
+    }
+
+    get url(): string {
+      return this.getAttribute("url")!;
```
TimWolla, line 189: I wonder if a `href` attribute would be more natural.

ts/WoltLabSuite/Core/Ui/Pagination/JumpToPage.ts
```diff
@@ -0,0 +1,50 @@
...
+          class="tiny"
+          type="number"
+          min="1"
+          max="${element.count}"
+          value="${element.count}"
+          aria-label="${getPhrase("wcf.page.jumpTo.pageNo")}"
+          placeholder="${getPhrase("wcf.page.jumpTo.pageNo")}"
+          required
+        >
+        <small>${getPhrase("wcf.page.jumpTo.description").replace(/#pages#/, element.count.toString())}</small>
```
TimWolla, line 30: I know the phrase is left unmodified, but I feel like we should migrate to a regular variable here for consistency:

```suggestion
        <small>${getPhrase("wcf.page.jumpTo.description", { pages: element.count })}</small>
```

ts/WoltLabSuite/WebComponent/woltlab-core-pagination.ts
BurntimeX: This would re-sort the parameters, which is not wanted here.

ts/WoltLabSuite/WebComponent/woltlab-core-pagination.ts
```diff
@@ -0,0 +1,318 @@
...
+      const nav = this.#getNavElement();
+      this.append(nav);
+
+      const previousLinkElement = this.#getPreviousLinkElement();
+      if (previousLinkElement) {
+        nav.append(previousLinkElement);
+      }
+
+      const ul = document.createElement("ul");
+      ul.classList.add(this.#className + "__list");
```
TimWolla, line 52:
```suggestion
      ul.classList.add(`${this.#className}__list);
```
Injecting `#className` is currently performed inconsistently. A template string is the more readable variant.

ts/WoltLabSuite/WebComponent/woltlab-core-pagination.ts
```diff
@@ -0,0 +1,318 @@
...
+      return li;
+    }
+
+    /**
+     * On smaller screens the ellipsis is shown when there
+     * is at least one page in-between the current page and
+     * the first or last page. On larger screens the ellipsis
+     * is only shown if it hides at least two numbers.
+     */
+    get thresholdForEllispsis(): number {
```
TimWolla, line 235:
```suggestion
    get thresholdForEllipsis(): number {
```

ts/WoltLabSuite/WebComponent/woltlab-core-pagination.ts
```diff
@@ -0,0 +1,318 @@
...
+
+      return button;
+    }
+
+    #getLinkItem(page: number): HTMLLIElement {
+      const li = document.createElement("li");
+      li.classList.add(this.#className + "__item");
+
+      const button = this.#getButtonElement(page);
+      button.ariaLabel = window.WoltLabLanguage.getPhrase("wcf.page.pageNo", { pageNo: page });
```
TimWolla, line 162: This phrase appears to be missing from JavaScript.

ts/WoltLabSuite/WebComponent/woltlab-core-pagination.ts
```diff
@@ -0,0 +1,318 @@
...
+      if (page === this.page) {
+        button.ariaCurrent = "page";
```
TimWolla, line 164: The `ariaCurrent` property is not supported by Firefox: https://developer.mozilla.org/en-US/docs/Web/API/Element/ariaCurrent. This will cause issues, because the CSS uses the `aria-current` attribute, instead of the `pagination__link--current` class.

ts/WoltLabSuite/Core/Ui/Pagination/JumpToPage.ts
```diff
@@ -0,0 +1,50 @@
...
+      <dt></dt>
+      <dd>
+        <input
+          class="tiny"
+          type="number"
+          min="1"
+          max="${element.count}"
+          value="${element.count}"
+          aria-label="${getPhrase("wcf.page.jumpTo.pageNo")}"
+          placeholder="${getPhrase("wcf.page.jumpTo.pageNo")}"
```
TimWolla, line 27: This placeholder is not really seen, because there is a value by default and because the `<input>` is `tiny`. Should this be put as the `<dt>` content instead?

wcfsetup/install/files/lib/system/template/plugin/PagesFunctionTemplatePlugin.class.php
```diff
@@ -160,6 +64,10 @@ public function execute($tagArgs, TemplateEngine $tplObj)
                 $parameters['application'] = $tagArgs['application'];
             }
 
+            // The previous implementation required the `pageNo=%d` placeholder
+            // to be present in the link argument.
+            $tagArgs['link'] = \preg_replace('~(^pageNo=%d&?|&pageNo=%d)~', '', $tagArgs['link']);
```
TimWolla, line 69:
```suggestion
            $tagArgs['link'] = \preg_replace('~(?:^pageNo=%d(?:&|$))|(?:&pageNo=%d(?=&|$))~', '', $tagArgs['link']);
```

This regular expression is a little more defensive.

wcfsetup/install/files/lib/system/template/plugin/PagesFunctionTemplatePlugin.class.php
```diff
@@ -172,95 +80,12 @@ public function execute($tagArgs, TemplateEngine $tplObj)
...
-            // last page
-            $html .= $this->makeLink($link, $tagArgs['pages'], $tagArgs['page'], $tagArgs['pages']);
-
-            // next page
-            $html .= $this->makeNextLink($link, $tagArgs['page'], $tagArgs['pages']);
-
-            // close div and ul
-            $html .= "</ul></nav>\n";
+            $html = \sprintf(
+                '<woltlab-core-pagination page="%d" count="%d" url="%s"></woltlab-core-pagination>',
```
TimWolla, line 84: I dislike it a little that the element vanishes entirely without JavaScript, without any indication that something is missing. Perhaps the `.textContent` could be some simple “Page 7/8” text? This makes it a little more obvious that something is amiss in case the JS breaks for whatever reason. Of course this would require undeprecating the template plugin or creating another one with a reasonable API.

dtdesign: This would add little to no value but at the same time increases the complexity. The JS breaking apart is most likely the symptom of a much greater issue, so this is acceptable. This does not cover a browser where JS is disabled, but at that point you should prepare yourself for a rough ride anyway.

TimWolla: I just remembered that `:defined` exists.

```
woltlab-core-pagination:not(:defined)::after {
    content: 'Page ' attr(page) '/' attr(count);
}
```

would possibly work, but I'll leave this decision up to you :smiley:

ts/WoltLabSuite/WebComponent/woltlab-core-pagination.ts
```diff
@@ -0,0 +1,318 @@
...
+          continue;
+        }
+
+        items.push(this.#getLinkItem(i));
+      }
+
+      return items;
+    }
+
+    #getEllipsesItem(): HTMLLIElement {
```
TimWolla, line 210:
```suggestion
    #getEllipsisItem(): HTMLLIElement {
```
?

Also applies to the `className`.

TimWolla (approved):
Overall this LGTM.

ts/WoltLabSuite/WebComponent/woltlab-core-pagination.ts
```diff
@@ -0,0 +1,316 @@
...
+      const li = document.createElement("li");
+      li.classList.add(`${this.#className}__item`);
+
+      const button = this.#getButtonElement(page);
+      button.setAttribute("aria-label", window.WoltLabLanguage.getPhrase("wcf.page.pageNo", { pageNo: page }));
+      if (page === this.page) {
+        button.setAttribute("aria-current", "page");
+        button.classList.add(`${this.#className}__link--current`);
+      }
+      button.textContent = page.toString();
```
TimWolla, line 165:
```suggestion
      button.textContent = page.toLocaleString(document.documentElement.lang);
```
For correct formatting of numbers > 1000.

PR #5159, opened 2022-12-03
Convert `*Category*Form` to form builder api

mutec:
WSC 6.0 is an ideal moment for converting old template-based forms to form builder-api.
Using the new api it's easier to extend forms with additional fields. There is also no need for using any template-scripting which means less and faster releases.


wcfsetup/install/files/lib/acp/form/ArticleCategoryAddForm.class.php
```diff
@@ -23,100 +27,91 @@ class ArticleCategoryAddForm extends AbstractCategoryAddForm
...
+        $this->form->getDataHandler()->addProcessor(
+            new CustomFormDataProcessor(
+                'articleAdditionalData',
+                static function (IFormDocument $document, array $parameters) {
+                    if (!isset($parameters['additionalData'])) {
+                        $parameters['additionalData'] = [];
+                    }
+
+                    // move sorting data to additional data
+                    $parameters['additionalData'] = $parameters['data']['sortField'];
```
Fighter456, line 92:
```suggestion
                    $parameters['additionalData']['sortField'] = $parameters['data']['sortField'];
```

wcfsetup/install/files/lib/acp/form/ArticleCategoryAddForm.class.php
```diff
@@ -23,100 +27,91 @@ class ArticleCategoryAddForm extends AbstractCategoryAddForm
...
+                    unset($parameters['data']['sortField']);
+                    $parameters['additionalData'] = $parameters['data']['sortOrder'];
```
Fighter456, line 94:
```suggestion
                    $parameters['additionalData']['sortOrder'] = $parameters['data']['sortOrder'];
```

Fighter456 (changes requested):
A very good change that I would definitely welcome. It worked perfectly for me in a first rudimentary test.

The storage of the additional data for the article categories is incorrect. It still has to be verified here whether the reading of the existing values and the display in the processing form then works.

wcfsetup/install/files/lib/acp/form/ArticleCategoryAddForm.class.php
mutec: Thanks, fixed both lines.

Fighter456 (approved):

SoftCreatR (approved):

TimWolla:
An additional note with regard to “process” with regard to the specific changes of this PR, but applicable more generally:

The commits that touch `SingleSelectionFormField` should be proposed in separate PRs and then reviewed independently, as they are not logically related to converting the *Category forms. The newly added functionality just _happen to be_ a required prerequisite. Ideally two PRs, because one is an independent cleanup, whereas the other is a new feature. A proper commit message explaining the *reasons* behind each commit is strongly recommended, unless it's purely “obviously correct” mechanical work (such as adding return types).

mutec:
> You’ve broken the backwards compatibility by modifying the base class(es), requiring changes to all deriving classes. This PR should rather introduce a new base class based on the form builder and make the existing implementations derive from that class instead. The existing base classes can be marked as deprecated and phased out in a future version.

The name of the "new" class would be changed to `AbstractCategoryManagementFrom` or whatever within seconds, so this isn't a show-stopper.

> Please reach out to us before making substantial changes. Simply open an issue so that we are able to discuss what you want to change and how, making it easier to figure out these things before you actually start investing your time.

The whole class was created for some other packages, so it was just renaming and writing the commit-stuff. I already learned not to waste any time for things I don't really need in the past. Mostly it's faster and less of pain to provide things within an own package than doing stuff over here. ;)

> The commits that touch `SingleSelectionFormField` should be proposed in separate PRs and then reviewed independently, as they are not logically related to converting the *Category forms.

I can separate that commit in case it's worth the time it takes - which means there is a guarantee that it will be merged. However it is required to be merged when converting things to FormBuilder in the future.

PR #5174, opened 2022-12-08
Clean up the API in Psr15DialogForm

TimWolla:
This commit adds PHPDoc explanation to the PSR-15 “API” of the DialogForm. It
also adjusts the return type to the generic ResponseInterface: The methods
guarantee that the returned response is appropriate for consumption by the JS
API, but should not make any further guarantees.

### DIFF
--- a/wcfsetup/install/files/lib/system/form/builder/Psr15DialogForm.class.php
+++ b/wcfsetup/install/files/lib/system/form/builder/Psr15DialogForm.class.php
@@ -34,18 +34,18 @@ final class Psr15DialogForm extends FormDocument
 
     /**
      * Processes the form using the request's parsed body. Returns 'null'
-     * if validation succeeded and the result of 'toJsonResponse()' otherwise.
+     * if validation succeeded and the result of 'toResponse()' otherwise.
      *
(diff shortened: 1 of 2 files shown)


wcfsetup/install/files/lib/system/form/builder/Psr15DialogForm.class.php
```diff
@@ -44,7 +51,10 @@ public function validatePsr7Request(ServerRequestInterface $request): ?JsonRespo
         return null;
     }
 
-    public function toJsonResponse(): JsonResponse
+    /**
+     * Returns a response that can be consumed by JavaScript's `dialogFactory().usingFormBuilder()`.
+     */
+    public function toJsonResponse(): ResponseInterface
```
TimWolla, line 57: Perhaps the methods should even be renamed to:

- validateRequest()

and

- toResponse()

respectively?

dtdesign: I have no strong preference here, I’m fine with the proposed names.

dtdesign (approved):

PR #5181, opened 2022-12-12
Add `CheckHttpMethod` middleware

### LINKED ISSUE #5177: Add middleware to validate the HTTP request verb
Only GET and POST should be allowed by default, other verbs may be enabled on an opt-in based using an attribute.

This avoids checks like:

https://github.com/WoltLab/WCF/blob/555dc71cd4bb0805ac5cf48e685fc25a80a69fd7/wcfsetup/install/files/lib/action/ModerationQueueAssignUserAction.class.php#L119-L121

TimWolla:
Resolves #5177

### DIFF
--- a/wcfsetup/install/files/lib/action/ModerationQueueAssignUserAction.class.php
+++ b/wcfsetup/install/files/lib/action/ModerationQueueAssignUserAction.class.php
@@ -74,7 +74,7 @@ final class ModerationQueueAssignUserAction implements RequestHandlerInterface
 
         $form = $this->getForm($moderationQueue);
 
-        if ($request->getMethod() === 'GET') {
+        if ($request->getMethod() === 'GET' || $request->getMethod() === 'HEAD') {
             return $form->toResponse();
         } elseif ($request->getMethod() === 'POST') {
             $response = $form->validateRequest($request);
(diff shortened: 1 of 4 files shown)


wcfsetup/install/files/lib/action/ModerationQueueAssignUserAction.class.php
```diff
@@ -75,7 +75,7 @@ public function handle(ServerRequestInterface $request): ResponseInterface
 
         $form = $this->getForm($moderationQueue);
 
-        if ($request->getMethod() === 'GET') {
+        if ($request->getMethod() === 'GET' || $request->getMethod() === 'HEAD') {
```
TimWolla, line 78: This is “not great”, but I'd rather not disallow `HEAD` by default. Not sure what the best course of action is here.

dtdesign:
I do not agree with the decision to permit `HEAD` by default. It has very limited use for dynamic endpoints and in my experience is almost exclusively used for DDoS attacks, because it creates a heavy load on the server without exhausting the bandwidth on the client.

We should make `HEAD` an opt-in verb in case an endpoint provides some useful output, for example, when serving dynamically generated assets.

wcfsetup/install/files/lib/http/middleware/CheckHttpMethod.class.php
```diff
@@ -0,0 +1,71 @@
...
+        if (\in_array($request->getMethod(), $validMethods, true)) {
+            return $handler->handle($request);
+        }
+
+        return new TextResponse('The used HTTP method is not allowed.', 405, [
+            'allow' => \implode(', ', $validMethods),
+        ]);
+    }
+
+    private function getValidMethods(Request $request)
```
TimWolla, line 55:
```suggestion
    private function getValidMethods(Request $request): array
```

TimWolla:
A look at access logs revealed that the HEAD verb is used for some legitimate use cases at least. Thus the middleware was updated to implicitly rewrite HEAD into GET, unless the controller explicitly opts into manual HEAD handling.

PR #5184, opened 2022-12-12
Merge `INTERNAL_HOSTNAMES` into list of allowed image domains

### LINKED ISSUE #5146: Implicitly merge “internal domains” into list of valid domains for external images

TimWolla:
Written for 5.5, but this is possibly a relevant behavioral change, so maybe
should be applied to 6.0 only? Feel free to adjust the target branch, it
applies cleanly to both 5.5 and 6.0.

-----------------

Fixes #5146


dtdesign (approved):

dtdesign:
It makes no sense that internal hostnames are not considered to be an allowed image source. This was overlooked when the internal hostname setting was introduced, so I consider this to be a reasonably serve bug that should be addressed in 5.5.

PR #5186, opened 2022-12-13
Fix validation in AbstractFileDeletePackageInstallationPlugin

TimWolla:
The query for logged files must not include the `packageID` in its condition,
as the entire purpose is retrieving all the `packageID`s for installed files to
compare them against the package that is currently installed or updated. Thus
by selecting only files for the current packageID the purpose is defeated.

see 672cd6166b684767a3fc8ee1fd6a4516d2061285


dtdesign:
I do not understand this change. The previous issue was that the unfiltered removal of files could cause issues if it was in the meantime added by another package. The PIP should only remove files that are currently owned by the package itself. Files that should be removed but are now owned by a different package, should be silently skipped.

TimWolla:
> The previous issue was that the unfiltered removal of files could cause issues if it was in the meantime added by another package.

Yes, and this issue still exists and this is what this PR fixes. I didn't get around to investigating this in detail, until now.

> Files that should be removed but are now owned by a different package, should be silently skipped.

Okay, will update.

dtdesign (approved):

PR #5199, opened 2023-01-02
fix creation of reason-field in `Confirmation.softDelete()`

mutec:
The integration of the optional reason-field when calling `softDelete(string, true)` is completely broken.
The current logic creates a `dl`-element containing the textarea, but only the textarea is appended to the dialog.
The second mistake is that the textarea (which should be the whole `dl`-element) get's appended to the upper dialog-element before everything else is even initialized, causing the textarea being hidden.

image
image


dtdesign (approved):

dtdesign:
Thanks!

PR #5200, opened 2023-01-03
fix handling of multi-selection fields as search parameters

mutec:
Using the following additional search parameter field:
```HTML
<select name="fooCategoryIDs" id="fooCategoryIDs" multiple>
    <option value="1">1</option>
    <option value="2">2</option>
    <option value="3">3</option>
</select>
```

Select 1, 2, 3.
Expectation: The request contains a comma-separated list of the values or an array containing the selected values.
Current behavior: The request contains only the last value (of the list => `3`), which means the submitted data is unreliable.

image
image
image

The change processes the form data before using it. In case one parameter-key appears more than once, the value is translated to an `Array<string>` instead of overriding the `string`-value every time the key appears:
image

### DIFF
--- a/ts/WoltLabSuite/Core/Ui/Search/Extended.ts
+++ b/ts/WoltLabSuite/Core/Ui/Search/Extended.ts
@@ -38,10 +38,6 @@ const enum SearchAction {
   Init,
 }
 
-type CombinedFormData = {
-  [key: string]: string | string[];
-}
-
 export class UiSearchExtended {
   private readonly form: HTMLFormElement;
   private readonly queryInput: HTMLInputElement;
@@ -147,29 +143,11 @@ export class UiSearchExtended {
 
     if (searchAction !== SearchAction.Navigation) {
       this.searchParameters = [];
-      const formData: CombinedFormData = {};
       new FormData(this.form).forEach((value, key) => {
         if (value.toString().trim()) {
-          if (formData[key] === undefined) {
-            formData[key] = value.toString().trim();
-          } else {
-            if (!Array.isArray(formData[key])) {
-              formData[key] = [formData[key] as string];
-            }
-
-            (formData[key] as Array<string>).push(value.toString().trim());
-          }
+          this.searchParameters.push([key, value.toString().trim()]);
         }
       });
-      for (const [key, value] of Object.entries(formData)) {
-        if (!Array.isArray(value)) {
-          this.searchParameters.push([key, value]);
-        } else {
-          value.forEach((itemValue) => {
-            this.searchParameters.push([key + "[]", itemValue]);
-          })
-        }
-      }
     }
     const parameters = this.searchParameters.slice();
 
@@ -189,14 +167,19 @@ export class UiSearchExtended {
     const data = {};
     new FormData(this.form).forEach((value, key) => {
       if (value.toString()) {
-        if (data[key] === undefined) {
+        const element = this.form.elements[key] as HTMLElement;
+        // represent values as array for select-fields with multiple-flag, multiple checkboxes
+        const isPlainValue = !(element instanceof HTMLSelectElement && element.multiple)
+          && !(element instanceof RadioNodeList && element[0] instanceof HTMLInputElement && element[0].type === "checkbox");
+
+        if (isPlainValue) {
           data[key] = value;
         } else {
-          if (!Array.isArray(data[key])) {
-            data[key] = [data[key]];
+          if (data[key] === undefined) {
+            data[key] = [];
           }
 
-          (data[key] as Array<any>).push(value);
+          (data[key] as any[]).push(value);
         }
       }
     });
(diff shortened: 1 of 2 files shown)


ts/WoltLabSuite/Core/Ui/Search/Extended.ts
```diff
@@ -143,11 +147,29 @@ export class UiSearchExtended {
...
         if (value.toString().trim()) {
-          this.searchParameters.push([key, value.toString().trim()]);
+          if (formData[key] === undefined) {
+            formData[key] = value.toString().trim();
+          } else {
+            if (!Array.isArray(formData[key])) {
+              formData[key] = [formData[key] as string];
+            }
+
+            (formData[key] as Array<string>).push(value.toString().trim());
```
TimWolla, line 160: Some general remark: `Array<T>` should be `T[]`.

ts/WoltLabSuite/Core/Ui/Search/Extended.ts
```diff
@@ -167,7 +189,15 @@ export class UiSearchExtended {
...
       if (value.toString()) {
-        data[key] = value;
+        if (data[key] === undefined) {
+          data[key] = value;
+        } else {
+          if (!Array.isArray(data[key])) {
+            data[key] = [data[key]];
+          }
+
+          (data[key] as Array<any>).push(value);
```
TimWolla, line 199: This likely does not do the correct thing:

If multiple items are selected, an array will be sent, but if only a single item is selected, a string will be sent. It should probably always be a (single-element) array if the element supports multiple selections.

In any case the amount of typechecks required here are not a good style, but this will likely solve itself with the previous paragraph.

TimWolla (commented):
I've had a first look at this:

- The change is incomplete, because values do not round-trip via the URL via `initQueryString()` (i.e. linking a search must do the right thing).
- Thus the change in `updateQueryString()` is not required, the query parameters do not need to have the trailing `[]` (and should not, because that makes the processing for the reloading much harder).

I defer to someone else for an in-depth and conceptual review, this is not my area of expertise. I recommend waiting for that review before making any further changes.

mutec:
@TimWolla I just figured out that the same issue exists (at the moment) for multiple selection when using checkboxes.

ts/WoltLabSuite/Core/Ui/Search/Extended.ts
mutec: The problem is to determine the exact case when it can/should be an array/multi-value property. The current logic (`FormData`) does not support getting the HTML-elements in order to determine the `multiple`-attribute, so we'd have to re-integrate the whole logic to force an array for those properties.

SoftCreatR:
FYI, I think, this change has to be ported to 5.5 as well:

```
Cannot assign string to property ticketsystem\system\search\TicketMessageSearch::$ticketCategoryIDs of type array


File:

/wsc/ticketsystem/lib/system/search/TicketMessageSearch.class.php in line 114


Stacktrace:

#0 /wsc/lib/system/search/SearchHandler.class.php(185): ticketsystem\system\search\TicketMessageSearch->getConditionBuilder(Array)
#1 /wsc/lib/system/search/SearchHandler.class.php(116): wcf\system\search\SearchHandler->buildTypeBasedConditionBuilders()
#2 /wsc/lib/system/search/SearchHandler.class.php(70): wcf\system\search\SearchHandler->buildConditions()
#3 /wsc/lib/data/search/SearchAction.class.php(105): wcf\system\search\SearchHandler->search()
#4 /wsc/lib/data/AbstractDatabaseObjectAction.class.php(216): wcf\data\search\SearchAction->search()
#5 /wsc/lib/action/AJAXProxyAction.class.php(92): wcf\data\AbstractDatabaseObjectAction->executeAction()
#6 /wsc/lib/action/AJAXInvokeAction.class.php(111): wcf\action\AJAXProxyAction->invoke()
#7 /wsc/lib/action/AbstractAction.class.php(57): wcf\action\AJAXInvokeAction->execute()
#8 /wsc/lib/action/AJAXInvokeAction.class.php(65): wcf\action\AbstractAction->__run()
#9 /wsc/lib/system/request/Request.class.php(90): wcf\action\AJAXInvokeAction->__run()
#10 /wsc/lib/system/request/RequestHandler.class.php(107): wcf\system\request\Request->execute()
#11 /wsc/index.php(11): wcf\system\request\RequestHandler->handle('wcf')
#12 {main}
```

dtdesign:
The search feature was subject of a major overhaul in WoltLab Suite 5.5 and one important change was the ability to store the entire search query in the URL. Unlike `POST` data, there are actual limits for `GET` parameters that are enforced in browsers, causing URLs to be cut off if they exceed a certain size. We have migrated away from multi select fields towards single selection fields while preserving the recursive search in child nodes.

This was also discussed in https://www.woltlab.com/community/thread/295658-mehrere-foren-durchsuchen/

mutec:
So we have to tell our customers the search stays broken, because you don't want to fix that bug?

dtdesign:
No. The API of the search feature has effectively changed in 5.5 and you need to update your implementation to work within the updated boundaries.

PR #5223, opened 2023-01-11
Add `allowCaching()` option to BackendRequest

TimWolla:
Not sure of this is placed correctly on the `BackendRequest`, it's not really
useful for `POST` requests and thus not providing it for
`prepareRequest().post()` might make sense. WDYT?

### DIFF
--- a/ts/WoltLabSuite/Core/Ajax/Backend.ts
+++ b/ts/WoltLabSuite/Core/Ajax/Backend.ts
@@ -33,8 +33,8 @@ class SetupRequest {
     this.url = url;
   }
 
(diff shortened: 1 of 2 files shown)


dtdesign (commented):
Overall LGTM, although I would strongly prefer to allow this only for `GET` requests, because all other request types are imho unsafe for caching by default.

TimWolla:
Okay, I've made a change. This was the simplest solution: The method is technically available, but not visible for POST requests, as enforced by tsc.

dtdesign (approved):

PR #5232, opened 2023-01-20
implement `PackageUpdateListChanged`

mutec:
This event indicated that there are some changes in the package update list. In case other packages need to grab information based on that list, they get a reliable event to update their data.


dtdesign (changes requested):
Please change the event to accept a single `PackageUpdateServer` instead and then fire one event per package server. From a performance standpoint this makes absolutely no difference and in return we gain strict typings instead of an arbitrary array parameter.

dtdesign:
Thanks!

PR #5233, opened 2023-01-21
assign `controllerName|controllerApplication` to tpl in `AbstractPage`

mutec:
This makes using ´form´-tags with actions to the same controller and using the new web component for pagination much easier and more generic:
`<woltlab-core-pagination page="1" count="10" url="{link application=$controllerAplication controller=$controllerName}{/link}"></woltlab-core-pagination>`

### DIFF
--- a/wcfsetup/install/files/lib/page/AbstractPage.class.php
+++ b/wcfsetup/install/files/lib/page/AbstractPage.class.php
@@ -199,11 +199,17 @@ abstract class AbstractPage implements IPage
         // call assignVariables event
         EventHandler::getInstance()->fireAction($this, 'assignVariables');
 
+        $classParts = \explode('\\', static::class);
+        $controllerName = \preg_replace('~(Form|Page)$~', '', \array_pop($classParts));
+        $controllerApplication = \array_shift($classParts);
+
         // assign parameters
         WCF::getTPL()->assign([
             'action' => $this->action,
             'templateName' => $this->templateName,
             'templateNameApplication' => $this->templateNameApplication,
+            'controllerName' => $controllerName,
+            'controllerApplication' => $controllerApplication,
             'canonicalURL' => $this->canonicalURL,
         ]);
     }


TimWolla:
ref: https://www.woltlab.com/community/thread/295153-abstractpage-aktuellen-controller-und-klassennamen-an-template-%C3%BCbergeben/

dtdesign:
I don’t like this for three reasons (in arbitrary order):

 1. We have already begun shifting the link generation to dedicated methods. Combined with `getControllerLink()` this approach allows it to conveniently discover all usages of a link. It also reduces the risks of having multiple links gradually become out of sync during changes.
 2. Manually supplying the name and application abbreviation feels cumbersome and error-prone. I think `{link}` really needs a way to support a class string instead, similar to how `getControllerLink()` works. But at this point we’re also close to (1.) above.
 3. We try to avoid introducing new _de facto_ global template variables unless we really have to, because those can easily collide with existing variables. I remember the days before we started sandboxing templates from boxes where we had all sorts of weird bugs caused by variables from boxes bleeding into the page’s variables. It was a true nightmare to debug those, because of the complexity of possible box configurations out there.

mutec:
Making `{link}` supporting the class strings would be great, but also wouldn't (and will never be able to) solve the problem, that everything in smarty-code does not get recognized by the IDE and won't show up as usage. Also it doesn't solve the problem, that the class string is not available by default, so one couldn't easily generate a pagination and re-use that template for several similar pages.

dtdesign:
In an ideal world, neither `{link}` nor 'LinkHandler::getInstance()->getLink()' would exist at all, because they are inherently bad. We cannot undo my past mistakes, but we can work on eradicating them going forward.

The problem with your suggestion is that the only argument here is that it’s convenient. I don’t even blame you, a nice developer experience is a good thing. However, it should not come at the expense of throwing more garbage on top of an already existing pile of bad design decisions.

PR #5237, opened 2023-01-23
Introduce `Helper::mapRequestParameters()` to reduce boiler plating

dtdesign:
The Valinor handling is pretty much the same for all PSR controllers and is pretty verbose when implemented correctly in a controller.

Using a helper method also reduces the risk of developers unintentionally misusing the API, for example, disallowing superfluous URL parameters. Simple patterns like cache busters would cause a strict parameter validation to fail, but this is not something a developer might take into consideration when implementing an API. It does make debugging a bit more painful, but in reality you only have to go through this once during development, whereas cache buster parameters are ubiquitous.

### DIFF
--- a/wcfsetup/install/files/lib/http/Helper.class.php
+++ b/wcfsetup/install/files/lib/http/Helper.class.php
@@ -27,7 +27,7 @@ final class Helper
     }
 
     /**
-     * Validates source parameters against the provided schema. Unknown
+     * Validates query parameters against the provided schema. Unknown
      * keys are skipped and values are gracefully converted into the
      * requested types.
      *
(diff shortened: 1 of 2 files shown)


wcfsetup/install/files/lib/http/Helper.class.php
```diff
@@ -23,6 +26,30 @@ public static function isAjaxRequest(ServerRequestInterface $request): bool
...
+     * keys are skipped and values are gracefully converted into the
+     * requested types.
+     *
+     * The returned array will contain only the values specified in the
+     * schema. Missing parameters or values that cannot be casted to the
+     * requested type will yield a `MappingError`.
+     *
+     * @throws MappingError
+     */
+    public static function mapRequestParameters(array $sourceParameters, string $schema): mixed
```
TimWolla, line 40: The implemented mapper is appropriate for query-string mapping and thus should be named as such. Within the request body we don't usually need to support superfluous keys and if JSON is used, flexible casting is not required either.

TimWolla (commented):
LGTM, except for the method name.

TimWolla (approved):

Issue #4472, opened 2021-08-23
Replace I18nHandler::remove() by a safer API


TimWolla:
- [ ] Take the package ID as a parameter.
- [ ] Possibly allow bulk deletions.

dtdesign:
The `I18nHandler` is a dead-end and we should focus on #6109 instead.

PR #5251, opened 2023-01-29
convert `CaptchaQuestionAddForm` to form builder api

mutec:


wcfsetup/install/files/lib/acp/form/CaptchaQuestionAddForm.class.php
```diff
@@ -3,198 +3,104 @@
...
-use wcf\system\WCF;
 use wcf\util\StringUtil;
 
 /**
  * Shows the form to create a new captcha question.
  *
- * @author  Matthias Schmidt
+ * @property    CaptchaQuestionAction   $objectAction
+ *
+ * @author  Matthias Schmidt, Florian Gail
```
TimWolla, line 21:
```suggestion
 * @author  Florian Gail, Matthias Schmidt
```
Alphabetic sort by last name.

wcfsetup/install/files/lib/acp/form/CaptchaQuestionAddForm.class.php
```diff
@@ -3,198 +3,104 @@
...
      */
     public $neededPermissions = ['admin.captcha.canManageCaptchaQuestion'];
 
-    /**
-     * @inheritDoc
-     */
-    public function assignVariables()
-    {
-        parent::assignVariables();
+    public $objectActionClass = CaptchaQuestionAction::class;
```
TimWolla, line 37: `@inheritDoc` for both of the new properties.

wcfsetup/install/files/lib/acp/form/CaptchaQuestionAddForm.class.php
```diff
@@ -3,198 +3,104 @@
...
+                        ->required()
+                        ->i18n()
+                        ->languageItemPattern('wcf.captcha.question.question.question\d+'),
+                    MultilineTextFormField::create('answers')
+                        ->label('wcf.acp.captcha.question.answers')
+                        ->required()
+                        ->i18n()
+                        ->languageItemPattern('wcf.captcha.question.answers.question\d+')
+                        ->addValidator(new FormFieldValidator('regex', function (MultilineTextFormField $formField) {
+                            if (!empty($formField->getValidationErrors())) {
```
TimWolla, line 63: `empty()`

wcfsetup/install/files/lib/acp/form/CaptchaQuestionAddForm.class.php
```diff
@@ -3,198 +3,104 @@
...
+                                return;
+                            }
+
+                            // returns `true` for a valid answer
+                            $answerValidator = function (string $text): bool {
+                                if (!\mb_substr($answer, 0, 1) == '~' || !\mb_substr($answer, -1, 1) == '~') {
```
TimWolla, line 69: `$answer` does not exist.

wcfsetup/install/files/lib/acp/form/CaptchaQuestionAddForm.class.php
```diff
@@ -3,198 +3,104 @@
...
+                        ->required()
+                        ->i18n()
+                        ->languageItemPattern('wcf.captcha.question.answers.question\d+')
+                        ->addValidator(new FormFieldValidator('regex', function (MultilineTextFormField $formField) {
+                            if (!empty($formField->getValidationErrors())) {
+                                return;
+                            }
+
+                            // returns `true` for a valid answer
+                            $answerValidator = function (string $text): bool {
```
TimWolla, line 68: Should be `static` if `$this` is not referenced.
```suggestion
                            $answerValidator = static function (string $text): bool {
```

wcfsetup/install/files/lib/acp/form/CaptchaQuestionEditForm.class.php
```diff
@@ -3,16 +3,14 @@
...
-use wcf\system\language\I18nHandler;
-use wcf\system\WCF;
 
 /**
  * Shows the form to edit an existing captcha question.
  *
- * @author  Matthias Schmidt
+ * @property    CaptchaQuestion $formObject
+ *
+ * @author  Matthias Schmidt, Florian Gail
```
TimWolla, line 13: ditto

wcfsetup/install/files/lib/acp/form/CaptchaQuestionEditForm.class.php
```diff
@@ -23,119 +21,21 @@ class CaptchaQuestionEditForm extends CaptchaQuestionAddForm
...
-            );
-            I18nHandler::getInstance()->setOptions(
-                'answers',
-                1,
-                $this->captchaQuestion->answers,
-                'wcf.captcha.question.question.answers\d+'
-            );
-
-            $this->isDisabled = $this->captchaQuestion->isDisabled;
+        if (\isset($_REQUEST['id']) && \is_numeric($_REQUEST['id'])) {
```
TimWolla, line 33: see other review regarding input validation.

wcfsetup/install/files/lib/data/TI18nDatabaseObjectAction.class.php
```diff
@@ -0,0 +1,92 @@
...
+namespace wcf\data;
+
+use wcf\system\language\I18nHandler;
+use wcf\util\StringUtil;
+
+/**
+ * Provides methods for automatic handling of i18n values.
+ *
+ * @author  Florian Gail
+ * @copyright   2001-2019 WoltLab GmbH
```
TimWolla, line 12:
```suggestion
 * @copyright   2001-2023 WoltLab GmbH
```

wcfsetup/install/files/lib/data/TI18nDatabaseObjectAction.class.php
```diff
@@ -0,0 +1,92 @@
...
+
+    /**
+     * Automatically saves the i18n data for `$propertyName` and updates the property value using the given pattern
+     * and object.
+     *
+     * @param string                                $propertyName
+     * @param string                                $languageItemPattern
+     * @param string                                $languageItemCategory
+     * @param integer                               $packageID
+     * @param DatabaseObject|DatabaseObjectEditor   $object
```
TimWolla, line 60: These should be dropped, they are redundant.

wcfsetup/install/files/lib/data/TI18nDatabaseObjectAction.class.php
```diff
@@ -0,0 +1,92 @@
...
+trait TI18nDatabaseObjectAction
+{
+    /**
+     * Automatically pre-fills `$propertyName` using the given pattern or a random string if none is given.
+     * This function is meant to be used for NOT NULL-columns without default value, which would cause an error
+     * when no value is given during the object's creation.
+     *
+     * @param string                $propertyName
+     * @param string                $languageItemPattern
+     * @param DatabaseObject|null   $object
```
TimWolla, line 24: ditto

wcfsetup/install/files/lib/acp/form/CaptchaQuestionEditForm.class.php
TimWolla, line 33:
```suggestion
        if (isset($_REQUEST['id']) && \is_numeric($_REQUEST['id'])) {
```

`isset()` is not a function.

TimWolla (changes requested):
Some remarks to the form implementation. I'll need to have a think about the adjustments to the DboAction (that's not fully reviewed yet). Please also take my review comments for the other PRs into account as far as they apply.

wcfsetup/install/files/lib/acp/form/CaptchaQuestionAddForm.class.php
mutec: Why? It is meant to validate in case the default validation has passed, which means it should do a early return in case there are validation errors. 🤔

TimWolla: Sorry, I was unclear here. My comment was meant to say “no empty please”, implicitly referencing the other review:

https://github.com/WoltLab/WCF/pull/5252#discussion_r1090336095

wcfsetup/install/files/lib/acp/form/CaptchaQuestionAddForm.class.php
```diff
@@ -46,155 +37,76 @@ class CaptchaQuestionAddForm extends AbstractForm
...
+            FormContainer::create('data')
+                ->appendChildren([
+                    TextFormField::create('question')
+                        ->label('wcf.acp.captcha.question.question')
+                        ->maximumLength(255)
+                        ->required()
+                        ->i18n()
+                        ->languageItemPattern('wcf.captcha.question.question.question\d+'),
+                    MultilineTextFormField::create('answers')
+                        ->label('wcf.acp.captcha.question.answers')
```
TimWolla, line 64: The description is no longer there.

wcfsetup/install/files/lib/acp/form/CaptchaQuestionAddForm.class.php
```diff
@@ -46,155 +37,76 @@ class CaptchaQuestionAddForm extends AbstractForm
...
+                        ->required()
+                        ->i18n()
+                        ->languageItemPattern('wcf.captcha.question.answers.question\d+')
+                        ->addValidator(new FormFieldValidator('regex', function (MultilineTextFormField $formField) {
+                            if ($formField->getValidationErrors() !== []) {
+                                return;
+                            }
+
+                            // returns `true` for a valid answer
+                            $answerValidator = static function (string $text): bool {
```
TimWolla, line 74: I was not a fan of the closure even in my initial review and now I'm even less of a fan. Please split this into a `private function` with an appropriate name. May I suggest: `private function validateAnswer(string $answer): bool`?

wcfsetup/install/files/lib/acp/form/CaptchaQuestionAddForm.class.php
```diff
@@ -46,155 +37,76 @@ class CaptchaQuestionAddForm extends AbstractForm
...
+                                $answers = [];
+                                foreach ($formField->getValue() as $value) {
+                                    $answers = \array_merge($answers, \explode("\n", StringUtil::unifyNewlines($value)));
+                                }
+                            } else {
+                                $answers = \explode("\n", StringUtil::unifyNewlines($formField->getValue()));
+                            }
+
+                            foreach ($answers as $answer) {
+                                if (!$answerValidator($answer)) {
```
TimWolla, line 94:
```suggestion
                                if (!$this->validateAnswer($answer)) {
```
see previous comment.

wcfsetup/install/files/lib/acp/form/CaptchaQuestionEditForm.class.php
```diff
@@ -23,119 +21,21 @@ class CaptchaQuestionEditForm extends CaptchaQuestionAddForm
...
-            'action' => 'edit',
-            'captchaQuestion' => $this->captchaQuestion,
-        ]);
-    }
-
-    /**
-     * @inheritDoc
-     */
-    public function readData()
+    public function readFormParameters()
```
TimWolla, line 29: This is the wrong method:
```suggestion
    public function readParameters()
```
(also update the parent call)

wcfsetup/install/files/lib/data/captcha/question/CaptchaQuestionAction.class.php
```diff
@@ -5,21 +5,22 @@
 use wcf\data\AbstractDatabaseObjectAction;
 use wcf\data\IToggleAction;
 use wcf\data\TDatabaseObjectToggle;
+use wcf\data\TI18nDatabaseObjectAction;
 
 /**
  * Executes captcha question-related actions.
  *
- * @author  Matthias Schmidt
+ * @author  Matthias Schmidt, Florian Gail
```
TimWolla, line 13: alphabetic order

wcfsetup/install/files/lib/data/TI18nDatabaseObjectAction.class.php
```diff
@@ -0,0 +1,82 @@
...
+use wcf\system\language\I18nHandler;
+use wcf\util\StringUtil;
+
+/**
+ * Provides methods for automatic handling of i18n values.
+ *
+ * @author  Florian Gail
+ * @copyright   2001-2019 WoltLab GmbH
+ * @license GNU Lesser General Public License <http://opensource.org/licenses/lgpl-license.php>
+ */
```
TimWolla, line 14: `@since`

TimWolla (changes requested):
Did a second pass with a functional test. I still didn't yet look into the updated DBOAction.

TimWolla:
I've now had a look at this once more: The form itself looks good to me, however the changes to the DBOAction are problematic:

The DBOAction implicitly relies on the values within the I18nHandler, thus it no longer is programmatically usable in a reusable way and this reliance on implicit global state violates the expectations of the user of the API.

I'm aware that the whole i18n handling API is “not great” (see also: #4472) – to put it mildly – but the change as proposed would make it much worse.

Instead the controller (i.e. the add / edit form) should perform the i18n saving, possibly adding a custom data processor to fill in some default values, like the old form did. The i18n values should then likely be saved within the `saved()` method.

Apart from that, I'd like to note the following issues with the current implementation:

- Localized answers are not stored at all.
- If the question is localized during creation, an exception is thrown: "SQLSTATE[HY000]: General error: 1364 Field 'question' doesn't have a default value".
- Phrases are not removed if the field changes from localized to non-localized.

/cc @dtdesign in case he has to add something with regard to i18n here.

dtdesign:
My only concern is that this PR adds code specific to the handling of i18n fields, but places it inside a trait designed to be used with DBOAction. I personally consider DBOAction to be borderline abandonware, because it is flawed in so many ways and we’re already looking of slowly migrating towards separate PSR-15 controllers and commands.

That said, I’m not really sure where we are heading with i18n fields at all. Offloading it to the language system was an easy solution without much of an afterthought, but this decision bit us more than once. I do believe that having separate data structures for these values is the better solution, similar to how we practice this for newer i18m components such as articles or pages.

At the same time, I do not see us migrating away from i18n fields in the foreseeable future. I do appreciate the efforts to reduce code duplication by introducing helper methods for phrase handling, but wouldn’t it be possible to move this into dedicated helper methods? And while we’re at it, the helper methods should be designed in a way that the developer can’t “hold it wrong”: Not removing phrases when switching from i18n to plain is a common error that such an API should automagically handle.

PR #5252, opened 2023-01-29
convert `Language*Form` to form builder

mutec:


wcfsetup/install/files/lib/acp/form/LanguageAddForm.class.php
```diff
@@ -103,149 +73,105 @@ public function readParameters()
...
-        if (empty($this->languageCode)) {
-            throw new UserInputException('languageCode');
-        }
-        if (LanguageFactory::getInstance()->getLanguageByCode($this->languageCode)) {
-            throw new UserInputException('languageCode', 'notUnique');
-        }
+        $locales = [
+            '0' => WCF::getLanguage()->get('wcf.global.noSelection'),
+        ];
+        $locales = \array_merge($locales, $this->locales);
```
TimWolla, line 85: Can you check whether the following works:

```
        $locales = [
            '' => WCF::getLanguage()->get('wcf.global.noSelection'),
            ...$this->locales,
        ];
```

I'd generally avoid reassigning variables, this makes types more clear for analysis tools.

wcfsetup/install/files/lib/acp/form/LanguageAddForm.class.php
```diff
@@ -103,149 +73,105 @@ public function readParameters()
...
+                    TextFormField::create('languageCode')
+                        ->label('wcf.acp.language.code')
+                        ->description('wcf.acp.language.code.description')
+                        ->maximumLength(20)
+                        ->required()
+                        ->addValidator(new FormFieldValidator('unique', function (TextFormField $formField) {
+                            if (!empty($formField->getValidationErrors())) {
+                                return;
+                            }
+                            if (
```
TimWolla, line 104:
```suggestion
                            }

                            if (
```

Consistent spacing.

wcfsetup/install/files/lib/acp/form/LanguageAddForm.class.php
```diff
@@ -103,149 +73,105 @@ public function readParameters()
...
+                        ->description('wcf.acp.language.name.description')
+                        ->maximumLength(255)
+                        ->required(),
+                    TextFormField::create('languageCode')
+                        ->label('wcf.acp.language.code')
+                        ->description('wcf.acp.language.code.description')
+                        ->maximumLength(20)
+                        ->required()
+                        ->addValidator(new FormFieldValidator('unique', function (TextFormField $formField) {
+                            if (!empty($formField->getValidationErrors())) {
```
TimWolla, line 101:
```suggestion
                            if ($formField->getValidationErrors() !== []) {
```

No `empty()`, please.

wcfsetup/install/files/lib/acp/form/LanguageAddForm.class.php
```diff
@@ -103,149 +73,105 @@ public function readParameters()
...
+                                return;
+                            }
+                            if (
+                                $this->formObject instanceof Language
+                                && \mb_strtolower($this->formObject->languageCode) === $formField->getValue()
+                            ) {
+                                return;
+                            }
+
+                            if (LanguageFactory::getInstance()->getLanguageByCode($this->languageCode)) {
```
TimWolla, line 111: `languageCode` is an undefined property.
```suggestion
                            if (LanguageFactory::getInstance()->getLanguageByCode($formField->getValue())) {
```
?

wcfsetup/install/files/lib/acp/form/LanguageAddForm.class.php
```diff
@@ -103,149 +73,105 @@ public function readParameters()
...
-
-    /**
-     * @inheritDoc
-     */
-    public function readData()
-    {
-        parent::readData();
+            // copy content
+            /** @var SingleSelectionFormField $sourceIDField */
+            $sourceIDField = $this->form->getNodeById('sourceLanguageID');
```
TimWolla, line 168:
```suggestion
            $sourceIDField = $this->form->getNodeById('sourceLanguageID');
            \assert($sourceIDField instanceof SingleSelectionFormField);
```

`assert()` achieves the same, but is also checked by PHP itself. The same for `$language`, please.

wcfsetup/install/files/lib/acp/form/LanguageEditForm.class.php
```diff
@@ -31,79 +25,12 @@ public function readParameters()
...
-    }
-
-    /**
-     * @inheritDoc
-     */
-    protected function validateLanguageCode()
-    {
-        if ($this->language->languageCode != \mb_strtolower($this->languageCode)) {
-            parent::validateLanguageCode();
+        if (isset($_REQUEST['id']) && \is_numeric($_REQUEST['id'])) {
```
TimWolla, line 28: Not technically new code, but we don't use `$_REQUEST` in new code. As the `id` should always be part of the query string, not the body, simply using `$_GET` should be sufficient. In this case please perform the validation like this:

```
        $parameters = Helper::mapQueryParameters(
            $_GET,
            <<<'EOT'
                array {
                    id?: positive-int
                }
                EOT
        );

        // $parameters['id'] is now either a positive int or does not exist, no need for "numeric" checks or manual casting.
```

TimWolla (changes requested):
Thank you. Some remarks with regard to style, but this looks good to me in general.

TimWolla:
Looking at #5251: `objectActionClass` and `objectEditLinkController` are not defined. Did you test this?

mutec:
> Looking at #5251: `objectActionClass` and `objectEditLinkController` are not defined. Did you test this?

I changed something after testing, maybe that part got lost. 😳

wcfsetup/install/files/lib/acp/form/LanguageAddForm.class.php
mutec: Why isn't that one catched by the codestyle-action? O.o

wcfsetup/install/files/lib/acp/form/LanguageEditForm.class.php
mutec: Do you want the `Helper` to read the ID? That seems a little bit overkill for preventing type-errors.

TimWolla: Yes, please use Valinor using `Helper::mapQueryParameters`. Valinor is the structual validator of choice beginning with WoltLab Suite 6.0. Consistency in validation is important, performance is not super critical in the ACP (and neither for the vast majority of frontend actions) and it's certainly much cleaner to use that than to repeat `$_REQUEST['id']` a total of 3 times.

wcfsetup/install/files/lib/acp/form/LanguageEditForm.class.php
```diff
@@ -3,26 +3,20 @@
...
  * @license GNU Lesser General Public License <http://opensource.org/licenses/lgpl-license.php>
  */
 class LanguageEditForm extends LanguageAddForm
 {
-    /**
-     * language id
-     * @var int
-     */
-    public $languageID = 0;
+    public $formAction = 'edit';
```
TimWolla, line 19: `@inheritDoc`

wcfsetup/install/files/lib/acp/form/LanguageAddForm.class.php
```diff
@@ -103,149 +83,107 @@ public function readParameters()
...
+                    TextFormField::create('countryCode')
+                        ->label('wcf.acp.language.countryCode')
+                        ->description('wcf.acp.language.countryCode.description')
+                        ->maximumLength(10)
+                        ->required(),
+                    SingleSelectionFormField::create('locale')
+                        ->label('wcf.acp.language.locale')
+                        ->description('wcf.acp.language.locale.description')
+                        ->options($locales)
+                        ->required(),
```
TimWolla, line 138: I wonder if we should validate the `locale` against the `languageCode` to prevent dumb combinations like “de” with “en_US” locale.

@dtdesign What do you think?

wcfsetup/install/files/lib/acp/form/LanguageAddForm.class.php
```diff
@@ -103,149 +83,107 @@ public function readParameters()
...
-
-        // source language id
-        $this->validateSource();
+        parent::createForm();
 
-        if (!isset($this->locales[$this->locale])) {
-            throw new UserInputException('locale');
-        }
-    }
+        $this->languages = LanguageFactory::getInstance()->getLanguages();
```
TimWolla, line 90: As this is only used within this method, this should be a local variable.

wcfsetup/install/files/lib/acp/form/LanguageAddForm.class.php
```diff
@@ -3,82 +3,62 @@
...
-     * @var Language
-     */
-    public $sourceLanguage;
-
-    /**
-     * source language id
-     * @var int
-     */
-    public $sourceLanguageID = 0;
+    public array $languages = [];
```
TimWolla, line 55: This can be dropped entirely then.

wcfsetup/install/files/lib/acp/form/LanguageAddForm.class.php
TimWolla, line 138: The `required()` should not be there: Not selecting anything is fine.

```suggestion
```

wcfsetup/install/files/lib/acp/form/LanguageAddForm.class.php
```diff
@@ -103,149 +83,107 @@ public function readParameters()
...
-        parent::save();
-
-        $this->language = LanguageEditor::create(\array_merge($this->additionalFields, [
-            'countryCode' => \mb_strtolower($this->countryCode),
-            'languageName' => $this->languageName,
-            'languageCode' => \mb_strtolower($this->languageCode),
-            'locale' => $this->locale,
-        ]));
-        $languageEditor = new LanguageEditor($this->sourceLanguage);
-        $languageEditor->copy($this->language);
```
TimWolla, line 198: You dropped this: Any newly created language will not have any phrases.

TimWolla (changes requested):
Now also did a functional test.

wcfsetup/install/files/lib/acp/form/LanguageAddForm.class.php
dtdesign: Yes, the `languageCode` and the `locale` must match, because the locale is just a more specific version of it.

mutec: It was required before, are you sure?
However this case would benefit from https://github.com/WoltLab/WCF/pull/5159/commits/e821b790d581d029384e07d4ab4c36c11febba3e

mutec: In that case we should also validate `languageCode` against ISO 639-1 and `countryCode` against ISO 3166-1. To be consequent, we could remove both fields (if a locale is selected) since they are defined by the locale.

dtdesign: > To be consequent, we could remove both fields (if a locale is selected) since they are defined by the locale.

No. `languageCode` is still referenced a lot of times in both our and 3rd party code, including the language XML files. We could derive the language code from the locale though, but this is a journey for another day.

The `countryCode` on the other hand is used to pick which flag to show. It’s a bit awkward for English since we use American English, but use UK’s flag. Maybe one day we can get rid of the country flags, because it just makes things quite difficult without adding much benefit.

wcfsetup/install/files/lib/acp/form/LanguageEditForm.class.php
```diff
@@ -31,79 +29,18 @@ public function readParameters()
...
-        if (isset($_REQUEST['id'])) {
-            $this->languageID = \intval($_REQUEST['id']);
-        }
-        $this->language = new Language($this->languageID);
-        if (!$this->language->languageID) {
+        $queryParameters = Helper::mapQueryParameters(
+            $_GET,
+            <<<'EOT'
+                array {
+                    id?: positive-int
```
TimWolla, line 36: I likely wasn't sufficiently clear: `id?` implies that the `id` parameter may be absent. You will still need to check `isset($queryParameters['id'])` or use `$queryParameters['id'] ?? 0` when constructing the DBO.

The reason I suggested making the `id` optional here and manually handling that case is that the Valinor mapping errors are not particularly pretty:

```
{
    "errors": [
        "Value *missing* is not a valid positive integer."
    ]
}
```

@dtdesign Which error handling would you prefer here? Making the `id` required within the array-shape and thus generating the unpretty Valinor error if the `id` is absent – or making the `id` optional, checking it manually and throwing `IllegalLinkException`?

wcfsetup/install/files/lib/acp/form/LanguageAddForm.class.php
TimWolla: Yes, I'm sure about this: The `required()` implies that the user has to make a selection, when actually “No Selection” is fine. Completely deleting the `<select>` from the HTML will result in a `null` return even if the field is `required()`, so this is definitely is a bug there. It appears that support for `required()` is not implemented at all. I'll file an issue for this.

I also agree that this would likely benefit from the commit you linked, however that should be reviewed and included separately. I have some concerns about the added API surface that we'd need to discuss, instead of making that a separate flag it should likely rely on `required()` to fix the behavior of `required()`. 6.0 is the right version to make this fix, as it can be slightly breaking. Would you be interested in making a PR for this?

TimWolla:
If the three open points:

- Validate locale against languageCode
- Remove `->required()` for locale.
- Fix the `id` validation.

are resolved and no other changes are made, then this can be merged. The other open-ended stuff (SingleSelectionField + requirement handling) can and should be resolved in a follow-up PR.

wcfsetup/install/files/lib/acp/form/LanguageEditForm.class.php
dtdesign: The `id` should not be optional, because it is a required part of the route. Making it optional for the sake of avoiding a less helpful error message generated by Valinor is the exact opposite of what I would do.

If you believe that the case of a missing parameter should yield a better error message, then I suggest raising an issue with Valinor to improve this instead.

TimWolla: > If you believe that the case of a missing parameter should yield a better error message, then I suggest raising an issue with Valinor to improve this instead.

@dtdesign 

To clarify, the error handling is our own code and I didn't yet check all the capabilities. It primarily exists to have *something*.

https://github.com/WoltLab/WCF/blob/a87fd5826477d6b3579abfb1fd6e01d15ca44be0/wcfsetup/install/files/lib/http/middleware/HandleValinorMappingErrors.class.php#L34-L44

In this case I was concerned about the output, because this is a non-API route, thus the URL is immediately visible in the URL bar and might accidentally be truncated by a user, thus resulting in the JSON output + HTTP 400 instead of the regular `IllegalLinkException` + HTTP 404.

We might be able to generate some well-formatted HTML output within that middleware by switching the output type based on the client's `Accept` header, but the status code cannot be handled generically.

dtdesign: I can understand your standpoint, so maybe Valinor is not the right choice there? At the end of the day, marking the parameter as optional here (when it actually is not!) is merely a bogus workaround. Consider this, you’re not marking it as optional because _it is_ optional, but because you want a nicer error handling.

TimWolla: @mutec We've discussed this in chat. Please use the following pattern:

```
        try {
            $queryParameters = Helper::mapQueryParameters(
                $_GET,
                <<<'EOT'
                    array {
                        id: positive-int
                    }
                    EOT
            );
            $this->formObject = new Language($queryParameters['id']);

            if (!$this->formObject->getObjectID()) {
                throw new IllegalLinkException();
            }
        } catch (MappingError) {
            throw new IllegalLinkException();
        }
```

It's a little on the verbose side due to the additional try-catch, but it avoids all the pitfalls and issues with parameter validation. If the try-catch is missing it fails “safely”: The error message won't be pretty, but the behavior will be correct.

mutec: Seeing that code makes me thinking about `DatabaseObject::construct()` throwing an exception in case the id is not `null` and no object is found. :D

wcfsetup/install/files/lib/acp/form/LanguageAddForm.class.php
mutec: What kind of validation would you expect? `^$languageCode_*` or `$languageCode === \explode('_', $locale)[0];`?
I don't see any proper way to do this without making the language and country code fields completely useless.

TimWolla: The `explode()` variant is fine. With the regular expression you possibly need to mess around with escaping.

> I don't see any proper way to do this without making the language and country code fields completely useless.

And yes, it's a little redundant for now, but this is something to incrementally improve once we have experience in the field. This form is mostly going to be used by power-users who should be able to deal with this.

The “locale” selection is based on PHP’s intl extension and thus ICU. While our non-scientific research within the Smalltalk thread in our forum shows that intl is generally available and the ICU version also tends to be reasonably up to date, we don't yet know if there’s any way to break those as a web hoster. By separating the `languageCode` from the `locale` one can provide the actual localization even if the language is not (yet) supported by one’s ICU version. This might be a non-issue in practice, but we’ll find that out. We generally tend to do this kind of incremental improvement nowadays, because it allows one to react on real-world feedback instead of making a bunch of assumptions. One, purely technical, example might be #5237.

wcfsetup/install/files/lib/acp/form/LanguageEditForm.class.php
dtdesign: In hindsight it would have been much better to implement a static method for DBOs that returns the constructed object or null if it was not found. The whole magic behind `$data` is just a mess and if we would design these today, we would simply use a `readonly class` that represents all fields as actual properties.

But this is a story for another day ;)

TimWolla:
@mutec Can you add the locale validation (https://github.com/WoltLab/WCF/pull/5252#discussion_r1094174253) to bring this across the finish line?

TimWolla:
@mutec ping

mutec:
@TimWolla pong
I'm trying to manage this on Saturday or Sunday; I'm pretty busy at the moment.

TimWolla:
Thank you. I've just wanted to make sure you've seen this, because there was no feedback on this one (not even an emoji reaction), but you commented on the package.json one.

mutec:
@TimWolla I just found a few minutes. I added the validation to the `languageCode`-field. The other issues should have been resolved already. Could you please have a look on the error-message?

TimWolla:
@mutec Thank you for taking the time. I'll have a look *right now* to not delay this further.

wcfsetup/install/files/lib/acp/form/LanguageAddForm.class.php
```diff
@@ -103,149 +75,123 @@ public function readParameters()
...
+                                            'localeMismatch',
+                                            'wcf.acp.language.add.languageCode.error.localeMismatch',
+                                            [
+                                                'locale' => $localeField->getValue(),
+                                            ]
+                                        )
+                                    );
+                                }
+                            }
+                        )),
```
TimWolla, line 145: I'm afraid I'm getting an exception here whenever I submit the form.

```
wcf\acp\form\LanguageAddForm::wcf\acp\form\{closure}(): Argument #1 ($formField) must be of type wcf\system\form\builder\field\SingleSelectionFormField, wcf\system\form\builder\field\TextFormField given, called in /var/www/html/lib/system/form/builder/field/validation/FormFieldValidator.class.php on line 71
```

The validator needs to be attached to the `locale` field, not the `languageCode` field.

wcfsetup/install/files/lib/system/language/event/CopyLanguageContent.php
TimWolla, line 1: Missing `.class` within the filename.

wcfsetup/install/files/lib/acp/form/LanguageAddForm.class.php
```diff
@@ -103,149 +75,123 @@ public function readParameters()
...
+                        }))
+                        ->addValidator(new FormFieldValidator(
+                            'locale',
+                            function (SingleSelectionFormField $formField) {
+                                if ($formField->getValidationErrors() !== []) {
+                                    return;
+                                }
+
+                                $localeField = $formField->getDocument()->getNodeById('locale');
+                                \assert($localeField instanceof SingleSelectionFormField);
```
TimWolla, line 127: And here it should be `languageCode` field then. I suspect you initially tested with the validator attached to the `languageCode` field and then incompletely moved it over, because it belong to the `locale` field?

This full implementation appears to work for me (but please double check):

```
                        ->addValidator(new FormFieldValidator(
                            'locale',
                            static function (SingleSelectionFormField $formField) {
                                if ($formField->getValidationErrors() !== []) {
                                    return;
                                }

                                if ($formField->getValue() === '') {
                                    return;
                                }

                                $languageCodeField = $formField->getDocument()->getNodeById('languageCode');
                                \assert($languageCodeField instanceof TextFormField);

                                $test = \explode('_', $formField->getValue());
                                if (\count($test) === 1 || $languageCodeField->getValue() !== $test[0]) {
                                    $formField->addValidationError(
                                        new FormFieldValidationError(
                                            'localeMismatch',
                                            'wcf.acp.language.add.languageCode.error.localeMismatch',
                                            [
                                                'locale' => $formField->getValue(),
                                            ]
                                        )
                                    );
                                }
                            }
                        )),
```

TimWolla (changes requested):

wcfsetup/install/files/lib/acp/form/LanguageAddForm.class.php
TimWolla: Oh, the error phrase name should contain `locale` instead of `languageCode`. That will need to be fixed in both the phrases and PHP.

wcfsetup/install/files/lib/acp/form/LanguageAddForm.class.php
mutec: Ah sorry, I moved the validator and didn't fix that up.

mutec: Since the user generated content is the language code, the language code should be shown as faulty, not the system given locale the user wouldn't be able to change.

TimWolla: It's a mismatch between both fields, so either would be an acceptable error target. However the select field comes later, so the issue is that the selection does not match the text input and the error should be attached to the select. Especially since the locale (region) selection is optional.

Please have a look at the proposed validator, I've tested it and it appears to match the desired behavior.

wcfsetup/install/files/lib/data/language/LanguageEditor.class.php
```diff
@@ -1054,5 +1056,10 @@ public static function copyLanguageContent($sourceLanguageID, $destinationLangua
         }
 
         PageEditor::resetCache();
+
+        EventHandler::getInstance()->fire(new CopyLanguageContent(
+            LanguageFactory::getInstance()->getLanguage($sourceLanguageID),
+            LanguageFactory::getInstance()->getLanguage($destinationLanguageID)
+        ));
```
TimWolla, line 1063:
```suggestion
        EventHandler::getInstance()->fire(new CopyLanguageContent(
            LanguageFactory::getInstance()->getLanguage($sourceLanguageID),
            LanguageFactory::getInstance()->getLanguage($destinationLanguageID)
        ));
```

This appears to throw an exception in `LanguageAddForm`, likely due to some caching issue?

> wcf\system\language\event\CopyLanguageContent::__construct(): Argument #2 ($targetLanguage) must be of type wcf\data\language\Language, null given, called in /var/www/html/lib/data/language/LanguageEditor.class.php on line 1047

The simple solution would be using:

```suggestion
        EventHandler::getInstance()->fire(new CopyLanguageContent(
            new Language($sourceLanguageID),
            new Language($destinationLanguageID)
        ));
```

wcfsetup/install/files/lib/acp/form/LanguageAddForm.class.php
TimWolla: I can move the Validator in a follow-up commit to save you some work, as the current one is at least functional, but the Exception when adding a language is a blocker (see the other comment). We want to avoid broken commits as much as possible.

wcfsetup/install/files/lib/acp/form/LanguageAddForm.class.php
```diff
@@ -103,149 +75,123 @@ public function readParameters()
...
+                            static function (SingleSelectionFormField $formField) {
+                                if ($formField->getValidationErrors() !== []) {
+                                    return;
+                                }
+
+                                $languageCodeField = $formField->getDocument()->getNodeById('languageCode');
+                                \assert($languageCodeField instanceof TextFormField);
+                                if ($languageCodeField->getValue() === '') {
+                                    return;
+                                }
```
TimWolla, line 139:
```suggestion
                                if ($formField->getValue() === '') {
                                    return;
                                }

                                $languageCodeField = $formField->getDocument()->getNodeById('languageCode');
                                \assert($languageCodeField instanceof TextFormField);
```

You didn't use the suggestion as-is: Currently the empty selection is rejected, but it should not (as it's not required).

wcfsetup/install/files/lib/acp/form/LanguageAddForm.class.php
```diff
@@ -103,149 +75,123 @@ public function readParameters()
...
+
+                                $test = \explode('_', $formField->getValue());
+                                if (\count($test) === 1 || $languageCodeField->getValue() !== $test[0]) {
```
TimWolla, line 142:
```suggestion
                                if ($languageCodeField->getValue() !== $test[0]) {
```
Is there a reason you rejected the locales without an actual region code? I feel it should be legal to select “Italian” when create the `it` language, even if it is not all that useful to select just the locale without a region.

wcfsetup/install/files/lib/system/language/event/CopyLanguageContent.class.php
```diff
@@ -0,0 +1,23 @@
...
+
+/**
+ * Indicates that the contents of a language should be copied to another one.
+ *
+ * @author Florian Gail
+ * @copyright 2001-2023 WoltLab GmbH
+ * @license GNU Lesser General Public License <http://opensource.org/licenses/lgpl-license.php>
+ * @since 6.0
+ */
+final class CopyLanguageContent implements IEvent
```
TimWolla, line 16: I realize it just now, but the event name does not follow our (I believe not yet documented) naming scheme. It should be: `LanguageContentCopying`.

The pattern is:
- The prefix describes the “object” that is affected (LanguageContent).
- The suffix describes the action that is happening (“Copy”)
- The tense (as in “past tense”) describes whether the operation is in progress or already happened (“-ing” = in progress).

TimWolla (changes requested):
Thank you. I can do all three of the remarks in follow-up commits. Let me know if you would like me to make the changes.

TimWolla:
@mutec Please see: https://github.com/WoltLab/WCF/pull/5252#pullrequestreview-1343176663

mutec:
@TimWolla sorry, I don't want to waste my time any more. Feel free to finalize that.

TimWolla:
Thank you for confirming, I'll reopen this so it stays on my dashboard until I get around to include these follow-ups.

TimWolla:
Now applied, thank you. I've noted in the commit messages that I made changes after the fact.

PR #5253, opened 2023-01-29
convert `PackageUpdateServerAddForm` to form builder api

mutec:


wcfsetup/install/files/acp/templates/packageUpdateServerAdd.tpl
```diff
@@ -14,57 +14,10 @@
 	</nav>
 </header>
 
-{if $packageUpdateServer|isset && $packageUpdateServer->errorMessage}
-	<p class="warning">{lang}wcf.acp.updateServer.lastErrorMessage{/lang}<br>{$packageUpdateServer->errorMessage}</p>
+{if !$formObject|empty && $formObject->errorMessage}
```
TimWolla, line 17:
```suggestion
{if $formObject|isset && $formObject->errorMessage}
```

wcfsetup/install/files/lib/acp/form/PackageUpdateServerAddForm.class.php
```diff
@@ -3,22 +3,32 @@
...
+use wcf\system\form\builder\field\validation\FormFieldValidator;
+use wcf\system\form\builder\IFormDocument;
 
 /**
  * Shows the server add form.
  *
- * @author  Marcel Werk
+ * @property    PackageUpdateServerAction   $objectAction
+ *
+ * @author  Marcel Werk, Florian Gail
```
TimWolla, line 27: Alphabetic sort by last name.

wcfsetup/install/files/lib/acp/form/PackageUpdateServerAddForm.class.php
```diff
@@ -31,153 +41,167 @@ class PackageUpdateServerAddForm extends AbstractForm
...
-        if (isset($_POST['loginUsername'])) {
-            $this->loginUsername = $_POST['loginUsername'];
-        }
-        if (isset($_POST['loginPassword'])) {
-            $this->loginPassword = $_POST['loginPassword'];
+        parent::createForm();
+
+        $passwordPlaceholder = '';
+        if (
+            $this->formAction !== IFormDocument::FORM_MODE_CREATE && $this->formObject instanceof PackageUpdateServer
```
TimWolla, line 62:
```suggestion
            $this->formAction !== IFormDocument::FORM_MODE_CREATE
            && $this->formObject instanceof PackageUpdateServer
```

wcfsetup/install/files/lib/acp/form/PackageUpdateServerAddForm.class.php
```diff
@@ -31,153 +41,167 @@ class PackageUpdateServerAddForm extends AbstractForm
...
+            && !empty($this->formObject->loginPassword)
```
TimWolla, line 63:
```suggestion
            && $this->formObject->loginPassword !== ''
```
no empty.

wcfsetup/install/files/lib/acp/form/PackageUpdateServerAddForm.class.php
```diff
@@ -31,153 +41,167 @@ class PackageUpdateServerAddForm extends AbstractForm
...
+            FormContainer::create('data')
+                ->appendChildren([
+                    UrlFormField::create('serverURL')
+                        ->label('wcf.acp.updateServer.serverURL')
+                        ->pattern('https://.+')
+                        ->required()
+                        ->maximumLength(255)
+                        ->immutable($this->formAction !== IFormDocument::FORM_MODE_CREATE)
+                        ->addValidator(new FormFieldValidator('valid', function (UrlFormField $formField) {
+                            if (!empty($formField->getValidationErrors())) {
```
TimWolla, line 78: `!== []`

wcfsetup/install/lang/en.xml
```diff
@@ -2724,7 +2724,7 @@ If you have <strong>already bought the licenses for the listed apps</strong>, th
 		<item name="wcf.acp.updateServer.serverURL.error.nonStandardPort"><![CDATA[The address must not use a non-standard port.]]></item>
 		<item name="wcf.acp.updateServer.serverURL.error.userinfo"><![CDATA[The address may not contain credentials.]]></item>
 		<item name="wcf.acp.updateServer.serverURL.error.woltlab"><![CDATA[The official servers are already set up and correctly configured.]]></item>
-		<item name="wcf.acp.updateServer.serverURL.error.duplicate"><![CDATA[The entered server already exists. You can <a href="{link controller="PackageUpdateServerEdit" id=$errorType[duplicate]->packageUpdateServerID}{/link}"><strong>edit this server instead</strong></a>.]]></item>
+		<item name="wcf.acp.updateServer.serverURL.error.duplicate"><![CDATA[The entered server already exists. You can <a href="{link controller="PackageUpdateServerEdit" id=$duplicate->getObjectID()}{/link}"><strong>edit this server instead</strong></a>.]]></item>
```
TimWolla, line 2727:
```suggestion
		<item name="wcf.acp.updateServer.serverURL.error.duplicate"><![CDATA[The entered server already exists. You can <a href="{link controller="PackageUpdateServerEdit" object=$duplicate}{/link}"><strong>edit this server instead</strong></a>.]]></item>
```

wcfsetup/install/lang/de.xml
```diff
@@ -2792,7 +2792,7 @@ Abschnitte dürfen nicht leer sein und nur folgende Zeichen enthalten: <kbd>[a-z
 		<item name="wcf.acp.updateServer.serverURL.error.nonStandardPort"><![CDATA[Die eingetragene Adresse darf keinen abweichenden Port verwenden.]]></item>
 		<item name="wcf.acp.updateServer.serverURL.error.userinfo"><![CDATA[Die eingetragene Adresse darf keine Login-Informationen enthalten.]]></item>
 		<item name="wcf.acp.updateServer.serverURL.error.woltlab"><![CDATA[Die offiziellen Paketserver sind bereits hinterlegt und korrekt konfiguriert.]]></item>
-		<item name="wcf.acp.updateServer.serverURL.error.duplicate"><![CDATA[Dieser Paketserver ist bereits hinterlegt. {if LANGUAGE_USE_INFORMAL_VARIANT}Du kannst{else}Sie können{/if} <a href="{link controller="PackageUpdateServerEdit" id=$errorType[duplicate]->packageUpdateServerID}{/link}"><strong>den bestehenden Paketserver bearbeiten</strong></a>.]]></item>
+		<item name="wcf.acp.updateServer.serverURL.error.duplicate"><![CDATA[Dieser Paketserver ist bereits hinterlegt. {if LANGUAGE_USE_INFORMAL_VARIANT}Du kannst{else}Sie können{/if} <a href="{link controller="PackageUpdateServerEdit" id=$duplicate->getObjectID()}{/link}"><strong>den bestehenden Paketserver bearbeiten</strong></a>.]]></item>
```
TimWolla, line 2795:
```suggestion
		<item name="wcf.acp.updateServer.serverURL.error.duplicate"><![CDATA[Dieser Paketserver ist bereits hinterlegt. {if LANGUAGE_USE_INFORMAL_VARIANT}Du kannst{else}Sie können{/if} <a href="{link controller="PackageUpdateServerEdit" object=$duplicate}{/link}"><strong>den bestehenden Paketserver bearbeiten</strong></a>.]]></item>
```

wcfsetup/install/files/lib/acp/form/PackageUpdateServerEditForm.class.php
```diff
@@ -41,75 +35,12 @@ public function readParameters()
     {
         parent::readParameters();
 
-        if (isset($_REQUEST['id'])) {
-            $this->packageUpdateServerID = \intval($_REQUEST['id']);
+        if (isset($_REQUEST['id']) && \is_numeric($_REQUEST['id'])) {
```
TimWolla, line 38: see other review for input handling.

wcfsetup/install/files/lib/acp/form/PackageUpdateServerAddForm.class.php
```diff
@@ -31,153 +41,167 @@ class PackageUpdateServerAddForm extends AbstractForm
...
+    protected function getDuplicateServer(string $serverUrl): ?PackageUpdateServer
     {
         $packageServerList = new PackageUpdateServerList();
         $packageServerList->readObjects();
         foreach ($packageServerList as $packageServer) {
-            if ($packageServer->serverURL == $this->serverURL) {
+            if ($packageServer->serverURL == $serverUrl) {
                 return $packageServer;
             }
         }
```
TimWolla, line 205: This requires an explicit `return null;`, because of the explicit return type.

wcfsetup/install/files/lib/acp/form/PackageUpdateServerAddForm.class.php
```diff
@@ -31,153 +41,167 @@ class PackageUpdateServerAddForm extends AbstractForm
...
+                                    'invalid',
+                                    'wcf.acp.updateServer.serverURL.error.invalid',
+                                ));
+                            }
+
+                            if (
+                                ($duplicate = $this->getDuplicateServer((string)$url))
+                                && ($this->formAction === IFormDocument::FORM_MODE_CREATE
+                                || ($this->formObject instanceof PackageUpdateServer
+                                && $this->formObject->getObjectID() !== $duplicate->getObjectID()))
```
TimWolla, line 125: The indentation is super wonky here. One additional indentation level per opening parentheses, please. I believe the following is correct:
```suggestion
                                ($duplicate = $this->getDuplicateServer((string)$url))
                                && (
                                    $this->formAction === IFormDocument::FORM_MODE_CREATE
                                    || (
                                        $this->formObject instanceof PackageUpdateServer
                                        && $this->formObject->getObjectID() !== $duplicate->getObjectID()
                                    )
                                )
```

wcfsetup/install/files/lib/acp/form/PackageUpdateServerAddForm.class.php
```diff
@@ -31,153 +41,167 @@ class PackageUpdateServerAddForm extends AbstractForm
...
-                if ($url->getScheme() !== 'https') {
-                    throw new UserInputException('serverURL', 'invalidScheme');
-                }
-                if ($url->getPort()) {
-                    throw new UserInputException('serverURL', 'nonStandardPort');
+                if ($username === '') {
+                    $parameters['data']['loginUsername'] = '';
+                    $parameters['data']['loginPassword'] = '';
+                } elseif (
+                    $username === $this->formObject->loginUsername && empty($parameters['data']['loginPassword'])
```
TimWolla, line 175:
```suggestion
                    $username === $this->formObject->loginUsername
                    && $parameters['data']['loginPassword'] === ''
```
If you already please the condition into multiple lines, then also split at the binary boolean operators. Also: `empty`.

wcfsetup/install/files/lib/acp/form/PackageUpdateServerAddForm.class.php
```diff
@@ -31,153 +41,167 @@ class PackageUpdateServerAddForm extends AbstractForm
...
+            'password',
+            function (IFormDocument $document, array $parameters) {
+                if ($this->formAction === IFormDocument::FORM_MODE_CREATE) {
+                    return $parameters;
+                }
 
-        try {
-            $url = new Uri($this->serverURL);
-            $this->serverURL = (string)$url;
+                $username = $parameters['data']['loginUsername'] ?? null;
```
TimWolla, line 168: Is `null` the correct fallback value?

TimWolla (changes requested):

wcfsetup/install/files/lib/acp/form/PackageUpdateServerAddForm.class.php
mutec: I thought about adding a dependency to the username-field, so there is no need for a fallback since the index should be set in every case.

wcfsetup/install/files/lib/acp/form/PackageUpdateServerAddForm.class.php
```diff
@@ -31,153 +41,176 @@ class PackageUpdateServerAddForm extends AbstractForm
...
-        }
-        if (isset($_POST['loginPassword'])) {
-            $this->loginPassword = $_POST['loginPassword'];
+        parent::createForm();
+
+        $passwordPlaceholder = '';
+        if (
+            $this->formAction !== IFormDocument::FORM_MODE_CREATE
+            && $this->formObject instanceof PackageUpdateServer
+            && $this->formObject->loginPassword !== ''
```
TimWolla, line 64:
```suggestion
            && $this->formObject->loginUsername !== ''
```
Previously the username was checked and this should remain like this to not leak anything about the password.

wcfsetup/install/files/lib/acp/form/PackageUpdateServerEditForm.class.php
```diff
@@ -26,13 +26,7 @@ class PackageUpdateServerEditForm extends PackageUpdateServerAddForm
      * update server id
      * @var int
      */
```
TimWolla, line 28: This comment should be updated to `@inheritDoc`.

wcfsetup/install/files/lib/acp/form/PackageUpdateServerAddForm.class.php
```diff
@@ -31,153 +41,176 @@ class PackageUpdateServerAddForm extends AbstractForm
...
+                        ->addValidator(new FormFieldValidator('valid', function (UrlFormField $formField) {
+                            if ($formField->getValidationErrors() !== []) {
+                                return;
+                            }
+
+                            try {
+                                $url = new Uri($formField->getValue());
+
+                                if (!$url->getHost()) {
+                                    throw new UserInputException('serverURL', 'invalid');
```
TimWolla, line 87: Should not throw.

wcfsetup/install/files/lib/acp/form/PackageUpdateServerAddForm.class.php
```diff
@@ -31,153 +41,176 @@ class PackageUpdateServerAddForm extends AbstractForm
...
+                                    $this->formAction === IFormDocument::FORM_MODE_CREATE
+                                    || (
+                                        $this->formObject instanceof PackageUpdateServer
+                                        && $this->formObject->getObjectID() !== $duplicate->getObjectID()
+                                    )
+                                )
+                            ) {
+                                $formField->addValidationError(new FormFieldValidationError(
+                                    'duplicate',
+                                    'wcf.acp.updateServer.serverURL.error.invalid',
```
TimWolla, line 134:
```suggestion
                                    'wcf.acp.updateServer.serverURL.error.duplicate',
```

wcfsetup/install/files/lib/acp/form/PackageUpdateServerAddForm.class.php
```diff
@@ -31,153 +41,176 @@ class PackageUpdateServerAddForm extends AbstractForm
...
+        ) {
+            $passwordPlaceholder = 'wcf.acp.updateServer.loginPassword.noChange';
         }
+
+        $this->form->appendChildren([
+            FormContainer::create('data')
+                ->appendChildren([
+                    UrlFormField::create('serverURL')
+                        ->label('wcf.acp.updateServer.serverURL')
+                        ->pattern('https://.+')
```
TimWolla, line 74:
```suggestion
```
Please remove the pattern. The error handling by the browser is “non-great”, making it totally non-obvious what the issue is (the lack of 's').

TimWolla (changes requested):
I made a functional test now. Apart from the line annotations, actually editing a package server does not work for me, because the “address” field will be cleared after submitting the form. This might be related to the `->immutable()` and perhaps a bug in form builder itself (if so, it should be fixed in a separate PR).

TimWolla:
> This might be related to the ->immutable() and perhaps a bug in form builder itself (if so, it should be fixed in a separate PR).

Ah, yes, reported here: https://www.woltlab.com/community/thread/294841-formbuilder-fehlverhalten-bei-nutzung-von-immutable/

wcfsetup/install/files/lib/acp/form/PackageUpdateServerAddForm.class.php
mutec: The processor `password` ensures that username and password will be set to an empty string in case no username is given while it removes the column's value when reading the object for editing it.
There should be no case a password is stored without a username. Also `password` shouldn't be empty in case a username is stored.

TimWolla (commented):
I realize that the whole “password on edit” logic is broken, because the password field is `required()`.

I wonder though, if this logic is actually necessary at all: The form only consists of URL, username and password. The URL is immutable when editing, thus the only fields that can be edited are the credentials. Thus the only case where the form will be submitted is when the credentials are intended to be changed.

PR #5254, opened 2023-01-29
implement `INullableFormField` to `TextFormField`

mutec:


TimWolla:
I don't see how this does anything if the nullability is never checked.

mutec:
> if the nullability is never checked.

That should be done by the i18n-trait, that is used:
https://github.com/WoltLab/WCF/blob/522eba74dab4475c04ec3735486dcc02831132a0/wcfsetup/install/files/lib/system/form/builder/field/TI18nFormField.class.php#L108-L110

TimWolla:
Oh wow is this non-obvious, thanks. I've requested a review from myself and will come back to this later. Until then: Do you have a use case for me that can be included in the commit message? It will also help review / testing.

mutec:
> Do you have a use case for me that can be included in the commit message? It will also help review / testing.

When we'd store data correctly, you'd have to store null in case no username is given for an updateserver object. I think I used that a lot in the shop package and I'm pretty sure it's used some times when converting more classes. I'll have a look later.

mutec:
@TimWolla for use-cases please have a look at the shop packages (e.g. core, customer, currency) - any 8.0-version I think.
The integration differs a little bit, but you could easily modify the extending class (or bypassing that class and merge the changes from this PR to the WSC, should do the same I think, but would be more original).
Searching for usages of `shop\system\form\builder\field\TextFormField` should do it.

TimWolla:
Thank you, this still is on my list. I plan to have another look at this, once the current pending changes to form builder have landed.

TimWolla:
I've now taken a look at this and decided not to merge this. It's not unlikely the INullableFormField interface will be deprecated entirely in the future (with the existing fields cleaned up / replaced by a fixed variant).

The reason for this is that it is not semantically sound to interpret an empty value as NULL depending on a flag. Instead the nullability is an inherent property of the field:

A field that is `->required()` can never return “explicitly no input”. Thus we must concern us only with optional fields in the argumentation.

For a pure text input (i.e. without any other UI elements), it is impossible to determine whether the user explicitly doesn't want to give any input (`null`) or wants to provide an empty input (`''`). In other words: The empty text is a valid text input. By interpreting an empty string as `null` you are making assumption with regard to the user’s intent and those assumptions might not hold. Instead an extra UI element is required to determine the difference (like it's done in UserSearchForm). This extra UI element already exists in the form of form field dependencies.

For a date input on the other hand, all dates are equally valid choices. However the empty string is not. Interpreting the “empty string” as the explicit absence of a choice is sound there. The UI also provides the user an explicit means of not making a choice by providing the “clear selection” button. The absence of a choice is appropriately represented as `null` for that field. However using unix timestamp `0` (i.e. 1970-01-01T00:00:00Z) as the “no choice” value is semantically unsound, because it can also be a valid selection. See also: https://www.woltlab.com/community/thread/295616-dateformfield-speichert-minuten-und-sekunden-bei-supportstime-false/?postID=1919574#post1919574

PR #5270, opened 2023-02-02
Clean up form data processors

TimWolla:


TimWolla:
> Have you considered marking those properties as `readonly`?

I did not, I always forget about that keyword existing. At least I remember to add the `: void` return type now. I'll create an issue so that all existing typed properties can be checked whether making them `readonly` makes sense, there are going to be many more in all those singleton classes.

PR #5271, opened 2023-02-02
Clean up I18nHandler

TimWolla:


wcfsetup/install/files/lib/system/language/I18nHandler.class.php
```diff
@@ -102,7 +98,7 @@ public function unregister($elementID)
     /**
      * Reads plain and i18n values from request data.
      *
-     * @param array|null $requestData used request data (if `null`, `$_POST` is used)
+     * @param $requestData used request data (if `null`, `$_POST` is used)
      * @throws  SystemException
      */
     public function readValues(?array $requestData = null)
```
dtdesign, line 104: Missing a `void` return type here.

dtdesign (changes requested):
Otherwise LGTM

PR #5272, opened 2023-02-02
Always load object data for immutable form fields

### LINKED ISSUE #4725: immutable does not work as expected in the form builder
ref https://www.woltlab.com/community/thread/294841-formbuilder-fehlverhalten-bei-nutzung-von-immutable/

TimWolla:
When reading the input values in immutable form fields are (correctly) skipped,
thus leaving whatever value was there in place. However when an existing object
is provided, the values will not be read from that object, thus resulting in
the default value (which is likely some empty value).

This is confusing and will actually result in bugs, because when validating
saving that default value will then be used.

Fix this by checking whether the field is immutable and then forcing the load
of the object data, even if loading is not requested (i.e. when the input is
non-empty).

I've opted to implement the check within the fields themselves, but it might
make more sense to implement it in FormDocument::updatedObject() directly,
possibly avoiding issues if some immutable field overrides updatedObject and
forgets the check.

There is an existing asymmetry anyway, because the loading of input values is
done in TFormParentNode::readValues() which is pulled into the FormDocument via
the trait, but the loading of object data is directly implemented in
FormDocument. The result of this is that nodes have deeper access to input
loading, compared to object loading. The readValues() method should likely be
removed from the trait and implemented in the document directly, like the
object loading. The method does not really belong on the nodes. This should be
evaluated in a follow-up issue and possibly the immutability check for object
loading then be moved into the FormDocument as well, as suggested above.

see https://www.woltlab.com/community/thread/294841-formbuilder-fehlverhalten-bei-nutzung-von-immutable/
Resolves #4725

### DIFF
--- a/wcfsetup/install/files/lib/system/form/builder/container/SuffixFormFieldContainer.class.php
+++ b/wcfsetup/install/files/lib/system/form/builder/container/SuffixFormFieldContainer.class.php
@@ -178,15 +178,17 @@ class SuffixFormFieldContainer extends FormContainer
      */
     public function suffixHasSelectableOptions()
     {
-        if ($this->getSuffixField() === null) {
+        $suffixField = $this->getSuffixField();
+
+        if ($suffixField === null) {
             return false;
         }
 
-        if ($this->getSuffixField() instanceof IImmutableFormField && $this->getSuffixField()->isImmutable()) {
+        if ($suffixField instanceof IImmutableFormField && $suffixField->isImmutable()) {
             return false;
         }
 
-        foreach ($this->getSuffixField()->getNestedOptions() as $option) {
+        foreach ($suffixField->getNestedOptions() as $option) {
             if ($option['isSelectable']) {
                 return true;
             }
--- a/wcfsetup/install/files/lib/system/form/builder/field/AbstractFormField.class.php
+++ b/wcfsetup/install/files/lib/system/form/builder/field/AbstractFormField.class.php
@@ -218,6 +218,10 @@ abstract class AbstractFormField implements IFormField
      */
     public function updatedObject(array $data, IStorableObject $object, $loadValues = true)
     {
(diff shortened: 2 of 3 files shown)


dtdesign (approved):

TimWolla:
Just adjusted the commit message to include a reference to #4725.

@mutec Do you have any comments regarding this?

TimWolla:
@mutec ping

mutec:
@TimWolla pong
I'm trying to check this on Saturday or Sunday; I'm pretty busy at the moment. After having a first look it seams like you're reading data from the object (during the post) and submitting them to the action/via form-data. I'll have a look at the packages I'm participating at for possible side-issues I thought about, but I'm pretty sure those are more theory than real cases.

TimWolla:
> After having a first look it seams like you're reading data from the object (during the post) and submitting them to the action/via form-data.

Yes, exactly. For immutable fields any browser-submitted values will be overwritten (or should be, I'll double check that to make sure) by whatever is in the database. These values will then be treated as if they have been submitted by the browser. They will go through validation and ultimately end up in the save value. If validation is undesired, then the validator can check the field properties to see if it is immutable. I considered it important that validation will actually happen in the general case, because it makes debugging easier. It also ensures that invalid values will not end up in the database in case the reloading from the database fails.

TimWolla:
@mutec I just performed a clean rebase onto the latest master to improve testing with the current development state.

TimWolla:
@mutec Thank you; so I may consider this a “approval” and you didn't find any obvious issues?

SoftCreatR (approved):

SoftCreatR:
We have been doing excessive testing over the past few days. So I personally would approve that.

TimWolla:
Thank you. I'm also reasonably confident that the change is correct (and dtdesign approved), but we didn't have any real-world test cases.

PR #5274, opened 2023-02-02
Add `Helper::getPreferredContentType`

TimWolla:
This needs to be carefully reviewed, it worked on the first attempt :grin:

### DIFF
--- a/wcfsetup/install/files/lib/http/Helper.class.php
+++ b/wcfsetup/install/files/lib/http/Helper.class.php
@@ -5,6 +5,7 @@ namespace wcf\http;
 use CuyZ\Valinor\Mapper\MappingError;
 use CuyZ\Valinor\Mapper\Source\Source;
 use CuyZ\Valinor\MapperBuilder;
+use GuzzleHttp\Psr7\Header;
 use Psr\Http\Message\RequestInterface;
 
 /**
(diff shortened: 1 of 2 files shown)


wcfsetup/install/files/lib/http/middleware/HandleValinorMappingErrors.class.php
```diff
@@ -38,8 +39,11 @@ public function process(ServerRequestInterface $request, RequestHandlerInterface
             $errors = Messages::flattenFromNode($e->node())
                 ->formatWith(new PrependPath());
 
-            if ($this->prefersJson($request)) {
-                return new JsonResponse(
+            return match (Helper::getPreferredContentType($request, [
+                'text/html',
+                'application/json',
+            ])) {
```
TimWolla, line 45: Not sure about the formatting here.

wcfsetup/install/files/lib/http/Helper.class.php
```diff
@@ -26,6 +27,61 @@ public static function isAjaxRequest(RequestInterface $request): bool
...
+    /**
+     * Returns the MIME type the client prefers most from the input list of
+     * $availableTypes. Checks the request's 'accept' header.
+     *
+     * @param list<string> $availableTypes
+     */
+    public static function getPreferredContentType(RequestInterface $request, array $availableTypes): string
+    {
+        if (!$request->hasHeader('accept')) {
+            return false;
```
TimWolla, line 39: Wrong type.

wcfsetup/install/files/lib/http/middleware/HandleValinorMappingErrors.class.php
dtdesign: I would split it into two separate statements. First assign `$acceptedType = Helper::…` and then match against the variable.

wcfsetup/install/files/lib/http/Helper.class.php
```diff
@@ -26,6 +27,62 @@ public static function isAjaxRequest(RequestInterface $request): bool
         return $request->getHeaderLine('x-requested-with') === 'XMLHttpRequest';
     }
 
+    /**
+     * Returns the MIME type the client prefers most from the input list of
+     * $availableTypes. Checks the request's 'accept' header.
```
dtdesign, line 32: I was about to point out that your `match()` statement is not exhaustive when there is a type mismatch. I think this methods is in need of an explanation of its behavior, because it actually guarantees to either return a match or the first value of `$availableTypes`.

dtdesign (changes requested):

wcfsetup/install/files/lib/http/Helper.class.php
TimWolla: Okay, I've expanded the explanation of the behavior. Technically I believe the existing description already explains that the return value is an element from the input list, but it could certainly be written more explicitly.

PR #5283, opened 2023-02-06
Abstract implementation for category box controller

BurntimeX:


wcfsetup/install/files/acp/templates/boxCategoryConditions.tpl
```diff
@@ -0,0 +1,15 @@
+<section class="section">
```
TimWolla, line 1: Mixed indentation in this file.

wcfsetup/install/files/lib/data/article/category/ArticleCategoryNode.class.php
```diff
@@ -28,12 +28,7 @@ class ArticleCategoryNode extends CategoryNode
      */
     protected $articles;
 
-    /**
-     * Returns number of articles in the category.
-     *
-     * @return  int
-     */
-    public function getArticles()
```
TimWolla, line 36: Should the old method be preserved with a `public function getArticles() { return $this->getItems(); }` implementation + deprecation? Genuine question, I don't know if this was used anywhere.

wcfsetup/install/files/lib/system/box/AbstractCategoriesBoxController.class.php
```diff
@@ -0,0 +1,133 @@
...
+     * @inheritDoc
+     */
+    public function validateConditions()
+    {
+    }
+
+    /**
+     * @inheritDoc
+     */
+    public function getConditionDefinition()
```
TimWolla, line 89: Please also add return types to inherited methods (where reasonably possible): Adding return types in child classes is safe from a Liskov substitution principle (“LSP”) perspective, because return types are covariant.

wcfsetup/install/files/lib/system/box/AbstractCategoriesBoxController.class.php
```diff
@@ -0,0 +1,133 @@
...
+    }
+
+    protected abstract function getNodeTree(): CategoryNodeTree;
+
+    protected function getActiveCategory(): ?AbstractDecoratedCategory
+    {
+        return null;
+    }
+
+    protected function getResetLink(): string
```
TimWolla, line 64: If this is new and part of the API (i.e. not private), it should have a (simple) PHPDoc explaining the purpose. Same applies to other new methods.

wcfsetup/install/files/lib/system/box/ArticleCategoriesBoxController.class.php
```diff
@@ -16,46 +19,29 @@
...
-                'wcf',
-                ['categoryList' => $categoryList, 'activeCategory' => $activeCategory],
-                true
-            );
         }
+
+        return $activeCategory;
+    }
+
+    protected function getResetLink(): string
```
TimWolla, line 43: And then an `@inheritDoc` here.

TimWolla (changes requested):

TimWolla (approved):
Diff LGTM.

PR #5284, opened 2023-02-07
Add debug autoloader

### LINKED ISSUE #5282: Validate class casing in autoloader
One common issue when testing on Windows (with a case insensitive file system) is using an incorrectly cased class name when referencing a class. Because both the file system and PHP class names are case insensitive, the logic works properly, but breaks on Linux (or any case sensitive file system).

The autoloader should validate the class case in debug mode:

```
        if (\class_exists($className)) {
            $reflection = new ReflectionClass($className);

            if ($className !== $reflection->getName()) {
                throw new \Exception('…');
            }
        }
```

TimWolla:
Please perform another functional test with a case insensitive file system.

--------------

Resolves #5282


dtdesign (approved):
I have confirmed it to be working as expected on a case-insensitive filesystem.

PR #5285, opened 2023-02-07
Update sebastian/diff composer dependency

TimWolla:


dtdesign (commented):
I don’t see an issue with this PR in itself, but I am rather concerned about the remarks in the linked issue regarding the security policy. If the library is not properly hardened (read: designed and tested) against malicious data, then this could be a liability issue for us.

TimWolla:
> If the library is not properly hardened (read: designed and tested) against malicious data, then this could be a liability issue for us.

I'm not concerned about this: The library is fairly simple and easily reviewed. In fact I had a quick look over the entire implementation when first including it. The library as used by use effectively only does some “data processing” to implement a longest common subsequence detection algorithm – and that is pretty much agnostic with regard to the input values. The only attack vector I can think of that would be in-scope for the library is resource exhaustion – and frankly a homegrown implementation would not fare any better there.

dtdesign:
Fair enough. Feel free to continue with this.

PR #5293, opened 2023-02-09
Modify `confirmationFactory().withReason()` to allow the reason to be mandatory

dtdesign:
This is a follow-up for https://github.com/WoltLab/WCF/commit/75c922056f57ebf5e011217ad348a52bf3096dbe#r99945184 which implemented this incorrectly. The reason can now be both optional and mandatory, plus it relies on a global phrase instead to reduce duplication.

Function overloads for `softDelete()` have been added to improve type safety.

### DIFF
--- a/ts/WoltLabSuite/Core/Component/Confirmation.ts
+++ b/ts/WoltLabSuite/Core/Component/Confirmation.ts
@@ -121,7 +121,6 @@ class ConfirmationPrefab {
   async withReason(question: string, isOptional: boolean): Promise<ResultConfirmationWithReason> {
     const dialog = dialogFactory().withoutContent().asConfirmation();
 
-    let reason: HTMLTextAreaElement | undefined = undefined;
     const id = DomUtil.getUniqueId();
     const label = getPhrase(isOptional ? "wcf.dialog.confirmation.reason.optional" : "wcf.dialog.confirmation.reason");
 
@@ -130,7 +129,7 @@ class ConfirmationPrefab {
       <dt><label for="${id}">${label}</label></dt>
       <dd><textarea id="${id}" cols="40" rows="3"></textarea></dd>
     `;
-    reason = dl.querySelector("textarea")!;
+    const reason = dl.querySelector("textarea")!;
 
     dialog.content.append(dl);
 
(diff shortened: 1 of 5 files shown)


ts/WoltLabSuite/Core/Component/Confirmation.ts
```diff
@@ -78,37 +81,69 @@ class ConfirmationPrefab {
...
-  async withReason(question: string, askForReason: boolean): Promise<ResultConfirmationWithReason> {
+  async withReason(question: string, isOptional: boolean): Promise<ResultConfirmationWithReason> {
     const dialog = dialogFactory().withoutContent().asConfirmation();
 
     let reason: HTMLTextAreaElement | undefined = undefined;
-    if (askForReason) {
-      const id = DomUtil.getUniqueId();
-      const label = getPhrase("wcf.dialog.confirmation.reason");
+    const id = DomUtil.getUniqueId();
+    const label = getPhrase(isOptional ? "wcf.dialog.confirmation.reason.optional" : "wcf.dialog.confirmation.reason");
```
TimWolla, line 126: Should this use the asterisk instead?

TimWolla (commented):
The API LGTM, the remark is about the visuals.

ts/WoltLabSuite/Core/Component/Confirmation.ts
dtdesign: No, the asterisk is a horrible design decision. It should have always been the other way around by marking optional fields with an explicit label.

ts/WoltLabSuite/Core/Component/Confirmation.ts
```diff
@@ -78,42 +81,66 @@ class ConfirmationPrefab {
...
+        });
+      });
+    });
   }
 
-  async withReason(question: string, askForReason: boolean): Promise<ResultConfirmationWithReason> {
+  async withReason(question: string, isOptional: boolean): Promise<ResultConfirmationWithReason> {
     const dialog = dialogFactory().withoutContent().asConfirmation();
 
     let reason: HTMLTextAreaElement | undefined = undefined;
```
TimWolla, line 124: I believe the type can be simplified to no longer allow undefined.

PR #5319, opened 2023-02-22
CKEditor Integration

TimWolla:


eveneme:
CKEditor version 5 this is a very good solution.

PR #5330, opened 2023-03-02
Fix the `addEventListener` definition for custom elements

TimWolla:
The previous definition violated the liskov substitution principle, because not all valid calls for the parent type `HTMLElement` are valid calls on the custom element. Specifically it was not allowed to e.g. bind a `click` (or any other “standard”) event onto a `woltlab-core-dialog` element.

This error was reported when consuming the `.d.ts` definitions in a TypeScript project with `strict: true`.

Also the `addEventListener` was needlessly emitted as a real function that just delegates to `super.addEventListener()`.

Fix both issues by moving the function signature definition into an `interface` (to avoid the code generation) and by intersecting our more specific definition with the `HTMLElement["addEventListener"]` definition to inherit all signatures from `HTMLElement`.

see microsoft/TypeScript#52916

### DIFF
--- a/ts/WoltLabSuite/Core/Element/woltlab-core-dialog-control.ts
+++ b/ts/WoltLabSuite/Core/Element/woltlab-core-dialog-control.ts
@@ -144,19 +144,16 @@ export class WoltlabCoreDialogControlElement extends HTMLElement {
       this.append(this.#extraButton);
     }
   }
+}
 
-  public addEventListener<T extends keyof WoltlabCoreDialogControlEventMap>(
-    type: T,
-    listener: (this: WoltlabCoreDialogControlElement, ev: WoltlabCoreDialogControlEventMap[T]) => any,
-    options?: boolean | AddEventListenerOptions,
-  ): void;
-  public addEventListener(
-    type: string,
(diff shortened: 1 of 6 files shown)


ts/WoltLabSuite/Core/Element/woltlab-core-dialog-control.ts
```diff
@@ -144,19 +144,16 @@ export class WoltlabCoreDialogControlElement extends HTMLElement {
...
-  public addEventListener<T extends keyof WoltlabCoreDialogControlEventMap>(
-    type: T,
-    listener: (this: WoltlabCoreDialogControlElement, ev: WoltlabCoreDialogControlEventMap[T]) => any,
-    options?: boolean | AddEventListenerOptions,
-  ): void;
-  public addEventListener(
-    type: string,
-    listener: (this: WoltlabCoreDialogControlElement, ev: Event) => any,
-    options?: boolean | AddEventListenerOptions,
-  ): void {
```
TimWolla, line 157: As additional context: Keep in mind that for overloaded functions, the actual signature is not visible to the outside, unless it is repeated.

dtdesign (approved):
Remember to rebuild the web component bundle before merging.

PR #5344, opened 2023-03-08
Resize Upright Videos

nice42q:
'vh' 1% of the viewport's height.

Responsive style for all devices


dtdesign:
I’m curious, the value `72vh` ist oddly specific. What is your rationale for that particular value?

dtdesign:
Thanks!

PR #5374, opened 2023-03-28
Add `userBBCodeTag` to the shared templates

SoftCreatR:
The missing acp template `userBBCodeTag` may lead to exceptions.


TimWolla:
> when using it e.g. in a background job.

Background jobs should not run within the ACP. This PR is likely correct, but that explanation does not fit.

dtdesign (approved):

dtdesign:
Thanks!

Issue #4652, opened 2022-01-28
[Meta] Generated HTML is inconsistent


dtdesign:
The HTML produced by the HtmlInputProcessor/HtmlOutputProcessor is highly inconsistent, because we only enforce the validity of the HTML, but not the representation. This causes all sort of issues with whitespace detection, comparisons, or generally being able to determine the deterministic output of a call.

* [ ] #3955 
* [ ] #4022 
* [ ] WoltLab/com.woltlab.wcf.elasticSearch#14

PR #5375, opened 2023-03-29
Force replace some HTML tags before sending messages to the search index

dtdesign:
Stripping the HTML can cause certain words to be accidentally joined when there is no symbol between them that is recognized by the tokenizer. Inserting a whitespace at tag positions that are known to be prone is a stop-gap solution until we find a more stable replacement strategy.

See #4652 and WoltLab/com.woltlab.wcf.elasticSearch#14

### DIFF
--- a/wcfsetup/install/files/lib/system/search/SearchIndexManager.class.php
+++ b/wcfsetup/install/files/lib/system/search/SearchIndexManager.class.php
@@ -122,6 +122,24 @@ class SearchIndexManager extends SingletonFactory implements ISearchIndexManager
         $languageID = null,
         $metaData = ''
     ) {
+        // Force replace certain tags with a whitespace to prevent words from adjacent
+        // lines to be glued together.
+        $message = \str_replace([
+            '<br>',
+            '</h1>',
+            '</h2>',
+            '</h3>',
+            '</h4>',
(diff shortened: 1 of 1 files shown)


TimWolla (commented):
It would be little more resilient to re-add the matched tag within the replacement (i.e. perform a regex replacement by `\1 ` or so). I would also not be surprised if a regex (`~(<(?:br|/h1|…)>)~`) would be faster than scanning a possibly large string several times for each replaced tag.

Currently this is making the assumption that `StringUtil::stripHTML()` does not care about the well-formedness of the original HTML, which I had to verify.

TimWolla (approved):

TimWolla:
Looks good to me now, though it makes me wonder if this would technically be a bug in `stripHTML()` then. Stripping the HTML should leave the “textContent“ as-is.

dtdesign:
Consider this code:

```html
<table><tbody><tr><td>Hello</td><td>World</td></tr></tbody></table>
```

This is perfectly valid HTML code, but once you strip the HTML tags, everything that is left is `HelloWorld`. This PR changes this behavior to insert a whitespace after certain elements (such as `</td>`) in order to keep the contents separated even when the tags are gone.

PR #5381, opened 2023-04-02
(Un)Escape apostrophe

Cyperghost:
If you use the `wcf\system\form\builder\field\MultilineTextFormField` with `->i18n()` and save the example value
```
That's a test
```
As soon as the form is opened again, the text will be displayed as follows:
```
That&#039;s a test
```

However, the storage in the database is done correctly

### DIFF
--- a/ts/WoltLabSuite/Core/StringUtil.ts
+++ b/ts/WoltLabSuite/Core/StringUtil.ts
@@ -26,7 +26,7 @@ export function addThousandsSeparator(number: number): string {
  * Escapes special HTML-characters within a string
  */
 export function escapeHTML(string: string): string {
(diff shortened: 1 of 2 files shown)


TimWolla:
wow is the current logic terrible. The JavaScript expects the value HTML encoded to manually decode it, instead of just providing the raw value. Furthermore the PHP code runs `encodeJS()` instead of delegating that to the `|encodeJS` modifier (https://github.com/WoltLab/WCF/blob/db5497501a19688760fb0432ba8eab35679f2237/wcfsetup/install/files/lib/system/language/I18nHandler.class.php#L431).

TimWolla:
The proposed fix is likely the correct choice for 5.5, but with 6.0 this should likely be cleaned up. I'm just concerned about possible vulnerabilities, if someone relies on the encoding happening within the PHP code.

dtdesign (approved):

TimWolla:
Manually picked to write a proper commit message. Thank you!

PR #5410, opened 2023-04-07
Add the localization files for CKEditor 5

dtdesign:
Fixes WoltLab/editor#1


wcfsetup/install/files/lib/system/bbcode/BBCodeHandler.class.php
```diff
@@ -191,4 +192,109 @@ public function getImageExternalSourceWhitelist()
...
+        if ($index !== false) {
+            $locale = \substr($locale, 0, $index);
+
+            if (\in_array($locale, $availableTranslations, true)) {
+                return \sprintf(
+                    '"ckeditor5-translation/%s",',
+                    $locale
+                );
+            }
+        }
```
TimWolla, line 295:
```suggestion
        [$languageCode] = \explode('-', $locale, 2);

        if (\in_array($languageCode, $availableTranslations, true)) {
            return \sprintf(
                '"ckeditor5-translation/%s",',
                $languageCode
            );
        }
```

This will perform a redundant `in_array` check for unknown locales without region code, but is much simpler to understand.

wcfsetup/install/files/acp/templates/wysiwyg.tpl
```diff
@@ -3,7 +3,12 @@
 {event name='wysiwyg'}
 
 <script data-relocate="true">
-	require(["WoltLabSuite/Core/Component/Ckeditor"], ({ setupCkeditor }) => {
+	require([
+		"WoltLabSuite/Core/Component/Ckeditor",
+		{@$__wcf->getBBCodeHandler()->getEditorLocalization()}
```
TimWolla, line 8:
```suggestion
		'{@$__wcf->getBBCodeHandler()->getEditorLocalization()|encodeJS}',
```
If the suggestion for the PHP code is applied.

TimWolla (changes requested):

wcfsetup/install/files/lib/system/bbcode/BBCodeHandler.class.php
```diff
@@ -191,4 +192,109 @@ public function getImageExternalSourceWhitelist()
...
+            'uz',
+            'vi',
+            'zh-cn',
+            'zh',
+        ];
+
+        $locale = WCF::getLanguage()->getBcp47();
+        if (\in_array($locale, $availableTranslations, true)) {
+            return \sprintf(
+                '"ckeditor5-translation/%s",',
```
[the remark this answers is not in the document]
dtdesign: CKEditor requires the translations for the default language to be statically baked into the bundle, in our case this is `en`. Your suggestion creates a lot of noise in the template, because when the UI language is English, an empty string is returned.

PR #5412, opened 2023-04-10
Fix reading boolean field values in form builder dialogs

BurntimeX:
For normal forms, the value of `BooleanFormField` is passed as a string. In form builder dialogs, however, it is passed as an int.


TimWolla:
Picked into 5.5.

PR #5413, opened 2023-04-11
Fix JSON encoding in IconPreFilterTemplatePlugin

SoftCreatR:
Instead of the actual HTML code of an icon, the value of `$encodeJson` is returned instead.


TimWolla:
Yeah, that was non-sense. Thank you.

PR #5418, opened 2023-04-12
Stop using Exceptions to return error responses outside of a controller

### LINKED ISSUE #5238: AJAX error handling is not prepared for MappingErrors
https://github.com/WoltLab/WCF/blob/c3ed5cad84271dfa57b5347bfb1ccb0665d02198/ts/WoltLabSuite/Core/Ajax/Error.ts#L66-L87

https://github.com/WoltLab/WCF/blob/c3ed5cad84271dfa57b5347bfb1ccb0665d02198/wcfsetup/install/files/lib/http/middleware/HandleValinorMappingErrors.class.php#L34-L44

### LINKED ISSUE #5268: Create easily reusable variant of userException.tpl
https://github.com/WoltLab/WCF/blob/9543cac566752134fe60069762749da590553701/wcfsetup/install/files/lib/http/middleware/HandleValinorMappingErrors.class.php#L55

It should not be required to manually set a large number of variables, instead a single exception parameter should be sufficient and everything else should be derived from that.

As for the stack trace that is shown in the debug mode, the new template should be able to print the full stack trace, including any “previous” exceptions that are chained to it.

### LINKED ISSUE #5414: PermissionDeniedException for PSR-15 AJAX controllers not properly handled within the JS
Either the JS should be prepared for HTML or it should be returned as JSON, similarly to the Valinor middleware. Same likely applies for IllegalLinkException and NamedUserException. 

see #5238
see #5268

TimWolla:
This PR removes the use of Exceptions for response generation within middlewares. Exceptions thrown from a controller are intercepted right before the controller and passed into appropriate (reusable) handlers.

With regard to error handling in middlewares, I've made the following decisions:

- Delegate to an appropriate handler as appropriate. The handlers will perform content negotiation (HTML / JSON) as appropriate.
- For errors where the non-AJAX response is a redirect, it is specifically checked whether it's an AJAX request and no content negotiation is performed, because redirects are not useful for AJAX.
- For “Misconfiguration” errors that do not fit an existing handler and where a JSON response is not useful (e.g. expired evaluation), the code simply sends HTML.
- For technical errors that indicate a programming error (e.g. incorrect HTTP verb) an end user will not see, the existing plaintext response remains.

-----------

Resolves #5414 
Resolves #5268 
Resolves #5238


wcfsetup/install/files/lib/http/error/ErrorDetail.class.php
```diff
@@ -0,0 +1,52 @@
...
+    {
+        return $this->throwable;
+    }
+
+    public static function fromThrowable(\Throwable $e): self
+    {
+        return new self(
+            throwable: $e,
+            message: $e->getMessage(),
+        );
```
dtdesign, line 40:
```suggestion
        return new self($e->getMessage(), $e);
```

wcfsetup/install/files/lib/http/error/NotFoundHandler.class.php
```diff
@@ -0,0 +1,55 @@
...
+ * @since   6.0
+ */
+final class NotFoundHandler implements RequestHandlerInterface
+{
+    public function handle(ServerRequestInterface $request): ResponseInterface
+    {
+        $errorDetail = ErrorDetail::fromRequest($request);
+        $message = $errorDetail?->getMessage() ?? WCF::getLanguage()->getDynamicVariable('wcf.page.error.illegalLink');
+
+        if (!\class_exists(WCFACP::class, false)) {
```
dtdesign, line 30: → `RequestHandler::isACPRequest()`

wcfsetup/install/files/lib/http/error/NotFoundHandler.class.php
```diff
@@ -0,0 +1,55 @@
...
+            BoxHandler::disablePageLayout();
+        }
+        SessionHandler::getInstance()->disableTracking();
+
+        $preferredType = Helper::getPreferredContentType($request, [
+            'text/html',
+            'application/json',
```
dtdesign, line 37: These should be swapped.
* A regular browser request will ask for HTML.
* AJAX requests will expect JSON (at least in case of our error handling).
* AJAX requests could ask for a different content type, but in case of an error we’re returning JSON.

For all other types of requests, where HTML was not requested, HTML is not a valid substitute either. For example, an image tag will refuse to load anything that is not an image, therefore we can return whatever we want, because it will not be a valid substitute either way.

wcfsetup/install/files/lib/http/error/ErrorDetail.class.php
```diff
@@ -0,0 +1,52 @@
...
+
+namespace wcf\http\error;
+
+use Psr\Http\Message\ServerRequestInterface;
+
+/**
+ * Provides various helper methods for PSR-7/PSR-15 request processing.
+ *
+ * @author  Tim Duesterhus
+ * @copyright   2001-2022 WoltLab GmbH
```
dtdesign, line 11: 2023. Also please get rid of these inconsistent whitespaces between the tag and the value. A single space is all that it takes.

wcfsetup/install/files/lib/http/error/PermissionDeniedHandler.class.php
```diff
@@ -0,0 +1,57 @@
...
+
+        if (!\class_exists(WCFACP::class, false)) {
+            BoxHandler::disablePageLayout();
+            NoticeHandler::disableNotices();
+        }
+        SessionHandler::getInstance()->disableTracking();
+
+        $preferredType = Helper::getPreferredContentType($request, [
+            'text/html',
+            'application/json',
```
dtdesign, line 39: See my remarks `NotFoundHandler`.

dtdesign (changes requested):

wcfsetup/install/files/lib/http/error/ErrorDetail.class.php
TimWolla: Fixed; That comment was outdated as well, copy and paste mistake.

com.woltlab.wcf/templates/error.tpl
```diff
@@ -0,0 +1,29 @@
+{if ENABLE_DEBUG_MODE}
+{if $exception !== null}
+<!--
+{* A comment may not contain double dashes. *}
+{@'--'|str_replace:'- -':$exception}
+-->
+{/if}
+{/if}
```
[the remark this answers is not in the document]
dtdesign: This might mess up the encoding of characters, because there is potentially UTF-8 content before the document starts. At that point you dive deep into “garbage in – garbage out” territory, because you’re violating the specs.

You could move it right after the include of the `header` template and add a unique string that serves as a marker like this:
https://github.com/WoltLab/WCF/blob/6b9ea31ed7959bad65793f5c28339f509005ef86/com.woltlab.wcf/templates/pageFooter.tpl#L2-L18

wcfsetup/install/files/lib/http/error/NotFoundHandler.class.php
TimWolla: I believe that an HTML response is the correct default if the client does not send a preference. I've added a FixAcceptHeader middleware instead that adds a preference for JSON for AJAX requests without any preference. What do you think?

dtdesign: You’re missing my point here: Unless the client explicitly (*) requests HTML, HTML is _never_ a valid substitute. You are then either dealing with requests for other media types, for example, images or videos, in which case JSON at least saves computing time and bandwidth on the server, or any other type of request that is most likely a programmatic request.

(*) Or implicitly through the proposed `FixAcceptHeader` middleware.

dtdesign (approved):
I still believe the response needs to be JSON unless HTML was explicitly requested.

PR #5419, opened 2023-04-13
added new articleLikeButtons template event to article template

darkwood-studios:
similar to the messageFooterButtons there should also be the possibility to extend the button list in articles via template listener.


TimWolla:
Just to make sure: Is this intended for 6.0 only?

darkwood-studios:
@TimWolla An implementation in 5.5 would be great, too. For 5.5 I have created this PR: https://github.com/WoltLab/WCF/pull/5420

dtdesign:
I have merged #5420 into 5.5 and then merged its changes into the main branch, making this PR superfluous.

PR #5425, opened 2023-04-14
Drop AMP support

BurntimeX:


TimWolla (approved):
LGTM. Due to the large number of removal, I recommend just mentioning the AMP removal in https://docs.woltlab.com/6.0/migration/wsc55/deprecations_removals/, instead of separately listing each method / template / class.

PR #5427, opened 2023-04-18
Apply small stylig optimization

SoftCreatR:
Adding the `userExceptionMessage` makes it look better. And what about the inline CSS? Is that intentional? I would prefer to replace it with a proper CSS class as well.

### DIFF
--- a/com.woltlab.wcf/templates/error.tpl
+++ b/com.woltlab.wcf/templates/error.tpl
@@ -20,7 +20,7 @@
 <div class="section">
 	<div class="box64 userException">
 		{icon size=64 name='circle-exclamation'}
-		<p>
+		<p class="userExceptionMessage">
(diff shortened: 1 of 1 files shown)


TimWolla:
> And what about the inline CSS? Is that intentional? I would prefer to replace it with a proper CSS class as well.

When it comes to styling I've got two left hands. The template effectively is a mix between userException.tpl and permissionDenied.tpl, but I've not taken over the majority of the existing styling / attributes to gain a “clean start” without historical cruft.

dtdesign planned to have another look at them, but this PR might be a good start. Do you have before and after screenshots at hand?

SoftCreatR:
Before:
image

After:
image

dtdesign (approved):

dtdesign:
Thanks!

PR #5434, opened 2023-04-20
Add UsernameValidating event

TimWolla:
This event is useful for these two plugins and also us to clean up our own
code:

https://www.woltlab.com/pluginstore/file/6708-regular-expression-for-validation-of-usernames/
https://www.woltlab.com/pluginstore/file/4784-one-time-username/


Krymonota:
Cool, that would definitely simplify a lot for my plugin!

dtdesign (approved):

PR #5437, opened 2023-04-21
Support randomized cronjobs in cronjob PIP

### LINKED ISSUE #5202: refreshPackageUpdate cronjob's execution time does not stay randomized when resyncing / updating cronjob.xml

TimWolla:
WCFSetup is not tested yet. 

With regard to the actual scheduling, previously it was ensured that the
package list updater and search robot updater are not be scheduled anywhere
near each other, now it might happen that they run in the same minute.

While both of them can be expensive, the chance for them to actually exceed the
maximum execution time should be rare nowadays, as PHP configurations in the
wild become increasingly less broken.

In any case the admin is able to manually execute them and the more important
package list updater runs daily, whereas the less important robot updater only
monthly, allowing the package list updater to execute successfully on most of
the days.

--------

Resolves #5202


dtdesign (approved):

TimWolla:
WCFSetup now tested + fixed.

PR #5440, opened 2023-04-21
Align the styling of CKEditor 5 with the regular style

dtdesign:
The builtin styling diverges too much from our style guide and especially the dark mode variant looked a lot like a foreign object. The CSS has been adjusted to replicate a “good enough” integration that looks familiar without diverging too much.

### DIFF
--- a/wcfsetup/install/files/style/ui/ckeditor.scss
+++ b/wcfsetup/install/files/style/ui/ckeditor.scss
@@ -167,6 +167,10 @@
 	--ck-color-labeled-field-label-background: var(--ck-color-base-background);
 }
 
+.ck-insert-table-dropdown__grid .ck-button {
+	border-radius: 0 !important;
+}
+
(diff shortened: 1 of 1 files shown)


TimWolla (approved):

SoftCreatR:
There's a small thing I noticed:

When hovering over "Insert Block" for example, everything works fine, and looks good. But hovering over "Insert image" shows a white line/separator:

image

dtdesign:
Yes, that is part of the default styling of CKEditor. It uses the default color to create a visual separator for these kind of split buttons. It works fine in the light theme, but is a bit too strong in the dark theme. Using the border color from the toolbar isn’t helpful here either, because it has a somewhat similar luminance.

Maybe we can use a generic separator based on #fff/#000 with some added transparency to create a separator that adapts to the button color…

SoftCreatR:
lgtm.

PR #5451, opened 2023-04-24
Remove FirstTimeSetupForm.class.php on upgrade

SoftCreatR:
Upon performing a software upgrade, an issue has been identified whereby the user becomes constrained within the previous iteration of the First Time Setup page. Despite the presence of a new version, the user is presented with the outdated version due to the prioritization of the FirstTimeSetupForm over the recently added FirstTimeSetupAction. This issue stems from the oversight of not removing the previous controller and template during the _upgrade_ from version 5.5, resulting in a conflict with the updated software. In order to resolve this issue, it is necessary to remove the outdated components using the *delete-PiPs. Failure to do so may result in users being unable to properly utilize the updated software.

### DIFF
--- a/com.woltlab.wcf/acpTemplateDelete.xml
+++ b/com.woltlab.wcf/acpTemplateDelete.xml
@@ -13,6 +13,7 @@
 		<template>dashboardList</template>
 		<template>dashboardOption</template>
 		<template>devtoolsFormBuilderTest</template>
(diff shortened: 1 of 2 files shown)


TimWolla:
Thank you, but the fooDelete PIPs will be updated with an automated script before the actual releases, thus there is no need to send a PR for those.

PR #5452, opened 2023-04-25
Fix visuals in devtoolsProjectSync if the first non-important PIP has no target

TimWolla:
The thicker border was missing, because the element having the border was
visually hidden.


dtdesign (approved):

dtdesign:
_I only barfed twice reading that selector._

TimWolla:
> _I only barfed twice reading that selector._

Only put this up for review, so you were able to enjoy that beauty :stuck_out_tongue_winking_eye:

PR #5454, opened 2023-04-25
Add default implementation for user notifications about comments

BurntimeX:
The goal of this PR is that implementing notifications about comments requires less boilerplating. At the same time, it creates a generic solution for better subjects of email notifications about comments.


wcfsetup/install/files/lib/system/user/notification/event/ModerationQueueCommentResponseUserNotificationEvent.class.php
```diff
@@ -48,9 +48,10 @@ class ModerationQueueCommentResponseUserNotificationEvent extends AbstractShared
     protected $moderationQueueLoaded = false;
 
     /**
-     * @inheritDoc
+     * language item for the type name
+     * @var string|null
      */
-    protected $stackable = true;
+    protected $typeName;
```
TimWolla, line 54: This should be typed as `protected string $typeName;` (with the `@var` removed).

wcfsetup/install/files/lib/system/user/notification/event/ModerationQueueCommentResponseUserNotificationEvent.class.php
```diff
@@ -234,4 +216,27 @@ protected static function getTestCommentObjectData(UserProfile $recipient, UserP
             'objectTypeID' => CommentHandler::getInstance()->getObjectTypeID('com.woltlab.wcf.moderation.queue'),
         ];
     }
+
+    /**
+     * @inheritDoc
+     */
+    protected function getTypeName(): string
+    {
+        if ($this->typeName === null) {
```
TimWolla, line 225:
```suggestion
        if (!isset($this->typeName)) {
```
`isset()` is required to work with typed properties that might be *uninitialized*.

wcfsetup/install/files/lib/system/user/notification/event/ModerationQueueCommentUserNotificationEvent.class.php
```diff
@@ -36,15 +36,16 @@ class ModerationQueueCommentUserNotificationEvent extends AbstractUserNotificati
     protected $languageItemPrefix = '';
 
     /**
-     * moderation queue object the notifications (indirectly) belong to
-     * @var ViewableModerationQueue
+     * language item for the type name
+     * @var string
      */
-    protected $moderationQueue;
+    protected $typeName = '';
```
TimWolla, line 42: Likewise this should be typed without a default, especially since the default probably isn't useful it's better to error out if it is used uninitialized.

TimWolla (changes requested):
Apart from the three remarks wrt property types this LGTM.

PR #5457, opened 2023-04-26
Remove deprecated support for `ISearchableObjectType`

BurntimeX:


TimWolla (commented):
I can't meaningfully comment on this, because I didn't really follow the search changes in 5.5. The diff looks not obviously wrong, but that might not mean much.

BurntimeX:
This should possibly be postponed to 6.1 or later, depending on how far the migration to "ISearchProvider" has already been completed by third-party developers.

PR #5461, opened 2023-04-27
Custom icon for search result items

### LINKED ISSUE #5109: Bring back custom icon for seach result items
In commit 3616f5d90ce4fb95e940a94fa2354d2fa2421de8, you have removed the custom icon for search result items. According to you, "this feature was entirely unused". This may be the case with WoltLab apps, but there are existing use cases for plugins as this feature enables customization and an improved user experience at a low cost.

With [Simple FAQ](https://www.woltlab.com/pluginstore/file/6515-simple-faq/), there's at least one plugin in the Plugin-Store that utilizes this feature: https://github.com/HanashiDev/dev.hanashi.wsc.faq/blob/ecbfcf13d51cd66640c8f479a80f25d9e17c8085/files/lib/data/faq/Question.class.php#L17

In addition, it has been considered to use this feature for products in the VieCode Shop in the future.

Of course, a generic file icon is still better than a guest user profile as in previous WSC versions, but in my opinion it would be a shame to lose this feature for no (apparent) reason. Therefore, I would advocate for reintroducing this feature.

BurntimeX:
Closes #5109

### DIFF
--- a/wcfsetup/install/files/lib/system/search/SearchResultHandler.class.php
+++ b/wcfsetup/install/files/lib/system/search/SearchResultHandler.class.php
@@ -139,7 +139,7 @@ final class SearchResultHandler
     }
 
     /**
-     * @return string[]
+     * @return array<string, string|null>
      */
     public function getCustomIcons(): array
     {


com.woltlab.wcf/templates/searchResultList.tpl
```diff
@@ -3,7 +3,9 @@
 		{foreach from=$objects item=message}
 			<li class="gridListItem gridListItemMessage">
 				<div class="gridListItemImage">
-					{if $message->getUserProfile()}
+					{if !$customIcons[$message->getObjectTypeName()]|empty}
```
TimWolla, line 6:
```suggestion
					{if $customIcons[$message->getObjectTypeName()]|isset}
```

wcfsetup/install/files/lib/system/search/ISearchProvider.class.php
```diff
@@ -102,4 +102,11 @@ public function getCustomSortField(string $sortField): string;
      * placeholder {WCF_SEARCH_INNER_JOIN} within an empty INNER JOIN() statement.
      */
     public function getFetchObjectsQuery(?PreparedStatementConditionBuilder $additionalConditions = null): string;
+
+    /**
+     * Provides the ability to show a custom icon in the list of search results.
+     * Returns either a FontAwesome icon name or an empty string. The latter will trigger the default behavior.
```
TimWolla, line 108: I'd prefer `null` for the default behavior.

wcfsetup/install/files/lib/system/search/SearchResultHandler.class.php
```diff
@@ -138,6 +138,23 @@ public function getTemplateName(): array
...
+     * @return string[]
+     */
+    public function getCustomIcons(): array
+    {
+        $customIcons = [];
+        foreach (SearchEngine::getInstance()->getAvailableObjectTypes() as $name => $type) {
+            if ($type instanceof ISearchProvider) {
+                if ($type->getCustomIconName()) {
+                    $customIcons[$name] = $type->getCustomIconName();
+                }
```
TimWolla, line 151:
```suggestion
                $customIcons[$name] = $type->getCustomIconName();
```
If null is used, the check isn't necessary, because the suggested `isset()` in the template will implicitly check for null.

TimWolla (approved):
I recommend a squash and merge.

PR #5465, opened 2023-04-28
Use the new poll API in inline editor, integrate with CKEditor 5

### LINKED ISSUE #5431: Poll data not transmitted in quick reply
Tested with com.woltlab.wbb, but likely applies to everything that uses quick reply and polls.

dtdesign:
Fixes #5431


TimWolla (approved):
Not my area of expertise, so this isn't an in-depth review, but the Diff doesn't look wrong to me.

PR #5467, opened 2023-05-01
Fix LanguageImportForm

SoftCreatR:
The LanguageImported listener was previously receiving an object of type `wcf\data\language\LanguageEditor`, instead of the required `wcf\data\language\Language`. This commit updates the code to ensure that the listener is provided with a Language object, allowing it to function properly.


TimWolla (approved):

TimWolla:
Thanks. Apparently the PHPDoc for the `->language` property is outdated / broken then …

PR #5495, opened 2023-05-08
Enforce the selection of a landing page for WoltLab Suite Core in applicationManagement

### LINKED ISSUE #4843: Landing Page UI
This is a follow-up for #4842:

- [x] ~Re-add the checkbox to pageAdd.tpl, but configure the value in wcf1_application with it.~
- [x] Enforce the selection of a landing page for `wcf` in ApplicationManagement
- [x] Modify the `ApplicationManagamentForm` to be accessible in enterprise mode but with disabled domain configuration.

TimWolla:
This is enforced in the template only, because the PHP logic already needs to
handle a NULL landing page, as the global landing page could be deleted or
uninstalled. It will implicitly fall back to the ArticleList then.

Resolves #4843

### DIFF
--- a/wcfsetup/install/files/acp/templates/applicationManagement.tpl
+++ b/wcfsetup/install/files/acp/templates/applicationManagement.tpl
@@ -100,7 +100,7 @@
 							<td class="columnText columnLandingPageID">
 								<select name="landingPageID[{$application->packageID}]" required>
 									<option value="">{lang}wcf.global.noSelection{/lang}</option>
-									<option value="0"{if $application->getAbbreviation() !== 'wcf'}{if $application->landingPageID === null} selected{/if}{else} disabled{/if}>{lang}wcf.acp.application.landingPage.default{/lang}</option>
+									<option value="0"{if $application->getAbbreviation() === 'wcf'} disabled{elseif $application->landingPageID === null} selected{/if}>{lang}wcf.acp.application.landingPage.default{/lang}</option>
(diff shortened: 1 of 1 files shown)


wcfsetup/install/files/acp/templates/applicationManagement.tpl
```diff
@@ -98,8 +98,9 @@
 								<small>{$application->getPageURL()}</small>
 							</td>
 							<td class="columnText columnLandingPageID">
-								<select name="landingPageID[{$application->packageID}]">
-									<option value="0">{lang}wcf.acp.application.landingPage.default{/lang}</option>
+								<select name="landingPageID[{$application->packageID}]" required>
+									<option value="">{lang}wcf.global.noSelection{/lang}</option>
+									<option value="0"{if $application->getAbbreviation() !== 'wcf'}{if $application->landingPageID === null} selected{/if}{else} disabled{/if}>{lang}wcf.acp.application.landingPage.default{/lang}</option>
```
dtdesign, line 103:
```suggestion
									<option value="0"{if $application->getAbbreviation() === 'wcf'} disabled{elseif $application->landingPageID === null} selected{/if}>{lang}wcf.acp.application.landingPage.default{/lang}</option>
```

I literally had to copy this nested `if` into an editor and reformat it to double check it.

dtdesign (changes requested):

PR #5498, opened 2023-05-09
Fix handling of WebP avatars

### LINKED ISSUE #5264: Fix the handling of WebP avatars
https://github.com/WoltLab/WCF/blob/2c4095c7a798c929476bb2a19eff9d837d0230f3/wcfsetup/install/files/lib/data/user/avatar/UserAvatarEditor.class.php#L105-L110

BurntimeX:
Closes #5264


TimWolla (approved):
It's probably reasonable to not update the hash here, it wouldn't surprise me if it already isn't up to date when the avatar needs to be resized.

PR #5510, opened 2023-05-16
Improve search form UX

BurntimeX:
If a user uses the search function within a certain area (e.g. subforum or thread), only the corresponding area will be searched. If the search does not return any results in this case, the "advanced filter" area in the search form is collapsed and therefore not directly visible to the user that advanced filters are active.


dtdesign (changes requested):
Otherwise LGTM

PR #5526, opened 2023-05-25
Allow time function to be used without timestamps

SoftCreatR:
As of now, there's always a time property required. So calling this function without a timestamp results in an exception message.

With this PR, a timestamp is optional. When omitted, it uses TIME_NOW instead.


TimWolla:
It should be fairly rare use-case to output the current time as a relative timestamp (`{time}`) and even for the other formats outputting the current time is likely a rare occurrence.

Adding the `time=TIME_NOW` if that's what is desired is trivial. However by not adding a default value incorrect usages (e.g. `{time timestamp=$foo->time}`, because the parameter was misremembered) are more easily debugged, far outweighting the benefit of having a default value.

SoftCreatR:
Okay, makes sense :)

PR #5551, opened 2023-06-25
Add options to suppress text formatting

dtdesign:
Three new options will allow font color, families and sizes to be disallowed and dynamically stripped. The stripping takes place when outputting the message, preventing the alteration of stored formatting unless the message is actively being added – rebuilding does not modify the formatting.

Partially related: WoltLab/editor#60


wcfsetup/install/files/lib/system/html/output/node/HtmlOutputNodeProcessor.class.php
```diff
@@ -289,4 +289,61 @@ protected function invokeHtmlNode(IHtmlNode $htmlNode)
...
+                }
+
+                return true;
+            });
+
+            $style = \implode(";", $values);
+            if ($style !== "") {
+                $element->setAttribute("style", $style);
+                continue;
+            }
```
TimWolla, line 338: Is an `else` missing here to remove the attribute if it's empty after processing?

dtdesign: Fixed in 0ed94adac902bb751fa7168f5367a6a88eb13d5a

PR #5559, opened 2023-06-28
Dont show button for the comment menu if the menu has no entries

BurntimeX:
ref https://www.woltlab.com/community/thread/300327-kommentarmen%C3%BC-f%C3%BCr-g%C3%A4ste-ohne-funktion/


TimWolla (approved):
Diff LGTM

PR #5560, opened 2023-06-29
Add dark mode variant for Prism highlighting

TimWolla:
see https://www.woltlab.com/community/thread/300315-syntax-highlighting-im-dark-mode-unpassend/

### DIFF
--- a/wcfsetup/install/files/style/bbcode/code.scss
+++ b/wcfsetup/install/files/style/bbcode/code.scss
@@ -151,7 +151,7 @@
 	position: static;
 }
 
-html:not([data-color-scheme="dark"]) {
+html[data-color-scheme="light"] {
 	/* PrismJS 1.29.0
 	https://prismjs.com/download.html#themes=prism */
 	/**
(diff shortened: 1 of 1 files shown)


wcfsetup/install/files/style/bbcode/code.scss
```diff
@@ -151,83 +151,168 @@
...
- * @author Lea Verou
- */
-
-.prism-token.prism-comment,
-.prism-token.prism-prolog,
-.prism-token.prism-doctype,
-.prism-token.prism-cdata {
-	color: slategray;
-}
+html:not([data-color-scheme="dark"]) {
```
dtdesign, line 154:
```suggestion
html[data-color-scheme="light"] {
```

dtdesign (changes requested):

wcfsetup/install/files/style/bbcode/code.scss
TimWolla: > Do you really need this at all?

The two Prism styles are completely independent stylesheets obtained from an external source. I don't want any light mode styling to leak into the dark mode styling, because that will likely result in non-sense.

PR #5562, opened 2023-06-30
Move replacement of placerholders in mail texts into `UserMailBox`

BurntimeX:
Also added support for an `{$email}` placeholder.

### DIFF
--- a/wcfsetup/install/files/lib/system/email/UserMailbox.class.php
+++ b/wcfsetup/install/files/lib/system/email/UserMailbox.class.php
@@ -47,4 +47,15 @@ class UserMailbox extends Mailbox implements IUserMailbox
 
         return $user;
     }
+
+    /**
+     * Takes the passed text and replaces the placeholders for username and email address with the user's values.
+     */
+    public function getPersonalizedText(string $text): string
+    {
+        $text = \str_replace('{$username}', $this->getUser()->username, $text);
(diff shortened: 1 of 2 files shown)


wcfsetup/install/files/lib/system/email/UserMailbox.class.php
```diff
@@ -47,4 +47,15 @@ public function getUser(): User
...
+
+    /**
+     * Takes the passed text and replaces the placeholders for username and email address with the user's values.
+     */
+    public function getPersonalizedText(string $text): string
+    {
+        $text = \str_replace('{$username}', $this->getUser()->username, $text);
+        $text = \str_replace('{$email}', $this->getUser()->email, $text);
+
+        return $text;
```
TimWolla, line 59:
```suggestion
        return \strtr(
            $text,
            [
                '{$username}' => $this->getUser()->username,
                '{$email}' => $this->getUser()->email,
            ]
        );
```

This is required for correctness, because `strtr()` will not touch replaced values once more (https://www.php.net/manual/en/function.strtr.php):

> Once a substring has been replaced, its new value will not be searched again.

The current variant will incorrectly replace usernames containing `{$email}`.

Please double-check my suggested replacement / test it locally, it's written in GitHub's markdown box and might contain errors.

TimWolla (approved):

PR #5578, opened 2023-07-12
Improve the contrast of colors in the light mode

dtdesign:
n/t


TimWolla (commented):
Just confirming that the diff looks correct. Of course I can't say anything about the choice of color :smile:

PR #5583, opened 2023-07-13
Improve detection of URL in texts

BurntimeX:
ref https://www.woltlab.com/community/thread/296180-fehlerhafte-automatische-umwandlung-der-urls-wenn-am-ende-eine-klammer/

### DIFF
--- a/wcfsetup/install/files/lib/system/html/input/node/HtmlInputNodeTextParser.class.php
+++ b/wcfsetup/install/files/lib/system/html/input/node/HtmlInputNodeTextParser.class.php
@@ -420,32 +420,31 @@ class HtmlInputNodeTextParser
      */
     protected function parseURL(\DOMText $text, $value, $allowURL, $allowMedia)
     {
-        static $urlPattern = '';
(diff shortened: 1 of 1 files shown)


wcfsetup/install/files/lib/system/html/input/node/HtmlInputNodeTextParser.class.php
```diff
@@ -422,24 +422,7 @@ protected function parseURL(\DOMText $text, $value, $allowURL, $allowMedia)
     {
         static $urlPattern = '';
         if ($urlPattern === '') {
```
TimWolla, line 424: This can be removed now and the regex inlined in the `preg_replace_callback` call, as the regex is a static string. The `$illegalChars` property can likely be deprecated.

BurntimeX: `$illegalChars` is still being used in `parseEmail()`

TimWolla (approved):

PR #5587, opened 2023-07-14
Refactor `AttachmentBBCode`

dtdesign:
The intention is to clean up some long deprecated functions as well as organizing the code a bit. The previous implementation used one large method with heavy nesting of conditions to generate the HTML output.

My main goal was to split up the code path into smaller chunks to make it possible to get to the responsible section depending on the type of attachment being displayed. I also changed the HTML generation from barely readable multiline string concatenation into a series of `\sprintf()` calls.

### DIFF
--- a/wcfsetup/install/files/lib/system/bbcode/AttachmentBBCode.class.php
+++ b/wcfsetup/install/files/lib/system/bbcode/AttachmentBBCode.class.php
@@ -35,45 +35,21 @@ final class AttachmentBBCode extends AbstractBBCode
         $outputType = $parser->getOutputType();
 
         if ($attachment->showAsImage() && $attachment->canViewPreview() && ($outputType == 'text/html' || $outputType == 'text/simplified-html')) {
-            return $this->renderAsImage($attachment, $outputType, $openingTag['attributes']);
+            return $this->showImage($attachment, $outputType, $openingTag['attributes']);
         } elseif (\substr($attachment->fileType, 0, 6) === 'video/' && $outputType == 'text/html') {
-            return $this->renderAsVideoPlayer($attachment);
+            return $this->showVideoPlayer($attachment);
         } elseif (\substr($attachment->fileType, 0, 6) === 'audio/' && $outputType == 'text/html') {
-            return $this->rederAsAudioPlayer($attachment);
+            return $this->showAudioPlayer($attachment);
         }
 
         return StringUtil::getAnchorTag($attachment->getLink(), $attachment->filename);
     }
 
-    private function renderAsImage(Attachment $attachment, string $outputType, array $attributes): string
+    private function showImage(Attachment $attachment, string $outputType, array $attributes): string
     {
-        // image
-        $alignment = ($attributes[1] ?? '');
-        $thumbnail = ($attributes[2] ?? false);
+        $alignment = $attributes[1] ?? '';
 
-        // backward compatibility, check if width is larger than thumbnail's width to display full version
-        if (\is_numeric($thumbnail)) {
-            if ($thumbnail == 0) {
-                $thumbnail = true;
-            } else {
-                // true if supplied width is smaller or equal to thumbnail's width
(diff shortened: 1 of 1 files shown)


wcfsetup/install/files/lib/system/bbcode/AttachmentBBCode.class.php
```diff
@@ -55,162 +62,156 @@ public function getParsedTag(array $openingTag, $content, array $closingTag, BBC
...
-        $attachment = MessageEmbeddedObjectManager::getInstance()->getObject(
-            'com.woltlab.wcf.attachment',
-            $attachmentID
-        );
-        if ($attachment === null) {
-            if (self::$attachmentList !== null) {
-                $attachments = self::$attachmentList->getGroupedObjects(self::$objectID);
-                if (isset($attachments[$attachmentID])) {
-                    $attachment = $attachments[$attachmentID];
+        if ($thumbnail === false) {
```
TimWolla, line 65: I find this confusing, because `$thumbnail` is guaranteed to be a bool. 
```suggestion
        if (!$thumbnail) {
```

wcfsetup/install/files/lib/system/bbcode/AttachmentBBCode.class.php
```diff
@@ -55,162 +62,156 @@ public function getParsedTag(array $openingTag, $content, array $closingTag, BBC
...
+            $class = '';
+            if ($alignment === 'left' || $alignment === 'right') {
+                $class = 'messageFloatObject' . \ucfirst($alignment);
+            }
```
TimWolla, line 69: This might be more readable as `$class = match ($alignment)` with explicit strings, instead of `ucfirst()`.

wcfsetup/install/files/lib/system/bbcode/AttachmentBBCode.class.php
```diff
@@ -55,162 +62,156 @@ public function getParsedTag(array $openingTag, $content, array $closingTag, BBC
...
+        }
 
-                // Force the use of the thumbnail if the user cannot access the full version.
-                if (!$thumbnail && !$attachment->canDownload()) {
-                    $thumbnail = true;
-                }
+        $class = '';
+        if ($alignment == 'left' || $alignment == 'right') {
+            $class = 'messageFloatObject' . \ucfirst($alignment);
+        }
```
TimWolla, line 121: Ditto

TimWolla (changes requested):
The rest appears to be reasonable, but it was not easy to review this PR. Feel free to disregard the suggestion for match() if you disagree.

wcfsetup/install/files/lib/system/bbcode/AttachmentBBCode.class.php
dtdesign: This was intentional, because I misread that condition as `if ($thumbnail)` more than once in the original code.

However, it might be better to simply swap places of both code paths so that the first path will output the thumbnail.

dtdesign:
The code is anything but lightweight, because there is a lot of logic going on.

I can imagine moving the actual logic into a dedicated `<woltlab-core-attachment>` web component that handles the entire HTML generation. What do you think, @TimWolla ?

dtdesign:
Thinking about this, I come to the conclusion that a web component does not make much sense at this point. It might be an interesting feature to tap into once we introduce the new upload pipeline. A custom web component would allow for state changes, such as being able to submit a message while thumbnails are still being processed etc.

TimWolla (approved):

TimWolla:
> Thinking about this, I come to the conclusion that a web component does not make much sense at this point.

Agree. I feel strongly about not using JavaScript, unless a brings an actual value-add in interactivity. HTML generation is PHP’s job and if “lightweight-ness” is a concern, then moving the HTML generation from a well-powered server onto cheap Android phones doesn't make sense.

PR #5615, opened 2023-08-10
Insert trim to remove empty values

Cyperghost:
If the value is an empty string, this class contains an array with an empty string value. And the array is not empty


dtdesign (approved):

dtdesign:
Thank you!

Issue #5512, opened 2023-05-17
Trim down the list of allowed template modifiers


dtdesign:
WoltLab Suite 6.0 restricts the usage of built-in PHP functions in templates to a safe subset. This list was previously only enforced in the enterprise mode and was a few times extended for compatibility reasons.

There are still too many functions that imply some heavy business logic and should not belong in the template. For reference, this is the list at the time of writing: https://github.com/WoltLab/WCF/blob/c853f0e78a07fca3620025a110dd80ce7c7e43f4/wcfsetup/install/files/lib/system/template/TemplateScriptingCompiler.class.php#L37C1-L134

## Candidates for Removal

| PHP Function | Reason |
|--------------|--------|
| `addslashes` | Not safe for escaping, there are other proper escaping functions available. |
| `array_diff` | Comparing complex values is a non trivial task. |
| `array_fill` | Data structures should not be created in templates. |
| `array_slice` | Data manipulations should take place in PHP code. |
| `constant` | Complex conditions should be pre-computed in PHP code. |
| `doubleval` | This is an alias for `floatval()`. |
| `file_exists` | File system operation, usually relying on deprecated directory constants. |
| `hash` | Computing complex hashes that go beyond a simple `md5()` or `sha1()` should not be performed in the template. |
| `htmlspecialchars` | Not safe for escaping, there are other proper escaping functions available. |
| `html_entity_decode` | Decoding of data should be performed in PHP code or offloaded to the JavaScript utility functions. |
| `http_build_query` | URLs should be constructed in PHP or using JavaScript’s `URL` object. |
| `json_encode` | Encode values in PHP code, invoke `json` or use JavaScript’s `JSON.stringify()`. |
| `lcfirst` | Unsafe for UTF-8 strings, should use the `StringUtil` helper methods in either PHP or JavaScript. |
| `mb_strpos` | Appears twice in the list, one occurrence should be removed. |
| `nl2br` | Unsafe operation that potentially emits unescaped HTML, should use `newlineToBreak` instead. |
| `number_format` | There is already the `#` prefix as well as the `currency` modifier. |
| `parse_url` | Unsafe operation for malformed URLs, should use the PHP helper `Url` or the native JavaScript `URL` constructor. |
| ~`print_r`~ | ~For debugging purposes there is already `wcfDebug`~ |
| ~`rawurlencode`~ | ~Encoding of data should be performed in PHP code or offloaded to the native JavaScript functions.~ |
[...]

PR #5621, opened 2023-08-13
fix usage of forbidden functions in templates

mutec:
see #5512


dtdesign (approved):

dtdesign:
Thank you for your help!

PR #5640, opened 2023-09-03
Prevent concurrent `click` event while the callback is still loading

dtdesign:
This is primarily motivated by https://www.woltlab.com/community/thread/301446-unexpected-behaviour-when-double-clicking-on-create-album-in-the-gallery/ which uses a double click to trigger concurrent requests while the first request is still in flight. This issue can be reproduced with a throttled network link, for example, being on a slow 3G network.

The intention is to provide a simple to use wrapper that prevents multiple clicks, either by mistake or by a lack of response from the UI due to a high network latency or a slow server response. It relies on a promise being passed in and is intended to be combined with things like the `dialogFactory()` and the form builder.

The `AssignUser` action in the moderation queue has been updated to use this helper as a proof of concept.

### DIFF
--- a/ts/WoltLabSuite/Core/Helper/ClickGuard.ts
+++ b/ts/WoltLabSuite/Core/Helper/ClickGuard.ts
@@ -1,0 +1,38 @@
+/**
+ * Prevents concurrent runs of the event handler for the click event by blocking
+ * the event while a previous call is still running.
+ *
+ * @author Alexander Ebert
+ * @copyright 2001-2023 WoltLab GmbH
+ * @license GNU Lesser General Public License <http://opensource.org/licenses/lgpl-license.php>
+ * @since 6.0
+ */
+
+export function clickGuard(element: HTMLElement, eventHandler: (event: MouseEvent) => Promise<void>): void {
+  let pending = false;
+
+  element.addEventListener("click", (event) => {
+    if (pending) {
+      event.preventDefault();
+
+      return;
+    }
+
+    pending = true;
+
+    void eventHandler(event)
+      .then(
+        () => {
+          pending = false;
+        },
+        () => {
+          pending = false;
+        },
+      )
+      .catch((reason) => {
+        pending = false;
+
+        throw reason;
+      });
+  });
+}
--- a/ts/WoltLabSuite/Core/Controller/Moderation/AssignUser.ts
+++ b/ts/WoltLabSuite/Core/Controller/Moderation/AssignUser.ts
@@ -7,6 +7,7 @@
  * @since 6.0
  */
 
+import { clickGuard } from "WoltLabSuite/Core/Helper/ClickGuard";
 import { dialogFactory } from "../../Component/Dialog";
 import { getPhrase } from "../../Language";
 import { show as showNotification } from "../../Ui/Notification";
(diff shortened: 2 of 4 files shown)


ts/WoltLabSuite/Core/Helper/ClickGuard.ts
```diff
@@ -0,0 +1,38 @@
...
+        },
+        () => {
+          pending = false;
+        },
+      )
+      .catch((reason) => {
+        pending = false;
+
+        throw reason;
+      });
```
TimWolla, line 36: This `.catch()` will never trigger, because the `onRejected` callback of the preceding `.then()` is infallible. This can likely be simplified to just `eventHandler(event).finally(() => { pending = false; });`.

ts/WoltLabSuite/Core/Helper/ClickGuard.ts
```diff
@@ -0,0 +1,38 @@
...
+
+export function clickGuard(element: HTMLElement, eventHandler: (event: MouseEvent) => Promise<void>): void {
+  let pending = false;
+
+  element.addEventListener("click", (event) => {
+    if (pending) {
+      event.preventDefault();
+
+      return;
+    }
```
TimWolla, line 19: I'm not sure if this function should also bind the event listener. It might be useful to make this more generically reusable:

```
function promiseMutex<T extends ((...args: any[]) => Promise<unknown>)>(promise: T): ((...args: Parameters<T>) => boolean)
{
    let pending = false;

    return function (...args: Parameters<T>): boolean {
        if (pending) {
            return false;
        }
        pending = true;
        promise(...args).finally(() => {
            pending = false;
        });

        return true;
    };
}
```

and then:

```
button.addEventListener("click", promiseMutex(() => showDialog(button.dataset.url!)));
// or
const callback = promiseMutex(() => showDialog(button.dataset.url!));
button.addEventListener("click", (event) => {
  if (!callback(event)) {
     // already running
     event.preventDefault();
  }
});
```

see:

https://www.typescriptlang.org/play?target=8&ssl=16&ssc=2&pln=1&pc=1#code/GYVwdgxgLglg9mABABwE5wLYwM4FMCyIUuAHgDwAqipxYAJtogBRMB07AhqgObYBciDmACeAbQC6ASkQBeAHyIACuix4y4ANZg4AdzBzJcpmkw5cAipIEt2rLrwGKuHDLmKpslA7IUAjOHAANrhCkgBQAN5hiDGIwVAouPQwYNyyiMAcgXgA3GHRsahuIKhIoJCwCMy29vxKzq7unhQGAv5BIUhRsT2IMMDMyEl0KdzS3b2TRVAlZVm5BZMAvou9Q8mp6VCoILh5kzEmqrhsnDzYkqzAKVmBwizS8ogTB7HrI5syGfN7qz1Lkjyf0KxVKiG2u32sSWeRWYQgCGwCQAwgAhdJHMyEYgkJgcbDCSAZcDQeBIJgANyyuwEYBAGF8uFQViUKjMZApcBgdAUL0QCLASJQ6TAuB0rNMeBYRWwQQpuAANIgigArXDQR684ExPBQCgwVxwIgPHzK3CywLyynU3CSJUARgADM7AasAUCegKLbhWIE4Nw8ToODAEshXe78l6Or7-Uw0Ux7ZJXVHgjGA-GAExJvK6-WG41MTXPfmI6N+9OopgAZmziCWDoArC68kA

PR #5642, opened 2023-09-05
Add a license page to install WoltLab and Plugin-Store products

dtdesign:
This page offers customers a convenient way to install their products and replaces the legacy page for Plugin-Store purchases. Furthermore a new page to edit the license data was added which replaces the edit form for the official package servers, enhancing the workflow by a proper validation.


wcfsetup/install/files/lib/acp/page/LicensePage.class.php
```diff
@@ -0,0 +1,641 @@
...
+
+            \uksort($this->licenseData[$type], function ($packageA, $packageB) {
+                $a = $this->installedPackages[$packageA] ?? $this->packageUpdates[$packageA];
+                $b = $this->installedPackages[$packageB] ?? $this->packageUpdates[$packageB];
+
+                if ($a->isApplication === $b->isApplication) {
+                    return $a->getName() <=> $b->getName();
+                }
+
+                return $b->isApplication <=> $a->isApplication;
```
TimWolla, line 97:
```suggestion
                return ($b->isApplication <=> $a->isApplication) ?: ($a->getName() <=> $b->getName());
```

Possibly with a line-break inserted before `?:`. I find this more readable, because the conditions read in the natural order priority, instead of the reverse order.

ts/WoltLabSuite/Core/Acp/Component/License.ts
```diff
@@ -0,0 +1,14 @@
...
+  const installation = new AcpUiPackagePrepareInstallation();
+  installation.start(button.dataset.package!, button.dataset.packageVersion!);
+}
+
+export function setup(): void {
+  document.querySelectorAll<HTMLButtonElement>(".jsInstallPackage").forEach((button) => {
+    button.addEventListener("click", () => {
+      installPackage(button);
+    });
+  });
```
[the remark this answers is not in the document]
dtdesign: Thanks for your suggestion. I’ve implemented it in de651d589cb6e8ba55d0cb0421b4f78397b7f9f5 and while it is not ideal (the workflow of the legacy dialogs is awful), it should catch pretty much everything.

wcfsetup/install/files/lib/acp/page/LicensePage.class.php
```diff
@@ -0,0 +1,637 @@
...
+
+        foreach (['woltlab', 'pluginstore'] as $type) {
+            $this->licenseData[$type] = \array_filter(
+                $this->licenseData[$type],
+                function (string $package) use ($identifiers) {
+                    if (isset($this->installedPackages[$package])) {
+                        return true;
+                    }
+
+                    return \in_array($package, $identifiers);
```
TimWolla, line 84: Should this check for `$this->installablePackages`, instead of `$identifiers`? The code could probably benefit from *some* comments here, because the exact semantics are not immediately clear from the variable names and method names alone, because I'm not sure which packages / versions are pruned by `removeUnknownPackages` and which by `getInstallablePackages()`.

TimWolla (commented):
LGTM, except for that remaining remark.

wcfsetup/install/files/lib/acp/page/LicensePage.class.php
dtdesign: The filter `removeUnknownPackages()` exists to remove packages that are returned by the API but are unknown to the package server. For example, the legacy styles offered during the Burning Board 3 era are still part of the licenses but these do not exist on any package server anymore.

I guess that `removeUnknownPackages()` is probably redundant at this point but it was required in an earlier iteration because there were some more steps between it and `getInstallablePackages()`.

PR #5661, opened 2023-09-19
PoC: Fix handling of comment notifications

TimWolla:
This is completely untested.


wcfsetup/install/files/lib/system/user/notification/object/type/IMultiRecipientCommentResponseOwnerUserNotificationObjectType.class.php
```diff
@@ -5,24 +5,8 @@
...
- * being sent to multiple recipients and has a `commentResponseOwner` event.
- *
- * This interface is only required, if you use the interface `IMultiRecipientCommentUserNotificationObjectType`.
- * If you use not this interface, the `commentResponseOwner` event is fired by default.
- *
- * @author  Joshua Ruesweg
- * @copyright   2001-2019 WoltLab GmbH
- * @license GNU Lesser General Public License <http://opensource.org/licenses/lgpl-license.php>
- * @since   5.2
+ * @deprecated
```
dtdesign, line 8: Please include the version number and a brief summary of what to do if one implements this interface.

wcfsetup/install/files/lib/data/comment/CommentAction.class.php
```diff
@@ -501,37 +501,28 @@ public function triggerPublication()
...
-                            $objectType->objectType . '.notification',
-                            $notificationObject,
-                            [$userID],
-                            ['objectUserID' => $userID]
-                        );
-                    }
+                    $recipientIDs = [];
                 }
+
+                $recipientIDs[] = $notificationObjectType->getOwnerID($comment->commentID);
```
dtdesign, line 515: This can legitimately return a `0`. Maybe its wise to filter the `$recipientIDs` parameter in `UserNotificationHandler::fireEvent()` and simply return early if there are no recipient?

com.woltlab.wcf/userNotificationEvent.xml
```diff
@@ -84,12 +84,6 @@
 			<permissions>admin.content.cms.canManagePage</permissions>
 			<preset>1</preset>
 		</event>
-		<event>
```
dtdesign, line 87: This does not actually delete the event.

dtdesign (commented):
I believe it would have been easier to split this up into more commits, because there are changes like the inclusion of the object owner which are split across multiple files.

That said, it looks somewhat reasonable to me although there is a limit of how far I can debug this in my head. The notification handling for comments is quite messy so I second your choice of targeting 6.1.

TimWolla:
@dtdesign Still untested, but I've now split this into commits. The resulting diff is almost identical, guaranteeing that I didn't lose an intended change along the way. See https://github.com/WoltLab/WCF/compare/fd653ff05c968a967e0f67319911b3db1366a223..6a04fafe9a77e597ee8a5a4d55601db9c2d8a96b for reference.

dtdesign (approved):
LGTM

PR #5662, opened 2023-09-25
Entirely hide the search for guests if login is forced

TimWolla:
I've opted to select 6.0 as the target branch, but the movement of the
`{include}` might not be appropriate for 6.0. Rendering appears to be fine,
though. Please retarget if you're not comfortable with this change at this
point.

--------

The search is non-functional and possibly leaks information about installed
packages by the list of search providers.


dtdesign:
This is an unexpected change that also severely breaks the compatibility by moving the actual template for the search around. There is no way I’m including such a change in a RC version and I don’t think this change is acceptable for 6.1 either.

I’m in for the idea to hide the search when it is effectively unavailable, but this would also include things like the maintenance mode. We can look into this in 6.1 to figure out if we can gracefully hide the search areas.

PR #5663, opened 2023-09-26
Google maps Geocoding typing fix

Cyperghost:


ts/WoltLabSuite/Core/Component/GoogleMaps/Geocoding.ts
```diff
@@ -116,7 +116,7 @@ class Geocoding {
     this.#element.dataset.googleMapsLng = lng.toString();
 
     if (this.#element.hasAttribute("data-google-maps-geocoding-store") && this.#element.form) {
-      this.#store(this.#element.dataset.googleMapsGeocdingStore || "", lat, lng);
+      this.#store(this.#element.dataset.googleMapsGeocodingStore || "", lat, lng);
     }
```
TimWolla, line 120: This whole logic looks a little odd anyway. Can you check whether the following works:
```
if ("googleMapsGeocodingStore" in this.#element.dataset && this.#element.form) {
  this.#store(this.#element.dataset.googleMapsGeocodingStore, lat, lng);
}
```
This should hopefully cleanly type check.

TimWolla (changes requested):
Please make sure to rebase onto 6.0 and change the target branch to 6.0.

TimWolla:
You'll need to perform an actual rebase, otherwise you're including the changes from master.

Use:

```
git rebase master --onto 6.0
# force push
```

That should do the right thing.

ts/WoltLabSuite/Core/Component/GoogleMaps/Geocoding.ts
```diff
@@ -115,8 +115,8 @@ class Geocoding {
     this.#element.dataset.googleMapsLat = lat.toString();
     this.#element.dataset.googleMapsLng = lng.toString();
 
-    if (this.#element.hasAttribute("data-google-maps-geocoding-store") && this.#element.form) {
-      this.#store(this.#element.dataset.googleMapsGeocdingStore || "", lat, lng);
+    if ("googleMapsGeocodingStore" in this.#element.dataset && this.#element.form) {
+      this.#store(this.#element.dataset.googleMapsGeocodingStore!, lat, lng);
```
TimWolla, line 119: Is the exclamation mark actually necessary here? I hoped that TypeScript is able to narrow down the type.

Cyperghost: My IDE flags this as a possible error, because the value could be undefined.

PR #5664, opened 2023-09-26
Fix storing the Geocoding results

Cyperghost:
Fix typing in Google Maps Geocoding


TimWolla:
I've just pushed a commit to clean up the type checking. Apparently using `in` doesn't work with DOMStringMap, but it reliably returns `undefined` for missing attributes and the empty string for existing, but value-less attributes.

Can you please check once more that my change did not break your use case?

Cyperghost:
> I've just pushed a commit to clean up the type checking. Apparently using `in` doesn't work with DOMStringMap, but it reliably returns `undefined` for missing attributes and the empty string for existing, but value-less attributes.
> 
> Can you please check once more that my change did not break your use case?

Yes, this change has not broken anything.

Issue #4672, opened 2022-02-16
Unify string ordering functions


TimWolla:
we currently have:

- `<=>`: https://github.com/WoltLab/WCF/blob/ce4b59bfbf770fb5a88105fd3205f60ea6039181/wcfsetup/install/files/lib/acp/form/LanguageImportForm.class.php#L174
- `strcmp`: https://github.com/WoltLab/WCF/blob/ce4b59bfbf770fb5a88105fd3205f60ea6039181/wcfsetup/install/files/lib/acp/form/BoxAddForm.class.php#L267-L270
- `strcasecmp`: https://github.com/WoltLab/WCF/blob/ce4b59bfbf770fb5a88105fd3205f60ea6039181/wcfsetup/install/files/lib/acp/form/DataImportForm.class.php#L138-L141

They should all be unified to use `<=>`. `strcmp` is a direct replacement. `strcasecmp` will need to either call `strtolower` or `mb_strtolower` depending on whether the strings hold Unicode values or not.

TimWolla:
Maybe a helper function should be introduced for the case-insensitive comparison. It could use `Collator::compare()` when intl is available.

TimWolla:
> when intl is available.

Intl is always available now and it likely makes sense to use Collator even for case-sensitive comparisons to correctly handle accented characters.

PR #5670, opened 2023-09-28
Use intl Collator to sort strings

BurntimeX:
See #4672


TimWolla (approved):
LGTM. This should probably be mentioned in the migration guide for 6.1. Please file an issue in docs.woltlab.com as a follow-up.

Issue #5385, opened 2023-04-03
Redesign of the admin panel home page


BurntimeX:
The admin panel home page currently offers little benefit. I could imagine a redesign to an admin panel dashboard with different boxes containing useful information. The boxes could be extended by plugins and the administrator could also compose the displayed boxes individually.

Useful boxes would be for example:
- [x]  WoltLab news
- [x]  Available updates
- [x]  Status messages
- [ ]  Activity of the last days (e.g. forum posts)
- [x]  Users that need to be activated
- [ ] Information about expiring licenses

PR #5675, opened 2023-10-01
Admin dashboard

BurntimeX:
See https://github.com/WoltLab/WCF/issues/5385


wcfsetup/install/files/acp/style/layout.scss
```diff
@@ -909,3 +909,26 @@ html[data-color-scheme="dark"] {
...
+	grid-auto-rows: minmax(300px, auto);
+	margin-top: 40px;
+}
+
+.acpDashboardBox {
+	background-color: var(--wcfContentContainerBackground);
+	border-radius: var(--wcfBorderRadius);
+	box-shadow:
+		0 0 3px rgba(0, 0, 0, 0.12),
+		0 1px 2px rgba(0, 0, 0, 0.24);
```
dtdesign, line 926: Please use either `var(--wcfBoxShadow)` or `var(--wcfBoxShadowCard)`.

wcfsetup/install/files/acp/style/layout.scss
```diff
@@ -909,3 +909,26 @@ html[data-color-scheme="dark"] {
...
+	padding: 20px;
+}
+
+.acpDashboardBox__title {
+	margin-bottom: 20px;
```
dtdesign, line 931: I would rather set `.acpDashboardBox` to `display: flex` with `flex-direction: column` and `row-gap: 20px`. It is a bit more verbose, but avoids the use of margins and moves the layout to the top-most block.

com.woltlab.wcf/templates/creditsAcpDashboardBox.tpl
```diff
@@ -0,0 +1,59 @@
+<dl>
```
dtdesign, line 1: You’ve also copied the template to the frontend, I assume this was a mistake?

wcfsetup/install/files/lib/system/acp/dashboard/event/AcpDashboardCollecting.class.php
```diff
@@ -0,0 +1,38 @@
...
+{
+    /**
+     * @var IAcpDashboardBox[]
+     */
+    private array $boxes = [];
+
+    /**
+     * Registers a new box.
+     */
+    public function register(string $name, IAcpDashboardBox $box): void
```
dtdesign, line 26: I think it would be better if there was `$box->getName()` so that everything about the box is in one place?

dtdesign (changes requested):

wcfsetup/install/files/lib/bootstrap/com.woltlab.wcf.php
```diff
@@ -68,6 +72,12 @@
 
     $eventHandler->register(PackageUpdateListChanged::class, PackageUpdateListChangedLicenseListener::class);
 
+    $eventHandler->register(AcpDashboardCollecting::class, static function (AcpDashboardCollecting $event) {
+        $event->register('com.woltlab.wcf.news', new NewsAcpDashboardBox());
+        $event->register('com.woltlab.wcf.systemInfo', new SystemInfoAcpDashboardBox());
+        $event->register('com.woltlab.wcf.credits', new CreditsAcpDashboardBox());
```
TimWolla, line 78: For this kind of bulk-registration it might make sense to use a fully-qualified import, like done for `RebuildWorkerCollecting`.

wcfsetup/install/files/lib/system/acp/dashboard/event/AcpDashboardCollecting.class.php
TimWolla: Agreed. In fact it appears the name is not used at all, except for implicit deduplication.

wcfsetup/install/files/lib/system/acp/dashboard/box/UsersAwaitingApprovalAcpDashboardBox.class.php
```diff
@@ -0,0 +1,81 @@
...
+final class UsersAwaitingApprovalAcpDashboardBox extends AbstractAcpDashboardBox
+{
+    private UserProfileList $userList;
+    private int $usersAwaitingApproval;
+
+    #[\Override]
+    public function isAccessible(): bool
+    {
+        return WCF::getSession()->getPermission('admin.user.canSearchUser')
+            && REGISTER_ACTIVATION_METHOD & User::REGISTER_ACTIVATION_ADMIN;
```
TimWolla, line 26:
```suggestion
            && (REGISTER_ACTIVATION_METHOD & User::REGISTER_ACTIVATION_ADMIN);
```
Probably just copy-pasted, but we should make precedence clear while we're at it.

dtdesign (approved):

PR #5684, opened 2023-10-08
Replace the label chooser with a new `LabelPicker`

dtdesign:
The existing label selection uses a legacy jQuery component and requires a lot of boilerplate code to integrate with both the frontend and backend. The new `LabelPicker` and `LabelPickerGroup` implementation is designed to simplify the API usage and to provide a “drop-in replacement” for existing usages.

The selection of a label is now backed by the `<woltlab-core-label-picker>` web component that autonomously handles state changes and is eagerly evaluated. The later allows the component to render the proper initial state without requiring an extra render pass, removing the visible state change on page load when a value was preselected.

This PR is the foundation for the introduction of multiple label picks per label group and is intended to provide an API that is transparent to the implementers regardless of the use of single or multiple selections.

See https://www.woltlab.com/community/thread/288676-mehrere-labels-aus-selber-labelgruppe/


wcfsetup/install/files/lib/system/label/LabelPickerGroup.class.php
```diff
@@ -0,0 +1,259 @@
...
+            }
+        }
+
+        return false;
+    }
+
+    /**
+     * Returns an unencoded query string for `labelIDs` for use in the LinkHandler.
+     *
+     * @return string
```
TimWolla, line 101:
```suggestion
```

wcfsetup/install/files/lib/system/label/LabelPicker.class.php
```diff
@@ -0,0 +1,141 @@
...
+    /**
+     * Controls the availability of the inverted selection, allowing to filter
+     * for items that are not assigned any label of a label group.
+     */
+    public readonly bool $invertible;
+
+    /**
+     * Name of the hidden input field.
+     */
+    public string $name = 'labelIDs';
```
TimWolla, line 29: It's probably not possible to make this class entirely readonly, due to `$selected`, but it feels odd that `$name` is mutable, but `$invertible` is not.

PHP does not yet support “clone-with”, which makes the whole ordeal a little more unwieldy, but you might consider adding:

```
public function withName(string $name)
{
  $copy = clone $this;
  $copy->name = $name;

  return $copy;
}
```

and expect users not to touch `$name` directly.

wcfsetup/install/files/lib/system/label/LabelPicker.class.php
```diff
@@ -0,0 +1,141 @@
...
+
+    /**
+     * Returns the unique element id of this label picker.
+     */
+    public function getElementID(): string
+    {
+        if (!isset($this->elementID)) {
+            $this->elementID = \sprintf(
+                '%s_labelGroup%d',
+                \substr(\md5($this->name), 0, 8),
```
TimWolla, line 120: Is it important that the elementID is reproducible or does it just need to be unique?

If it does not need to be reproducible, just use: `bin2hex(random_bytes(16))`.

If it does need to be reproducible, the `md5()` call is an unnecessary red flag and can still conflict, e.g. when more than one form exists on a page?

wcfsetup/install/files/lib/system/label/LabelHandler.class.php
```diff
@@ -425,4 +425,37 @@ public function removeLabels($objectTypeID, array $objectIDs)
         $statement = WCF::getDB()->prepareStatement($sql);
         $statement->execute($conditions->getParameters());
     }
+
+    /**
+     * @param int[] $groupIDs
+     * @param bool $invertible
```
TimWolla, line 431:
```suggestion
```

wcfsetup/install/files/lib/data/article/category/ArticleCategory.class.php
```diff
@@ -195,6 +196,27 @@ public function getLabelGroups(string $permission = 'canSetLabel'): array
...
+        $accessibleCategoryIDs = self::getAccessibleCategoryIDs();
+
+        $groupIDs = [];
+        foreach ($labelGroupsToCategories as $categoryID => $__groupIDs) {
+            if (\in_array($categoryID, $accessibleCategoryIDs)) {
+                $groupIDs = \array_merge($groupIDs, $__groupIDs);
+            }
+        }
+
+        return LabelPickerGroup::fromGroupIDs(\array_unique($groupIDs), false);
```
TimWolla, line 217: Is it ever useful to handle duplicate `groupIDs`? If not, I'd push the `array_unique` call down into `fromGroupIDs()` to avoid repeating (and forgetting it) everywhere.

wcfsetup/install/files/lib/system/label/LabelPicker.class.php
```diff
@@ -0,0 +1,141 @@
...
+
+namespace wcf\system\label;
+
+use wcf\data\label\group\ViewableLabelGroup;
+use wcf\data\label\Label;
+use wcf\util\JSON;
+use wcf\util\StringUtil;
+
+/**
+ * Provides helper methods to interact with the label group.
```
TimWolla, line 11: Incorrect description?

TimWolla (commented):
I did a first pass through the PR, but did not yet look into all the files.

wcfsetup/install/files/lib/system/label/LabelPicker.class.php
dtdesign: This is a short-coming of the existing implementation and the new label picker is intended to be slotted right in with minimal friction. Unfortunately because of how the form builder works, the actual name is not known ahead of time so it must be mutable during its runtime.

wcfsetup/install/files/lib/system/form/builder/field/label/LabelFormField.class.php
```diff
@@ -36,6 +38,8 @@ final class LabelFormField extends AbstractFormField implements IObjectTypeFormN
      */
     protected $labelGroup;
 
+    protected LabelPickerGroup $labelPickerGroup;
```
TimWolla, line 41: I don't quite understand the addition of the `labelPickerGroup` here, since the whole logic just needs to turn it back into the singular `LabelPicker` anyways. Is this in anticipation of some future change or does `LabelPicker` itself not provide the necessary logic?

dtdesign: The form field itself is super awkward because it should have never been a label selection itself and rather a collection of label groups. The intended design is to use `LabelPickerGroup` to delegate all tasks but since the form field focuses on a single field I have to do some code gymnastics to make it work.

TimWolla (commented):
I don't have any further remarks.

PR #5686, opened 2023-10-13
Execute some middleware for early routing responses

TimWolla:


wcfsetup/install/files/lib/system/request/RequestHandler.class.php
```diff
@@ -146,8 +148,18 @@ public function handle(string $application = 'wcf', bool $isACPRequest = false):
...
+                \assert($builtRequest instanceof RequestHandlerInterface);
+
+                $pipeline = new Pipeline([
+                    new HandleStartupErrors(),
+                    new PreventMimeSniffing(),
+                    new AddAcpSecurityHeaders(),
+                    new EnforceCacheControlPrivate(),
+                    new EnforceNoCacheForTemporaryRedirects(),
+                    new EnforceFrameOptions(),
+                ]);
```
dtdesign, line 160: Aren’t those handlers now registered twice because I don’t see a change to the existing pipeline?

TimWolla: The regular middleware pipeline is created within the “then” part of the `if()`, this one is in the “else” part and completely independent.

The `Pipeline` class itself is immutable, thus adding additional middlewares is currently not possible, making a little duplication the easiest solution for now.

dtdesign (approved):

TimWolla:
Ah, crap. Incorrect target branch :weary:

TimWolla:
Now cherry-picked.

PR #5702, opened 2023-11-11
API to configure ACP menu items via bootstrap script

### LINKED ISSUE #5667: API to configure ACP menu items via bootstrap script

BurntimeX:
Closes #5667


dtdesign (commented):
Looks good, but please rename the new (!) classes to use `Acp` as the prefix so that we can eventually get rid of the `ACP` spelling.

Issue #5695, opened 2023-10-25
Overhaul the UX of the authentication process


BurntimeX:
All pages and forms that are part of the login and registration process should use a simplified layout/template that is different from the rest of the pages and where the form is the main focus.

We could omit the main menu, the user panel and the breadcrumbs.

<img width="1062" alt="Bildschirmfoto 2023-10-25 um 16 41 01" src="https://github.com/WoltLab/WCF/assets/915475/d134a405-6c8d-44d8-b143-64b62562b20f">

### Login Form
- [x]  Inputs for username/email+password
- [x]  Link to "Forgot Password?" form
- [x]  Link to Registration form
- [x]  Buttons for 3rd party logins

### Registration Form
- [x]  Skip disclaimer page and display an "Accept Terms" checkbox in the form instead.
- [x]  Remove inputs for "confirm email" and "confirm password".
- [x]  Overall make it more compact / remove the number of sections.
- [x] Link to Login form
- [x]  Buttons for 3rd party registration
- [x] Don't use a section for the reCaptcha

### Header
- [x] Separate links/buttons for login and register

eveneme:
You can also center the logo.

eveneme:
What do you think about adding the input name in input like in the photo? It would also save space.

![image](https://github.com/WoltLab/WCF/assets/85762221/6fcaff45-002e-4464-8156-b9146d4b6576)

cadeyrn:
> We could omit the main menu, the user panel and the breadcrumbs.

I don't agree that it should be made harder for users to reach other sites just because you're on the registration or login page. Especially, the main menu may contain important links that should be accessible from anywhere.

eveneme:
> > We could omit the main menu, the user panel and the breadcrumbs.
> 
> I don't agree that it should be made harder for users to reach other sites just because you're on the registration or login page. Especially, the main menu may contain important links that should be accessible from anywhere. Also, I can't see any advantage of using a different look and feel for these pages.

On the login or registration page, links to other pages are not necessary because these pages serve different purposes. Why do you need a menu on the login page? You can simply click on the logo, and it will take you to the homepage. Take a look at the login or registration pages of Facebook, Twitter, and so on. They usually have a well-organized UI/UX design.

cadeyrn:
> On the login or registration page, links to other pages are not necessary

[...]

PR #5716, opened 2023-11-24
Authentication process overhaul

BurntimeX:
see https://github.com/WoltLab/WCF/issues/5695


wcfsetup/install/files/lib/action/EmailValidationAction.class.php
```diff
@@ -0,0 +1,64 @@
...
+            }
+
+            if (User::getUserByEmail($bodyParameters['email'])->userID) {
+                $result = [
+                    'ok' => false,
+                    'error' => 'notUnique',
+                ];
+            }
+
+            return new JsonResponse($result);
```
dtdesign, line 59: This is a bit of an odd control flow. It would make more sense to immediately return a `JsonResponse` when a validation fails and at the very end report `ok => true`.

wcfsetup/install/files/lib/action/UsernameValidationAction.class.php
```diff
@@ -0,0 +1,64 @@
...
+            }
+
+            if (User::getUserByUsername($bodyParameters['username'])->userID) {
+                $result = [
+                    'ok' => false,
+                    'error' => 'notUnique',
+                ];
+            }
+
+            return new JsonResponse($result);
```
dtdesign, line 59: This is the same weird control flow as above.

wcfsetup/install/files/style/layout/form.scss
```diff
@@ -225,6 +225,20 @@ html[data-color-scheme="dark"] .formError {
 	--color-error: #ff4747;
 }
 
+.formSuccess {
+	--color-success: rgb(46, 139, 87);
```
dtdesign, line 229: This might require an additional declaration for the dark mode:
```css
html[data-color-scheme="dark"] .formSuccess {
    --color-success: rgb(some, color, here);
}
```

wcfsetup/install/files/lib/system/user/multifactor/totp/CodeFormField.class.php
```diff
@@ -74,6 +74,6 @@ public function getSaveValue(): array
      */
     protected static function getDefaultId(): string
     {
-        return 'onetimecode';
+        return 'code';
```
TimWolla, line 77: This effectively reverts #4392.

dtdesign (approved):

PR #5727, opened 2023-12-13
User authentication configuration

BurntimeX:
A central location for the configuration should make it easier to disable certain unwanted functions in SSO plugins.


dtdesign (changes requested):
Please add `@since 6.1` to each doc block, otherwise LGTM.

PR #5731, opened 2023-12-14
Use session variable to store login redirect URL

BurntimeX:
The goal of this PR is to make the redirection after login reliable by saving the URL in the session. This means that the redirection is also available after registration or after logging in via a third-party provider.


dtdesign (approved):
This also fixes https://www.woltlab.com/community/thread/303508-mehrmaliger-login-aufruf-erzeugt-immer-l%C3%A4ngere-url/

PR #5736, opened 2023-12-21
No margin-top after float object

Cyperghost:
See: https://www.woltlab.com/community/thread/303716-incorrect-margin-with-a-floated-object-at-the-start-of-a-message/

### DIFF
--- a/wcfsetup/install/files/style/layout/global.scss
+++ b/wcfsetup/install/files/style/layout/global.scss
@@ -197,6 +197,12 @@ ol.nativeList {
 			margin: 0 0 20px 20px;
 		}
 	}
+
+	.messageFloatObjectLeft, .image-style-side-left, .messageFloatObjectRight, .image-style-side {
+		+ p, + ul {
+			margin-top: 0 !important;
+		}
+	}
 }
 
 .containerContent,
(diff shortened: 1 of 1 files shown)


wcfsetup/install/files/style/layout/global.scss
```diff
@@ -197,6 +197,12 @@ ol.nativeList {
 			margin: 0 0 20px 20px;
 		}
 	}
+
+	.messageFloatObjectLeft, .image-style-side-left, .messageFloatObjectRight, .image-style-side {
+		+ p, + ul, + ol, + h1, + h2, + h3, + h4, + h5, + h6 {
```
dtdesign, line 202: This creates a lot of selectors, you should be able to simplify this with [`:is()`](https://developer.mozilla.org/en-US/docs/Web/CSS/:is).

dtdesign: Also, this selector is wrong, because it unconditionally targets the floating objects regardless of their position within the flow. You may only strip the top margin if this is actually the first direct child of the message text.

dtdesign (changes requested):

dtdesign (approved):

PR #5737, opened 2023-12-22
Reload content after save

Cyperghost:
See: 
- https://www.woltlab.com/community/thread/303729-fehlerhafte-bilddateneinbettung/
- https://www.woltlab.com/community/thread/303730-editor-eingebettete-medien-mit-bildunterschrift-sind-fehlerhaft/


dtdesign (changes requested):
These changes are incorrect. The preceding code allows for the input processor not being set which is achieved through the null coalescing operator (`??`) but you are using a simple ternary expression.

PS: “Ensue” is missing an “r” plus the comment does not end with a period.

dtdesign (approved):

PR #5741, opened 2024-01-02
Keep links for media available

Cyperghost:
If there is a media image element inside a `figure' tag and a link is added to that image, that link will not be saved.
This happened when a media image element marked as left-aligned and a link was added.

### DIFF
--- a/wcfsetup/install/files/lib/system/html/input/node/HtmlInputNodeImg.class.php
+++ b/wcfsetup/install/files/lib/system/html/input/node/HtmlInputNodeImg.class.php
@@ -71,6 +71,7 @@ class HtmlInputNodeImg extends AbstractHtmlInputNode
         foreach ($elements as $element) {
             $this->mirrorWidthAttribute($element);
             $this->moveClassNameFromFigureToImage($element);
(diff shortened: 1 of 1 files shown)


wcfsetup/install/files/lib/system/html/input/node/HtmlInputNodeImg.class.php
```diff
@@ -365,4 +366,29 @@ private function getParentFigure(\DOMElement $img): ?\DOMElement
...
+    private function moveFigureIntoATag(\DomElement $img): void
+    {
+        $figure = $this->getParentFigure($img);
+        $parent = $img->parentElement;
+        \assert($parent instanceof \DOMElement);
+        if ($figure === null || $parent->nodeName !== 'a') {
+            return;
+        }
+        $figure->parentNode->insertBefore($parent, $figure);
+        $parent->removeChild($img);
```
dtdesign, line 390:
```suggestion
```

This is redundant and possibly misleading. DOM nodes can only have one parent at a time, appending it somewhere else will move the entire node, regardless of where it was before.

dtdesign (changes requested):

PR #5747, opened 2024-01-08
Migrate follow/unfollow to typescript

BurntimeX:


ts/WoltLabSuite/Core/Component/User/Follow.ts
```diff
@@ -0,0 +1,41 @@
...
+    void prepareRequest(button.dataset.endpoint!)
+      .post({
+        action: "unfollow",
+      })
+      .fetchAsResponse();
+  }
+}
+
+export function setup(): void {
+  document.querySelectorAll<HTMLButtonElement>(".jsFollowButton").forEach((button) => {
```
dtdesign, line 36: First of all, `setup()` is never invoked, but could very well be dynamically invoked through a `whenFirstSeen()` call in `BootstrapFrontend`?

Second, I recommend using `wheneverFirstSeen()` here because it makes sure that any dynamically registered instances, such as through a popover, are still registered just fine.

Third: Using `.jsFollowButton` is unsafe because this is the previously used class name that did not have the `data-endpoint` set. It is wise to use a different class name or possibly even an attribute that also serves as the property for the endpoint:

```smarty
<button
    type="button"
    data-follow-user="{link controller='UserFollow' id=$targetUser->userID}{/link}"
>…</button>
```

ts/WoltLabSuite/Core/Component/User/Follow.ts
```diff
@@ -0,0 +1,41 @@
...
+ * @copyright 2001-2024 WoltLab GmbH
+ * @license GNU Lesser General Public License <http://opensource.org/licenses/lgpl-license.php>
+ * @since 6.1
+ */
+
+import { prepareRequest } from "WoltLabSuite/Core/Ajax/Backend";
+import { getPhrase } from "WoltLabSuite/Core/Language";
+
+function toggleFollow(button: HTMLButtonElement): void {
+  if (button.dataset.following != "1") {
```
dtdesign, line 14:
```suggestion
  if (button.dataset.following !== "1") {
```

wcfsetup/install/files/lib/action/UserFollowAction.class.php
```diff
@@ -0,0 +1,110 @@
...
+            $bodyParameters = Helper::mapRequestBody(
+                $request->getParsedBody(),
+                <<<'EOT'
+                    array {
+                        action: "follow" | "unfollow"
+                    }
+                    EOT
+            );
+
+            if ($bodyParameters['action'] == 'follow') {
```
dtdesign, line 59:
```suggestion
            if ($bodyParameters['action'] === 'follow') {
```

wcfsetup/install/files/lib/action/UserFollowAction.class.php
```diff
@@ -0,0 +1,110 @@
...
+        }
+    }
+
+    private function assertUserIsNotIgnored(User $target): void
+    {
+        $sql = "SELECT  ignoreID
+                FROM    wcf" . WCF_N . "_user_ignore
+                WHERE   userID = ?
+                    AND ignoreUserID = ?";
+        $statement = WCF::getDB()->prepareStatement($sql);
```
dtdesign, line 88:
```suggestion
        $sql = "SELECT  ignoreID
                FROM    wcf1_user_ignore
                WHERE   userID = ?
                    AND ignoreUserID = ?";
        $statement = WCF::getDB()->prepare($sql);
```

ts/WoltLabSuite/Core/Component/User/Follow.ts
```diff
@@ -0,0 +1,41 @@
...
+        action: "unfollow",
+      })
+      .fetchAsResponse();
+  }
+}
+
+export function setup(): void {
+  document.querySelectorAll<HTMLButtonElement>(".jsFollowButton").forEach((button) => {
+    button.addEventListener("click", () => {
+      toggleFollow(button);
```
dtdesign, line 38: You want to use `promiseMutex(() => toggleFollow(button));` here which will block concurrent requests by spamming the button.

We can then turn `toggleFollow()` into an `async` and invert the logic inside by first `await`ing the request before updating the UI. This sounds like a lot but is actually just a small change but makes sure that the UI stays in sync.

dtdesign (changes requested):

dtdesign (approved):

PR #5765, opened 2024-01-18
Allow access to the menu item in menu page handler

BurntimeX:


wcfsetup/install/files/lib/system/page/handler/AbstractMenuPageHandler.class.php
```diff
@@ -37,4 +44,20 @@ public function isVisible($objectID = null)
...
+    {
+        $this->menuItem = $menuItem;
+    }
+
+    /**
+     * @inheritDoc
+     */
+    public function getMenuItem(): MenuItem
+    {
+        return $this->menuItem;
```
dtdesign, line 61: This will fail hard if `$this->menuItem` is not set yet. You _could_ work around this by changing the return type to `?MenuItem` and change the body to `return $this->menuItem ?? null;`.

PR #5767, opened 2024-01-20
Fix inconsistent event handling in `woltlab-core-dialog.ts`

max-m:
Previously some events that should have been cancelable were not.
The [documentation](https://docs.woltlab.com/6.0/javascript/components_dialog/#event-access) says that `close`, `cancel` and `validate` can be canceled. (the documentation also misses the cancelable `backdrop` event)

I hope I tested all the event combinations / dialog options, but it’s possible I missed some edge cases. :worried: 

Summary of the changes:
- if `validate` failed it unconditionally closed the dialog, not emitting the `close` event, now it does and checks the event state
- if `validate` has no callbacks attached the `close` event gets emitted and if the dialog should not be closed the `primary` event is emitted manually
- the `cancel` form button did not emit the `close` event, now it does and checks the event state
- the `closeButton` unconditionally closed the dialog, not emitting the `close` event, now it does and checks the event state
- the dialog’s native `cancel` event (emitted for example when a user presses the `Escape` key or clicks on the backdrop) unconditionally closed the dialog while emitting a non-cancelable `cancel` event. Now the custom `cancel` event is cancelable, like the `cancel` form button event, and the event state is being checked.
- the custom `close` event used to check whether the dialog should be closed by checking the `event.defaultPrevented` property was in fact not cancelable, now it is

Note: The `backdrop` event is cancelable but does not have any effect in Firefox and Chromium. `event.preventDefault()` and `event.stopPropagation()` on the `mousedown` event did not have any effect and the browser still emitted the `cancel` event on the native dialog element.

### DIFF
--- a/ts/WoltLabSuite/Core/Element/woltlab-core-dialog.ts
+++ b/ts/WoltLabSuite/Core/Element/woltlab-core-dialog.ts
@@ -197,9 +197,18 @@ export class WoltlabCoreDialogElement extends HTMLElement {
             // `validate` to run again, causing an infinite loop.
             this.#dispatchPrimaryEvent();
 
-            this.close();
+            if (this.#shouldClose()) {
+              this.close();
+            }
           }
         });
+      }
+
+      if (!this.#shouldClose()) {
+        // Prevent the browser from closing the dialog
+        event.preventDefault();
+        // but dispatch the `primary` event
+        this.#dispatchPrimaryEvent();
       }
     });
 
@@ -217,7 +226,7 @@ export class WoltlabCoreDialogElement extends HTMLElement {
       const event = new CustomEvent("cancel", { cancelable: true });
       this.dispatchEvent(event);
 
-      if (!event.defaultPrevented) {
+      if (!event.defaultPrevented && this.#shouldClose()) {
         this.close();
       }
     });
@@ -248,7 +257,9 @@ export class WoltlabCoreDialogElement extends HTMLElement {
       closeButton.classList.add("dialog__closeButton", "jsTooltip");
       closeButton.title = Language.get("wcf.dialog.button.close");
       closeButton.addEventListener("click", () => {
-        this.close();
+        if (this.#shouldClose()) {
+          this.close();
+        }
       });
     }
 
@@ -276,11 +287,22 @@ export class WoltlabCoreDialogElement extends HTMLElement {
     this.#dialog.classList.add("dialog");
     this.#dialog.setAttribute("aria-labelledby", DomUtil.identify(this.#title));
 
-    this.#dialog.addEventListener("cancel", () => {
-      const event = new CustomEvent("cancel");
-      this.dispatchEvent(event);
+    this.#dialog.addEventListener("cancel", (event) => {
+      const evt = new CustomEvent("cancel", { cancelable: true });
+      this.dispatchEvent(evt);
 
-      this.#detachDialog();
+      if (evt.defaultPrevented) {
+        event.preventDefault();
+        return;
+      }
+
+      if (this.#shouldClose()) {
+        this.#detachDialog();
+      }
+      else {
+        // Prevent the browser from closing the dialog.
+        event.preventDefault();
+      }
     });
 
     // Close the dialog by clicking on the backdrop.
@@ -290,9 +312,9 @@ export class WoltlabCoreDialogElement extends HTMLElement {
     // dialog and then releasing it on the backdrop.
     this.#dialog.addEventListener("mousedown", (event) => {
       if (event.target === this.#dialog) {
-        const event = new CustomEvent("backdrop", { cancelable: true });
-        this.dispatchEvent(event);
-        if (event.defaultPrevented) {
+        const evt = new CustomEvent("backdrop", { cancelable: true });
+        this.dispatchEvent(evt);
+        if (evt.defaultPrevented) {
           return;
         }
 
@@ -306,7 +328,7 @@ export class WoltlabCoreDialogElement extends HTMLElement {
   }
 
   #shouldClose(): boolean {
-    const event = new CustomEvent("close");
+    const event = new CustomEvent("close", { cancelable: true });
     this.dispatchEvent(event);
 
     return event.defaultPrevented === false;
--- a/ts/WoltLabSuite/Core/Element/woltlab-core-dialog.ts
+++ b/ts/WoltLabSuite/Core/Element/woltlab-core-dialog.ts
@@ -173,8 +173,10 @@ export class WoltlabCoreDialogElement extends HTMLElement {
       });
       this.dispatchEvent(evt);
 
+      // Canceling this event is interpreted as a form validation failure.
       if (evt.defaultPrevented) {
         event.preventDefault();
+        return;
       }
 
       if (evt.detail.length > 0) {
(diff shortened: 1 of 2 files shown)


dtdesign:
About the dialog’s [`close`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLDialogElement/close_event) event:

> This event is not cancelable and does not bubble.

This means that while it is possible to invoke `preventDefault`, it does nothing. When the native `close` event is being dispatched, the dialog will close no matter what. How did that work in your testing, are you using a browser that behaves differently here?

max-m:
Sorry for the late reply, but I think I found some more issues.
With this patch a bare dialog / dialog without controls does not fire the cancel event when the backdrop is clicked, while prompts, alerts and confirmations do. I assume the browser doesn’t emit the native `cancel` event when there’s no `form` inside the dialog. That’s something I’ll have to investigate.
I’ve also noticed an issue with dialogs that cancel the `validate` event: Right now a `close` event gets emitted when the validation has failed.
I’ll try to fix those issues and update the PR.

Regarding the `close` event: As far as I understand the spec and the MDN (as well as browser behavior), the native `HTMLDialogElement`’s `close` event gets emitted after the dialog has been closed, similar to the custom `afterClose` event.

max-m:
I’ve pushed a commit that should address the `validate` issue.

Using the following CMS page I checked a couple of things.
- When `validate` gets cancelled the dialog stays open and no further events are fired. This seems sound.
- Cancelling `close` works. All user actions I could think of (:x:-button, primary button, secondary button, backdrop, escape, enter in a text input) fire the close event at last and cancelling it prevents the dialog from closing.
- Dialogs with an attached `WoltlabCoreDialogControlElement` and a “Cancel“ button fire the `cancel` event when the backdrop is clicked because the `WoltlabCoreDialogControlElement` listens for the `backdrop` event and clicks its cancel button. The other dialogs fire the close event when the backdrop is clicked, but they fire the `cancel` event when the escape key is used to close the dialog, because the browser fires its native `cancel` event on the `HTMLDialogElement` in that case.

```smarty
{foreach from=' '|explode:'prompt alert confirmation bare' key=key item=value}
	<p>
		<h3>{$value|ucfirst}</h3>
		<button class="button buttonPrimary" id="{$value}1" data-from="dialog1" data-type="{$value}">Default</button>
		<button class="button" id="{$value}2" data-from="dialog1" data-type="{$value}">Cancels <kbd>validate</kbd></button>
		<button class="button" id="{$value}3" data-from="dialog1" data-type="{$value}">Cancels <kbd>cancel</kbd></button>
		<button class="button" id="{$value}4" data-from="dialog1" data-type="{$value}">Cancels <kbd>close</kbd></button>
		<button class="button" id="{$value}5" data-from="dialog1" data-type="{$value}">Fails <kbd>validate</kbd></button>
	</p>
{/foreach}

<h3>Confirmation Presets</h3>
<button class="button" id="preset1">Soft delete</button>
<button class="button" id="preset2">Restore</button>
<button class="button" id="preset3">Delete</button>
<button class="button" id="preset4">Custom</button>

<template id="dialog1">
	<dl>
		<dt>
			<label for="myInput">Title</label>
		</dt>
		<dd>
			<input type="text" name="myInput" id="myInput" value="" />
		</dd>
	</dl>
</template>

{literal}
<script>
	setTimeout(() => {
		require(["WoltLabSuite/Core/Component/Dialog", "WoltLabSuite/Core/Component/Confirmation"], function (Dialog, Confirmation) {
			const dialogFactory = Dialog.dialogFactory;
			const confirmationFactory = Confirmation.confirmationFactory;

			const logEvents = (dialog) => {
				[
					"backdrop", // Fired when the backdrop has been clicked / touched.

					"validate", // Fired when a user submits the dialog’s `<form>`, before `primary` (DialogControls must be attached)
					"cancel", // Fired when dialog
					"close", // Fired **before** the dialog has been closed, can cancel the closing of the dialog

					// Non-cancelable
					"afterClose", // Fired **after** the dialog has been closed
					"primary", // Fired when a user submits the dialog’s `<form>` (DialogControls must be attached)
					"extra", // Fired when the user clicks the dialog’s `extra` button (DialogControls must be attached)
				].forEach((name) => {
					dialog.addEventListener(name, (event) => {
						console.log(name, event);
					});
				});
			};

			const createDialog = (id, callback) => {
				document
					.getElementById(id)
					.addEventListener("click", (event) => {
						const fromId = event.currentTarget.dataset.from;
						const type = event.currentTarget.dataset.type;

						let dialog = dialogFactory().fromId(fromId);
						switch (type) {
							case "prompt":
								dialog = dialog.asPrompt({
									extra: "Force Close",
								});
								dialog.addEventListener("extra", () =>
									dialog.close(),
								);
								break;
							case "alert":
								dialog = dialog.asAlert();
								break;
							case "confirmation":
								dialog = dialog.asConfirmation();
								break;
							case "bare":
								dialog = dialog.withoutControls();
								break;
						}

						if (typeof callback === "function") {
							callback(dialog);
						}

						logEvents(dialog);
						dialog.show(event.currentTarget.textContent);
					});
			};

			["prompt", "alert", "confirmation", "bare"].forEach((id) => {
				createDialog(`${id}1`);

				createDialog(`${id}2`, (dialog) => {
					dialog.addEventListener("validate", (e) =>
						e.preventDefault(),
					);
				});

				createDialog(`${id}3`, (dialog) => {
					dialog.addEventListener("cancel", (e) =>
						e.preventDefault(),
					);
				});

				createDialog(`${id}4`, (dialog) => {
					dialog.addEventListener("close", (e) => e.preventDefault());
				});

				createDialog(`${id}5`, (dialog) => {
					dialog.addEventListener("validate", (e) => e.detail.push(Promise.resolve(false)));
				});
			});

			document.getElementById('preset1').addEventListener('click', async e => {
				const { result, reason } = await confirmationFactory().softDelete("some object", true);
				console.log(result, reason);
			})

			document.getElementById('preset2').addEventListener('click', async e => {
				const result = await confirmationFactory().restore("some object");
				console.log(result);
			})

			document.getElementById('preset3').addEventListener('click', async e => {
				const result = await confirmationFactory().delete("some object");
				console.log(result);
			})

			document.getElementById('preset4').addEventListener('click', async e => {
				const result = await confirmationFactory().custom("Custom question").message("Custom message");
				console.log(result);
			})
		});
	}, 0);
</script>
{/literal}

```

max-m:
I just learned another peculiarity of Chrome:
When you explicitly click on the `HTMLDialogElement` (backdrop or content), pressing the escape key **does** emit the `cancel` event on the `HTMLDialogElement`, but when you press the escape key again without re-focusing the dialog element before, Chrome **will** close the dialog and emit the native `close` event.
Calling `this.#dialog.focus()` or `this.#dialog.click();` after `this.#dialog.showModal();` didn’t change the behavior, so I assume there’s no way sensible way to coerce Chrome into not forcefully closing dialog elements.

In theory you can intercept the key events for the escape key and prevent their default behavior, but that’s probably better done by the user of `WoltLabSuite/Core/Component/Dialog`.

```js
const preventEscape = (event) =>
    if (event.key == 'Escape') event.preventDefault();
};

dialog = ...

document.addEventListener('keydown', preventEscape);
dialog.addEventListener("close", (e) => e.preventDefault());
// in case your dialog has a handler that calls `dialog.close()`
dialog.addEventListener("afterClose", e => document.removeEventListener('keydown', preventEscape));

dialog.show('Foo');
```

dtdesign:
Thank you for your amazing work. It is truly painful to go through the whole logic because there exist so many different paths to dismiss a dialog.

Issue #4778, opened 2022-05-07
Bulk processing should automatically process large amounts of data in multiple batches


BurntimeX:
In basically the same way as the rebuild data function.

BurntimeX:
The primary issue is that all objects are determined, fetched and processed at once. We can possibly solve this issue by splitting this process up:

1. Determine all affected object ids and store them _somewhere_.
2. Start a worker process which processes `n` items and execute the action.

The batch processing is most likely only required for users, threads and post thus we don’t need an abstract implementation for this. A simple copy & paste job is sufficient here.

PR #5770, opened 2024-01-23
Run bulk processing in worker

Cyperghost:
See https://github.com/WoltLab/WCF/issues/4778


BurntimeX (changes requested):
Please add `@since 6.1` tags for the new methods / classes.

PR #5771, opened 2024-01-24
ACP user add/edit signature attachments

Cyperghost:
See https://www.woltlab.com/community/thread/304178-signatur-dateianhang-verschwindet-beim-speichern-oder-andere-komische-effekte/

In the ACP, add message tabs for smileys and attachments to the signature.

### DIFF
--- a/wcfsetup/install/files/lib/acp/form/UserAddForm.class.php
+++ b/wcfsetup/install/files/lib/acp/form/UserAddForm.class.php
@@ -128,7 +128,7 @@ class UserAddForm extends UserOptionListForm
      * @var array
      */
     public $optionTree = [];
-    public ?AttachmentHandler $attachmentHandler;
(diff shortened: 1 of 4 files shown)


wcfsetup/install/files/lib/acp/form/UserAddForm.class.php
```diff
@@ -124,6 +128,35 @@ class UserAddForm extends UserOptionListForm
      * @var array
      */
     public $optionTree = [];
+    public ?AttachmentHandler $attachmentHandler;
```
dtdesign, line 131: This makes no sense. The handler is either set or not, at no point is a `value` being assigned. Please keep in mind that typed properties without an initializer are not set at all, which is different to untyped properties that will implicitly initialized with `null`.

dtdesign (changes requested):

PR #5781, opened 2024-01-27
Lowercase the hostname for use in wcf1_unfurl_url

### LINKED ISSUE #5651: Lowercase the hostname for use in wcf1_unfurl_url
… allowing to deduplicate URLs that only differ in the hostname's casing.

BurntimeX:
Closes #5651

### DIFF
--- a/wcfsetup/install/files/lib/system/html/node/HtmlNodeUnfurlLink.class.php
+++ b/wcfsetup/install/files/lib/system/html/node/HtmlNodeUnfurlLink.class.php
@@ -2,6 +2,7 @@
 
 namespace wcf\system\html\node;
 
+use Laminas\Diactoros\Uri;
 use wcf\data\unfurl\url\UnfurlUrlAction;
 use wcf\util\DOMUtil;
 use wcf\util\Url;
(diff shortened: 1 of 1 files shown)


dtdesign (approved):

wcfsetup/install/files/lib/system/html/node/HtmlNodeUnfurlLink.class.php
```diff
@@ -63,4 +60,24 @@ private static function removeStyling(HtmlNodePlainLink $element): void
 
         $element->topLevelParent->appendChild($element->link);
     }
+
+    private static function lowercaseHostname(string $url): string
+    {
+        $uri = new Uri($url);
+        $uri->withHost(\mb_strtolower($uri->getHost()));
```
TimWolla, line 67: The Uri class is immutable, thus this line is a noop. (generally speaking: `with*()` methods return a new object)

BurntimeX: Thanks for pointing this out.

PR #5787, opened 2024-01-31
Html upcast processor

Cyperghost:


wcfsetup/install/files/lib/acp/form/ArticleAddForm.class.php
```diff
@@ -645,6 +646,23 @@ public function assignVariables()
 
         SmileyCache::getInstance()->assignVariables();
 
+        // readData not work in edit form
```
dtdesign, line 649: What does this mean?

wcfsetup/install/files/lib/system/html/input/node/HtmlInputNodeImg.class.php
```diff
@@ -149,6 +152,104 @@ protected function handleAttachment(\DOMElement $element, $class)
         DOMUtil::replaceElement($replaceElement, $newElement, false);
     }
 
+    /**
+     * Replaces image element with media metacode element.
+     *
+     * @param \DOMElement $element
+     * @param string $class
```
dtdesign, line 159:
```suggestion
```

dtdesign (changes requested):
Please mark all new classes as `final` by default and turn `protected` properties/methods into `private` ones. Unless they inherit from an existing implementation and the properties/methods are `protected` before.

wcfsetup/install/files/lib/acp/form/ArticleAddForm.class.php
Cyperghost: It was just a comment to me that I can't implement the code below in the readData function. Because in the ArticleEditForm the content is changed with the content from the edit object, after this code will be executed.

dtdesign (approved):

PR #5789, opened 2024-02-01
Check if recipientIDs empty

Cyperghost:
When `$recipientIDs` is empty the call `$conditions->add("userID IN (?)", [$recipientIDs]);` will thrown an exception.


Cyperghost:
> This is actually quite late in the code flow and takes place after the somewhat broken event is being fired. Since `$recipientIDs` isn’t mutated before, we can move this to the very top.
> 
> Is there any particular reason why you placed it so far down?

Yes, that is correct. I didn't realise that the event was already marked as deprecated.

dtdesign (approved):

Issue #5571, opened 2023-07-07
Remove code for TOC from RSS feed


BurntimeX:
ref https://www.woltlab.com/community/thread/295378-toc-und-js-in-artikel-feed/

PR #5790, opened 2024-02-01
Add API for RSS feed generation

BurntimeX:
ref https://github.com/WoltLab/WCF/issues/5571

### DIFF
--- a/wcfsetup/install/files/lib/system/rssFeed/RssFeed.class.php
+++ b/wcfsetup/install/files/lib/system/rssFeed/RssFeed.class.php
@@ -1,0 +1,53 @@
+<?php
+
+namespace wcf\system\rssFeed;
+
+final class RssFeed
+{
+    /**
+     * @var RssFeedChannel[]
+     */
+    private array $channels = [];
+
+    public function channel(RssFeedChannel $channel): static
+    {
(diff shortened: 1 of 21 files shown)


wcfsetup/install/files/lib/system/rssFeed/RssFeed.class.php
```diff
@@ -0,0 +1,53 @@
...
+    {
+        $header = <<<'EOT'
+            <?xml version="1.0" encoding="UTF-8"?>
+            <rss version="2.0"
+                xmlns:atom="http://www.w3.org/2005/Atom"
+                xmlns:content="http://purl.org/rss/1.0/modules/content/"
+                xmlns:dc="http://purl.org/dc/elements/1.1/"
+                xmlns:slash="http://purl.org/rss/1.0/modules/slash/"
+            >
+        EOT;
```
dtdesign, line 29:
```suggestion
            EOT;
```

The indentation matters here. PHP will trim all lines equally based on how the closing tag is indented. Currently the code outputs four spaces in front of the XML declaration, but indenting `EOT;` one more step will get rid of it.

dtdesign (changes requested):

Issue #3960, opened 2021-02-08
Use service workers for push notifications


dtdesign:
The current implementation for push notifications is implemented through a period polling with increasing delays during inactivity. This does create some load on the server (although very little even for large communities) and is anything but "instant".

We should look into service workers and figure out what the data security looks like.

TimWolla:
Possibly related to #2777.

dtdesign:
Depends on #4295, enabling us to use the `localStorage` to reliably keep multiple tabs in sync.

darsto:
Service workers only work with HTTPS [1], so even if you decide to use them there should probably be a legacy fallback. Also, I'm already using a service worker on a WCF site for cache management so I would probably end up disabling WCF's one anyway. (because there can be only one service worker per domain).

[1] From MDN: Service workers only run over HTTPS, for security reasons. Having modified network requests, wide open to man in the middle attacks would be really bad. In Firefox, Service Worker APIs are also hidden and cannot be used when the user is in private browsing mode. https://developer.mozilla.org/en-US/docs/Web/API/Service_Worker_API

cadeyrn:
> Service workers only work with HTTPS [1], so even if you decide to use them there should probably be a legacy fallback.

Honestly I do not see that need. Why should any administrator run a website without HTTPS in the year 2022 (and later)? There are zero reasons. And even if an administrator decides so - push notifications are not strictly needed and it's an explicit decision to deny the use of HTTPS. So why bother with a legacy implementation?

darsto:
If not intentional HTTP then testing or just internal installations come to mind. Maybe they are minor cases, I don't know. But it's exactly the case of 'why bother'. The current solution works, this one has drawbacks, imo no need to touch it.

cadeyrn:
> If not intentional HTTP then testing or just internal installations come to mind.

There is no reason why you couldn't use HTTPS in testing or for internal installations.

Cyperghost:
- A public and private key is required as [p256ecdsa](https://developer.mozilla.org/en-US/docs/Web/API/PushSubscriptionOptions/applicationServerKey)
  - The public key must have a length of 65 bytes.
  - The private key must have a length of 32 bytes.
  - If these are changed, the database with all endpoints must be cleared, as they can no longer be used.
- A script that register the `serviceWorker.js` https://developer.mozilla.org/en-US/docs/Web/API/ServiceWorkerContainer/register#examples
  - This script need our server public key
[...]

PR #5805, opened 2024-02-16
Add service worker for push notifications

Cyperghost:
Closes https://github.com/WoltLab/WCF/issues/3960


ts/WoltLabSuite/Core/Notification/ServiceWorker.ts
```diff
@@ -0,0 +1,134 @@
...
+  if (!("showNotification" in ServiceWorkerRegistration.prototype)) {
+    return false;
+  }
+  if ("Notification" in window && Notification.permission === "denied") {
+    return false;
+  }
+  return true;
+}
+
+export function init(publicKey: string, serviceWorkerJsUrl: string, registerUrl: string): void {
```
dtdesign, line 122: Please use `setup()` to keep the naming predictable.

ts/WoltLabSuite/Core/Notification/ServiceWorker.ts
```diff
@@ -0,0 +1,134 @@
+/**
+ * @author      Olaf Braun
+ * @copyright   2001-2024 WoltLab GmbH
+ * @license     GNU Lesser General Public License <http://opensource.org/licenses/lgpl-license.php>
```
dtdesign, line 4:
```suggestion
 * @license     GNU Lesser General Public License <http://opensource.org/licenses/lgpl-license.php>
 * @since 6.1
```

ts/WoltLabSuite/Core/Notification/ServiceWorker.ts
```diff
@@ -0,0 +1,134 @@
...
+
+import { prepareRequest } from "WoltLabSuite/Core/Ajax/Backend";
+
+let _serviceWorker: ServiceWorker | null = null;
+
+class ServiceWorker {
+  private readonly publicKey: string;
+  private readonly serviceWorkerJsUrl: string;
+  private readonly registerUrl: string;
+  private readonly serviceWorkerRegistration: Promise<ServiceWorkerRegistration>;
```
dtdesign, line 16: Please use actual private fields.

wcfsetup/install/files/lib/action/RegisterServiceWorkerAction.class.php
```diff
@@ -0,0 +1,82 @@
...
+            $request->getParsedBody(),
+            <<<'EOT'
+                array {
+                    remove: bool,
+                    endpoint: non-empty-string,
+                    publicKey: non-empty-string,
+                    authToken: non-empty-string,
+                    contentEncoding: "aesgcm" | "aes128gcm",
+                }
+            EOT,
```
dtdesign, line 44:
```suggestion
                EOT,
```

Indentation matters here. Currently all lines are intended by four additional spaces, because PHP only trims whitespace that aligns with the identation of the ending marker.

wcfsetup/install/files/lib/action/RegisterServiceWorkerAction.class.php
```diff
@@ -0,0 +1,82 @@
...
+final class RegisterServiceWorkerAction implements RequestHandlerInterface
+{
+    #[\Override]
+    public function handle(ServerRequestInterface $request): ResponseInterface
+    {
+        if ($request->getMethod() !== 'POST') {
+            return new TextResponse('Unsupported', 400);
+        }
+        if (!WCF::getUser()->userID) {
+            // Notification not sent to guests
```
dtdesign, line 31:
```suggestion
            // Notifications are not supported for guests.
```

wcfsetup/install/files/lib/action/RegisterServiceWorkerAction.class.php
```diff
@@ -0,0 +1,82 @@
...
+        $serviceWorkerList = new ServiceWorkerList();
+        $serviceWorkerList->getConditionBuilder()->add('userID = ?', [WCF::getUser()->userID]);
+        if ($parameters["remove"]) {
+            $serviceWorkerList->getConditionBuilder()->add('endpoint = ?', [$parameters['endpoint']]);
+            $serviceWorkerList->getConditionBuilder()->add('publicKey = ?', [$parameters['publicKey']]);
+            $serviceWorkerList->getConditionBuilder()->add('authToken = ?', [$parameters['authToken']]);
+            $serviceWorkerList->readObjectIDs();
+            ServiceWorkerEditor::deleteAll($serviceWorkerList->getObjectIDs());
+
+            return new EmptyResponse(204);
```
dtdesign, line 55:
```suggestion
            return new EmptyResponse();
```

Please stop doing that, `204` is the default value.

wcfsetup/install/files/lib/action/RegisterServiceWorkerAction.class.php
```diff
@@ -0,0 +1,82 @@
...
+        foreach ($serviceWorkerList as $serviceWorker) {
+            if ($serviceWorker->endpoint === $parameters['endpoint']) {
+                // Update existing service worker
+                $editor = new ServiceWorkerEditor($serviceWorker);
+                $editor->update([
+                    'publicKey' => $parameters['publicKey'],
+                    'authToken' => $parameters['authToken'],
+                    'contentEncoding' => $parameters['contentEncoding'],
+                ]);
+                return new EmptyResponse(204);
```
dtdesign, line 69:
```suggestion
                return new EmptyResponse();
```

wcfsetup/install/files/lib/action/RegisterServiceWorkerAction.class.php
```diff
@@ -0,0 +1,82 @@
...
+        }
+        ServiceWorkerEditor::fastCreate([
+            'userID' => WCF::getUser()->userID,
+            'endpoint' => $parameters['endpoint'],
+            'publicKey' => $parameters['publicKey'],
+            'authToken' => $parameters['authToken'],
+            'contentEncoding' => $parameters['contentEncoding'],
+        ]);
+
+        return new EmptyResponse(204);
```
dtdesign, line 80:
```suggestion
        return new EmptyResponse();
```

wcfsetup/install/files/lib/data/service/worker/ServiceWorker.class.php
```diff
@@ -0,0 +1,34 @@
...
+ */
+final class ServiceWorker extends DatabaseObject
+{
+    public const CONTENT_ENCODING_AESGCM = 'aesgcm';
+    public const CONTENT_ENCODING_AES128GCM = 'aes128gcm';
+
+    /**
+     * Parses the endpoint and returns the scheme and host.
+     *
+     * @return string
```
dtdesign, line 28:
```suggestion
```

wcfsetup/setup/db/install.sql
```diff
@@ -1255,6 +1255,17 @@ CREATE TABLE wcf1_search_keyword (
 	KEY (searches, lastSearchTime)
 );
 
+DROP TABLE IF EXISTS wcf1_service_worker;
+CREATE TABLE wcf1_service_worker (
+	workerID INT(10) NOT NULL AUTO_INCREMENT PRIMARY KEY,
+	userID INT(10) NOT NULL,
+	endpoint TEXT NOT NULL,
+	publicKey VARCHAR(88) DEFAULT '' NOT NULL,
+	authToken VARCHAR(24) DEFAULT '' NOT NULL,
```
dtdesign, line 1264:
```suggestion
	publicKey VARCHAR(88) NOT NULL,
	authToken VARCHAR(24) NOT NULL,
```

That default value makes no sense, these fields must never be empty.

wcfsetup/setup/db/install.sql
```diff
@@ -1255,6 +1255,17 @@ CREATE TABLE wcf1_search_keyword (
...
+	contentEncoding ENUM ('aes128gcm', 'aesgcm') DEFAULT 'aes128gcm' NOT NULL,
```
dtdesign, line 1265:
```suggestion
	contentEncoding VARCHAR(40) NOT NULL,
```

Do not use `ENUM` in SQL, it’s a dumpster fire.

dtdesign (changes requested):

wcfsetup/install/files/lib/system/service/worker/Util.class.php
```diff
@@ -0,0 +1,79 @@
...
+        if (Binary::safeSubstr($data, 0, 2) !== '04') {
+            throw new \InvalidArgumentException('Invalid public key format');
+        }
+        $data = Binary::safeSubstr($data, 2);
+        $dataLength = Binary::safeStrlen($data);
+
+        return [
+            'x' => \hex2bin(Binary::safeSubstr($data, 0, $dataLength / 2)),
+            'y' => \hex2bin(Binary::safeSubstr($data, $dataLength / 2)),
+        ];
```
TimWolla, line 35: I don't understand the purpose of the hexadecimal encoding here. I believe the logic would be much simpler and easier to follow if it directly dealt with the binary data.

Side note: `Binary::safe*()` is not required here. The misfeature this protects against was removed in PHP 8.0.

Untested:

```suggestion
        if ($data[0] !== "\x04") {
            throw new \InvalidArgumentException('Invalid public key format');
        }
        $data = \substr($data, 1);
        $dataLength = \strlen($data);
        if ($dataLength !== /* whatever length is expected */) {
            throw new \InvalidArgumentException('Invalid public key format');
        }
        
        [$x, $y] = \str_split($data, \intdiv($dataLength, 2));

        return [
            'x' => $x,
            'y' => $y,
        ];
```

wcfsetup/install/files/lib/system/service/worker/Util.class.php
```diff
@@ -0,0 +1,79 @@
...
+     *
+     * @param string $x
+     * @param string $y
+     * @return string
+     */
+    public static function serializePublicKey(string $x, string $y): string
+    {
+        $hexString = '04';
+        $hexString .= \str_pad(\bin2hex(Base64UrlSafe::decode($x, true)), 64, '0', STR_PAD_LEFT);
+        return $hexString . \str_pad(\bin2hex(Base64UrlSafe::decode($y, true)), 64, '0', STR_PAD_LEFT);
```
TimWolla, line 49: Likewise this probably should return the binary representation, as `ServiceWorkerHandler` decodes the hexadecimal representation to binary anyway. Cryptography is much simpler when dealing with the binary representation.

wcfsetup/install/files/lib/system/service/worker/ServiceWorkerHandler.class.php
```diff
@@ -0,0 +1,102 @@
...
+
+    /**
+     * Send the given payload to the service worker.
+     *
+     * @param ServiceWorker $serviceWorker
+     * @param string $payload
+     */
+    public function sendToServiceWorker(ServiceWorker $serviceWorker, string $payload): void
+    {
+        if (Binary::safeStrlen($payload) > self::MAX_PAYLOAD_LENGTH) {
```
TimWolla, line 66:
```suggestion
        if (\strlen($payload) > self::MAX_PAYLOAD_LENGTH) {
```
(same below, see the note in the other comment for the reasoning)

wcfsetup/install/files/lib/system/service/worker/ServiceWorkerHandler.class.php
```diff
@@ -0,0 +1,102 @@
...
+        $jwk = JWKFactory::createECKey(Encryption::CURVE_ALGORITHM);
+        $binaryPublicKey = \hex2bin(Util::serializePublicKey($jwk->get('x'), $jwk->get('y')));
+        $binaryPrivateKey = \hex2bin(
+            \str_pad(
+                \bin2hex(Base64UrlSafe::decode($jwk->get('d'), true)),
+                2 * VAPID::PRIVATE_KEY_LENGTH,
+                '0',
+                STR_PAD_LEFT
+            )
+        );
```
TimWolla, line 51: Untested:

```suggestion
        $binaryPrivateKey = \str_pad(
            Base64UrlSafe::decode($jwk->get('d'), true),
            VAPID::PRIVATE_KEY_LENGTH,
            "\x00",
            STR_PAD_LEFT,
        );
```

Wondering though: Doesn't the JWK class return the padded key by itself?

wcfsetup/install/files/lib/system/service/worker/Encryption.class.php
```diff
@@ -0,0 +1,187 @@
...
+        string $clientPublicKey,
+        string $serverPublicKey,
+        string $contentEncoding
+    ): ?string {
+        if ($contentEncoding === ServiceWorker::CONTENT_ENCODING_AES128GCM) {
+            return null;
+        }
+        \assert(Binary::safeStrlen($clientPublicKey) === 65);
+
+        $len = "\0A"; // 65 as Uint16BE
```
TimWolla, line 160:
```suggestion
        $len = \pack('n', 65);
```

1. Not using hexadecimal escape sequences is incredibly confusing. This looked like `\x0A` at a first glance.
2. Using `pack()` makes the intent clear even without a comment.

wcfsetup/install/files/lib/system/service/worker/Encryption.class.php
```diff
@@ -0,0 +1,187 @@
...
+    }
+
+    private static function getIKM(
+        ServiceWorker $serviceWorker,
+        #[\SensitiveParameter] string $sharedSecret,
+        string $userPublicKey,
+        string $newPublicKey
+    ): string {
+        if ($serviceWorker->contentEncoding === ServiceWorker::CONTENT_ENCODING_AESGCM) {
+            $info = "Content-Encoding: auth\0";
```
TimWolla, line 172:
```suggestion
            $info = "Content-Encoding: auth\x00";
```
same below (and above).

wcfsetup/install/files/lib/system/service/worker/Encryption.class.php
```diff
@@ -0,0 +1,187 @@
...
+
+    private static function getEncryptionContentCodingHeader(
+        string $contentEncoding,
+        string $salt,
+        string $publicKey
+    ): string {
+        if ($contentEncoding === ServiceWorker::CONTENT_ENCODING_AES128GCM) {
+            return $salt
+                . \pack('N*', 4096)
+                . \pack('C*', Binary::safeStrlen($publicKey))
```
TimWolla, line 138: These two `pack` calls and especially the second one look fishy to me.

wcfsetup/install/files/lib/system/service/worker/Encryption.class.php
```diff
@@ -0,0 +1,187 @@
...
+     * Return the encrypted payload and necessary headers to send a push message to the push-service.
+     * {@link https://www.rfc-editor.org/rfc/rfc8291}
+     *
+     * @param ServiceWorker $serviceWorker
+     * @param string $payload
+     * @param array $headers to this array the encryption headers will be added
+     * @return string the encrypted payload
+     */
+    public static function encrypt(ServiceWorker $serviceWorker, string $payload, array &$headers): string
+    {
```
TimWolla, line 34: As the `payload` is going to be encrypted, it should be marked as sensitive:
```suggestion
    public static function encrypt(
        ServiceWorker $serviceWorker,
        #[\SensitiveParameter]
        string $payload,
        array &$headers
    ): string {
```

wcfsetup/install/files/lib/system/service/worker/VAPID.class.php
```diff
@@ -0,0 +1,73 @@
...
+        ['x' => $x, 'y' => $y] = Util::unserializePublicKey($rawPublicKey);
+
+        $header = [
+            'typ' => 'JWT',
+            'alg' => 'ES256',
+        ];
+        $payload = JSON::encode([
+            'aud' => $serviceWorker->getEndpoint(),
+            'exp' => TIME_NOW + 43200, // 12h
+            'sub' => "mailto:" . MAIL_ADMIN_ADDRESS,
```
TimWolla, line 48: Should the email address be `rawurlencode`d? Thinking about email addresses such as `admin+myforum@example.com`.

TimWolla (commented):
It also might make sense to drop the reverted dependency-related commits entirely, for two reasons:

1. Less bloat for the repository, because they will not be part of `master` after being merged.
2. Less garbage for the fileDelete.xml generation script to work through. See 04f17cd5859f20f9c976c608b0f3ca2c06dd21c3. That script will add anything to fileDelete.xml that was previously part of the `master` branch, but no longer is part of it.

wcfsetup/install/files/lib/system/service/worker/VAPID.class.php
Cyperghost: This is not needed. The value `mailto:admin+myforum@example.com` works fine.

wcfsetup/install/files/lib/system/service/worker/Encryption.class.php
Cyperghost: But this is exactly what is needed.
> salt (16) | rs (4) | idlen (1) | keyid (idlen)

https://datatracker.ietf.org/doc/html/rfc8188#section-2.1

wcfsetup/install/files/lib/system/service/worker/ServiceWorkerHandler.class.php
```diff
@@ -0,0 +1,116 @@
...
+            'Content-Encoding' => $serviceWorker->contentEncoding,
+            'TTL' => ServiceWorkerHandler::TTL,
+        ];
+        $content = Encryption::encrypt(
+            $serviceWorker,
+            $payload,
+            $headers
+        );
+        $headers['Content-Length'] = \mb_strlen($content, '8bit');
+        VAPID::addHeader($serviceWorker, $headers);
```
TimWolla, line 97: Small nit: I would generally recommend defining headers in all-lowercase in new code, as that's the canonical representation anyway (but also see my other comments first).

```suggestion
        $headers = [
            'content-type' => 'application/octet-stream',
            'content-encoding' => $serviceWorker->contentEncoding,
            'ttl' => ServiceWorkerHandler::TTL,
        ];
        $content = Encryption::encrypt(
            $serviceWorker,
            $payload,
            $headers
        );
        $headers['content-length'] = \mb_strlen($content, '8bit');
        VAPID::addHeader($serviceWorker, $headers);
```

wcfsetup/install/files/lib/system/service/worker/VAPID.class.php
```diff
@@ -0,0 +1,72 @@
...
+        $jwt = $compactSerializer->serialize($jws, 0);
+
+        if ($serviceWorker->contentEncoding === ServiceWorker::CONTENT_ENCODING_AESGCM) {
+            $headers['Authorization'] = 'WebPush ' . $jwt;
+            $headers['Crypto-Key'] .= ';p256ecdsa=' . SERVICE_WORKER_PUBLIC_KEY;
+        } elseif ($serviceWorker->contentEncoding === ServiceWorker::CONTENT_ENCODING_AES128GCM) {
+            $headers['Authorization'] = 'vapid t=' . $jwt . ', k=' . SERVICE_WORKER_PUBLIC_KEY;
+        } else {
+            throw new \InvalidArgumentException('Invalid content encoding: "' . $serviceWorker->contentEncoding . '"');
+        }
```
TimWolla, line 70: The way the headers are merged is potentially dangerous, because casing is relevant: You might get duplicate headers if the caller writes the name differently. Also out parameters (i.e. reference parameters) are pretty intransparent.

This is especially relevant for the `crypto-key` which is combined in two entirely different classes, which makes it hard for me to follow the data-flow.

Instead of passing the header array, I would suggest passing the `RequestInterface` and using `->withHeader('authorization', …)` instead. The class will then take care of properly handling duplicates. Note of caution: The RequestInterface class is immutable, thus you would need to return the updated object, similar to `HeaderUtil::withNoCacheHeaders()`.

This code with possibly also benefit from making use of either string interpolation or `sprintf()`.

wcfsetup/install/files/lib/system/service/worker/Encryption.class.php
```diff
@@ -0,0 +1,194 @@
...
+
+    private static function createInfo(string $type, ?string $context, string $contentEncoding)
+    {
+        if ($contentEncoding === ServiceWorker::CONTENT_ENCODING_AESGCM) {
+            \assert($context !== null);
+            \assert(\mb_strlen($context, '8bit') === 135);
+
+            return 'Content-Encoding: ' . $type . "\x00" . Encryption::CURVE_ALGORITHM . $context;
+        } elseif ($contentEncoding === ServiceWorker::CONTENT_ENCODING_AES128GCM) {
+            return 'Content-Encoding: ' . $type . "\x00";
```
TimWolla, line 129: I believe one of the following would be more readable.
```suggestion
            return "Content-Encoding: {$type}\x00";
```
```suggestion
            return \sprintf("Content-Encoding: %s\x00", $type);
```

wcfsetup/install/files/lib/system/service/worker/Encryption.class.php
```diff
@@ -0,0 +1,194 @@
...
+        $len = \pack('n', 65);
+
+        return "\x00" . $len . $clientPublicKey . $len . $serverPublicKey;
+    }
+
+    private static function getIKM(
+        ServiceWorker $serviceWorker,
+        #[\SensitiveParameter] string $sharedSecret,
+        string $userPublicKey,
+        string $newPublicKey
```
TimWolla, line 176: I feel it would help to understand the code if instead of passing the serialized key, the `JWK` object would be passed, because the amount of logic in `encrypt()` would be reduced and type safety would be improved.

wcfsetup/install/files/lib/system/service/worker/Encryption.class.php
```diff
@@ -0,0 +1,194 @@
...
+        string $contentEncoding
+    ): ?string {
+        if ($contentEncoding === ServiceWorker::CONTENT_ENCODING_AES128GCM) {
+            return null;
+        }
+        \assert(\mb_strlen($clientPublicKey, '8bit') === VAPID::PUBLIC_KEY_LENGTH);
+
+        $len = \pack('n', 65);
+
+        return "\x00" . $len . $clientPublicKey . $len . $serverPublicKey;
```
TimWolla, line 169: Looking at this once more, I believe the following would be more resistant to introducing bugs in a future refactoring, because it actually computes the string length. I've also taken the liberty to merge this into a single pack call, similarly to the session encoder: https://github.com/WoltLab/WCF/blob/e745d4785d7054655d0f036c4b9c6471ef0bd69c/wcfsetup/install/files/lib/system/session/SessionHandler.class.php#L289-L294

```suggestion
        return \pack(
            'CnA*nA*',
            0,
            \strlen($clientPublicKey),
            $clientPublicKey,
            \strlen($serverPublicKey),
            $serverPublicKey,
        );
```

wcfsetup/install/files/lib/system/service/worker/Encryption.class.php
TimWolla: I see. The asterisk was confusing to me (because there is only a single parameter) and I expected more than a single byte for the length. I'll make a suggestion to make this easier to follow.

wcfsetup/install/files/lib/system/service/worker/Encryption.class.php
```diff
@@ -0,0 +1,194 @@
...
+        int $length,
+        string $salt,
+        string $publicKey
+    ): string {
+        if ($contentEncoding === ServiceWorker::CONTENT_ENCODING_AES128GCM) {
+            /** {@link https://datatracker.ietf.org/doc/html/rfc8188#section-2.1} */
+            return $salt
+                . \pack('N*', $length)
+                . \pack('C*', \mb_strlen($publicKey, '8bit'))
+                . $publicKey;
```
TimWolla, line 146:
```suggestion
            return \pack(
                'A*NCA*,
                $salt,
                $length,
                \strlen($publicKey)
                $publicKey
            );
```

Just `\strlen()` is fine with PHP 8.0+.

wcfsetup/install/files/lib/system/service/worker/Encryption.class.php
```diff
@@ -0,0 +1,194 @@
...
+        $length = \mb_strlen($payload, '8bit');
+        $paddingLength = ServiceWorkerHandler::MAX_PAYLOAD_LENGTH - $length;
+
+        if ($contentEncoding === ServiceWorker::CONTENT_ENCODING_AES128GCM) {
+            return \str_pad($payload . "\x02", $paddingLength + $length, "\x00", STR_PAD_RIGHT);
+        } elseif ($contentEncoding === ServiceWorker::CONTENT_ENCODING_AESGCM) {
+            return \pack('n*', $paddingLength) . \str_pad($payload, $paddingLength + $length, "\x00", STR_PAD_LEFT);
+        } else {
+            throw new \RuntimeException('Unknown content encoding "' . $contentEncoding . '"');
+        }
```
TimWolla, line 118: I can't help but wonder if making the two possible content encodings an enum and adding a padding method would make this more readable. Something like this. Also note the use of `str_repeat()`:

```php
enum Encoding {
  case Aes128Gcm;
  case AesGcm;

  public function pad(#[\SensitiveParameter] string $payload) {
    $length = \mb_strlen($payload, '8bit');
    $paddingLength = ServiceWorkerHandler::MAX_PAYLOAD_LENGTH - $length;
    $padding = \str_repeat("\x00", $paddingLength);

    return match ($this) {
      self::Aes128Gcm => $payload . "\x02" . $padding,
      self::AesGcm => \pack('n', $paddingLength) . $padding . $payload,
    };
  }
}

File: wcfsetup/install/files/lib/system/service/worker/Util.class.php [hunk 28]
```diff
@@ -0,0 +1,80 @@
...
+    /**
+     * Creates a JWK with EC P-256 curve.
+     *
+     * @param string|null $x non-encode base64 x coordinate
+     * @param string|null $y non-encoded base64 y coordinate
+     * @param string|null $d base64 encoded private key
+     *
+     * @return JWK
+     */
+    public static function createJWK(?string $x = null, string $y = null, #[\SensitiveParameter] ?string $d = null): JWK
```
TimWolla, line 62: see: https://wiki.php.net/rfc/deprecate-implicitly-nullable-types
```suggestion
    public static function createJWK(?string $x = null, ?string $y = null, #[\SensitiveParameter] ?string $d = null): JWK
```

I wonder though: Are the `= null` default values even meaningful? Having an `x` without a `y` does not really make sense to me.

wcfsetup/install/files/lib/system/user/notification/UserNotificationHandler.class.php
```diff
@@ -303,7 +304,28 @@ public function fireEvent(
...
+                    continue;
+                }
+                \assert($notification instanceof UserNotification);
+                foreach ($workerIDs as $workerID) {
+                    $jobs[] = new ServiceWorkerDeliveryBackgroundJob($workerID, $notification->notificationID);
+                }
+            }
+            if ($jobs !== []) {
+                BackgroundQueueHandler::getInstance()->enqueueIn($jobs);
+            }
```
TimWolla, line 328: it's legal to pass an empty array.
```suggestion
            BackgroundQueueHandler::getInstance()->enqueueIn($jobs);
```

TimWolla (commented):
Some more remarks regarding the cryptography. I did not test any of the code snippets, so please disregard the stuff that does not make sense to you / that you disagree with. It's already much better with the `hex2bin`/`bin2hex` gone.

wcfsetup/install/files/lib/system/service/worker/Encryption.class.php
Cyperghost: However, only the serialised key is needed, not the JWK object. I do not think it is helpful to serialise the JWK object multiple times in different functions.

ts/WoltLabSuite/Core/Notification/Handler.ts
```diff
@@ -90,18 +99,18 @@ class NotificationHandler {
...
     if (inactiveMinutes < 15) {
-      return 5;
+      return 1;
     } else if (inactiveMinutes < 30) {
-      return 10;
+      return 1;
     }
 
-    return 15;
+    return 1;
```
dtdesign, line 113: What is the purpose of this change?

wcfsetup/install/files/lib/system/service/worker/Encoding.class.php
```diff
@@ -0,0 +1,127 @@
...
+        string $userPublicKey,
+        string $newPublicKey
+    ): string {
+        if ($this === Encoding::AesGcm) {
+            $info = "Content-Encoding: auth\x00";
+        } elseif ($this === Encoding::Aes128Gcm) {
+            $info = "WebPush: info\x00{$userPublicKey}{$newPublicKey}";
+        } else {
+            throw new \LogicException('Unreachable');
+        }
```
TimWolla, line 109: This would give you the error handling for free:
```suggestion
        $info = match ($this) {
            self::AesGcm => "Content-Encoding: auth\x00",
            self::Aes128Gcm => "WebPush: info\x00{$userPublicKey}{$newPublicKey}",
        };
```

wcfsetup/install/files/lib/system/service/worker/Util.class.php
```diff
@@ -0,0 +1,110 @@
...
+     */
+    public static function updateCryptoKeyHeader(
+        RequestInterface $request,
+        string $name,
+        string $value
+    ): RequestInterface {
+        return $request->withHeader('crypto-key', \implode(';', [
+            ...$request->getHeader('crypto-key'),
+            \sprintf('%s=%s', $name, $value)
+        ]));
```
TimWolla, line 108: This is not correct from an HTTP semantics perspective: If a header is duplicated, this means that they will be implicitly concatenated with a comma.

You might want to check that `$request->getHeader('crypto-key')` returns a single header only, somewhat like this:

```
$headers = $request->getHeader('crypto-key');
if (\count($headers) !== 1) {
    throw new \InvalidArgumentException('...');
}

return $request->withHeader('crypto-key', \sprintf('%s; %s=%s', $headers[0], $name, $value));
```

TimWolla (commented):
I still don't quite follow the entire cryptographic flow, but that might be because I didn't actually read the underlying specification. That said, it looks reasonable to me now.

You might want to consider running `phpcbf` and `php-cs-fixer` on `lib/system/service/`. Some of the spacing (with regard to empty lines) doesn't match the code style that is generally used and made it a little jarring while attempting to grok the code.

wcfsetup/install/files/lib/system/service/worker/Encryption.class.php
TimWolla: Cryptographic code is hard to understand even for an experienced developer. I would recommend to write the code as misuse-resistant as possible. Leveraging the type system to assist in that is probably a good thing.

Depending on the architecture this might lead to needing to serialize the JWK multiple times, but the serialization is not really expensive, the code is not sitting in a hot path and the actual cryptographic operations and IO will likely take up the bulk of the time.

ts/WoltLabSuite/Core/Notification/Handler.ts
Cyperghost: This should not be changed and is for testing purposes only. Was a wrong commit.

wcfsetup/install/files/lib/system/api/composer/installed.php
```diff
@@ -133,6 +151,24 @@
...
         ),
+        'paragonie/random_compat' => array(
+            'pretty_version' => 'v9.99.100',
+            'version' => '9.99.100.0',
+            'reference' => '996434e5492cb4c3edcb9168db6fbb1359ef965a',
+            'type' => 'library',
+            'install_path' => __DIR__ . '/../paragonie/random_compat',
+            'aliases' => array(),
+            'dev_requirement' => false,
+        ),
```
TimWolla, line 162: Should be added to the replace section.

wcfsetup/install/files/lib/system/api/composer/installed.php
```diff
@@ -297,6 +357,21 @@
...
+        'symfony/polyfill-mbstring' => array(
+            'pretty_version' => 'v1.29.0',
+            'version' => '1.29.0.0',
+            'reference' => '9773676c8a1bb1f8d4340a62efe641cf76eda7ec',
+            'type' => 'library',
+            'install_path' => __DIR__ . '/../symfony/polyfill-mbstring',
+            'aliases' => array(),
+            'dev_requirement' => false,
+        ),
```
TimWolla, line 374: Should be added to the replace section.

Cyperghost:
Waiting for composer packages
- `minishlink/web-push` version `9.0.0`
- `web-token/jwt-library` version `4.0.0`

dtdesign:
> minishlink/web-push

https://github.com/web-push-libs/web-push-php/releases has v9.0.0-rc2 out for a few days already and we can possibly expect a release to arrive in July.

> web-token/jwt-library

The version 4.0 appears to be still under development as of now. However, its `composer.json` [requires a minimum PHP version of 8.3](https://github.com/web-token/jwt-library/blob/4.0.x/composer.json) which exceeds our current minimum version of PHP 8.1. @Cyperghost can we somehow backport the necessary changes in the v3.4 tree to work around this?

dtdesign (approved):

PR #5814, opened 2024-02-21
Unique background job

Cyperghost:
Background jobs can be marked as "unique".
These background jobs can only exist once in the database to avoid blocking the background queue with too many jobs.

These jobs are automatically requeued when there is more work to do.

### DIFF
--- a/wcfsetup/install/files/lib/system/background/job/AbstractUniqueBackgroundJob.class.php
+++ b/wcfsetup/install/files/lib/system/background/job/AbstractUniqueBackgroundJob.class.php
@@ -2,8 +2,6 @@
 
 namespace wcf\system\background\job;
 
-use wcf\system\background\BackgroundQueueHandler;
-
 /**
  * @author      Olaf Braun
  * @copyright   2001-2024 WoltLab GmbH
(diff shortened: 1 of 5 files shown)


wcfsetup/install/files/lib/system/background/job/AbstractUniqueBackgroundJob.class.php
```diff
@@ -0,0 +1,62 @@
...
+        return static::class;
+    }
+
+    /**
+     * Returns a new instance of this job to be queued again.
+     * This will reset the fail counter.
+     *
+     * @return AbstractUniqueBackgroundJob
+     */
+    public function newInstance(): AbstractUniqueBackgroundJob
```
dtdesign, line 37:
```suggestion
     */
    public function newInstance(): static
```

wcfsetup/install/files/lib/system/background/job/AbstractUniqueBackgroundJob.class.php
```diff
@@ -0,0 +1,62 @@
...
+{
+    /**
+     * @inheritDoc
+     */
+    final public const MAX_FAILURES = 0;
+
+    /**
+     * Returns a unique identifier for this job.
+     *
+     * @return string
```
dtdesign, line 24:
```suggestion
```

wcfsetup/install/files/lib/system/background/job/AbstractUniqueBackgroundJob.class.php
```diff
@@ -0,0 +1,62 @@
...
+     */
+    public function newInstance(): AbstractUniqueBackgroundJob
+    {
+        return new static();
+    }
+
+    /**
+     * Returns whether this job should be queued again because it has more to do.
+     *
+     * @return bool
```
dtdesign, line 45:
```suggestion
```

dtdesign (changes requested):
Otherwise LGTM

dtdesign:
On second though, the implementation of `AbstractUniqueBackgroundQueue::identifier()` makes no sense. This will always be a semi-static string because `queueAgain()` will initialize an empty object thus there is no available data for any kind of custom identifier.

PR #5828, opened 2024-03-01
Move reCAPTCHA widget overlay for dialogs to `pageOverlayContainer`

Cyperghost:
See https://www.woltlab.com/community/thread/304714-recaptcha-widget-erscheint-hinter-dem-gastkommentar-dialog/

### DIFF
--- a/ts/WoltLabSuite/Core/Bootstrap.ts
+++ b/ts/WoltLabSuite/Core/Bootstrap.ts
@@ -178,7 +178,7 @@ export function setup(options: BoostrapOptions): void {
           continue;
         }
 
-        if (node.querySelectorAll(".g-recaptcha-bubble-arrow").length === 0) {
+        if (node.querySelector(".g-recaptcha-bubble-arrow") === null) {
           return;
         }
 
--- a/ts/WoltLabSuite/Core/Bootstrap.ts
+++ b/ts/WoltLabSuite/Core/Bootstrap.ts
@@ -179,21 +179,21 @@ export function setup(options: BoostrapOptions): void {
         }
 
         if (node.querySelector(".g-recaptcha-bubble-arrow") === null) {
-          return;
+          continue;
         }
 
         const iframe = node.querySelector("iframe");
         if (!iframe) {
-          return;
+          continue;
         }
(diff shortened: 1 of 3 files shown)


ts/WoltLabSuite/Core/Bootstrap.ts
```diff
@@ -168,4 +168,40 @@ export function setup(options: BoostrapOptions): void {
...
+  // Move the reCAPTCHA widget overlay to the `pageOverlayContainer`
+  // when widget form elements are placed in a dialog.
+  const observer = new MutationObserver((mutations) => {
+    for (const mutation of mutations) {
+      for (const node of mutation.addedNodes) {
+        if (!(node instanceof HTMLElement)) {
+          continue;
+        }
+
+        if (node.querySelectorAll(".g-recaptcha-bubble-arrow").length === 0) {
```
dtdesign, line 181:
```suggestion
        if (node.querySelector(".g-recaptcha-bubble-arrow") === null) {
```

This is a bit more efficient because it will stop searching after the first match.

ts/WoltLabSuite/Core/Bootstrap.ts
```diff
@@ -168,4 +168,40 @@ export function setup(options: BoostrapOptions): void {
...
+          return;
```
dtdesign, line 182:
```suggestion
          continue;
```

The same applies to the other `return` below, because each invocation of the observer can contain multiple mutations.

dtdesign (changes requested):

dtdesign (approved):

Krymonota:
To be honest, I haven't spent any time looking into the problem yet. But am I right in assuming that other captcha plugins (I can think of at least 3 plugins including mine) must implement an analogous solution because there's no way the problem could be fixed centrally for all captchas in WSC 6.0?

dtdesign:
@Krymonota Not necessarily, this is only an issue if elements are placed inside the document’s body instead of using the page overlay container. If your code uses the page overlay container for absolute / fixed positioning, then no change is required at all, because that container is implicitly adopted and released by dialogs.

PR #5835, opened 2024-03-06
Custom event to update the counter for user menu entries

Cyperghost:


dtdesign (approved):

dtdesign:
@Cyperghost This should target 6.1!

PR #5841, opened 2024-03-09
PoC: REST/RPC API

### LINKED ISSUE #5840: Draft: REST/RPC API
We have already introduced support for PSR-7 and PSR-15 with past releases and started implementing PSR-7 based controllers with the recently released version 6.0. Following our incremental approach for introducing new features to the framework, we’re now working towards the next step by implementing a baseline interface for a REST / RPC API.

We do not strive to offer a general and/or complete API at this point but instead migrate existing PSR-7 controllers as well as adding new routes. This helps us to evaluate the effectiveness of the API and paves the way for a gradual transition of existing HTTP-based APIs.

# Starting With RPC Only

The API should use a single controller to receive and dispatch requests, forwarding controllers to previously registered routes. This uniquely allows us to support both a RPC (cookie based authentication) and REST API (token based authentication) without having to make a distinction in the actual route implementations.

We also do not intend to implement a REST API at this point but to focus entirely on the RPC part but allowing the REST API to be implemented later. The centralized API endpoint allows us to add this without requiring any changes to the actual routes.

# Predictable API

## Idempotency Whenever Reasonably Possible

Some endpoints will perform a specific action, such as following a user or reacting to a message. These endpoints SHOULD be idempotent whenever possible, treating the request to be the “should” state.

For example, following a user for the first time should produce the same response as attempting to follow them again while still following. Reacting to a message that has already been reacted to using a difference reaction should implicitly revoke the previous reaction. If the user has already reacted using the same reaction then no change should be made. However, in both cases the response should be indistinguishable from reacting to a message for the first time.

## HTTP Verbs

We will only support three basic verbs: `GET`, `POST` and `DELETE`.

Some implementations also make use of `PATCH` and `PUT`, but this would make it much more complex and adds little benefit plus its quite verbose since the verb implies the semantics of the endpoint. The same can be achieved by using distinct endpoints and using `POST` instead. Plus historically the support for `PATCH` and `PUT` was rather poor.

## Response Format

The PSR messaging interface allows for a host of useful response types, such as `204 No Content`, returning a plain “HTML” response or any format that is suitable. While this is more efficient, it also makes the API much more complicated and less predictable, requiring different response handling based on the endpoint being targeted.

## HTTP Status Codes

| Code | Name | Meaning |
| --- | --- | --- |
| 200 | OK | The request was successful. |
| 400 | Bad Request | The provided parameters are insufficient. |
| 403 | Permission Denied | The callee is not authorized to execute this request. | 
| 404 | Not Found | The requested endpoint does not exist. |
| 405 | Method Not Allowed | Using any HTTP verb other than `GET`, `POST` or `DELETE`. |
| 500 | Internal Server Error | The server failed to process the request. |
| 503 | Service Unavailable | The API is currently unavailable. |

## Error Format

Whenever a request is rejected with the error code 400, the response will match the following data structure:

```ts
type ErrorResponse = {
  type: "api_error" | "invalid_request_error";
  code: string;
  message: string;
  param: string;
};
```

A typical response could look like this:

```json
{
  "type": "invalid_request_error",
  "code": "value_too_short",
  "message": "",
  "param": "username"
}
```

### `type`

The error type is used to tell apart issues caused by the request being made by the client or by unexpected errors taking place on the server side processing.

### `code`

The error code is all lowercase and using snake case to describe the type of error, for example, `missing_api_key`.

### `message`

The message can be empty but when present contains a non-localized string explaining the cause of the error with the intention of assisting a developer to resolve the issue.

You MUST NOT present this message to the end user.

### `param`

Validation errors may refer to a specific parameter that has caused the request to be rejected. A common use case is to point to a specific input field, allowing for contextual error message presented to the user.

# Namespaces for Endpoints

Endpoints are defined using a strict rule set:

* The first path segment is a namespace that holds other objects.
* The namespaces `core`, `forum`, `blog`, `filebase` and `gallery` are reserved and MUST NOT be used by third parties.
* Segments MUST be lowercase ASCII and use the plural form.
* Parameters MAY appear starting with the third path segment and are defined with a leading colon.
* The name of a parameter MUST be unique within one endpoint.

## Placeholders

The route implementation uses [FastRoute](https://github.com/nikic/FastRoute) which supports named placeholders through the `{name}` syntax. Optionally, a validation pattern can be specified to further narrow down the valid value of the placeholder: `{id:\d+}`

## Examples

The endpoints below may or may not exist at any point and are only used for illustration purposes.

| Verb | Endpoint | Explanation |
| --- | --- | --- |
| `GET` | `/core/users` | Retrieve a list of users, may include query parameters to control the response. |
| `GET` | `/core/users/{id:\d+}` | Retrieve a user, may include query parameters to control the response. |
| `GET` | `/core/users/{id:\d+}/following` | Retrieve a list of users that the target user is following. |
| - | - | - |
| `POST` | `/core/users` | Creates a new user, parameters are always presented in the body. |
| `POST` | `/core/users/{id:\d+}` | (Partially) updates a user. |
| `POST` | `/core/moderationqueues/{id:\d+}/assign` | Assigns a user. |
| - | - | - |
| `DELETE` | `/forum/threads/{id:\d+}` | Deletes a thread. |

dtdesign:
Closes #5840


wcfsetup/install/files/lib/system/endpoint/controller/core/messages/MentionSuggestions.class.php
```diff
@@ -0,0 +1,112 @@
...
+/**
+ * Retrieves the list of users and groups that can be mentioned.
+ *
+ * @author Alexander Ebert
+ * @copyright 2001-2024 WoltLab GmbH
+ * @license GNU Lesser General Public License <http://opensource.org/licenses/lgpl-license.php>
+ * @since 6.1
+ */
+#[GetRequest('/core/messages/mentionsuggestions')]
+final class MentionSuggestions implements IController
```
BurntimeX, line 25: Rename to `GetMentionSuggestions`

BurntimeX (changes requested):

BurntimeX (approved):

PR #5850, opened 2024-03-18
Use language variable for `errorType` in PollManager

Cyperghost:
See https://www.woltlab.com/community/thread/305317-fehlende-sprachvariabel-bei-der-generierung-von-umfragen-in-der-vergangenheit/


dtdesign:
This change is bad, the properties of a `UserInputException` should not be phrases. If you need to handle a specific phrase suffix, like the “famous” handling of `empty`, that target that phrase in particular.

dtdesign (approved):
Please squash when merging

PR #5851, opened 2024-03-18
Add an empty upcast converter that handles custom bb-codes

Cyperghost:
Convert all BBCodes that are not from the original system to their plain value representation.
Also check whether it is a source code BB code

### DIFF
--- a/wcfsetup/install/files/lib/system/html/metacode/upcast/EmptyMetacodeUpcast.class.php
+++ b/wcfsetup/install/files/lib/system/html/metacode/upcast/EmptyMetacodeUpcast.class.php
@@ -3,7 +3,7 @@
 namespace wcf\system\html\metacode\upcast;
 
 /**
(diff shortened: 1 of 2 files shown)


wcfsetup/install/files/lib/system/html/metacode/upcast/EmptyMetacodeUpcast.class.php
```diff
@@ -0,0 +1,32 @@
+<?php
+
+namespace wcf\system\html\metacode\upcast;
+
+/**
+ * An empty implementation for metacode upcast, that always let the metacdoe convert to his plain text representation.
```
dtdesign, line 6:
```suggestion
 * Always forces the passed BBCode to be converted into its plain text representation.
```

dtdesign (changes requested):

dtdesign (approved):

PR #5852, opened 2024-03-20
Moderation notification report

### LINKED ISSUE #5281: Notification to moderators when content has been reported

Cyperghost:
Closes #5281

- Added new interface `ICommentPermissionManager `
- Deprecated the old `canModerate` function in `ICommentManager`


wcfsetup/install/files/lib/system/comment/manager/AbstractCommentManager.class.php
```diff
@@ -153,10 +154,14 @@ public function canDeleteResponse(CommentResponse $response)
     }
 
     /**
+     * @param UserProfile|null $user
```
dtdesign, line 157: Do not add PhpDoc that is the same as the parameter type, that’s simply redundant and at risk of getting out of sync. Also, the parameter should be typed in the interface so it makes even less sense.

wcfsetup/install/lang/de.xml
```diff
@@ -4389,6 +4389,10 @@ Dateianhänge:
...
 		<item name="wcf.moderation.report.changeJustifiedStatus.confirmMessage"><![CDATA[{if LANGUAGE_USE_INFORMAL_VARIANT}Willst du{else}Wollen Sie{/if} den „Berechtigt“-Status dieser Meldung wirklich ändern?]]></item>
+		<item name="wcf.moderation.report.notification.title"><![CDATA[Neue Meldung]]></item>
+		<item name="wcf.moderation.report.notification.message"><![CDATA[{if $author->userID}<strong>{$author}</strong>{else}Ein Gast{/if} hat <strong>{$moderationQueue->getTitle()}</strong>({lang}wcf.moderation.type.{@$moderationQueue->getObjectTypeName()}{/lang}) gemeldet.]]></item>
+		<item name="wcf.moderation.report.notification.mail.plaintext"><![CDATA[{if $author->userID}{$author}{else}Ein Gast{/if} hat {$moderationQueue->getTitle()} ({lang}wcf.moderation.type.{@$moderationQueue->getObjectTypeName()}{/lang}) gemeldet [URL:{$moderationQueue->getLink()}].]]></item>
+		<item name="wcf.moderation.report.notification.mail.html"><![CDATA[<p>{if $author->userID}{$author}{else}Ein Gast{/if} hat <a href="{$moderationQueue->getLink()}">{$moderationQueue->getTitle()}</a> ({lang}wcf.moderation.type.{@$moderationQueue->getObjectTypeName()}{/lang}) gemeldet.</p>]]></item>
```
dtdesign, line 4395: Please simplify those phrases by passing in simple variables for each value. For example, the first phrase could read this instead:

```smarty
{if $author}<strong>{$author}</strong>{else}Ein gast{/if} hat <strong>{$title}</strong> ({$typeName}) gemeldet.
```

This also makes it obvious that your phrase lacks a space between the title and the type name.

Also, do not use `{lang}` inside phrases ever.

wcfsetup/install/lang/en.xml
```diff
@@ -4341,6 +4341,10 @@ Attachments:
...
 		<item name="wcf.moderation.report.changeJustifiedStatus.confirmMessage"><![CDATA[Do you really want to change the “Justified” status of this report?]]></item>
+		<item name="wcf.moderation.report.notification.title"><![CDATA[New report]]></item>
+		<item name="wcf.moderation.report.notification.message"><![CDATA[{if $author->userID}<strong>{$author}</strong>{else}A guest{/if} has reported <strong>{$moderationQueue->getTitle()}</strong>({lang}wcf.moderation.type.{@$moderationQueue->getObjectTypeName()}{/lang}).]]></item>
+		<item name="wcf.moderation.report.notification.mail.plaintext"><![CDATA[{if $author->userID}{$author}{else}A guest{/if} has reported {$moderationQueue->getTitle()} ({lang}wcf.moderation.type.{@$moderationQueue->getObjectTypeName()}{/lang}) [URL:{$moderationQueue->getLink()}].]]></item>
+		<item name="wcf.moderation.report.notification.mail.html"><![CDATA[<p>{if $author->userID}{$author}{else}A guest{/if} has reported <a href="{$moderationQueue->getLink()}">{$moderationQueue->getTitle()}</a> ({lang}wcf.moderation.type.{@$moderationQueue->getObjectTypeName()}{/lang}):</p>]]></item>
```
dtdesign, line 4347: See the remarks on `de.xml`.

dtdesign (changes requested):

dtdesign (changes requested):
I don’t know how I should feel about this change.

You are completely ignoring the logic that exists in `canModerate()` which usually is far more complex than just checking the permission. Aside from allowing to check the permission for a different user, the new check is effectively more permissive than the existing `canModerate()` implementation. Doesn’t this mean that you would now be notifying users that don’t actually have access, because the permission itself is only one part of the story?

Second, why would you ever allow `SessionHandler` as a substitute for the permission check, that is far too broad. The method should never accept anything else but `UserProfile` for `$user`, just go the extra mile of fetching that object.

wcfsetup/install/files/lib/system/comment/manager/ICommentManager.class.php
```diff
@@ -132,6 +131,11 @@ public function getTitle($objectTypeID, $objectID, $isResponse = false);
      */
     public function isAccessible($objectID, $validateWritePermission = false);
 
+    /**
+     * Returns true if the user may read content identified by object type id and object id.
+     */
+    public function canViewObject(int $objectID, UserProfile $user): bool;
```
dtdesign, line 137: There is no default implementation for this method, because it cannot be provided meaningfully, thus is break of the backwards compatibility. You MUST use a new interface for this!

wcfsetup/install/files/lib/system/moderation/queue/AbstractCommentResponseModerationQueueHandler.class.php
```diff
@@ -163,4 +164,26 @@ public function removeContent(ModerationQueue $queue, $message)
             $responseAction->executeAction();
         }
     }
+
+    #[\Override]
+    public function isAffectedUser(ModerationQueue $queue, $userID)
+    {
+        if (!AbstractModerationQueueHandler::isAffectedUser($queue, $userID)) {
```
dtdesign, line 171:
```suggestion
        if (!parent::isAffectedUser($queue, $userID)) {
```

dtdesign (changes requested):
Please double-check if there are runtime caches available instead of fetching each object separately.

wcfsetup/install/files/lib/system/moderation/queue/AbstractCommentResponseModerationQueueHandler.class.php
Cyperghost: I cannot use this because `AbstractCommentResponseModerationQueueHandler` extends `AbstractCommentCommentModerationQueueHandler` and in the function `isAffectedUser` `$queue->objectID` is not a comment id it is the comment response id.

wcfsetup/install/files/lib/system/comment/manager/ICommentPermissionManager.class.php
```diff
@@ -0,0 +1,21 @@
...
+namespace wcf\system\comment\manager;
+
+use wcf\data\user\UserProfile;
+
+/**
+ * Interface for comment managers that provide permission checks.
+ *
+ * @author      Olaf Braun
+ * @copyright   2001-2024 WoltLab GmbH
+ * @license     GNU Lesser General Public License <http://opensource.org/licenses/lgpl-license.php>
```
dtdesign, line 12:
```suggestion
 * @license     GNU Lesser General Public License <http://opensource.org/licenses/lgpl-license.php>
 * @since 6.1
```

dtdesign (changes requested):
Otherwise LGTM

PR #5860, opened 2024-03-26
Modernization of the ACL list

### LINKED ISSUE #4715: Transmit ACL in JSON format
A large number of user groups can currently cause max_input_vars to be exceeded when saving the ACL.

ref https://www.woltlab.com/community/thread/288318-benutzergruppen-berechtigungen-werden-bei-53-gruppen-nicht-gespeichert/

### LINKED ISSUE #5236: Improve the ACL UI
Often the board permissions are configured unnecessarily complex, not making use of inheritance and uselessly specifying permissions when the group permission is already appropriate.

The ACL UI likely is at fault and should be improved:

- [x] Remove the “Full Access” checkbox: It's there and it's attractive to deny access to “Everyone” or “Guests”. However it's almost never the right choice, because it requires configuring all permissions for the groups that should have access instead of inheriting the group permissions.
- [x] Change the UI from two checkboxes (that cannot actually be checked both) to a real radio button with an explicit “Inherit” option that works as if no checkbox would be checked. This should make it clearer that permissions are inherited from somewhere else.

Cyperghost:
Closes #4715 
Closes #5236


dtdesign (approved):
Please keep in mind to update the referenced thread (in the issue).

PR #5883, opened 2024-04-16
Opt out rocket loader

### LINKED ISSUE #5872: Add a flag to script tags to opt out of Cloudflare’s Rocket Loader
Cloudflare’s Rocket Loader has been nothing but a pain for the past decade and in generally messes up a lot of things. It is possible to explicitly opt out of this loader by flagging script tags with [`data-cfasync="false"`](https://developers.cloudflare.com/speed/optimization/content/rocket-loader/ignore-javascripts/).

It looks like that we can add this flag on runtime in `HeaderUtil::parseOutput()` which already processes all script tags.

Cyperghost:
Closes #5872


wcfsetup/install/files/lib/util/HeaderUtil.class.php
```diff
@@ -163,10 +168,14 @@ public static function parseOutput($output)
         // move script tags to the bottom of the page
         $javascript = [];
         self::$output = \preg_replace_callback(
-            '~<script data-relocate="true"(?P<script>.*?</script>)\s*~s',
+            '~<script data-relocate="true"(?<attributes>[^>]*)>(?P<script>.*?)</script>\s*~s',
```
dtdesign, line 171:
```suggestion
            '~<script data-relocate="true"(?<attributes>[^>]*+)>(?P<script>.*?)</script>\s*~s',
```

Since we’re using a character class, we can take advantage of a possessive quantifier to avoid backtracking.

dtdesign (changes requested):

PR #5884, opened 2024-04-17
Ignore `searchable`-flag in acp

### LINKED ISSUE #5882: `wcf1_user_option.searchable` has an unintended effect on functions in the admin panel
Options for which `searchable` is set to `false` cannot be searched in the admin panel and cannot be used as conditions for automatic user group assignment and other similar functions.

`searchable` should only have an effect on the frontend member search.

Cyperghost:
Closes #5882


BurntimeX:
The PR is formally correct, but results in all options being visible in the search, even those that are generally not desired there.

PR #5885, opened 2024-04-19
Replace privacy policy (required) confirmation

SoftCreatR:
See https://www.it-recht-kanzlei.de/edsa-zustimmung-in-datenschutzerklaerung-versto%C3%9F-gegen-treu-und-glauben.html


BurntimeX:
This change requires an adjustment to the privacy policy:
https://github.com/WoltLab/WCF/pull/5886

PR #5890, opened 2024-04-22
Add classes to IP addresses

SoftCreatR:
Working with IP addresses or user agents can be challenging due to the difficulty in efficiently locating them. This pull request addresses this issue by introducing specific classes to the fields where these elements appear, making them readily accessible. For instance, you can now easily manipulate or retrieve these values using JavaScript, streamlining the process and enhancing performance.

Additionally, there are scenarios, such as in the WBB/WSF (postIpAddress template), where applying this change could be beneficial or even necessary. However, for obvious reasons, those suggestions fall outside the scope of what I can propose in this context. :)

This change is relatively minor and carries no risk of causing side effects. Therefore, it should be safe to merge.

### DIFF
--- a/com.woltlab.wcf/templates/user.tpl
+++ b/com.woltlab.wcf/templates/user.tpl
@@ -226,7 +226,7 @@
 								{if $user->getCurrentLocation()}<li>{@$user->getCurrentLocation()}</li>{/if}
 							{/if}
 							{if $__wcf->session->getPermission('admin.user.canViewIpAddress') && $user->registrationIpAddress}
-								<li>{lang}wcf.user.registrationIpAddress{/lang}: <span class="userRegistrationIpAddress">{@$user->getRegistrationIpAddress()|ipSearch}</span></li>
+								<li class="ipAddress">{lang}wcf.user.registrationIpAddress{/lang}: <span class="userRegistrationIpAddress">{@$user->getRegistrationIpAddress()|ipSearch}</span></li>
 							{/if}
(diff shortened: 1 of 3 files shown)


com.woltlab.wcf/templates/user.tpl
```diff
@@ -226,7 +226,7 @@
 								{if $user->getCurrentLocation()}<li>{@$user->getCurrentLocation()}</li>{/if}
 							{/if}
 							{if $__wcf->session->getPermission('admin.user.canViewIpAddress') && $user->registrationIpAddress}
-								<li>{lang}wcf.user.registrationIpAddress{/lang}: <span class="userRegistrationIpAddress">{@$user->getRegistrationIpAddress()|ipSearch}</span></li>
+								<li class="ipAddress">{lang}wcf.user.registrationIpAddress{/lang}: <span class="userRegistrationIpAddress">{@$user->getRegistrationIpAddress()|ipSearch}</span></li>
```
SoftCreatR, line 229: I've overseen that there is already a span with an appropriate class. However, adding a class to the surrounding element doesn't hurt :)

SoftCreatR:
@dtdesign Any reason why?

dtdesign:
Sorry, I didn’t submit my message:

This is a pretty generic class name and introducing it in a stable version can cause unwanted side effects in case that class name is already in use somewhere else (plugins, styles, etc.). We can add it to 6.1 though if you want.

PR #5892, opened 2024-04-25
New form field `MultilineItemListFormField`

Cyperghost:
Adds the new form field `MultilineItemListFormField` that allows you to have a list of items that is not in a single input text field like the `ItemListFormField`. 
Instead, it spans multiple lines and can also be filtered. The items can be edited.

### DIFF
--- a/ts/WoltLabSuite/Core/Form/Builder/Field/MultilineItemList.ts
+++ b/ts/WoltLabSuite/Core/Form/Builder/Field/MultilineItemList.ts
@@ -1,0 +1,22 @@
+/**
+ * Data handler for an multiline item list form builder field in an Ajax form.
+ *
+ * @author      Olaf Braun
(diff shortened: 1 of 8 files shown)


ts/WoltLabSuite/Core/Form/Builder/Field/MultilineItemList.ts
```diff
@@ -0,0 +1,22 @@
...
+
+class MultilineItemList extends Field {
+  protected _getData(): FormBuilderData {
+    return {
+      [this._fieldId]: getValues(this._fieldId),
+    };
+  }
+}
+
+export = MultilineItemList;
```
dtdesign, line 22: Please use proper exports, this is a horrible workaround for some very old JS modules.

dtdesign (changes requested):

Issue #4031, opened 2021-02-26
Allow in-browser scaling for non-animated GIFs only / preserve the animation when scaling


TimWolla:
see #4030

PR #5905, opened 2024-05-05
Upload Pipeline v2

### LINKED ISSUE #5668: Unified upload pipeline
File uploads have been implemented differently several times in the software so far. The unified upload pipeline is intended to standardize these cases and make upload-relevant features available throughout:
* Chunked uploads
* Client-side image scaling
* Server-side image scaling
* User defined image processing (clipping, rotating, focal point for thumbnails, etc.)
* Thumbnail generation

ref
https://www.woltlab.com/community/thread/278320-automatische-bildskalierung-auch-in-der-galerie/
https://www.woltlab.com/community/thread/266577-chunked-upload/
https://www.woltlab.com/community/thread/266059-titelbild-positionieren/
https://www.woltlab.com/community/thread/291287-%C3%BCberpr%C3%BCfung-der-erlaubten-dateiendung/

dtdesign:
Missing parts:
* [x] Document the interface for file processors.
* [x] Proper clean-up of files and thumbnails.
* [x] Handling of orphaned file uploads that are no longer associated to an `objectTypeID`.
* [x] Resizing of GIFs in the browser (see #4031).
* [x] Enable downloads again for non-image attachments.
* [x] Add the missing phrases for the attachment action buttons.
* [x] Implement a proper error handling for server-side errors, we’re currently just dumping the `message` property. This might also require some updated CSS to work nicely with the grid layout.
* [x] Localized error messages for the pre-upload check.
* [x] Abstract implementation for `IFileProcessor` that allows us to easily expand the interface later.
* [x] It is currently only possible to adopt a file but not specific thumbnails. Do we need some kind of callback to register thumbnails in case they’re being referenced directly?
* [x] `Attachment/List.ts` handles too much logic and should offload some into `Attachment/Entry.ts` which is unused right now.
* [x] There is currently no upload progress because both Firefox and Safari do not support `ReadableStream` in the request body of `fetch()`. This means that we cannot track how much data has been sent to the server already. We can still track the chunks but that might be a bit light and potentially misleading due to how different the size of chunks can be for smaller uploads. (UPDATE: The decision was made to ignore this for now. Due to the extreme differences in the values for `post_max_size` this can differ widely.)
* [x] `<woltlab-core-file-upload>` does not support limiting the number of concurrently uploaded files yet.
* [ ] HTTPS is now a hard requirement and must be added to the system requirements. (see #5919)
* [x] Add support for a general upper limit for uploaded files that can be configured on the server side / environment.
* [x] Thumbnails are currently generated as WebP only, do we want to support garbage browsers without WebP support? (UPDATE: The decision was made to not make an exception for ancient browsers.)
* [ ] Write the API documentation for the dev docs. (see WoltLab/docs.woltlab.com#438)
* [x] Block uploads that exceed the file size limit before posting data to the server.
* [x] Add a worker to rebuild files / thumbnails.
* [x] Store the configuration used to generate a thumbnail to skip it if its unchanged and the file hash matches.
* [x] Migrate the existing attachments to the new system.
* [x] Remove the rebuild worker for attachments.
* [x] Add a static WebP variant of the original images for both PNG and JPEG images.
* [x] Remove the files (PHP+JS), methods and templates for the old attachment system.

Closes #5668


ts/WoltLabSuite/Core/Ajax/Backend.ts
```diff
@@ -135,14 +143,19 @@ class BackendRequest {
       init.method = "POST";
 
       if (this.#payload) {
-        if (this.#payload instanceof FormData) {
+        if (this.#payload instanceof Blob) {
+          init.headers!["Content-Type"] = "application/octet-stream";
```
TimWolla, line 147: Is the `!` actually necessary here? `init.headers` appears to always be initialized.

ts/WoltLabSuite/Core/Component/File/Upload.ts
```diff
@@ -0,0 +1,249 @@
...
+
+function validateFile(element: WoltlabCoreFileUploadElement, file: File): boolean {
+  const fileExtensions = (element.dataset.fileExtensions || "*").split(",");
+  for (const fileExtension of fileExtensions) {
+    if (fileExtension === "*") {
+      return true;
+    } else if (file.name.endsWith(fileExtension)) {
+      return true;
+    }
+  }
```
TimWolla, line 187: This could possibly be cleaner by leveraging `Array.some`:

```
const hasAllowedExtension = fileExtensions.some(…);
if (hasAllowedExtension) {
  return true;
} else {
  innerError(…);
  return false;
}
```

wcfsetup/install/files/lib/action/FileDownloadAction.class.php
```diff
@@ -0,0 +1,82 @@
...
+
+        $filename = $file->getPathname();
+        $response = new Response(
+            new Stream($filename),
+        );
+
+        $mimeType = FileUtil::getMimeType($filename);
+
+        // TODO: This should use `FileReader` instead.
+
```
TimWolla, line 59: Should it? The PSR-7/-15 implementation should already support streaming: That's the whole purpose of the `$response->getBody()` being a stream.

wcfsetup/install/files/lib/action/FileDownloadAction.class.php
```diff
@@ -0,0 +1,82 @@
...
+        $dispositionType = \in_array($mimeType, $inlineMimeTypes) ? 'inline' : 'attachment';
+
+        return $response->withHeader('content-type', $mimeType)
+            ->withHeader(
+                'content-disposition',
+                \sprintf(
+                    '%s; filename="%s"',
+                    $dispositionType,
+                    $file->filename,
+                ),
```
TimWolla, line 79: This should use https://github.com/WoltLab/WCF/blob/master/wcfsetup/install/files/lib/http/ContentDisposition.class.php to correctly handle Unicode filenames.

wcfsetup/install/files/lib/data/attachment/Attachment.class.php
```diff
@@ -190,11 +200,7 @@ public function getThumbnailLocation($size = '')
      */
     public function migrateStorage()
     {
-        foreach ([
-            $this->getLocation(),
-            $this->getThumbnailLocation(),
-            $this->getThumbnailLocation('tiny'),
-        ] as $location) {
+        foreach ([$this->getLocation(), $this->getThumbnailLocation(), $this->getThumbnailLocation('tiny'),] as $location) {
```
TimWolla, line 203: Intentional change?

wcfsetup/install/files/lib/data/file/File.class.php
```diff
@@ -0,0 +1,192 @@
...
+                    data-meta-data="%s"
+                    data-link="%s"
+                ></woltlab-core-file>
+                EOT,
+            $this->fileID,
+            StringUtil::encodeHTML($this->filename),
+            $this->fileSize,
+            StringUtil::encodeHTML($this->mimeType),
+            StringUtil::encodeHTML(\json_encode($thumbnails)),
+            StringUtil::encodeHTML(\json_encode($metaData)),
```
TimWolla, line 162: This should include the `JSON_THROW_ON_ERROR` flag if `JSON::encode()` is not used.

wcfsetup/install/files/lib/data/file/File.class.php
```diff
@@ -0,0 +1,192 @@
...
+        $fileExtension = \array_search($mimeType, self::SAFE_FILE_EXTENSIONS, true);
+        if (\is_string($fileExtension)) {
+            return $fileExtension;
+        }
+
+        if (\str_contains($filename, '.')) {
+            $fileExtension = \mb_substr(
+                $filename,
+                \mb_strrpos($filename, '.') + 1
+            );
```
TimWolla, line 183: `pathinfo($filename, PATHINFO_EXTENSION)`

wcfsetup/install/files/lib/data/file/temporary/FileTemporary.class.php
```diff
@@ -0,0 +1,88 @@
...
+
+    public function getChunkSize(): int
+    {
+        return \ceil($this->fileSize / $this->getChunkCount());
+    }
+
+    public function hasChunk(int $sequenceNo): bool
+    {
+        return $this->chunks[$sequenceNo] === '1';
+    }
```
TimWolla, line 44: Should this method check `$sequenceNo < $this->getChunkCount()`?

wcfsetup/install/files/lib/data/file/temporary/FileTemporary.class.php
```diff
@@ -0,0 +1,88 @@
...
+
+    public function getContext(): array
+    {
+        return JSON::decode($this->context);
+    }
+
+    public static function getNumberOfChunks(int $fileSize): int
+    {
+        return \ceil($fileSize / self::getOptimalChunkSize());
+    }
```
TimWolla, line 76: The result of this function will change if `post_max_size` changes. Will this cause problems?

wcfsetup/install/files/lib/system/endpoint/controller/core/files/upload/PostChunk.class.php
```diff
@@ -0,0 +1,144 @@
...
+     * See https://stackoverflow.com/a/61997147
+     */
+    private const FREAD_BUFFER_SIZE = 10 * 1_024 * 1_024;
+
+    public function __invoke(ServerRequestInterface $request, array $variables): ResponseInterface
+    {
+        $checksum = \current($request->getHeader('chunk-checksum-sha256'));
+        if ($checksum === false) {
+            throw new UserInputException('chunk-checksum-sha256');
+        }
```
TimWolla, line 30: `$request->getHeaderLine()` avoids dealing with arrays.

TimWolla (commented):
Some remarks from looking at the diff. Didn't attempt to understand the relation between the different parts. Feel free to disregard anything that doesn't make sense or anything you disagree with.

ts/WoltLabSuite/Core/Ajax/Backend.ts
dtdesign: It is defined as `HeadersInit | undefined` 🤷

ts/WoltLabSuite/Core/Component/File/Upload.ts
dtdesign: Tried it but apart from the nice variable name it just adds more clutter.

wcfsetup/install/files/lib/action/FileDownloadAction.class.php
dtdesign: I’m not sure, the `FileReader` supports all sorts of different headers and range handling, this is a very early comment too. I will look into it, thanks.

wcfsetup/install/files/lib/data/file/temporary/FileTemporary.class.php
dtdesign: Yes, but only if the value decreases while the chunks are being uploaded. There is nothing we can do about this because the client needs to know what the acceptable values are.

wcfsetup/install/files/lib/action/FileDownloadAction.class.php
TimWolla: Header handling is much cleaner and easier with plain PSR-7. As for the Range header support, I recommend:

1. Create a [helper function](https://github.com/WoltLab/WCF/blob/master/wcfsetup/install/files/lib/http/Helper.class.php) to parse the header, returning $offset + $length.
2. Possibly perform some validation of the returned offset + length against the filesize.
3. Set the appropriate response headers.
4. Wrap the body Stream into [Guzzle's LimitStream](https://github.com/guzzle/psr7/blob/2.6/src/LimitStream.php).

In fact all 4 steps could be wrapped into a single function that takes a `ResponseInterface` and returns an updated `ResponseInterface` with headers applied and body wrapped.

wcfsetup/install/files/lib/data/file/temporary/FileTemporary.class.php
TimWolla: Okay, if it's only during the upload, then it's probably acceptable. However I suggest to check the failure mode at least: Will the upload appear stuck to the user? Will it “gracefully” report an error message that the upload needs to be restarted? Will it result in a corrupt file, without any indication of an error?

dtdesign: The upload will fail in this case because the checksum will mismatch which triggers the error handler. This is the beauty of strict checksum checks for both the individual chunks and the complete file, guarding against corrupted files.

wcfsetup/install/files/lib/action/FileDownloadAction.class.php
```diff
@@ -0,0 +1,118 @@
...
+            throw new PermissionDeniedException();
+        }
+
+        $eTag = \sprintf(
+            '"%d-%s"',
+            $file->fileID,
+            \substr($file->fileHash, 0, 8),
+        );
+
+        $httpIfNoneMatch = $_SERVER['HTTP_IF_NONE_MATCH'] ?? '';
```
TimWolla, line 59: - At a minimum this should use `$request->getHeaderLine()`, instead of `$_SERVER`.
- A more correct solution would be leveraging Guzzle’s `Header::splitList()` + an in_array check (see the example in the README: https://github.com/guzzle/psr7?tab=readme-ov-file#guzzlehttppsr7headersplitlist). 
- The most correct solution would be actually using the “weak comparison function” to compare two ETags, as per [RFC 9110#13.1.2](https://www.rfc-editor.org/rfc/rfc9110#name-if-none-match). Basically an ETag would be a match for purposes of `if-none-match`, if it is identical after stripping the `W/` prefix from both ETags.

The last bullet point would be relevant if the web server gzips the attachment, thus turning the strong ETag by WoltLab Suite into a weak one.

PS: Technically there's also the asterisk to handle :smile:

wcfsetup/install/files/lib/action/FileDownloadAction.class.php
```diff
@@ -0,0 +1,118 @@
...
+        if ($httpIfNoneMatch === $eTag) {
+            return new EmptyResponse(304);
```
TimWolla, line 61: The 304 should include the caching headers (cache-control + expires): https://www.rfc-editor.org/rfc/rfc9110#name-304-not-modified. The easy solution might be including all of them by means of a helper function.

wcfsetup/install/files/lib/action/FileDownloadAction.class.php
```diff
@@ -0,0 +1,118 @@
...
+        $response = $response
+            ->withHeader('content-security-policy', "default-src 'none'; style-src 'unsafe-inline';")
+            ->withHeader('x-content-type-options', 'nosniff');
+
+        $lifetimeInSeconds = $processor->getFileCacheDuration($file)->lifetimeInSeconds;
+        if ($lifetimeInSeconds !== null) {
+            $expiresAt = \sprintf(
+                '%s GMT',
+                \gmdate('D, d M Y H:i:s', $lifetimeInSeconds)
+            );
```
TimWolla, line 99: I believe this is not be correct: The lifetime needs to be added to the current date.

I would suggest the following, avoiding manually specifying the format. RFC 7231 is the (old) HTTP/1.1 RFC.

```
var_dump((new DateTimeImmutable('now', new DateTimeZone('UTC')))->modify("+{$maxAge} seconds")->format(DateTimeImmutable::RFC7231));
```

wcfsetup/install/files/lib/action/FileDownloadAction.class.php
```diff
@@ -0,0 +1,118 @@
...
+            $maxAge = \sprintf(
+                'max-age=%d, private',
+                $lifetimeInSeconds ?: 0,
+            );
+
+            $response = $response
+                ->withHeader('Expires', $expiresAt)
+                ->withHeader('Cache-control', $maxAge);
```
TimWolla, line 107: I suggest to define the header names in lowercase for consistency (see also: ETag below).

wcfsetup/install/files/lib/data/file/FileEditor.class.php
```diff
@@ -0,0 +1,105 @@
...
+        if ($isImage) {
+            [$width, $height] = \getimagesize($pathname);
+        }
+
+        $fileAction = new FileAction([], 'create', ['data' => [
+            'filename' => $fileTemporary->filename,
+            'fileSize' => $fileTemporary->fileSize,
+            'fileHash' => $fileTemporary->fileHash,
+            'fileExtension' => File::getSafeFileExtension($mimeType, $fileTemporary->filename),
+            'secret' => \bin2hex(\random_bytes(10)),
```
TimWolla, line 84: What is this secret for? 10 bytes are 80 Bits of entropy. This is generally not considered secure enough to be unguessable. It should be 16 (i.e. 128 bits) for that and unless you have a good reason for it to be shorter, I would suggest that.

dtdesign: It’s entire point is to not allow the storage location to be guessable in case you know the id and SHA-256 hash of the file. I have increased this to 16 bytes as per your recommendation.

PR #5906, opened 2024-05-05
Move methods that are used for the configuration of boxes during package installation to commands

### LINKED ISSUE #5904: Move methods that are used for the configuration of boxes during package installation to commands
Calling `BoxHandler::createBoxCondition()` and `BoxHandler::addBoxToPageAssignments()` causes the boxes to be loaded (`BoxHandler::init()`), which is unnecessary during package installation and can potentially lead to errors.

BurntimeX:
Closes #5904


dtdesign (approved):
@BurntimeX Please check if this needs to be also called in existing code in the core too.

PR #5909, opened 2024-05-07
Force all media to use frontend route

### LINKED ISSUE #5902: Force all embedded media to use the frontend route
Historically we had two different sets of sessions for the access to the frontend and the backend, requiring embedded media to point to a different route depending on the context. However, since the unification of the session system in 5.4, a user logged in to the admin panel is **always** logged-in to the frontend at the same time.

* [ ] Embedded media should always be referencing the frontend controller.
* [ ] Redirect any request to the ACP media controller to the frontend, regardless of the login state.
  * [ ] Generated URL must reuse the parameters of the request itself to prevent information leak by generating the URL through the `object` key.

Cyperghost:
Closes #5902


wcfsetup/install/files/lib/http/middleware/RedirectMediaToFrontend.class.php
```diff
@@ -0,0 +1,47 @@
+<?php
```
dtdesign, line 1: Why are you using an extra middleware for a basic redirect that could just be implemented in the existing `wcf\acp\page\MediaPage`?

dtdesign (changes requested):

dtdesign (approved):

PR #5929, opened 2024-05-29
Use an event for the spam check during registration

BurntimeX:


dtdesign (approved):

wcfsetup/install/files/lib/event/page/ContactFormSpamChecking.class.php
```diff
@@ -0,0 +1,26 @@
...
+ * @license     GNU Lesser General Public License <http://opensource.org/licenses/lgpl-license.php>
+ * @since       6.1
+ */
+final class ContactFormSpamChecking implements IInterruptableEvent
+{
+    use TInterruptableEvent;
+
+    public function __construct(
+        public readonly string $email,
+        public readonly string $ipAddress,
```
dtdesign, line 23: The message body for the contact form can be disabled, but shouldn’t we still try to at least find and provide any values for input fields that use the message type?

dtdesign (approved):

PR #5932, opened 2024-06-01
Disable log directory recursion

SoftCreatR:
The exception log overview page currently displays all files that follow the same naming scheme, which is appropriate. However, it is possible for other log files with the same or similar naming schemes to be present in the log folder. For instance, in our case, we store log files with the same naming scheme in a subdirectory. This proposal addresses the issue because, at present, these files are included in the overview since the software reads the log directory recursively.


dtdesign (approved):

dtdesign:
Thanks!

PR #5944, opened 2024-06-14
Comment backend overhaul

BurntimeX:
This PR migrates the backend of the comment system to RPC controllers and commands and is intended to serve as a template for the future modernization of other backends.


wcfsetup/install/files/lib/system/endpoint/controller/core/comments/UpdateComment.class.php
```diff
@@ -0,0 +1,80 @@
...
+            $comment,
+            $htmlInputProcessor,
+        ))();
+
+        return new JsonResponse([]);
+    }
+
+    private function assertCommentIsEditable(Comment $comment): void
+    {
+        $processor = CommentHandler::getInstance()->getObjectType($comment->objectTypeID)->getProcessor();
```
dtdesign, line 64: This can be replaced with the new helper method.

wcfsetup/install/files/lib/system/endpoint/controller/core/comments/responses/EnableResponse.class.php
```diff
@@ -0,0 +1,53 @@
...
+        (new \wcf\system\comment\response\command\PublishResponse($response))();
+
+        return new JsonResponse([]);
+    }
+
+    private function assertResponseCanBeEnabled(CommentResponse $response): void
+    {
+        if (!$response->isDisabled) {
+            throw new IllegalLinkException();
+        }
```
dtdesign, line 44: This action should be idempotent, instead add a guard around `PublishResponse()`.

wcfsetup/install/files/lib/system/endpoint/controller/core/comments/responses/RenderResponse.class.php
```diff
@@ -0,0 +1,112 @@
...
+{
+    #[\Override]
+    public function __invoke(ServerRequestInterface $request, array $variables): ResponseInterface
+    {
+        $response = Helper::fetchObjectFromRequestParameter($variables['id'], CommentResponse::class);
+        $parameters = Helper::mapQueryParameters(
+            $request->getQueryParams(),
+            <<<'EOT'
+                array {
+                    messageOnly: null|bool,
```
dtdesign, line 39: Have you tried if `messageOnly?: bool` works?

dtdesign (changes requested):
After finishing the review I question some of my remarks. In particular commands should be designed to be only called with clean data, otherwise following the “garbage in, garbage out” principle. They shouldn’t perform any validation whatsoever because this is the callee’s job.

Therefore please disregard my comments that suggest additional validations for commands and strip the existing validation unless you feel it is really important because it would otherwise cause serious issues.

Also, some of my comments are merely remarks so please feel free to disregard my comments at your discretion.

PR #5945, opened 2024-06-17
Extra check that the user also has the right to upload attachment when editing a user

Cyperghost:
See https://www.woltlab.com/community/thread/306725-fehler-bei-dateianhang-hochladen-in-signatur-und-konversation/

If an administrator is currently editing the signature of himself or another user, the tab for attachments is displayed. This is also the case if the right to upload attachments in the signature has been withdrawn from all user groups.


dtdesign (approved):

dtdesign:
@Cyperghost Please validate that this does not cause issues when there already is an uploaded attachment, but the current user lacks the permission.

PR #5951, opened 2024-06-26
New upload form field

Cyperghost:
Adds a new upload form field with the new upload pipeline (https://github.com/WoltLab/WCF/issues/5668)


dtdesign (commented):
A good start, but I would like to see the logic between attachments and the upload form field to be deduplicated. Do not use inheritance, but instead check if you can provide some general purpose functions that offload the work. Especially the DOM and upload progress are pretty much set in stone anyway and can easily be handled by a common function.

ts/WoltLabSuite/Core/Form/Builder/Field/Controller/FileProcessor.ts
```diff
@@ -8,8 +8,16 @@
...
 import DomChangeListener from "WoltLabSuite/Core/Dom/Change/Listener";
+import {
+  fileInitializationFailed,
+  insertFileInformation,
+  removeUploadProgress,
+  trackUploadProgress,
+} from "WoltLabSuite/Core/Component/File/File";
+import { clearPreviousErrors } from "WoltLabSuite/Core/Component/File/Upload";
+
+const _data = new Map<string, FileProcessor>();
```
dtdesign, line 20:
```suggestion
type FileId = string;
const fileProcessors = new Map<FieldId, FileProcessor>();
```

This avoid the generic name and makes it immediately clear what it will contain.

ts/WoltLabSuite/Core/Component/File/File.ts
```diff
@@ -0,0 +1,90 @@
+import WoltlabCoreFileElement from "WoltLabSuite/Core/Component/File/woltlab-core-file";
```
dtdesign, line 1: The file name is too generic and it implies that it represents a fiel when it really just exposes a bunch of helper methods. It should be renamed to `Component/File/Helper.ts`.

Please remember to clean up the then obsolete JS artifact.

wcfsetup/install/files/lib/system/event/listener/PreloadPhrasesCollectingListener.class.php
```diff
@@ -109,6 +109,7 @@ public function __invoke(PreloadPhrasesCollecting $event): void
         $event->preload('wcf.global.success');
         $event->preload('wcf.global.success.add');
         $event->preload('wcf.global.success.edit');
+        $event->preload('wcf.global.button.replace');
```
dtdesign, line 112: Please move this up, the phrases are sorted and grouped together.

wcfsetup/install/files/lib/system/file/processor/FileProcessor.class.php
```diff
@@ -240,6 +240,11 @@ public function delete(array $files): void
 
     public function hasReachedUploadLimit(IFileProcessor $fileProcessor, array $context): bool
     {
+        // Replace button from `FileProcessorFormField` is clicked.
+        if (!empty($context["__replace"])) {
+            return false;
+        }
```
dtdesign, line 246:
```suggestion
        $isReplacement = $context['__replace'] ?? false;
        if ($isReplacement) {
            return false;
        }
```

Do not use `empty()`, it does not make clear if the key may be absent or if you’re simply checking if the value is falsy. I much prefer being verbose here. Even better: If the key may be absent, you can rewrite this while also getting rid of the comment.

dtdesign (changes requested):

dtdesign (approved):

PR #5963, opened 2024-07-10
Add `validateUpload` function to `IFileProcessor`

Cyperghost:
Add the new function `validateUpload` to `IFileProcessor` which is called to check if the uploaded file is allowed. 
This can be used to validate the image width/height, for example


dtdesign:
How does this differ from `acceptUpload()`?

Cyperghost:
> How does this differ from `acceptUpload()`?

This function does not provide any information about the file itself. It is therefore not possible to check whether an image has a minimum width or height.

dtdesign (approved):

PR #5973, opened 2024-08-09
Stop using HTTPRequest for package update server requests

BurntimeX:


wcfsetup/install/files/lib/system/package/PackageUpdateDispatcher.class.php
```diff
@@ -215,27 +222,27 @@ private function getPackageUpdateXML(PackageUpdateServer $updateServer)
...
-            if (empty($reply['httpHeaders']['etag']) && empty($reply['httpHeaders']['last-modified'])) {
+            if (empty($response->getHeaders()['etag']) && empty($response->getHeaders()['last-modified'])) {
                 throw new SystemException("Missing required HTTP headers 'etag' and 'last-modified'.");
             }
 
             $metaData['list'] = [];
-            if (!empty($reply['httpHeaders']['etag'])) {
-                $metaData['list']['etag'] = \reset($reply['httpHeaders']['etag']);
+            if (!empty($response->getHeaders()['etag'])) {
+                $metaData['list']['etag'] = \reset($response->getHeaders()['etag']);
```
TimWolla, line 237: Careful here (and in the other places where `getHeaders()` is used): The casing depends on the casing returned by the server [1], which is not necessarily lowercased [2]. 

The correct solution is `$response->getHeaderLine('etag')` (returns string, duplicate headers are comma-separated) or `$response->getHeader('etag')` (returns array) and possibly `$response->hasHeader('etag')`.

[1] https://github.com/WoltLab/WCF/blob/998bd9fecd3f562d710e3bfa9a02fec133f70e3e/wcfsetup/install/files/lib/system/api/guzzlehttp/psr7/src/MessageTrait.php#L15-L16
[2] HTTP/2 guarantees lowercase and recent HAProxy versions emit lowercase by default even with HTTP/1.1.

PR #6003, opened 2024-09-30
Displays an error message if the license information couldn't be fetched

Cyperghost:
See https://www.woltlab.com/community/thread/308438-curl-error-28-resolving-timed-out-after-5013-milliseconds/


BurntimeX (approved):
LGTM

PR #6006, opened 2024-10-01
Implement `canAdopt()` in `IFileProcessor` and `FileProcessor`

Cyperghost:
This function checks whether a file may be assigned to the `$context`. In addition, `FileProcessor::canAdopt()` checks whether the `objectType` is correct.

### DIFF
--- a/wcfsetup/install/files/lib/system/file/processor/AttachmentFileProcessor.class.php
+++ b/wcfsetup/install/files/lib/system/file/processor/AttachmentFileProcessor.class.php
@@ -41,6 +41,18 @@ final class AttachmentFileProcessor extends AbstractFileProcessor
     }
 
     #[\Override]
+    public function canAdopt(File $file, array $context): bool
+    {
+        $attachment = Attachment::findByFileID($file->fileID);
+
+        if ($attachment === null) {
+            return true;
+        }
+
+        return false;
+    }
+
+    #[\Override]
     public function adopt(File $file, array $context): void
     {
         $attachmentHandler = $this->getAttachmentHandlerFromContext($context);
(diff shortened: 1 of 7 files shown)


wcfsetup/install/files/lib/system/file/processor/AttachmentFileProcessor.class.php
```diff
@@ -40,6 +40,18 @@ public function getAllowedFileExtensions(array $context): array
...
+    #[\Override]
+    public function canAdopt(File $file, array $context): bool
+    {
+        $attachment = Attachment::findByFileID($file->fileID);
+
+        if ($attachment === null) {
+            return true;
+        }
+
+        return false;
```
dtdesign, line 52:
```suggestion
        return Attachment::findByFileID($file->fileID) === null;
```

wcfsetup/install/files/lib/system/form/builder/field/FileProcessorFormField.class.php
```diff
@@ -203,6 +203,18 @@ public function validate()
             );
         }
 
+        foreach ($this->files as $file) {
+            if (!FileProcessor::getInstance()->canAdopt($fileProcessor, $file, $this->context)) {
+                $this->addValidationError(
+                    new FormFieldValidationError(
+                        'adopt',
+                        'wcf.form.field.fileProcessor.error.adopt',
```
dtdesign, line 211: Should this be a more generic error message because it could happen outside of the form builder too?

dtdesign (changes requested):

dtdesign (approved):

PR #6008, opened 2024-10-07
Use `dialogFactory()` to set media categories

Cyperghost:
See https://www.woltlab.com/community/thread/308498-cms-medien-kategorie-bearbeiten-beim-zweiten-mal-ohne-funktion/


dtdesign (approved):
Please also merge this into the master afterwards because of the shared template.

PR #6024, opened 2024-10-16
Improve attachment support in form builder forms

BurntimeX:


wcfsetup/install/files/lib/system/form/builder/field/wysiwyg/WysiwygAttachmentFormField.class.php
```diff
@@ -64,7 +64,13 @@ public function attachmentHandler(?AttachmentHandler $attachmentHandler = null)
     {
         if ($attachmentHandler !== null) {
             if ($this->attachmentHandler === null) {
-                $tmpHash = StringUtil::getRandomID();
+                $tmpHash = \sha1(\implode("\0", [
+                    $this->getId(),
+                    $attachmentHandler->getObjectType()->objectType,
+                    $attachmentHandler->getParentObjectID(),
```
dtdesign, line 70:
```suggestion
                    $attachmentHandler->getParentObjectID(),
                    $attachmentHandler->getObjectID(),
```

You must include the object id otherwise the generated hash is identical for both existing objects and those that are about to be created.

dtdesign (changes requested):

Issue #5969, opened 2024-08-02
Emoji picker


BurntimeX:
A picker for emoji as a tab below the editor to simplify the insertion of emojis (especially on non-mobile devices).

The graphics of the respective operating system should be used to display the emojis so that we do not have to rely on third-party libraries.

As emojis are somewhat redundant to the existing smiley function, they could replace the smiley function in the long term. For the time being, I prefer to set the smiley function to "off" in the default installation.

Krymonota:
Personally, I'm a fan of [emoji-picker-element](https://www.npmjs.com/package/emoji-picker-element). You might want to take a look at this emoji picker – be it for integration or inspiration purposes 🙂

SoftCreatR:
I hope, this will be done developer friendly, so that we are able to replace them with custom graphics, when desired. This will be interesting for those, who purchased emojione for example.

PR #6026, opened 2024-10-18
Emoji picker

Cyperghost:
Required https://github.com/WoltLab/editor/pull/139
Closes https://github.com/WoltLab/WCF/issues/5969

Deactivates the smiley system for new installations.


ts/WoltLabSuite/Core/Component/Ckeditor.ts
```diff
@@ -309,6 +309,14 @@ export async function setupCkeditor(
     configuration.licenseKey = licenseKey;
   }
 
+  const { DATABASE_FOR_AUTO_COMPLETE } = await import("./EmojiPicker/woltlab-core-emoji-picker");
+  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+  // @ts-expect-error
+  // TODO remove eslint-disable
```
dtdesign: `// TODO`

Also, this module should import `woltlab-core-emoji-picker` as a side-effect and thus preemptively loading the emoji data.

extra/update-emoji-picker-element.ts
```diff
@@ -0,0 +1,120 @@
...
+  { local: "pl", i18n: pl },
+  { local: "pt", i18n: pt_PT },
+  { local: "ru", i18n: ru_RU },
+  { local: "es", i18n: es },
+  { local: "sv" },
+  { local: "uk" },
+];
+
+(async () => {
+  let localisationContent = `import { I18n } from "emoji-picker-element/shared";
```
dtdesign: I strongly suggest adding a comment at the very top that states that this file is auto-generated and where to find the file to rebuild it. Make use of inline comments so that this is stripped from the bundled version.

dtdesign (changes requested):
You should also include module comments in the TS files.

dtdesign (approved):

Issue #5981, opened 2024-08-21
Add support for the password strength meter to `PasswordFormField`


BurntimeX:
see https://github.com/WoltLab/WCF/blob/master/ts/WoltLabSuite/Core/Ui/User/PasswordStrength.ts

PR #6037, opened 2024-10-24
Add strength meter support to `PasswordFormField`

Cyperghost:
Closes https://github.com/WoltLab/WCF/issues/5981

### DIFF
--- a/wcfsetup/install/files/lib/system/form/builder/field/PasswordFormField.class.php
+++ b/wcfsetup/install/files/lib/system/form/builder/field/PasswordFormField.class.php
@@ -46,6 +46,16 @@ class PasswordFormField extends AbstractFormField implements
      */
     protected $templateName = 'shared_passwordFormField';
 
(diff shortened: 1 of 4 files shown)


wcfsetup/install/files/lib/system/form/builder/field/PasswordFormField.class.php
```diff
@@ -46,6 +46,16 @@ class PasswordFormField extends AbstractFormField implements
      */
     protected $templateName = 'shared_passwordFormField';
 
+    protected bool $strengthMeter = true;
+    /**
+     * @var IFormField[]
+     */
+    protected array $relatedFields = [];
+    /**
```
dtdesign:
```suggestion
    protected bool $strengthMeter = true;

    /**
     * @var IFormField[]
     */
    protected array $relatedFields = [];

    /**
```

dtdesign (changes requested):

PR #6047, opened 2024-11-06
File processor image cropping

Cyperghost:
Lets you crop images in two ways:

### Min Max

- Allows the user to crop the image to a specified minimum and maximum size
- Minimum and maximum must have the same aspect ratio

### Exact

- Alles sizes must have the same aspect ratio
- If the image already corresponds to the correct dimensions, no dialog is displayed and the image is uploaded directly
- The cropped image will be resized to the largest possible size


dtdesign (commented):
Looks good so far but the final verdict depends on the implementation of the cover photo so that we can validate the MinMax case with a real use case.

PR #6063, opened 2024-11-12
Update reactions.scss

eveneme:
Adding top and bottom margins to the popover for reactions so it doesn’t appear stuck to the buttons.

### DIFF
--- a/wcfsetup/install/files/style/ui/reactions.scss
+++ b/wcfsetup/install/files/style/ui/reactions.scss
@@ -11,6 +11,7 @@
 	vertical-align: middle;
 	visibility: hidden;
 	z-index: 60;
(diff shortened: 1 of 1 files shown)


eveneme:
@SoftCreatR You’re adding a negative reaction to my suggestion, yet on your own forum, you have a margin between the button and reactions. Paranoia.  😅

SoftCreatR:
@eveneme Misclick. I clicked on the existing reaction.

dtdesign:
This is highly inconsistent with all other uses of drop-down menus or other kind of overlays.

Furthermore, the positioning is controlled through a JS component (`Ui/Alignment`) that understands the concept of added gaps through `verticalOffset` without the need of CSS.

Issue #6014, opened 2024-10-10
Replacement for `WCF.Sortable.List`


BurntimeX:
`WCF.Sortable.List` is based on jQuery. We need a replacement for drag & drop sorting that is not based on jQuery.

PR #6124, opened 2024-11-28
Migrate `WCF.Sortable.List` to Typescript

Cyperghost:
Closes https://github.com/WoltLab/WCF/issues/6014


ts/WoltLabSuite/Core/Ui/Sortable/List.ts
```diff
@@ -35,52 +57,157 @@ class UiSortableList {
         containerId: "",
         className: "",
         offset: 0,
-        options: {},
+        maxNestingLevel: undefined,
+        toleranceElement: "span",
```
dtdesign, line 61: This is a breaking change despite the functionality remaining the same. Please add a check after the option initialization if `typeof options.toleranceElement === "string"` and then copy the property over.

That does cause the TS compiler to complain about it but existing JS artifacts remain functional.

ts/WoltLabSuite/Core/Ui/Sortable/List.ts
```diff
@@ -35,52 +57,157 @@ class UiSortableList {
...
+        options: {
+          animation: 150,
+          fallbackOnBody: true,
+          dataIdAttr: "data-object-id",
+          chosenClass: "sortablePlaceholder",
+          ghostClass: "",
+          draggable: "li",
+          filter: (event: Event | TouchEvent, target: HTMLElement) => {
```
dtdesign: Please move this callback and the ones following into separate methods.

ts/WoltLabSuite/Core/Ui/Sortable/List.ts
```diff
@@ -35,52 +57,157 @@ class UiSortableList {
...
+      data: {
+        actionName: "updatePosition",
+        className: this._options.className,
+        interfaceName: "wcf\\data\\ISortableAction",
+        parameters: parameters,
+      },
+      success: () => {
+        showNotification(getPhrase("wcf.global.success.edit"));
+      },
+    });
```
dtdesign: Please replace this with `dboAction()`.

dtdesign (changes requested):

dtdesign (approved):

Issue #5723, opened 2023-12-11
 Adjusting, cropping a profile picture, background image.


eveneme:
Adding the ability to customize and crop both the profile picture and the background image.

aethior:
This suggestion has also been discussed in the ideas and suggestions forum:

- Cover photo: https://www.woltlab.com/community/thread/266059-titelbild-positionieren/
- Avatar: https://www.woltlab.com/community/thread/292657-avatar-anpassung-w%C3%A4hrend-des-hochladens/

PR #6127, opened 2024-12-02
Use file processor for user cover photos

Cyperghost:
Required https://github.com/WoltLab/WCF/pull/6047
See https://github.com/WoltLab/WCF/issues/5723

### TODO

- [x] Change the base branch to `6.2` after https://github.com/WoltLab/WCF/pull/6047 has been merged
- [ ] Update docs for deprecated and removed function/classes


ts/WoltLabSuite/Core/Bootstrap.ts
```diff
@@ -52,6 +52,7 @@ window.__wcf_bc_eventHandler = EventHandler;
 export interface BoostrapOptions {
   dynamicColorScheme: boolean;
   enableMobileMenu: boolean;
+  defaultCoverPhoto?: string;
```
dtdesign: This doesn’t belong here, instead it should be provided on the button to edit the cover photo instead.

ts/WoltLabSuite/Core/Component/User/CoverPhoto.ts
```diff
@@ -0,0 +1,89 @@
...
+
+  dialog.addEventListener("afterClose", () => {
+    const file = dialog.querySelector<WoltlabCoreFile>("woltlab-core-file");
+    const coverPhotoUrl = file?.link ?? defaultCoverPhoto;
+
+    if (FormBuilderManager.hasForm(json.formId)) {
+      FormBuilderManager.unregisterForm(json.formId);
+    }
+
+    if (oldCoverPhoto === coverPhotoUrl || oldCoverPhoto === `url("${coverPhotoUrl}")`) {
```
dtdesign: The URL needs to be encoded. Ideally the entire string is stored in a variable because you are constructing it two more times below.

wcfsetup/install/files/lib/system/user/command/SetCoverPhoto.class.php
```diff
@@ -0,0 +1,76 @@
...
+    ) {
+    }
+
+    public function __invoke(): void
+    {
+        if ($this->file === null && $this->user->coverPhotoFileID !== null) {
+            (new FileAction([$this->user->coverPhotoFileID], 'delete'))->executeAction();
+        }
+
+        // Delete the old cover photo if it exists.
```
dtdesign: Why? The existing cover photo should be migrated as part of the `UserRebuildDataWorker`, therefore this code is either harmful or does nothing?

wcfsetup/install/files/lib/system/worker/UserRebuildDataWorker.class.php
```diff
@@ -289,48 +288,19 @@ public function execute()
...
-                        if ($result === false) {
-                            $data['coverPhotoExtension'] = 'jpg';
-                        }
-
-                        $editor->update($data);
-                    }
-                }
+            foreach ($userProfiles as $user) {
+                $file = FileEditor::createFromExistingFile(
+                    SetCoverPhoto::getCoverPhotoLocation($user, false),
```
dtdesign: This is an odd way of doing it, commands should not have publicly callable methods. If you need a helper method for the legacy location then add those to the `UserCoverPhoto` class as `getLegacyLocation(…)`.

dtdesign (changes requested):

PR #6132, opened 2024-12-03
Add support for images in recent activities

BurntimeX:


com.woltlab.wcf/templates/recentActivityListItem.tpl
```diff
@@ -22,6 +23,12 @@
 			</div>
 		{/if}
 
+		{if $event->getImage()}
+			<div class="recentActivityListItem__image">
+				<img src="{$event->getImage()}" loading="lazy">
```
dtdesign: `loading="lazy"` requires the `height` and `width` attribute to be set, otherwise the browser cannot reserve the required space in advance.

dtdesign (changes requested):

wcfsetup/install/files/lib/data/file/File.class.php
```diff
@@ -240,4 +242,30 @@ public static function getSafeFileExtension(string $mimeType, string $filename):
...
+                }
+                if ($minHeight !== null && $minHeight > $thumbnail->height) {
+                    continue;
+                }
+
+                return new ImageData($thumbnail->getLink(), $thumbnail->width, $thumbnail->height);
+            }
+        }
+
+        return new ImageData($this->getLink(), $this->width, $this->height);
```
dtdesign, line 276: This will unconditionally return the image even if it does not satisfy the `$minWidth` and/or `$minHeight` constraints.

wcfsetup/install/files/lib/data/media/Media.class.php
```diff
@@ -310,4 +312,23 @@ public static function getThumbnailSizes()
     {
         return static::$thumbnailSizes;
     }
+
+    #[\Override]
+    public function getImageData(?int $minWidth = null, ?int $minHeight = null): ?ImageData
+    {
+        if ($minWidth !== null || $minHeight !== null) {
```
dtdesign, line 327: You’re missing a check if this media is an image at all.

wcfsetup/install/files/lib/data/media/Media.class.php
```diff
@@ -310,4 +312,23 @@ public static function getThumbnailSizes()
...
+                }
+                if ($minHeight !== null && $minHeight > $this->getThumbnailHeight($size)) {
+                    continue;
+                }
+
+                return new ImageData($this->getThumbnailLink($size), $this->getThumbnailWidth($size), $this->getThumbnailHeight($size));
+            }
+        }
+
+        return new ImageData($this->getLink(), $this->width, $this->height);
```
dtdesign, line 347: This has the same issue as the attachment implementation: It will return an image even if it does not satisfy the width and/or height constraints.

dtdesign (changes requested):

dtdesign (approved):

PR #6147, opened 2024-12-20
Show last activity in user card

BurntimeX:


dtdesign (approved):

cadeyrn:
Thank you for bringing this information back! Unfortunately it doesn't look right if the location is a bit longer:

Bildschirmfoto 2024-12-21 um 23 22 12

dtdesign:
@cadeyrn This is the intended behavior when one of the columns overflows. This is using an existing layout that is aware of both sides being able to grow. Some languages are much more verbose than others and thus the “key” could become quite lengthy, forcing the “value” to wrap several times if we force them to stay in one row.

Instead this layout will allow the columns to wrap if there is not enough space to display them side-by-side without wrapping the contents.

PS: This update will also change the German term in this context from “Ort” to “Aufenthaltsort” to differentiate it from the “location” profile field. Keeping the columns next to each other could easily cause the value in your example to wrap 3 times.

PR #6148, opened 2024-12-20
Remove live-check of the `canHideOnlineStatus` permission

BurntimeX:
The live check means that the user's group assignments and permissions had to be loaded. This is particularly unfavorable in places where many users are displayed (such as the member list).

### DIFF
--- a/wcfsetup/install/files/lib/system/worker/UserRebuildDataWorker.class.php
+++ b/wcfsetup/install/files/lib/system/worker/UserRebuildDataWorker.class.php
@@ -349,7 +349,7 @@ final class UserRebuildDataWorker extends AbstractLinearRebuildDataWorker
     private function updateUserOnlineStatus(array $users): void
     {
         foreach ($users as $user) {
-            if (!$user->canViewOnlineStatus) {
+            if ($user->canViewOnlineStatus == UserProfile::ACCESS_EVERYONE) {
                 continue;
             }
             $userProfile = new UserProfile($user->getDecoratedObject());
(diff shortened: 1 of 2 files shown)


wcfsetup/install/files/lib/system/worker/UserRebuildDataWorker.class.php
```diff
@@ -334,4 +338,28 @@ public function execute()
...
+     * This method checks whether a user has restricted the visibility of their online status in the past,
+     * but has since lost the permission for it.
+     * In this case, the visibility of the online status is automatically set to default.
+     *
+     * @param UserEditor[] $users
+     */
+    private function updateUserOnlineStatus(array $users): void
+    {
+        foreach ($users as $user) {
+            if (!$user->canViewOnlineStatus) {
```
dtdesign: Please make this check compare against `UserProfile::ACCESS_EVERYONE`.

wcfsetup/install/files/lib/system/worker/UserRebuildDataWorker.class.php
```diff
@@ -334,4 +338,28 @@ public function execute()
...
+                continue;
+            }
+            $userProfile = new UserProfile($user->getDecoratedObject());
+            if ($userProfile->getPermission('user.profile.canHideOnlineStatus')) {
+                continue;
+            }
+
+            $user->updateUserOptions([
+                User::getUserOptionID('canViewOnlineStatus') => 0,
```
dtdesign: Please set the value to `UserProfile::ACCESS_EVERYONE` instead.

dtdesign (changes requested):

PR #6159, opened 2025-01-11
Fix the permission check for email notifications for new moderation queues

dtdesign:
The existing implementation implicitly validated the permissions using the current user. This causes issues when the reporting user does not have access to the moderation queue which would yield an empty object.

See https://www.woltlab.com/community/thread/310285-fehlermeldung-bei-reportmoderationqueueusernotificationevent-class-php/


BurntimeX (approved):
LGTM

PR #6168, opened 2025-01-17
Implement a JS WebP encoder/decoder and a PHP WebP decoder

dtdesign:
WoltLab Suite 6.1 introduced a new upload pipeline that yields resized images in the much more efficient WebP graphics format.

Part of the resizing process is to strip the EXIF meta data because the underlying nodeca/pica library does not support metadata and discards it. This not only means that the EXIF data is lost but also that any possible image rotation (which is stored in EXIF) is discarded alongside with it. We solved this issue in earlier versions by removing the EXIF data and embedded it into the generated JPEG again.

Unfortunately, there is no library around to accomplish the same in JS for WebP without relying on some CLI binaries.

This PR adds `WoltLab/Core/Image/WebP`, a pure JavaScript decoder/encoder for WebP (with focus on VP8X encoding) based on the official specs at https://developers.google.com/speed/webp/docs/riff_container.

To add insult to injury, PHP doesn’t support reading EXIF metadata from WebP either, let alone process it in any way. This PR also adds `wcf\system\image\WebPDecoder` which is a very basic port of our JS WebP decoder but focused on extracting the EXIF chunk only. Parsing the EXIF data is done by creating an empty 100 byte JPEG that serves as the host for the EXIF metadata, allowing us to utilize `\exif_read_data()`.

Fixes https://www.woltlab.com/community/thread/310622-bild-liegt-auf-der-seite/

### DIFF
--- a/ts/WoltLabSuite/Core/Image/ExifUtil.ts
+++ b/ts/WoltLabSuite/Core/Image/ExifUtil.ts
@@ -7,6 +7,8 @@
  * @woltlabExcludeBundle tiny
  */
 
+import { parseWebPFromBuffer } from "./WebP";
+
 const Tag = {
   SOI: 0xd8, // Start of image
   APP0: 0xe0, // JFIF tag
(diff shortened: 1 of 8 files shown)


BurntimeX (approved):
LGTM

dtdesign:
Replaced by

Issue #5911, opened 2024-05-07
RFC: Overhaul Caching: Eager and Async Rebuilds


dtdesign:
The current caching strategy has three major issues on runtime:

1. Resetting a cache causes the next request that needs this cache to trigger a synchronous rebuild.
2. Non-critical caches can sometimes be expensive to generate, causing dips in response times.
3. The same rebuild can take place simultaneously by concurrent requests. This is especially bad for expensive caches (see 2.).

Caches reset by (1.) are always the result of a change caused by the current request. Those requests also have a higher response time due to the amount of work being performed, for example, creating a new message, compared to typical read-heavy requests.

A prime example for the second issue (2.) is the statistics cache on the forum list that – among other things – counts the number of posts. This is an expensive query because in InnoDB this number is not denormalized, causing significant query runtimes for forums with a large number of posts.

The issue (3.) is mostly about efficiency.

## Eager Cache Rebuilds

We can solve (1.) and (3.) by immediately triggering a rebuild of a cache whenever it is being reset. As explained above, those requests are already slow in comparison due to the amount of work already been done.

Generally speaking, users don’t expect those actions to be lightning fast and adding 10, 20, 50 or even 100ms of latency doesn’t do much. At the same time, adding 100ms of latency to a page that usually loads in 50ms makes a huge difference.

If the cache does not exist, we must fall back to a synchronous rebuild of the cache just like before. This should only happen for legacy cache implementations or hard cache resets.

### “Pure” Cache Builder

An eager cache rebuild requires the implementation of the cache builder to be “pure”. This means that it MUST NOT rely on any (runtime) cache at any point because those could be stale. Any data must be fetched live from the _source of truth_, usually the database.

## Async Rebuilds

This is a **special type of cache** that differs in two ways from a regular cache:

1. Loading a stale cache is always acceptable.
2. A stale cache should trigger a unique background job to rebuild that cache.

We should consider a “refresh ahead” pattern for this type of cache that will preemptively queue up a cache rebuild when the cache is nearing its lifetime. For example, a cache with a TTL of 5 minutes could be rebuild after 4 minutes of its existence, reducing the likelihood of it becoming stale.

Cyperghost:
# Possible solutions

### Eager Cache

#### Interface Definition
```php
interface IEagerCache
{
    public function getCache(): mixed;
    public function reset(): void;
}
```

#### Abstract Implementation
```php
abstract class AbstractEagerCache implements IEagerCache
{
[...]

PR #6236, opened 2025-02-20
Eager caching

Cyperghost:
See #5911


wcfsetup/install/files/lib/system/cache/eager/data/LanguageCacheData.class.php
```diff
@@ -0,0 +1,67 @@
...
+        public readonly array $languages,
+        public readonly int $default,
+        /** @var array<string, LanguageCategory> */
+        public readonly array $categories,
+        /** @var array<int, string> */
+        public readonly array $categoryIDs,
+        public readonly bool $multilingualismEnabled
+    ) {
+        \assert($this->default > 0);
+        \assert(\array_key_exists($this->default, $this->languages));
```
dtdesign: This is merely a record class, it shouldn’t bother with the data making sense. Instead have the data validated in the cache itself including throwing custom exceptions when dealing with invalid states.

wcfsetup/install/files/lib/system/cache/eager/data/LanguageCacheData.class.php
```diff
@@ -0,0 +1,67 @@
...
+    {
+        return $this->categories[$categoryName] ?? null;
+    }
+
+    /**
+     * Returns the language category with the given category id.
+     */
+    public function getLanguageCategoryByID(int $languageCategoryID): ?LanguageCategory
+    {
+        return $this->categories[$this->categoryIDs[$languageCategoryID] ?? null] ?? null;
```
dtdesign: Please split this into two lines, this is too much clever code. Also it makes no difference to use a temporary variable because that is what is happening anyway for the index although implicit.

wcfsetup/install/files/lib/system/language/LanguageFactory.class.php
```diff
@@ -23,15 +23,8 @@ class LanguageFactory extends SingletonFactory
...
-     * @var mixed[]
      */
-    protected $cache;
-
-    /**
-     * initialized languages
-     * @var Language[]
-     */
-    protected $languages = [];
+    protected LanguageCache $cache;
```
dtdesign: Why would I want to deal with the `LanguageCache` if all I ever care for is `LanguageCacheData`? Just use that instead, it also makes the whole calls to it way less awkward to write (and read!).

Considering how the eager caches work there is some benefit of using a dynamic lookup but you end up paying for an extra function call with every access and given the fundamental importance of this class, changing the return values on runtime is possibly even dangerous.

On hindsight, I would even argue that holding a reference to the cache builder instead of the cached values is almost always a huge mistake. Therefore this should also be changed for the core objects!

wcfsetup/install/files/lib/system/language/LanguageFactory.class.php
```diff
@@ -96,7 +78,7 @@ public function getUserLanguage($languageID = 0)
     public function getLanguageByCode($languageCode)
     {
         // called within WCFSetup
-        if (empty($this->cache['codes'])) {
+        if (empty($this->cache->getCache()->codes)) {
```
dtdesign: This is always an opportunity to make improvements to the code, like throwing `empty()` into the garbage bin it belongs.

wcfsetup/install/files/lib/system/language/LanguageFactory.class.php
```diff
@@ -106,8 +88,8 @@ public function getLanguageByCode($languageCode)
             if (isset($row['languageID'])) {
                 return new Language($row['languageID']);
             }
-        } elseif (isset($this->cache['codes'][$languageCode])) {
-            return $this->getLanguage($this->cache['codes'][$languageCode]);
+        } elseif (isset($this->cache->getCache()->codes[$languageCode])) {
```
dtdesign: You already added helper methods to the cache, why not use them to handle this instead of manually poking in arrays structures?

wcfsetup/install/files/lib/system/cache/eager/LanguageCache.class.php
```diff
@@ -0,0 +1,76 @@
...
+        $languageList->readObjects();
+
+        $languages = $languageList->getObjects();
+        $default = 0;
+        $multilingualismEnabled = false;
+        $codes = [];
+        $countryCodes = [];
+
+        foreach ($languageList->getObjects() as $language) {
+            // default language
```
dtdesign: When copying code, please kick out pointless comments like this one that don’t add any value. In fact, all current comments in this functions are pointless.

dtdesign (changes requested):
I’m not entirely sure about the removal of the existing cache builders. Are you able to modify them so that they implicitly map to the new cache instead and thus preserving the backwards compatibility?

wcfsetup/install/files/lib/system/cache/eager/LanguageCache.class.php
```diff
@@ -63,6 +58,10 @@ protected function getCacheData(): LanguageCacheData
             $categoryIDs[$languageCategory->languageCategoryID] = $languageCategory->languageCategory;
         }
 
+        if (!isset($languages[$default])) {
+            throw new SystemException('No default language defined!');
```
dtdesign: Do not use `SystemException` for anything, it’s a bad exception to use. Since this is not an actionable error, it’s fine to use `\RuntimeException` instead.

wcfsetup/install/files/lib/system/cache/eager/data/LanguageCacheData.class.php
```diff
@@ -62,6 +60,68 @@ public function getLanguageCategory(string $categoryName): ?LanguageCategory
...
+            return null;
+        }
+
+        return $this->getLanguageCategory($categoryName);
+    }
+
+    /**
+     * Returns `true` if the language category with the given category name exists.
+     */
+    public function languageCategoryExists(string $categoryName): bool
```
dtdesign:
```suggestion
    public function hasLanguageCategory(string $categoryName): bool
```

wcfsetup/install/files/lib/system/cache/eager/data/LanguageCacheData.class.php
```diff
@@ -62,6 +60,68 @@ public function getLanguageCategory(string $categoryName): ?LanguageCategory
...
+    {
+        return \array_key_exists($categoryName, $this->categories);
+    }
+
+    /**
+     * Return all content languages.
+     *
+     * @return list<Language>
```
dtdesign:
```suggestion
     * @return array<int, Language>
```

The `list<…>` is incorrect for two reasons:

1. `$this->languages` is actually a mapping of the language id to the object.
2. `\array_filter()` preserves the keys so even if `$this->languages` was a zero-based array without holes, it would not longer be guaranteed afterwards.

`list<…>` indicates that the array only has numeric indices, starts at zero and has no holes in it. This is meant to provide type safety when iterating over it using a simple `for` loop.

wcfsetup/install/files/lib/system/cache/eager/data/LanguageCacheData.class.php
```diff
@@ -62,6 +60,68 @@ public function getLanguageCategory(string $categoryName): ?LanguageCategory
...
+     * Return all content languages IDs.
+     *
+     * @return list<int>
+     */
+    public function getContentLanguageIDs(): array
+    {
+        return \array_map(
+            static fn(Language $language) => $language->languageID,
+            $this->getContentLanguages()
+        );
```
dtdesign:
```suggestion
        return \array_keys($this->getContentLanguages());
```

wcfsetup/install/files/lib/system/cache/eager/data/LanguageCacheData.class.php
```diff
@@ -62,6 +60,68 @@ public function getLanguageCategory(string $categoryName): ?LanguageCategory
...
+        if ($categoryName === null) {
+            return null;
+        }
+
+        return $this->getLanguageCategory($categoryName);
+    }
+
+    /**
+     * Returns `true` if the language category with the given category name exists.
+     */
```
dtdesign:
```suggestion
```

This comment is redundant because this is literally what the name of the method and its return type implies.

wcfsetup/install/files/lib/system/language/LanguageFactory.class.php
```diff
@@ -45,119 +47,98 @@ protected function init()
...
+            $language = $this->cache->getLanguage($languageID);
             if ($language !== null) {
                 return $language;
             }
         }
 
         $languageID = $this->findPreferredLanguage();
 
-        return $this->getLanguage($languageID);
+        return $this->cache->getLanguage($languageID);
```
dtdesign, line 66: This code can be simplified by inverting the condition in line 58. This way the language only has to be looked up once.

wcfsetup/install/files/lib/system/language/LanguageFactory.class.php
```diff
@@ -45,119 +47,98 @@ protected function init()
...
      */
-    public function getLanguageByCode($languageCode)
+    public function getLanguageByCode(string $languageCode): ?Language
     {
-        // called within WCFSetup
-        if (empty($this->cache->getCache()->codes)) {
-            $sql = "SELECT  languageID
+        $language = $this->getLanguageByCode($languageCode);
+        if ($language === null) {
+            // called within WCFSetup
```
dtdesign: I don’t think this change is correct. This will now trigger a query whenever this method is called with an unknown code on runtime, regardless of the setup process.

wcfsetup/install/files/lib/system/language/LanguageFactory.class.php
```diff
@@ -45,119 +47,98 @@ protected function init()
...
+        $languageCode = self::getPreferredLanguage($this->cache->getLanguageCodes(), $defaultLanguageCode);
 
         // get language id of preferred language
-        foreach ($this->cache->getCache()->languages as $key => $language) {
+        foreach ($this->cache->languages as $key => $language) {
             if ($language->languageCode == $languageCode) {
                 return $key;
             }
         }
+
```
dtdesign, line 138: Don’t use `SystemException`, a `\RuntimeException` works here too.

wcfsetup/install/files/lib/system/cache/eager/data/LanguageCacheData.class.php
```diff
@@ -62,6 +60,68 @@ public function getLanguageCategory(string $categoryName): ?LanguageCategory
      */
     public function getLanguageCategoryByID(int $languageCategoryID): ?LanguageCategory
```
dtdesign: After checking the `LanguageFactory`, I think it’s safe to drop the `*Language*` from the methods for categories. It’s the language cache, what else should `getCategory()` return? :-)

wcfsetup/install/files/lib/system/cache/builder/AbstractLegacyCacheBuilder.class.php
```diff
@@ -0,0 +1,55 @@
...
+
+/**
+ * Legacy implementation of the ICacheBuilder interface that has been migrated to a new EagerCache or AsyncCache.
+ *
+ * @author      Olaf Braun
+ * @copyright   2001-2025 WoltLab GmbH
+ * @license     GNU Lesser General Public License <http://opensource.org/licenses/lgpl-license.php>
+ *
+ * @since       6.2
+ * @deprecated  6.2
```
dtdesign: No need for the indentation, it’s messed up anyway and nobody cares. Save yourself some time and just use a single space to separate values.

wcfsetup/install/files/lib/system/cache/builder/AbstractLegacyCacheBuilder.class.php
```diff
@@ -0,0 +1,55 @@
...
+ */
+abstract class AbstractLegacyCacheBuilder extends SingletonFactory implements ICacheBuilder
+{
+    protected array $cache = [];
```
dtdesign: _*screams in PHPStan @ level 6*_

dtdesign (changes requested):

wcfsetup/install/files/lib/system/language/LanguageFactory.class.php
Cyperghost: Won't this cause that if a languageID is passed to the function which does not exist, it will change its current behavior?

dtdesign: Uh, you’re correct. Please disregard my remark.

dtdesign (approved):

PR #6237, opened 2025-02-21
Eager and Tolerant Caching

Cyperghost:
Closes https://github.com/WoltLab/WCF/issues/5911


wcfsetup/install/files/lib/system/background/job/TolerantCacheRebuildBackgroundJob.class.php
```diff
@@ -0,0 +1,66 @@
...
+    #[\Override]
+    public function newInstance(): static
+    {
+        return new TolerantCacheRebuildBackgroundJob($this->cacheClass, $this->parameters);
+    }
+
+    #[\Override]
+    public function queueAgain(): bool
+    {
+        return \class_exists($this->cacheClass);
```
dtdesign: This kind of defeats the purpose of the tolerant cache principle because the `BackgroundQueueHandler` would queue it for execution in `retryAfter()` seconds. I don’t think it should ever re-queue itself and instead rely on being asked to refresh itself when necessary.

The whole idea is that this cache could go stale but still being served while it would be put at the end of the queue to be refreshed (unless it already exists of course).

The difference is that your current approach actively queues this cache for a rebuild, while always return `false` makes this an opt-in approach instead. This reduces the amount of times the cache is being rebuilt (allowing it to go stale more often) and reduces the load on the background queue.

wcfsetup/install/files/lib/system/background/job/TolerantCacheRebuildBackgroundJob.class.php
```diff
@@ -0,0 +1,66 @@
...
+    }
+
+    #[\Override]
+    public function retryAfter()
+    {
+        return (new $this->cacheClass(...$this->parameters))->getLifetime();
```
dtdesign: I don’t understand this decision, this value is meant to gracefully recover from transient failures by queueing it shortly afterwards. You’re basically removing this feature by setting the retry period to equal the maximum lifetime of the cache.

It’s arguably if the default value of 60 seconds is a proper choice here, maybe even something shorter like 30 seconds could work. But using the maximum lifetime doesn’t feel right here.

wcfsetup/install/files/lib/system/cache/builder/AbstractLegacyCacheBuilder.class.php
```diff
@@ -7,7 +7,7 @@
 use wcf\system\SingletonFactory;
 
 /**
- * Legacy implementation of the ICacheBuilder interface that has been migrated to a new EagerCache or AsyncCache.
+ * Legacy implementation of the ICacheBuilder interface that has been migrated to a new Eager-Cache or Tolerant-Cache.
```
dtdesign:
```suggestion
 * Provides a backwards compatible interface for ICacheBuilder classes that have been migrated to the eager or tolerant cache implementations.
```

wcfsetup/install/files/lib/system/cache/source/DiskCacheSource.class.php
```diff
@@ -157,4 +157,20 @@ private function readCache(string $filename): mixed
...
+
+        if (!\file_exists($filename)) {
+            return null;
+        }
+
+        if (!@\filesize($filename)) {
+            return null;
+        }
+
+        return \filemtime($filename);
```
dtdesign, line 174: How does the `\filemtime()` value represent the maximum lifetime of a file? You’re returning a timestamp here but in Redis the TTL in seconds is provided instead.

dtdesign (changes requested):

dtdesign (approved):

PR #6238, opened 2025-02-24
Overhaul caching

### LINKED ISSUE #5911: RFC: Overhaul Caching: Eager and Async Rebuilds

Cyperghost:
Closes #5911


dtdesign:
The Symfony cache component was considered as a replacement for our own implementation but falls short in a few ways:

- It does not permit us to use a stale cache, for example, serving slightly outdated stats is better than penalizing the response time for that users.
- The cache stampede protection is almost pointless because the algorithm it uses does not really yield a meaningful jitter for sub-second delta values. This actually defeats the purpose of an async rebuild ahead of time because the “early” rebuild will happen pretty much within <1s of its expiration.
- Symfony’s cache component is battle-tested but so is ours. There is no gain for us in that direction because we don’t start at zero here.

Overall the the code of Symfony’s cache component is a mess of excessive inheritance and traits stacked on top of each other. It sure saves a keystroke or two but at the expense of a complex behemoth for the arguably small scope of cache backends we’re supporting.

Issue #6181, opened 2025-01-27
Migrate existing list pages to grid views


BurntimeX:
ACP
- [x] `ACPSessionLogPage`
- [x] `AdListPage`
- [x] `ArticleListPage`
- [x] `AttachmentListPage`
- [x] `BBCodeListPage`
- [x] `BBCodeMediaProviderListPage`
- [x] `BoxListPage`
- [x] `CaptchaQuestionListPage`
- [x] `CronjobListPage`
- [x] `EmailLogListPage`
- [x] `LabelGroupListPage`
- [x] `LabelListPage`
- [x] `LanguageItemListPage`
- [x] `LanguageListPage`
- [x] #6288
- [x] `MenuListPage`
- [x] `NoticeListPage`
- [x] `PackageListPage` https://github.com/WoltLab/WCF/pull/6283
- [x] `PackageUpdateServerListPage`
- [x] `PageListPage`
- [x] `PaidSubscriptionListPage`
- [x] `PaidSubscriptionTransactionLogListPage`
- [x] `PaidSubscriptionUserListPage`
- [x] `ReactionTypeListPage`
- [x] `SmileyListPage`
- [x] `StyleListPage`
- [x] `TagListPage`
- [x] `TemplateGroupListPage`
- [x] `TemplateListPage`
- [x] `TrophyListPage`
- [x] `UserAuthenticationFailureListPage`
- [x] `UserGroupAssignmentListPage`
- [x] `UserGroupListPage`
- [x] `UserTrophyListPage`

Non ACP
- [x] `ModerationListPage`

ref https://github.com/WoltLab/WCF/pull/5990

PR #6248, opened 2025-03-18
Migrate `ReactionTypeListPage` to grid view

Cyperghost:
See https://github.com/WoltLab/WCF/issues/6181

### DIFF
--- a/wcfsetup/install/files/lib/bootstrap/com.woltlab.wcf.php
+++ b/wcfsetup/install/files/lib/bootstrap/com.woltlab.wcf.php
@@ -227,6 +227,11 @@ return static function (): void {
             $event->register(new \wcf\system\endpoint\controller\core\notices\DeleteNotice());
             $event->register(new \wcf\system\endpoint\controller\core\notices\GetShowOrder());
             $event->register(new \wcf\system\endpoint\controller\core\notices\ChangeShowOrder());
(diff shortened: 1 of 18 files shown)


wcfsetup/install/files/lib/bootstrap/com.woltlab.wcf.php
```diff
@@ -227,6 +227,11 @@ static function (\wcf\event\endpoint\ControllerCollecting $event) {
             $event->register(new \wcf\system\endpoint\controller\core\notices\DeleteNotice());
             $event->register(new \wcf\system\endpoint\controller\core\notices\GetShowOrder());
             $event->register(new \wcf\system\endpoint\controller\core\notices\ChangeShowOrder());
+            $event->register(new \wcf\system\endpoint\controller\core\reactions\types\Assignable());
+            $event->register(new \wcf\system\endpoint\controller\core\reactions\types\NotAssignable());
```
BurntimeX:
```suggestion
            $event->register(new \wcf\system\endpoint\controller\core\reactions\types\EnableType());
            $event->register(new \wcf\system\endpoint\controller\core\reactions\types\DisableType());
```
I would prefer these names so that it is consistent with other toggle actions.

BurntimeX (changes requested):

PR #6249, opened 2025-03-19
Migrate `UserTrophyListPage` to grid view

Cyperghost:
See https://github.com/WoltLab/WCF/issues/6181


wcfsetup/install/files/lib/system/gridView/admin/UserTrophyGridView.class.php
```diff
@@ -0,0 +1,180 @@
...
+            GridViewColumn::for("trophyID")
+                ->label("wcf.global.title")
+                ->renderer(
+                    new class extends DefaultColumnRenderer {
+                        #[\Override]
+                        public function render(mixed $value, DatabaseObject $row): string
+                        {
+                            \assert($row instanceof UserTrophy);
+
+                            return $row->getTrophy()->getTitle();
```
BurntimeX:
```suggestion
                            return StringUtil::encodeHTML($row->getTrophy()->getTitle());
```

BurntimeX (changes requested):

Issue #6133, opened 2024-12-03
Allow sorting of attachments via drag and drop


BurntimeX:
ref https://www.woltlab.com/community/thread/307681-dateianh%C3%A4nge-per-drag-and-drop-wieder-sortierbar-machen/

PR #6254, opened 2025-03-25
Implement sorting for attachments

Cyperghost:
Closes https://github.com/WoltLab/WCF/issues/6133

### DIFF
--- a/ts/WoltLabSuite/Core/Component/Attachment/Entry.ts
+++ b/ts/WoltLabSuite/Core/Component/Attachment/Entry.ts
@@ -34,6 +34,7 @@ function fileInitializationCompleted(element: HTMLElement, file: WoltlabCoreFile
       },
     });
   }
+  element.dataset.fileId = fileId.toString();
 
   const extraButtons: HTMLButtonElement[] = [];
 
--- a/ts/WoltLabSuite/Core/Component/Attachment/List.ts
+++ b/ts/WoltLabSuite/Core/Component/Attachment/List.ts
@@ -3,6 +3,9 @@ import { CkeditorDropEvent } from "../File/Upload";
 import { createAttachmentFromFile } from "./Entry";
 import { listenToCkeditor } from "../Ckeditor/Event";
 import { getTabMenu } from "../Message/MessageTabMenu";
+import Sortable from "sortablejs";
+import { showDefaultSuccessSnackbar } from "WoltLabSuite/Core/Component/Snackbar";
+import { postObject } from "WoltLabSuite/Core/Api/PostObject";
 
 function fileToAttachment(fileList: HTMLElement, file: WoltlabCoreFileElement, editor: HTMLElement): void {
   fileList.append(createAttachmentFromFile(file, editor));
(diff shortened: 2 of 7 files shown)


ts/WoltLabSuite/Core/Component/Attachment/Entry.ts
```diff
@@ -34,6 +34,7 @@ function fileInitializationCompleted(element: HTMLElement, file: WoltlabCoreFile
       },
     });
   }
+  element.dataset.fileId = fileId.toString();
```
dtdesign, line 37: Are you not able to access the `fileId` field of  `<woltlab-core-file>`?

ts/WoltLabSuite/Core/Component/Attachment/List.ts
```diff
@@ -44,6 +47,35 @@ export function setup(editorId: string): void {
...
+      if (event.oldIndex === event.newIndex) {
+        return;
+      }
+
+      const fileIDs = sortable.toArray().map(Number);
+      const context = JSON.parse(uploadButton.dataset.context!);
+
+      void postObject(`${window.WSC_RPC_API_URL}core/attachments/show-order`, { ...context, fileIDs }).then(() => {
+        showDefaultSuccessSnackbar();
+      });
```
dtdesign, line 75: There are a lot of things wrong with this:

- This will update the show order every time the order is changed.
- It does so even when the message editing is aborted thus without the user actually submitting anything.
- It’s not using a proper function for this like the remainder of the RPC API.
- You do not abort in-flight requests, possibly causing the show order to be incorrect on unstable networks.

ts/WoltLabSuite/Core/Component/Attachment/List.ts
```diff
@@ -44,6 +47,35 @@ export function setup(editorId: string): void {
     uploadButton.insertAdjacentElement("afterend", fileList);
   }
 
+  const sortable = new Sortable(fileList, {
+    direction: "vertical",
+    dataIdAttr: "data-file-id",
```
dtdesign, line 52: Why do you focus in the `fileID` here? You are sorting attachments, not files, the actual underlying file is an implementation detail.

wcfsetup/install/files/lib/system/endpoint/controller/core/attachments/ChangeShowOrder.class.php
```diff
@@ -0,0 +1,99 @@
...
+        if (!$attachmentHandler->canUpload()) {
+            throw new PermissionDeniedException();
+        }
+
+        foreach ($attachmentHandler->getAttachmentList()->getObjects() as $attachment) {
+            if (!\in_array($attachment->fileID, $fileIDs)) {
+                throw new PermissionDeniedException();
+            }
+        }
+    }
```
dtdesign, line 59: Your validation falls short because you do not validate if the user is allowed to edit these attachments at all.

wcfsetup/install/files/lib/system/endpoint/controller/core/attachments/ChangeShowOrder.class.php
```diff
@@ -0,0 +1,99 @@
...
+{
+    public function __construct(
+        /** @var non-empty-string */
+        public readonly string $objectType,
+        /** @var non-negative-int */
+        public readonly int $objectID,
+        /** @var non-negative-int */
+        public readonly int $parentObjectID,
+        public readonly string $tmpHash,
+        /** @var int[] */
```
dtdesign, line 95:
```suggestion
        /** @var list<positive-int> */
```

Your code relies on the `fileIDs` being a zero-based list of integers without gaps in line 73.

wcfsetup/install/files/style/ui/fileList.scss
```diff
@@ -106,3 +106,7 @@ woltlab-core-file img {
 .woltlabCoreFileUpload__input::-webkit-file-upload-button {
 	cursor: pointer;
 }
+
+.attachment__item:hover {
+	cursor: pointer;
+}
```
dtdesign, line 112: This is not gated by `@media (hover: hover)`, also why is this a “pointer” cursor? You are moving items around, this should be `cursor: move`.

That said, attachments should ideally have a drag handle because otherwise it will interfere with the buttons. Also, touch users on mobile will have a bad time because browsers make it difficult to distinguish between a swipe to scroll and a swipe to drag an item around.

dtdesign (changes requested):
I am not happy with the state of this PR. There are quite a few design decisions I cannot get behind and I don’t believe that AJAX is the correct call here. This should be submitted alongside the edited message like before which resolves a lot of the issues currently being present.

I suggest opening a new branch/PR to keep this separated.

PR #6255, opened 2025-03-26
Migrate `SortedUserCache` to TolerantCache

Cyperghost:


wcfsetup/install/files/lib/system/box/UserListBoxController.class.php
```diff
@@ -28,13 +27,19 @@ class UserListBoxController extends AbstractDatabaseObjectListBoxController
...
-        'activityPoints' => MostActiveMembersCacheBuilder::class,
-        'likesReceived' => MostLikedMembersCacheBuilder::class,
-        'registrationDate' => NewestMembersCacheBuilder::class,
-    ];
+    public $cacheBuilders = [];
+
+    /**
+     * maps special sort fields to tolerant caches
+     *
+     * @var array<string, callable(int $limt, string $sortOrder, Condition[] $conditions): AbstractTolerantCache<list<int>>>
```
dtdesign:
```suggestion
     * @var array<string, callable(int $limit, string $sortOrder, Condition[] $conditions): AbstractTolerantCache<list<int>>>
```

dtdesign (changes requested):
Otherwise LGTM

PR #6256, opened 2025-03-27
Migrate `TagCache` and `TypedTagCache` to TolerantCache

Cyperghost:


wcfsetup/install/files/lib/system/cache/tolerant/TagCloudCache.class.php
```diff
@@ -0,0 +1,137 @@
...
+    private static function compareTags(TagCloudTag $tagA, TagCloudTag $tagB): int
+    {
+        if ($tagA->counter > $tagB->counter) {
+            return -1;
+        }
+        if ($tagA->counter < $tagB->counter) {
+            return 1;
+        }
+
+        return 0;
```
dtdesign:
```suggestion
        return $tagB->counter <=> $tagA->counter;
```

wcfsetup/install/files/lib/system/cache/tolerant/TagCloudCache.class.php
```diff
@@ -0,0 +1,137 @@
...
+        if ($this->objectTypeIDs === []) {
+            return [];
+        }
+
+        return $this->fetchTags($this->fetchTagIDs());
+    }
+
+    /**
+     * @param list<int> $tagIDs
+     *
```
dtdesign:
```suggestion
     * @param array<int, int> $tagIDs
```

A `list<…>` is a zero-based numeric array without gaps. The return value of `fetchTagIDs()` is a map, not a list anyway. Despite that, `\array_keys()` in line 70 would have be an odd choice for a zero-based array.

wcfsetup/install/files/lib/system/cache/tolerant/TagCloudCache.class.php
```diff
@@ -0,0 +1,137 @@
...
+        $statement->execute($conditionBuilder->getParameters());
+
+        $tags = [];
+        while ($row = $statement->fetchArray()) {
+            $row['counter'] = $tagIDs[$row['tagID']];
+            $tags[$row['name']] = new TagCloudTag(new Tag(null, $row));
+        }
+
+        // sort by counter
+        \uasort($tags, self::compareTags(...));
```
dtdesign: If you need a comment for this, then the name of the method is poorly chosen. Maybe rename it to `sortTagsByCounter()` and get rid of the comment?

dtdesign (changes requested):

dtdesign (approved):

Issue #6268, opened 2025-04-15
Quote buttons have wrong position


BurntimeX:
The buttons are now always positioned on the left below the entire paragraph. Previously, they were positioned directly below the selected text passage.

<img width="1064" alt="Image" src="https://github.com/user-attachments/assets/c4c45c2d-2e00-448a-bd99-203c7b027414" />

PR #6269, opened 2025-04-16
Align the quote buttons directly below the selected text

Cyperghost:
Closes https://github.com/WoltLab/WCF/issues/6268

### DIFF
--- a/ts/WoltLabSuite/Core/Component/Quote/Message.ts
+++ b/ts/WoltLabSuite/Core/Component/Quote/Message.ts
@@ -500,7 +500,7 @@ function removeSelection(): void {
 }
 
 function alignQuoteButtons(content: HTMLElement): void {
-  const coordinates = getElementBoundaries(window.getSelection())!;
+  const coordinates = getElementBoundaries(window.getSelection());
   const dimensions = { height: copyQuote.offsetHeight, width: copyQuote.offsetWidth };
   let left = (coordinates.right - coordinates.left) / 2 - dimensions.width / 2 + coordinates.left;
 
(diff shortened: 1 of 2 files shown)


dtdesign (commented):
I don’t understand the motivation behind this change. Why don’t you pass `{ horizontal: "center" }` as the 3rd parameter to `UiAlignment.set()` instead which does exactly the same?

Cyperghost:
> I don’t understand the motivation behind this change. Why don’t you pass `{ horizontal: "center" }` as the 3rd parameter to `UiAlignment.set()` instead which does exactly the same?

Not exactly, we cannot pass a direct a text reference to `UiAlignment`, only `HTMLElement`.
That is, if the selected text is at the top of a `p` that contains a long text, the button will not be displayed directly below the selected text.
Which is the case in version 6.1.

ts/WoltLabSuite/Core/Component/Quote/Message.ts
```diff
@@ -492,3 +498,42 @@ function removeSelection(): void {
     selection.removeAllRanges();
   }
 }
+
+function alignQuoteButtons(content: HTMLElement): void {
+  const coordinates = getElementBoundaries(window.getSelection())!;
```
dtdesign: What is the point of the not-null-assertion if there are cases in which this could indeed be `null`? Either you should handle this or `getElementBoundaries()` should throw if this is considered to be an invalid state.

dtdesign (changes requested):

dtdesign (approved):

PR #6272, opened 2025-04-17
Allow empty value in `RadioButtonFormField`

### LINKED ISSUE #6025: `RadioButtonFormField` is always required
Expected behaviour: When a `RadioButtonFormField` is not required it should allow an empty value.

Cyperghost:
Closes #6025

### DIFF
--- a/com.woltlab.wcf/templates/shared_radioButtonFormField.tpl
+++ b/com.woltlab.wcf/templates/shared_radioButtonFormField.tpl
@@ -1,3 +1,16 @@
+{if !$field->isRequired()}
+	<label>
+		<input {*
+			*}type="radio" {*
+			*}name="{$field->getPrefixedId()}" {*
+			*}value="" {*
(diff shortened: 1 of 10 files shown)


com.woltlab.wcf/templates/shared_radioButtonFormField.tpl
```diff
@@ -1,3 +1,16 @@
+{if !$field->isRequired()}
+	<label>
+		<input {*
+			*}type="radio" {*
+			*}name="{$field->getPrefixedId()}" {*
+			*}value="" {*
+			*}{if !$field->getFieldClasses()|empty} class="{implode from=$field->getFieldClasses() item='class' glue=' '}{$class}{/implode}"{/if}{*
+			*}{if $field->getValue() !== null && $field->getValue() == ''} checked{/if}{*
```
dtdesign: Right now the `null` value is not considered in any case, shouldn’t the code look like this instead? (I haven’t tested it, please verify first!)

```suggestion
			*}{if $field->getValue() === null || $field->getValue() === ''} checked{/if}{*
```

dtdesign (changes requested):

PR #6275, opened 2025-04-29
Multilingual Content (PoC)

### LINKED ISSUE #6109: Abstraction for Multilingual Content
There are multiple different implementations for multilingual content and they are all either flawed to some extent or quite limited or both. Some store there values in phrases, others use separate tables but are inconsistent on how the ids are mapped. Furthermore being able to mix monolingual and multilingual content creates a lot of issues when sorting, filtering or generally trying to work with those values because it involves some kind of magic, for example, the value could be a plain value or the name of a phrase.

The only solution forward is a consistent abstraction that handles this in a uniform way that does not rely on phrases (which was a stop-gap solution in itself) and provides convenient helper methods to work with them. It needs to solve the following problems:

* Any localized value must be stored separately regardless of the number of available languages.
* Monolingual content must be represented as a multilingual content with only a single localization.
* Displaying, sorting and filtering by values must be deterministic.
  * Available localizations can be inconsistent, the value must be determined in this order:
    * Requested language.
    * Default language.
    * Values for the language with the lowest `languageID`.

Determining the values can be done through a sub select using either `CASE … THEN` or by assigning the `languageID` a value by preference. The latter could be achieved by setting `-2` for preferred language, `-1` for default language and the actual languageID for everything else followed by then selecting the lowest value.

This requires extra tables for any such content and should be entirely managed like it is already the case with the `*_search_index` tables. `objectID` and `languageID` need to be fixed and all other columns could be represented through a list of `IDatabaseTableColumn`. Additionally this would set up a foreign key for both the `languageID` as well as the `objectID` with the latter referencing the original object.

We need to explore if this is feasible through specialized helper methods to avoid API consumers having to write custom queries to work with these values.

Cyperghost:
Closes #6109

### TODO

- [x] Migration script for the current `rankTitle` into `wcf1_user_rank_content`
  - [ ] Check if `rankTitle` is still used somewhere
  - [ ] Can we remove the old column `rankTitle`?
- [ ] Ensure that `UserRankCache` is used to output the user rank if possible
- [x] `BadgeColorFormField` must be changed to work with more then one field
- [x] ~Check if the changes in `UserProfile::getRank()` do not cause any major problems in external dependencies/plugins~
- [ ] Migration script for the current `buttonLabel` into `wcf1_bbcode_content`
  - [ ] Check if `buttonLabel` is still used somewhere
  - [ ] Can we remove the old column `buttonLabel`?
- [ ] Add eager cache for BB Codes
- [ ] Rework for the `shared_multipleLanguageInputJavascript` template?


wcfsetup/install/files/lib/data/TMultilingualContentList.class.php
```diff
@@ -0,0 +1,110 @@
...
+ * @copyright 2001-2025 WoltLab GmbH
+ * @license GNU Lesser General Public License <http://opensource.org/licenses/lgpl-license.php>
+ * @since 6.2
+ *
+ * @template-covariant TDatabaseObject of DatabaseObject|DatabaseObjectDecorator<DatabaseObject>
+ * @mixin DatabaseObjectList<TDatabaseObject>
+ */
+trait TMultilingualContentList
+{
+    public readonly int $langaugeID;
```
dtdesign: Properties in Traits are a complete mess and must be avoided like the plague. In general traits should be completely self-contained because otherwise it creates all sort of strange external dependencies that are not immediately visible.

I dislike everything about this trait. It hides important logic behind an innocent use statement but touches pretty much all relevant methods. I am aware that we have this mess already present in some other parts of the software but I rather cut back the weeds and not add more of it.

PS: There is a typo, it should read `lang>ua<geID`.

wcfsetup/install/files/lib/acp/form/UserRankAddForm.class.php
```diff
@@ -126,15 +143,39 @@ protected function finalizeForm()
...
 
         $this->form->getDataHandler()
+            ->addProcessor(
+                new class('content', ['title']) extends MultilingualFormDataProcessor {
+                    #[\Override]
+                    public function processObjectData(IFormDocument $document, array $data, IStorableObject $object)
+                    {
+                        \assert($object instanceof UserRank);
+
+                        $sql = "SELECT    content.title, language.languageCode
```
dtdesign: Why do you have to join the language table, we already have all the information readily available in `LanguageFactory::getInstance()`?

wcfsetup/install/files/lib/acp/form/UserRankAddForm.class.php
```diff
@@ -60,20 +63,34 @@ public function createForm()
...
-                    TextFormField::create('rankTitle')
-                        ->label('wcf.acp.user.rank.title')
-                        ->i18n()
-                        ->languageItemPattern('wcf.user.rank.\w+')
-                        ->required(),
                     BadgeColorFormField::create('cssClassName')
                         ->label('wcf.acp.user.rank.cssClassName')
                         ->description('wcf.acp.user.rank.cssClassName.description')
-                        ->textReferenceNodeId('rankTitle')
+                        // TODO this must be changed to work with the new `MultilingualContainer`
```
dtdesign: TODO

wcfsetup/install/files/acp/database/update_com.woltlab.wcf_6.2.php
```diff
@@ -46,5 +51,35 @@
...
+        ]),
+    PartialDatabaseTable::create('wcf1_user_rank')
+        ->columns([
+            DefaultFalseBooleanDatabaseTableColumn::create('isMultilingual'),
+        ]),
+    DatabaseTable::create('wcf1_user_rank_content')
+        ->columns([
+            NotNullInt10DatabaseTableColumn::create('rankID'),
+            IntDatabaseTableColumn::create('languageID')
+                ->length(10)
```
dtdesign:
```suggestion
```

Do not specify the length of _new_ integer fields, it is meaningless these days and is something we will eventually get rid of entirely.

wcfsetup/install/files/acp/database/update_com.woltlab.wcf_6.2.php
```diff
@@ -46,5 +51,35 @@
...
+                ->defaultValue(null),
+            NotNullVarchar255DatabaseTableColumn::create('title')
+                ->defaultValue(''),
```
dtdesign: What is the point of this default value?

wcfsetup/install/files/acp/database/update_com.woltlab.wcf_6.2.php
```diff
@@ -46,5 +51,35 @@
...
         ])
+        ->indices([
+            DatabaseTableIndex::create('id')
+                ->columns(['rankID', 'languageID'])
+                ->type(DatabaseTableIndex::UNIQUE_TYPE),
```
dtdesign: This doesn’t work with nullable columns.

Also, this should rather be the primary key, because every table needs one and MySQL will implicitly assign one if it is missing. A primary key is unique by its very definition so that solves this requirement too.

wcfsetup/install/files/acp/database/update_com.woltlab.wcf_6.2.php
```diff
@@ -46,5 +51,35 @@
...
+    PartialDatabaseTable::create('wcf1_user_rank')
+        ->columns([
+            DefaultFalseBooleanDatabaseTableColumn::create('isMultilingual'),
+        ]),
+    DatabaseTable::create('wcf1_user_rank_content')
+        ->columns([
+            NotNullInt10DatabaseTableColumn::create('rankID'),
+            IntDatabaseTableColumn::create('languageID')
+                ->length(10)
+                ->defaultValue(null),
```
dtdesign: I don’t want this to be nullable (it also breaks your unique key, see below) and I don’t see why it should be nullable at all.

If the value should be monolingual then we can simply assign it to the default language id and call it a day. The query will make it work and we will have a clean transition between both modes without having to deal with NULL values.

wcfsetup/install/files/lib/data/user/rank/UserRank.class.php
```diff
@@ -24,11 +25,17 @@
  * @property-read   int $repeatImage        number of times the rank image is displayed
  * @property-read   int $requiredGender     numeric representation of the user's gender required for the user rank (see `UserProfile::GENDER_*` constants) or 0 if no specific gender is required
  * @property-read   int $hideTitle      hides the generic title of the rank, but not custom titles, `0` to show the title at all times
+ * @property-read   int $isMultilingual
```
dtdesign: Do we _really_ need this? Following up my remarks about the database structure, we can remove this distinction by just relying on the available localizations. If there is just one row to play with, then its monolingual, otherwise it inherently is multilingual?

wcfsetup/install/files/lib/data/user/rank/UserRank.class.php
```diff
@@ -54,7 +61,39 @@ public function getImage()
...
+    {
+        if (isset($this->content)) {
+            return;
+        }
+
+        $sql = "SELECT languageID, title 
+                FROM   wcf1_user_rank_content
+                WHERE  rankID = ?";
+
+        $statement = WCF::getDB()->prepare($sql, 1);
```
dtdesign: There is a lot wrong with this one:

1. You are limiting the query to at most one result, how is it supposed to ever return more than one value?
2. You are pulling an arbitrary row from the database because you specify no conditions to rank the candidates.
3. Why do you query the data live anyway, this will at worst cause _n_ extra queries per page given _n_ visible unique ranks. Why are you not using the `UserRankCache` for this?

wcfsetup/install/files/lib/data/user/rank/UserRankEditor.class.php
```diff
@@ -30,5 +32,9 @@ class UserRankEditor extends DatabaseObjectEditor implements IEditableCachedObje
     public static function resetCache()
     {
         UserRankCacheBuilder::getInstance()->reset();
+
+        foreach (LanguageFactory::getInstance()->getLanguages() as $language) {
+            (new UserRankCache($language->languageID))->rebuild();
```
dtdesign: This will create a lot of different caches and when applying this strategy to all sorts of i18n content, this will quickly yield a metric ton of unique cache items.

It also causes the weird cases where a localization is not available in the target language which requires all sort of shenanigans to resolve. The overhead of storing the localized values within the cache is rather small and at the same time solves all issues when resolving values at runtime?

wcfsetup/install/files/lib/data/user/rank/UserRankList.class.php
```diff
@@ -11,6 +12,13 @@
...
+ * @template-covariant TDatabaseObject of UserRank|DatabaseObjectDecorator<UserRank> = UserRank
+ * @extends DatabaseObjectList<TDatabaseObject>
  */
-class UserRankList extends DatabaseObjectList {}
+class UserRankList extends DatabaseObjectList
+{
+    /**
+     * @inheritDoc
+     */
+    public $className = UserRank::class;
```
dtdesign: The class name is implicitly derived on runtime, why do we need this?

wcfsetup/install/files/lib/data/user/rank/ViewableUserRank.class.php
```diff
@@ -0,0 +1,32 @@
...
+ * @author Olaf Braun
+ * @copyright 2001-2025 WoltLab GmbH
+ * @license GNU Lesser General Public License <http://opensource.org/licenses/lgpl-license.php>
+ *
+ * @property-read ?string $title
+ * @property-read ?int $languageID
+ * @mixin   UserRank
+ * @extends DatabaseObjectDecorator<UserRank>
+ */
+class ViewableUserRank extends DatabaseObjectDecorator implements ITitledObject
```
dtdesign: I don’t understand the point of this class? Why are you unable to merge this logic into the DBO itself?

wcfsetup/install/files/lib/data/user/rank/ViewableUserRankList.class.php
```diff
@@ -0,0 +1,68 @@
...
+
+/**
+ * @author Olaf Braun
+ * @copyright 2001-2026 WoltLab GmbH
+ * @license GNU Lesser General Public License <http://opensource.org/licenses/lgpl-license.php>
+ * @since 6.2
+ *
+ * @extends UserRankList<ViewableUserRank>
+ */
+class ViewableUserRankList extends UserRankList
```
dtdesign: I believe this whole class is pointless if you instead add all the localization data into the cache and then dynamically select the proper value. Removes the entire need to a dedicated list class and I strongly dislike those `Viewable*` classes in general, they create more problems than they solve.

dtdesign (changes requested):

PR #6277, opened 2025-04-30
Form options

BurntimeX:
Form options are similar to custom option but are based entirely on FormBuilder fields and also allow filtering in the list view.

As a reference implementation, the input fields in the contact form in this PR have been converted accordingly.


wcfsetup/install/files/acp/database/update_com.woltlab.wcf_6.2_step1.php
```diff
@@ -46,5 +49,14 @@
                 ->referencedTable('wcf1_file')
                 ->referencedColumns(['fileID'])
                 ->onDelete('SET NULL'),
-        ])
+        ]),
+    PartialDatabaseTable::create('wcf1_contact_option')
+        ->columns([
+            MediumtextDatabaseTableColumn::create('configurationData'),
```
dtdesign: Is there any reason not to call this `configuration`, the `Data` suffix sounds a bit redundant to me?

wcfsetup/install/files/acp/database/update_com.woltlab.wcf_6.2_step1.php
```diff
@@ -46,5 +49,14 @@
...
+        ]),
+    PartialDatabaseTable::create('wcf1_file')
+        ->columns([
+            IntDatabaseTableColumn::create('uploadTime')
+                ->length(10),
```
dtdesign: Please do not specify a length. This is a deprecated feature for `*INT`, all `INT` have an implicit width of 11 regardless of what you specify.

wcfsetup/install/files/acp/update_com.woltlab.wcf_6.2_contactOptions.php
```diff
@@ -0,0 +1,57 @@
...
+ * This script converts existing contact options to the new form option system.
+ */
+
+use wcf\data\contact\option\ContactOptionEditor;
+use wcf\data\contact\option\ContactOptionList;
+use wcf\util\JSON;
+use wcf\util\OptionUtil;
+
+$contactOptionList = new ContactOptionList();
+$contactOptionList->readObjects();
```
dtdesign, line 13: There is no condition to check if there is an configuration already set, which violates the principle of making those scripts idempotent.

This is a problem if the upgrade fails later on and is restarted because now the original columns may no longer exists and this script yields garbage data.

wcfsetup/install/files/lib/acp/form/AbstractFormOptionAddForm.class.php
```diff
@@ -0,0 +1,121 @@
...
+        }
+
+        return $formFields;
+    }
+
+    protected function getOptionTypeFormField(): SelectFormField
+    {
+        return SelectFormField::create('optionType')
+            ->label('wcf.form.option.optionType')
+            ->immutable($this->formAction != 'create')
```
dtdesign:
```suggestion
            ->immutable($this->formAction !== 'create')
```

wcfsetup/install/files/lib/system/form/option/formatter/BooleanFormatter.class.php
```diff
@@ -0,0 +1,24 @@
...
+final class BooleanFormatter implements IFormOptionFormatter
+{
+    #[\Override]
+    public function format(string $value, int $languageID, array $configurationData): string
+    {
+        if (!$value) {
+            return '';
+        }
+
+        return '✔️';
```
dtdesign, line 22: This looks a bit like a placeholder or is it intentional?

wcfsetup/install/files/lib/system/form/option/BooleanFormOption.class.php
```diff
@@ -0,0 +1,42 @@
...
+    }
+
+    #[\Override]
+    public function getFormField(string $id, array $configurationData = []): AbstractFormField
+    {
+        return BooleanFormField::create($id);
+    }
+
+    #[\Override]
+    public function getFormatter(): BooleanFormatter
```
dtdesign: Narrowing the return type is dangerous if the class is not final. Here (and in all other `*FormOption`) the formatter provided resolves to a class marked as final. It is therefore impossible to return anything else in derived classes.

If you do not intend to lock down the formatter class, you should avoid narrowing it down and just use the interface as the return type.

wcfsetup/install/files/lib/system/form/option/formatter/CurrencyFormatter.class.php
```diff
@@ -0,0 +1,36 @@
...
+ * @copyright   2001-2025 WoltLab GmbH
+ * @license     GNU Lesser General Public License <http://opensource.org/licenses/lgpl-license.php>
+ * @since       6.2
+ */
+class CurrencyFormatter implements IFormOptionFormatter
+{
+    #[\Override]
+    public function format(string $value, int $languageID, array $configurationData): string
+    {
+        $showDecimals = $value % 100 != 0;
```
dtdesign:
```suggestion
        $showDecimals = $value % 100 !== 0;
```

Besides, I’m not sure if this is a good decision at all. Having fixed decimal places is common for currencies because now you end up `99.99`, `100` and `100.01` mixed together?

dtdesign (changes requested):

PR #6283, opened 2025-05-07
Migrate package list to grid view

BurntimeX:
see #6181

### DIFF
--- a/wcfsetup/install/files/acp/js/WCF.ACP.js
+++ b/wcfsetup/install/files/acp/js/WCF.ACP.js
@@ -474,9 +474,7 @@ WCF.ACP.Package.Uninstallation = WCF.ACP.Package.Installation.extend({
 		this._elements = elements;
 		this._packageID = 0;
 		
-		//if (this._elements !== undefined && this._elements.length) {
-			this._super(0, 'UninstallPackage');
-		//}
+		this._super(0, 'UninstallPackage');
 	},
 	
 	/**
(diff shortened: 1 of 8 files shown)


wcfsetup/install/files/lib/system/gridView/admin/PackageGridView.class.php
```diff
@@ -0,0 +1,135 @@
...
+                        {
+                            $renderedValue = parent::render($value, $row);
+
+                            \assert($row instanceof Package);
+
+                            if ($row->isTainted()) {
+                                $title = WCF::getLanguage()->getDynamicVariable("wcf.acp.package.application.isTainted");
+                                $renderedValue .= <<<HTML
+                                    <span class="jsTooltip" title="{$title}">
+                                        <fa-icon name="triangle-exclamation"></fa-icon>
```
dtdesign: Please avoid using `<fa-icon>` directly, this should be created using the `FontAwesomeIcon` helper class.

wcfsetup/install/files/acp/js/WCF.ACP.js
```diff
@@ -474,9 +474,9 @@ WCF.ACP.Package.Uninstallation = WCF.ACP.Package.Installation.extend({
 		this._elements = elements;
 		this._packageID = 0;
 		
-		if (this._elements !== undefined && this._elements.length) {
+		//if (this._elements !== undefined && this._elements.length) {
```
dtdesign: This code (and in line 479, 501) is commented out. You should probably just remove it.

dtdesign (changes requested):

PR #6284, opened 2025-05-08
Use time() instead of TIME_NOW

ChristopherWalz:
Sending hundreds of emails in a long running task might take some time (depending on the provider and other work in the task) which could result in some emails having an "old" timestamp. An email might arrive at 12:53 (actual time) but have a timestamp of 12:50, because of the long running task.

### DIFF
--- a/wcfsetup/install/files/lib/system/email/Email.class.php
+++ b/wcfsetup/install/files/lib/system/email/Email.class.php
@@ -170,7 +170,7 @@ class Email
     public function getDate()
     {
         if ($this->date === null) {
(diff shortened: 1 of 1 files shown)


dtdesign:
I can see the point you’re trying to make but I don’t understand what kind of problem is being solved here. Let’s imagine that the email spends some time in the queue of your mail server for whatever reason. Your timestamp might now be more recent, but when the email is received it is still behind some time.

This is especially true for email servers that are required to be tolerant about reaching the destination server and usually use an exponential back-off to delay the delivery.

PS: `$this->date = new \DateTime('now');` would be more reasonable.

PR #6285, opened 2025-05-08
Allow the display of deleted items via plugin

BurntimeX:


com.woltlab.wcf/templates/deletedItemsBox.tpl
```diff
@@ -0,0 +1,15 @@
+<section class="box" data-static-box-identifier="com.woltlab.wcf.DeletedContentListMenu">
+	<h2 class="boxTitle">{lang}wcf.moderation.deletedContent.objectTypes{/lang}</h2>
+	
+	<div class="boxContent">
+		<nav>
+			<ul class="boxMenu">
+				{foreach from=$types item=type}
+					<li{if $type->id == $activeId} class="active"{/if}>
```
dtdesign:
```suggestion
					<li{if $type->id === $activeId} class="active"{/if}>
```

wcfsetup/install/files/lib/event/moderation/DeletedItemsCollecting.class.php
```diff
@@ -0,0 +1,61 @@
...
+ *
+ * @author      Marcel Werk
+ * @copyright   2001-2025 WoltLab GmbH
+ * @license     GNU Lesser General Public License <http://opensource.org/licenses/lgpl-license.php>
+ * @since       6.2
+ */
+final class DeletedItemsCollecting implements IPsr14Event
+{
+    /**
+     * @var DeletedItems[]
```
dtdesign:
```suggestion
     * @var list<DeletedItems>
```

The difference is that `list<…>` guarantees that this is a continuous zero-based array.

wcfsetup/install/files/lib/event/moderation/DeletedItemsCollecting.class.php
```diff
@@ -0,0 +1,61 @@
...
+        $this->loadLegacyProviders();
+    }
+
+    public function register(DeletedItems $type): void
+    {
+        $this->types[] = $type;
+    }
+
+    /**
+     * @return DeletedItems[]
```
dtdesign:
```suggestion
     * @return list<DeletedItems>
```

wcfsetup/install/files/lib/system/moderation/DeletedItemsBoxComponent.class.php
```diff
@@ -0,0 +1,45 @@
...
+            'deletedItemsBox',
+            [
+                'types' => $this->getTypes(),
+                'activeId' => $this->activeId,
+            ],
+        );
+    }
+
+    /**
+     * @return DeletedItems[]
```
dtdesign:
```suggestion
     * @return list<DeletedItems>
```

dtdesign (changes requested):

PR #6282, opened 2025-05-07
Multilingual Content (PoC)

Cyperghost:
Closes #6109

## TODO
- [ ] Add a helper function for the subquery

## Performance tests

Different queries were tested, these have different performance with larger tables.
`LATERAL` is currently not available in MariaDB and has also shown poor performance in MySQL and was therefore not investigated further.
An `UNION` query returned a slightly worse result than `ROW_NUMBER` and was not investigated further.
The best results were achieved with simple subqueries on the content table `wcf1_test_content`.

Test Table
```sql
CREATE TABLE wcf1_test (
    testID INT NOT NULL AUTO_INCREMENT,
    time INT NOT NULL,
    PRIMARY KEY (testID)
);

CREATE TABLE wcf1_test_content (
    contentID INT NOT NULL AUTO_INCREMENT,
    testID INT NOT NULL,
    languageID INT,
    title VARCHAR(255) NOT NULL,
    content VARCHAR(255) NOT NULL,

    PRIMARY KEY(contentID),
   KEY id (testID, languageID)
);

SELECT COUNT(*) FROM wcf1_test;
 -- 100000
SELECT COUNT(*) FROM wcf1_test_content;
 -- 149860
```

<details>
    <summary>Queries only sorted by title</summary>

Query with simple subqueries
```SQL
SELECT   test.*
FROM     wcf1_test test
ORDER BY (
    SELECT   title
    FROM     wcf1_test_content
    WHERE    testID = test.testID
    ORDER BY CASE
        WHEN languageID = 1 THEN -2
        WHEN languageID = 2 THEN -1
        ELSE languageID
    END ASC
    LIMIT    1
) DESC, test.testID ASC
LIMIT    20;
```

Output
```
-> Limit: 20 row(s)  (cost=10084 rows=20) (actual time=307..307 rows=20 loops=1)
    -> Sort: (select #2) DESC, test.testID, limit input to 20 row(s) per chunk  (cost=10084 rows=100275) (actual time=307..307 rows=20 loops=1)
        -> Table scan on test  (cost=10084 rows=100275) (actual time=2.54..13.5 rows=100000 loops=1)
-> Select #2 (subquery in projection; dependent)
    -> Limit: 1 row(s)  (cost=0.524 rows=1) (actual time=0.00264..0.00266 rows=1 loops=100000)
[...]

PR #6287, opened 2025-05-09
Multilingual Content table for BBCodes

Cyperghost:
Required #6282

### TODO
- [ ] Migrate `BBCodeCacheBuilder` to eager cache

### DIFF
--- a/XSD/bbcode.xsd
+++ b/XSD/bbcode.xsd
@@ -32,29 +32,29 @@
 	<xs:complexType name="bbcode_delete">
 		<xs:attribute name="name" type="woltlab_varchar" use="required" />
 	</xs:complexType>
-
-	<xs:complexType name="bbcode_icon">
-		<xs:simpleContent>
-			<xs:extension base="woltlab_varchar">
-				<xs:attribute name="solid" type="xs:boolean" />
-			</xs:extension>
-		</xs:simpleContent>
-	</xs:complexType>
 	
 	<!-- imported/updated page menu item element -->
 	<xs:complexType name="bbcode_import">
 		<xs:complexContent>
 			<xs:extension base="bbcode_delete">
-				<xs:all>
+				<xs:choice maxOccurs="unbounded">
 					<xs:element name="htmlopen" type="woltlab_varchar_nullable" minOccurs="0" />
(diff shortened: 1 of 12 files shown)


XSD/bbcode.xsd
```diff
@@ -32,29 +32,29 @@
...
-				<xs:attribute name="solid" type="xs:boolean" />
-			</xs:extension>
-		</xs:simpleContent>
-	</xs:complexType>
 	
 	<!-- imported/updated page menu item element -->
 	<xs:complexType name="bbcode_import">
 		<xs:complexContent>
 			<xs:extension base="bbcode_delete">
-				<xs:all>
```
dtdesign, line 48: This looks incorrect because those other elements are not supposed to appear multiple times?

wcfsetup/install/files/lib/acp/form/BBCodeAddForm.class.php
```diff
@@ -329,4 +314,17 @@ public function assignVariables()
...
         ]);
     }
+
+    protected function saveButtonLabel(int $bbcodeID): void
+    {
+        if ($this->showButton) {
+            if (\is_array($this->buttonLabel)) {
+                $buttonLabels = $this->buttonLabel;
+            } else {
+                $buttonLabels = [0 => $this->buttonLabel];
```
dtdesign, line 324: I think it makes more sense to explicitly use `NULL` instead of relying on `0` being a magic number.

wcfsetup/install/files/lib/data/bbcode/BBCode.class.php
```diff
@@ -36,6 +37,11 @@ class BBCode extends ProcessibleDatabaseObject implements IRouteController
      */
     protected $attributes;
 
+    /**
+     * @var array<int, string>
```
dtdesign, line 41: This signature is incorrect, the key is nullable.

dtdesign (changes requested):

wcfsetup/install/files/lib/acp/form/BBCodeAddForm.class.php
Cyperghost: Since a `null` array key is stored as an empty string, this is not implemented and the magic `0` for monolingual content is still used.

PR #6304, opened 2025-05-25
(PoC) Overhaul condition UI

### LINKED ISSUE #5984: Overhaul condition UI
Conditions are used at various points in the software (e.g. automatic group assignment). Currently, all available conditions are listed in the form, resulting in a very bloated and confusing user interface. Especially when editing existing conditions, it is very difficult to find out which conditions are active.

Instead, only the active conditions should be listed in the form. A button can then be used to select and add further conditions from a dropdown. The same condition could also be set multiple times, which would allow us to omit the use of multi-selects for some conditions.

Bildschirmfoto 2024-08-30 um 14 54 18

Cyperghost:
Closes #5984

## TODO:

- [ ] Deprecate the 'old' `ConditionHandler` functions
- [x] Add migration functions
- [x] Add `wcf1_user_group_assignment .conditions` to `install.sql`
- [ ] Remove old ObjectTypes and ObjectTypeDefinitions for UserGroupAssignment
- [ ] Handle delete condition types
- [ ] Handle when a value is deleted from a `SelectFormField` so that no exception is thrown
- [ ] https://github.com/WoltLab/WCF/issues/6318


wcfsetup/install/files/lib/system/condition/type/IObjectConditionType.class.php
```diff
@@ -0,0 +1,23 @@
...
+ * @extends  IConditionType<TFilter>
+ */
+interface IObjectConditionType extends IConditionType
+{
+    /**
+     * Returns `true` if the given object matches the filter, `false` otherwise.
+     *
+     * @param T $object
+     */
+    public function match(object $object): bool;
```
dtdesign: I suggest renaming this to `matches()` and then drop the obvious comment.

wcfsetup/install/files/lib/system/condition/type/user/UserUsernameConditionType.class.php
```diff
@@ -0,0 +1,86 @@
...
+
+    /**
+     * @return array<string, string>
+     */
+    private function getConditions(): array
+    {
+        return [
+            "%_" => "wcf.condition.startsWith",
+            "%_%" => "wcf.condition.contains",
+            "_%" => "wcf.condition.endsWith",
```
dtdesign: This looks like you have confused the placeholder position (applies above to): `_%` is what I would consider a “starts with”, because the placeholder is at the end.

The logic itself is correct, but the internal identifier appears to be reversed.

dtdesign (changes requested):

com.woltlab.wcf/templates/shared_prefixFormFieldContainer.tpl
```diff
@@ -0,0 +1,60 @@
+<dl id="{$container->getField()->getPrefixedId()}Container"{*
```
dtdesign: This template is barely readable due to the excessive use of `->` accessors. Can you please dumb it down using simple variables to make this easier to comprehend.

com.woltlab.wcf/templates/shared_prefixFormFieldContainer.tpl
```diff
@@ -0,0 +1,60 @@
...
+				{unsafe:$validationError->getHtml()}
+			{/foreach}
+		{/if}
+
+		{include file='shared_formFieldDependencies' field=$container->getField()}
+		{include file='shared_formFieldDataHandler' field=$container->getField()}
+	</dd>
+</dl>
+
+{if $container->getPrefixField() !== null && $container->getPrefixField()->isAvailable() && !$container->getPrefixField()->isImmutable() && $container->prefixHasSelectableOptions()}
```
dtdesign: We do support optional chaining, combined with a proper variable this can be shortened to just read something like `{if $prefixField?->isAvailable() && …}`

ts/WoltLabSuite/Core/Form/Builder/Container/ConditionFormField.ts
```diff
@@ -35,10 +36,26 @@ export class ConditionFormField {
       }),
     );
 
-    wheneverFirstSeen(`#${containerId}Container .condition-remove`, (element: HTMLButtonElement) => {
-      element.addEventListener("click", () => {
-        element.parentElement?.remove();
+    wheneverFirstSeen(`#${containerId}Container .condition-container`, (container: HTMLElement) => {
+      const deleteButton = document.createElement("button");
+      deleteButton.type = "button";
+      deleteButton.classList.add("button", "small", "jsTooltip", "condition-button-remove");
```
dtdesign: What kind of class name is `condition-button-remove`?! Why don’t you just use a data attribute instead, that makes it much simpler and does not rely on arbitrary class names.

wcfsetup/install/files/lib/system/form/builder/container/PrefixConditionFormFieldContainer.class.php
```diff
@@ -0,0 +1,211 @@
...
+{
+    /**
+     * form field to which the prefix selection is added
+     */
+    protected IFormField $field;
+
+    /**
+     * selection form field containing the prefix options
+     */
+    protected ?ISelectionFormField $prefixField;
```
dtdesign: This contain exists for the sole purpose of adding a selection field in front of another form field. Why is it nullable?

What not use a regular form field when there is no prefix and only use this field when there is one, greatly simplifying the logic? Is there any reason to why this is allowed to be optional?

dtdesign (changes requested):

PR #6305, opened 2025-05-25
Article Overview Improvement

jpk6789:
This PR introduces two minor improvements to the article overview for article managers. The first improvement changes the background color of unpublished articles, providing better visual separation between published and unpublished content than the "unpublished" tag alone. 

The second improvement adds view counts to the overview, making it easier for article managers to see how well received the articles are.

Disclaimer: I hope the locations of the changes are correct. I tested these modifications only in our forum using an updated template file and style changes.

WCF_Disabled_Article_bright
WCF_Disabled_Article_dark

### DIFF
--- a/com.woltlab.wcf/templates/articleListItems.tpl
+++ b/com.woltlab.wcf/templates/articleListItems.tpl
@@ -64,6 +64,12 @@
 							{include file='shared_topReaction' cachedReactions=$article->cachedReactions render='short'}
 						</div>
 					{/if}
+					{if $__wcf->getSession()->getPermission('admin.content.article.canManageArticle') || $__wcf->getSession()->getPermission('admin.content.article.canContributeArticle')}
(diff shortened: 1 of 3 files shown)


BurntimeX:
Thank you for your participation.

However, I cannot merge the PR for the following reasons:
1. We generally do not make such changes as part of bugfix releases, which is why the 6.1 branch is the wrong target.
2. There is already a good visible badge for the disabled state. In my opinion, the changed background color offers hardly any additional value. In addition, pure color changes are not barrier-free and should therefore not be used to provide information.
3. Checking the permissions for displaying the views is not necessary in my view, as this information is also accessible to normal users.
4. Personally, I find the display of views as this location rather unnecessary. The views counter is extremely easy to manipulate and is therefore not particularly meaningful. We are thinking about removing this feature completely from the software in the medium term.

PR #6306, opened 2025-05-25
Article layout overhaul

BurntimeX:


com.woltlab.wcf/templates/article.tpl
```diff
@@ -121,76 +62,72 @@
...
-					<div class="articleLikesSummery">
-						{include file="reactionSummaryList" reactionData=$articleLikeData objectType="com.woltlab.wcf.likeableArticle" objectID=$article->articleID}
-					</div>
-				</div>
-			{/if}
-			
-			<div class="col-xs-12 col-md-6 col-md{if !(MODULE_LIKE && ARTICLE_ENABLE_LIKE && $__wcf->session->getPermission('user.like.canViewLike'))} col-md-offset-6{/if}">
-				<ul class="articleButtons buttonGroup buttonList smallButtons">
+		{hascontent}
+			<ul class="entry__footerButtons buttonGroup buttonList smallButtons" role="menu">
```
dtdesign: We’re introducing new classes anyway and change up the layout. This is quite the opportunity to think about this kind of construct, especially with all the historic baggage that `buttonGroup` and `buttonList` come with.

The `<li>` are assigned the role `presentation` which indicates that the markup is not ideal.

I’m not saying that we have to change this but we should at least consider if it makes any sense to swap out the HTML at this point.

com.woltlab.wcf/templates/article.tpl
```diff
@@ -202,48 +139,62 @@
...
 					{/if}
 					{if MODULE_LIKE && ARTICLE_ENABLE_LIKE && $__wcf->session->getPermission('user.like.canLike') && $article->userID != $__wcf->user->userID}
-						<li class="jsOnly"><span class="button jsTooltip reactButton{if $articleLikeData[$article->articleID]|isset && $articleLikeData[$article->articleID]->reactionTypeID} active{/if}" title="{lang}wcf.reactions.react{/lang}" data-reaction-type-id="{if $articleLikeData[$article->articleID]|isset && $articleLikeData[$article->articleID]->reactionTypeID}{$articleLikeData[$article->articleID]->reactionTypeID}{else}0{/if}">{icon name='face-smile'} <span class="invisible">{lang}wcf.reactions.react{/lang}</span></span></li>
+						<li role="presentation">
+							<button
+								type="button"
+								role="menuitem"
+								class="button jsTooltip reactButton{if $articleLikeData[$article->articleID]|isset && $articleLikeData[$article->articleID]->reactionTypeID} active{/if}"
+								title="{lang}wcf.reactions.react{/lang}"
+								data-reaction-type-id="{if $articleLikeData[$article->articleID]|isset && $articleLikeData[$article->articleID]->reactionTypeID}{$articleLikeData[$article->articleID]->reactionTypeID}{else}0{/if}"
```
dtdesign: Can we somehow generate this button in a less awkward way?

com.woltlab.wcf/templates/entryTags.tpl
```diff
@@ -0,0 +1,14 @@
+{if !$tags|empty}
+	<section class="entry_tags" aria-label="{lang}wcf.tagging.tags{/lang}">
+		<ul class="tagList">
+			{foreach from=$tags item=tag}
+				<li>
+					<a href="{link controller='Tagged' object=$tag}objectType={$objectType}{/link}" class="tag">
```
dtdesign:
```suggestion
					<a href="{link controller='Tagged' object=$tag objectType=$objectType}{/link}" class="tag">
```

wcfsetup/install/files/lib/system/endpoint/controller/core/articles/contents/GetArticleContentHeaderTitle.class.php
```diff
@@ -0,0 +1,53 @@
...
+/**
+ * API endpoint for the rendering of the article content header title.
+ *
+ * @author      Marcel Werk
+ * @copyright   2001-2025 WoltLab GmbH
+ * @license     GNU Lesser General Public License <http://opensource.org/licenses/lgpl-license.php>
+ * @since       6.2
+ */
+#[GetRequest('/core/articles/contents/{id:\d+}/content-header-title')]
+final class GetArticleContentHeaderTitle implements IController
```
dtdesign, line 25: Isn't this technically pretty much the entire header sans the buttons? “Title” implies the pure title which it also is not.

wcfsetup/install/files/lib/system/listView/AbstractListView.class.php
```diff
@@ -81,6 +85,22 @@ public function setItemsPerPage(int $itemsPerPage): void
         $this->itemsPerPage = $itemsPerPage;
     }
 
+    /**
+     * Gets the maximum number of items shown.
+     */
+    public function getMaxItems(): int
```
dtdesign: I don’t like the naming, I was about to comment on line 203 that there is a bug when no explicit number is set but this is not the case. I think it is easy to confuse this with the customizable limit of object lists where this really means that only up to a maximum of X items should ever be shown.

Maybe use something like `fixedNumberOfItems` instead?

PR #6311, opened 2025-06-03
Refactor box-templates to use `unsafe` and remove unnecessary `@`

Cyperghost:


com.woltlab.wcf/templates/boxRecentActivity.tpl
```diff
@@ -44,6 +44,6 @@
 		{jsphrase name='wcf.user.recentActivity.more'}
 		{jsphrase name='wcf.user.recentActivity.noMoreEntries'}
 
-		setup(document.getElementById('boxRecentActivity{unsafe:$boxID|encodeJS}'));
```
dtdesign, line 47: This change is incorrect, the rules for encoding values in JS strings differ from HTML. Please revert this back to the original, correct way of handling the value.

wcfsetup/install/files/acp/templates/pageBoxOrder.tpl
```diff
@@ -131,11 +131,11 @@
 		const boxes = new Map();
 		{foreach from=$boxes key=position item=boxData}
 			{if $position != 'mainMenu'}
-				boxes.set('{$position}', [{implode from=$boxData item=box}{ boxId: {@$box->boxID}, name: '{$box->name|encodeJS}', isDisabled: {if $box->isDisabled}true{else}false{/if} }{/implode}]);
```
dtdesign, line 134: `|encodeJS` requires the use of `unsafe:`

dtdesign (changes requested):

PR #6315, opened 2025-06-04
Add a suffix to prevent key collisions with parameterized caches

### LINKED ISSUE #6314: Parameterized cache builders can cause key collisions in Redis
We’re storing parameterized cache builders using hash lists using `HSET` and access specific caches using `HGET`. This reduces the top level keys which probably was the motivation at some point.

However, this causes a problem when there is a cache both without and with parameters because the cache key is the same. One such example can be found with the trophies:

https://github.com/WoltLab/WCF/blob/e33bc1e91ce7866574dc1a94f7c72479ce55f4a5/wcfsetup/install/files/lib/data/trophy/TrophyCache.class.php#L41-L42

The second cache will always be an implicit cache miss because the key is already set as a string. Trying to access offsets using `HSET` fails with a `WRONGTYPE` error which never surfaced because all errors are implicitly suppressed by not checking `$redis->getLastError()`.

Following the example above, this will trigger 2 Redis errors (one for the lookup and one for the attempt to set it) that go unnoticed plus one extra `SELECT` query for the second cache.

dtdesign:
Fixes #6314

### DIFF
--- a/wcfsetup/install/files/lib/system/cache/source/RedisCacheSource.class.php
+++ b/wcfsetup/install/files/lib/system/cache/source/RedisCacheSource.class.php
@@ -46,10 +46,10 @@ class RedisCacheSource implements ICacheSource
         if (isset($parts[1])) {
             if ($useWildcard) {
                 // delete the complete hashset
-                $this->redis->del($this->getCacheName($parts[0]));
+                $this->redis->del($this->getCacheName($parts[0], true));
             } else {
                 // delete the specified key from the hashset
-                $this->redis->hDel($this->getCacheName($parts[0]), $parts[1]);
+                $this->redis->hDel($this->getCacheName($parts[0], true), $parts[1]);
             }
         } else {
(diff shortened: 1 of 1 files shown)


wcfsetup/install/files/lib/system/cache/source/RedisCacheSource.class.php
```diff
@@ -46,10 +46,10 @@ public function flush($cacheName, $useWildcard)
         if (isset($parts[1])) {
             if ($useWildcard) {
                 // delete the complete hashset
-                $this->redis->del($this->getCacheName($parts[0]));
+                $this->redis->del($this->getCacheName($parts[0], true));
```
TimWolla: A wildcard deletion for the DiskCacheSource will delete both the parameterized as well as the default cache, we must replicate this:
```suggestion
                $this->redis->del(
                    $this->getCacheName($parts[0], true),
                    $this->getCacheName($parts[0]),
                );
```

PR #6317, opened 2025-06-05
Get the `templateGroupID` for the shared templates from the cache

### LINKED ISSUE #6316: The `templateGroupID` of the shared templates is fetched on every request
https://github.com/WoltLab/WCF/blob/e33bc1e91ce7866574dc1a94f7c72479ce55f4a5/wcfsetup/install/files/lib/system/template/TemplateEngine.class.php#L1045-L1057

This fetches the `templateGroupID` with every request but since this is a static lookup, there is no reason not to cache the result. Furthermore, `TemplateEngine` [already uses a cache for the template groups](https://github.com/WoltLab/WCF/blob/e33bc1e91ce7866574dc1a94f7c72479ce55f4a5/wcfsetup/install/files/lib/system/template/TemplateEngine.class.php#L858) which means we can just look it up in the cache.

Cyperghost:
Closes #6316

### DIFF
--- a/wcfsetup/install/files/lib/system/template/TemplateEngine.class.php
+++ b/wcfsetup/install/files/lib/system/template/TemplateEngine.class.php
@@ -1046,7 +1046,7 @@ class TemplateEngine extends SingletonFactory
     {
         if (!isset($this->sharedTemplateGroupID)) {
             $templateGroup = \array_find(
-                TemplateGroupCacheBuilder::getInstance()->getData(),
+                $this->templateGroupCache,
                 static fn(TemplateGroup $group) => $group->templateGroupFolderName === '_wcf_shared/'
             );
 
--- a/wcfsetup/install/files/lib/system/template/TemplateEngine.class.php
+++ b/wcfsetup/install/files/lib/system/template/TemplateEngine.class.php
@@ -1044,15 +1044,16 @@ class TemplateEngine extends SingletonFactory
 
     private function getSharedTemplateGroupID(): int
     {
-        if (!isset($this->sharedTemplateGroupID)) {
-            /** @var TemplateGroup $templateGroup */
-            foreach (TemplateGroupCacheBuilder::getInstance()->getData() as $templateGroup) {
-                if ($templateGroup->templateGroupFolderName === '_wcf_shared/') {
-                    $this->sharedTemplateGroupID = $templateGroup->templateGroupID;
-                    break;
-                }
-            }
+        $templateGroup = \array_find(
+            TemplateGroupCacheBuilder::getInstance()->getData(),
+            static fn (TemplateGroup $group) => $group->templateGroupFolderName === '_wcf_shared/'
+        );
+
+        if ($templateGroup === null) {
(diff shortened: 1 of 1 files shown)


wcfsetup/install/files/lib/system/template/TemplateEngine.class.php
```diff
@@ -1045,14 +1045,15 @@ protected function getCompileFilePrefix(string $templateName): string
...
     {
         if (!isset($this->sharedTemplateGroupID)) {
-            $sql = "SELECT  templateGroupID
-                    FROM    wcf1_template_group
-                    WHERE   templateGroupFolderName = ?";
-            $statement = WCF::getDB()->prepare($sql);
-            $statement->execute(['_wcf_shared/']);
-
-            $this->sharedTemplateGroupID = $statement->fetchSingleColumn();
+            /** @var TemplateGroup $templateGroup */
```
dtdesign: I much prefer an `\array_find()` combined with a check that will throw a `RuntimeException` if the result is `null`. Right now your code will read from an uninitialized property if there is no match.

dtdesign (changes requested):

wcfsetup/install/files/lib/system/template/TemplateEngine.class.php
```diff
@@ -1044,15 +1044,17 @@ protected function getCompileFilePrefix(string $templateName): string
...
     {
-        if (!isset($this->sharedTemplateGroupID)) {
-            $sql = "SELECT  templateGroupID
-                    FROM    wcf1_template_group
-                    WHERE   templateGroupFolderName = ?";
-            $statement = WCF::getDB()->prepare($sql);
-            $statement->execute(['_wcf_shared/']);
-
-            $this->sharedTemplateGroupID = $statement->fetchSingleColumn();
+        $templateGroup = \array_find(
```
dtdesign: This will now execute on every invocation of the method because you removed the `isset()` wrapper.

dtdesign (changes requested):

wcfsetup/install/files/lib/system/template/TemplateEngine.class.php
```diff
@@ -1045,14 +1045,18 @@ protected function getCompileFilePrefix(string $templateName): string
     private function getSharedTemplateGroupID(): int
     {
         if (!isset($this->sharedTemplateGroupID)) {
-            $sql = "SELECT  templateGroupID
-                    FROM    wcf1_template_group
-                    WHERE   templateGroupFolderName = ?";
-            $statement = WCF::getDB()->prepare($sql);
-            $statement->execute(['_wcf_shared/']);
+            $templateGroup = \array_find(
+                TemplateGroupCacheBuilder::getInstance()->getData(),
```
dtdesign: Is there any reason why you don’t just use `$this->templateGroupCache` here? I even _explicitly_ pointed that out in #6316.

dtdesign (changes requested):

dtdesign (approved):

Issue #6310, opened 2025-06-02
Migration of the condition data


Cyperghost:
Ref https://github.com/WoltLab/WCF/issues/5984

With the overhaul of the condition system, the current data must be migrated to the new format.
This means that condition types of plugins that have already been created, e.g. for *Automatic user group assignment* or *Notes*, cannot be migrated correctly. At the time of the upgrade, the plugins are not up to date with the new condition system. It may also be the case that condition types that were added by plugins are no longer offered or have not yet been migrated.
This would mean that the filtering by the condition system would no longer work as before the update, e.g. users could be added to a user group that were not previously added.

PR #6330, opened 2025-06-10
Migrate "old" conditions

Cyperghost:
Ref https://github.com/WoltLab/WCF/pull/6304
See https://github.com/WoltLab/WCF/issues/6310

### TODO

- [ ] Change branch to `6.3-condition`


wcfsetup/install/files/acp/database/update_com.woltlab.wcf_6.3_step1.php
```diff
@@ -8,12 +8,14 @@
...
 
+use wcf\system\database\table\column\DefaultFalseBooleanDatabaseTableColumn;
 use wcf\system\database\table\column\MediumtextDatabaseTableColumn;
 use wcf\system\database\table\PartialDatabaseTable;
 
 return [
     PartialDatabaseTable::create('wcf1_user_group_assignment')
         ->columns([
             MediumtextDatabaseTableColumn::create('conditions'),
+            DefaultFalseBooleanDatabaseTableColumn::create('needMigration'),
```
dtdesign: I would prefer something like `isLegacy` which is more in line with our general naming schema.

wcfsetup/install/files/lib/system/user/group/assignment/command/UserGroupAssignmentMigrateCondition.class.php
```diff
@@ -0,0 +1,42 @@
...
+use wcf\system\condition\provider\UserConditionProvider;
+use wcf\util\JSON;
+
+/**
+ * @author Olaf Braun
+ * @copyright 2001-2025 WoltLab GmbH
+ * @license GNU Lesser General Public License <http://opensource.org/licenses/lgpl-license.php>
+ * @since 6.3
+ */
+final class UserGroupAssignmentMigrateCondition
```
dtdesign: This completely violates the naming schema for commands.

wcfsetup/install/files/lib/system/cache/eager/UserGroupAssignmentCache.class.php
```diff
@@ -0,0 +1,50 @@
...
+            if ($assignment->needMigration) {
+                (new UserGroupAssignmentMigrateCondition($assignment))();
+                $migrationDone = true;
+            }
+        }
+
+        if ($migrationDone) {
+            // Reload the list to ensure that no disabled assignments are included
+            return $this->getUserGroupAssignments()->getObjects();
+        } else {
```
dtdesign: This is a very odd place to perform a migration, business logic doesn’t belong here.

Migrations are only ever relevant after a package update has completed because that is the only ingress point for new implementations.

wcfsetup/install/files/lib/system/condition/ConditionHandler.class.php
```diff
@@ -172,4 +176,82 @@ public function getConditionsWithFilter(AbstractConditionProvider $provider, arr
...
+    public function exportConditions(string $definitionName): array
+    {
+        $objectTypes = ObjectTypeCache::getInstance()->getObjectTypes($definitionName);
+        if ($objectTypes === []) {
+            return [];
+        }
+
+        $conditionBuilder = new PreparedStatementConditionBuilder();
+        $conditionBuilder->add('objectTypeID IN (?)', [
+            \array_map(static fn (ObjectType $objectType): int => $objectType->objectTypeID, $objectTypes),
```
dtdesign:
```suggestion
            \array_column($objectTypes, 'objectTypeID'),
```

wcfsetup/install/files/lib/system/condition/ConditionHandler.class.php
```diff
@@ -172,4 +176,82 @@ public function getConditionsWithFilter(AbstractConditionProvider $provider, arr
...
+                    $migratedData,
+                    ...$conditionType->migrateConditionData($condition)
+                );
+
+                if ($condition === []) {
+                    unset($conditionData[$objectType]);
+                    break;
+                }
+            }
+        }
```
dtdesign, line 252:
```suggestion
        }
        \unset($condition);
```

Please always unset variables bound by referenc. This can easily lead to unintended behavior later on because the reference escapes the scope of the `foreach`.

wcfsetup/install/files/lib/system/condition/provider/ConditionMigration.class.php
```diff
@@ -0,0 +1,38 @@
+<?php
+
+namespace wcf\system\condition\provider;
```
dtdesign: What is the rationale of placing this in the `provider` sub namespace?

wcfsetup/install/files/lib/system/condition/provider/ConditionMigration.class.php
```diff
@@ -0,0 +1,38 @@
...
+    ) {
+    }
+
+    /**
+     * Creates a new ConditionMigration instance based on condition data and conditions.
+     *
+     * @param array{identifier: string, value: mixed}[] $previousConditionData
+     * @param array{identifier: string, value: mixed}[] $migratedConditionData
+     */
+    public static function for(array $previousConditionData, array $migratedConditionData): self
```
dtdesign: `fromData()` perhaps is a more fitting name.

wcfsetup/install/files/lib/system/condition/provider/ConditionMigration.class.php
```diff
@@ -0,0 +1,38 @@
...
+    {
+        return new self($previousConditionData === [], $migratedConditionData);
+    }
+
+    /**
+     * Creates a new ConditionMigration instance for empty data.
+     */
+    public static function forNoConditions(): self
```
dtdesign: `withoutData()` makes it less susceptible for an incorrect auto completion and is in line with other constructs like the `dialogFactory()` in TypeScript.

wcfsetup/install/files/lib/system/condition/provider/ConditionMigration.class.php
```diff
@@ -0,0 +1,38 @@
...
+/**
+ * @author Olaf Braun
+ * @copyright 2001-2025 WoltLab GmbH
+ * @license GNU Lesser General Public License <http://opensource.org/licenses/lgpl-license.php>
+ * @since 6.2
+ */
+final class ConditionMigration
+{
+    private function __construct(
+        public readonly bool $isFullMigrated,
```
dtdesign: I don’t understand the point of this variable (besides it missing a `y`), isn’t the presence of conditions sufficient?

wcfsetup/install/files/lib/system/condition/provider/ConditionMigration.class.php
```diff
@@ -0,0 +1,38 @@
...
+        /** @var array{identifier: string, value: mixed}[] */
+        public readonly array $conditions = [],
```
dtdesign: Please don’t use default values together with fixed construct methods. Instead have the methods explicitly define all parameters.

wcfsetup/install/files/lib/system/condition/provider/UserConditionProvider.class.php
```diff
@@ -35,25 +35,25 @@ final class UserConditionProvider extends AbstractConditionProvider
     public function __construct()
     {
         $this->addConditions([
-            new class("username", "username") extends AbstractUserStringConditionType {},
-            new class("email", "email") extends AbstractUserStringConditionType {},
+            new class("username", "username", "username", 'com.woltlab.wcf.user.username') extends AbstractUserStringConditionType {},
```
dtdesign: This (among others) crosses the threshold for named parameters. Maybe even wrap the parameters to make it an easy to parse `key: "value"` list.

wcfsetup/install/files/lib/system/condition/type/IMigrateConditionType.class.php
```diff
@@ -0,0 +1,23 @@
...
+    /**
+     * Converts the old condition structure to the new one. All migrated values must be removed from the `$conditionData` array.
+     *
+     * @param array<string, mixed> $conditionData
+     *
+     * @return array{identifier: string, value: mixed}[]
+     */
+    public function migrateConditionData(array &$conditionData): array;
+
+    public function canMigrateConditionData(string $objectType): bool;
```
dtdesign, line 34: This lacks a description that helps the developer understand what this method is supposed to validate. When is it reasonable not to support a migration?

wcfsetup/install/files/lib/system/condition/type/IMigrateConditionType.class.php
```diff
@@ -0,0 +1,23 @@
...
+interface IMigrateConditionType
+{
+    /**
+     * Converts the old condition structure to the new one. All migrated values must be removed from the `$conditionData` array.
+     *
+     * @param array<string, mixed> $conditionData
+     *
+     * @return array{identifier: string, value: mixed}[]
+     */
+    public function migrateConditionData(array &$conditionData): array;
```
dtdesign: This entire methods makes me feel uneasy:

1. References in parameters should be avoided like the plague because PHP does not require the call-side to mark references which makes these things quite obscure.
2. If the implementation is supposed to strip data from `$conditionData` then what is expected to be returned from it? The migrated data? And if yes, why bother removing it from `$conditionData`?
3. `array{…}[]` is an awful way to document the type, please use `[]` only for simple expressions like `int[]`. This should be `list<array{…}>` or something like that.

wcfsetup/install/files/lib/system/condition/type/user/AbstractUserBooleanConditionType.class.php
```diff
@@ -65,4 +68,28 @@ public function matches(object $object): bool
...
+                'identifier' => $this->identifier,
+                'value' => \boolval($value),
+            ],
+        ];
+    }
+
+    #[\Override]
+    public function canMigrateConditionData(string $objectType): bool
+    {
+        return $this->migrateConditionObjectType === null || $this->migrateConditionObjectType === $objectType;
```
dtdesign: Why does `$this->migrateConditionObjectType === null` qualify this type to perform a migration? What is that property even supposed to do, it is not used in `migrateConditionData()` at all?

wcfsetup/install/files/lib/system/condition/type/user/AbstractUserBooleanConditionType.class.php
```diff
@@ -65,4 +68,28 @@ public function matches(object $object): bool
...
     }
+
+    #[\Override]
+    public function migrateConditionData(array &$conditionData): array
+    {
+        if ($this->migrateKeyName === null || !isset($conditionData[$this->migrateKeyName])) {
+            return [];
+        }
+
+        $value = $conditionData[$this->migrateKeyName];
```
dtdesign: This can be simplified as `$value = $conditionData[$this->migrateKeyName] ?? null;` and then to return when `$value` is `null`.

This works because `null` when used as an array key is interpreted as an empty string and `isset()` performs a null-check already, thus `null` is never a valid value.

wcfsetup/install/files/lib/system/condition/type/user/AbstractUserIntegerConditionType.class.php
```diff
@@ -84,8 +86,39 @@ public function matches(object $object): bool
...
-    private function getConditions(): array
+    protected function getConditions(): array
     {
         return ["=", ">", "<", ">=", "<="];
     }
+
+    #[\Override]
+    public function canMigrateConditionData(string $objectType): bool
+    {
+        return $this->migrateConditionObjectType !== null && $objectType === $this->migrateConditionObjectType;
```
dtdesign:
```suggestion
        return $objectType === $this->migrateConditionObjectType;
```

`$objectType` can never be `null` thus a strict equality check will always fail when `$this->migrateConditionObjectType` is `null`.

Also, why does `null` prevent a migration here but allows it for `AbstractUserBooleanConditionType`?

wcfsetup/install/files/lib/system/condition/type/user/AbstractUserStringConditionType.class.php
```diff
@@ -103,4 +106,31 @@ private function getConditions(): array
             "%_" => "wcf.condition.endsWith",
         ];
     }
+
+    #[\Override]
+    public function canMigrateConditionData(string $objectType): bool
+    {
+        return $this->migrateConditionObjectType !== null && $objectType === $this->migrateConditionObjectType;
```
dtdesign: See `AbstractUserBooleanConditionType`.

wcfsetup/install/files/lib/system/condition/type/user/AbstractUserStringConditionType.class.php
```diff
@@ -103,4 +106,31 @@ private function getConditions(): array
...
+    }
+
+    #[\Override]
+    public function migrateConditionData(array &$conditionData): array
+    {
+        if ($this->migrateKeyName === null || !isset($conditionData[$this->migrateKeyName])) {
+            return [];
+        }
+
+        $value = $conditionData[$this->migrateKeyName];
```
dtdesign: See `AbstractUserBooleanConditionType`.

wcfsetup/install/files/lib/system/condition/type/user/UserRegistrationDateConditionType.class.php
```diff
@@ -85,4 +86,49 @@ private function getConditions(): array
...
+
+        unset($conditionData['registrationDateStart'], $conditionData['registrationDateEnd']);
+
+        return $conditions;
+    }
+
+    private function convertDateStringTimestamp(string $date, int $hour, int $minute, int $seconds): int
+    {
+        $dateTime = new \DateTime($date, new \DateTimeZone(TIMEZONE));
+        $dateTime->setTime($hour, $minute, $seconds);
```
dtdesign: Please never use `\DateTime` and pick `\DateTimeImmutable` instead, otherwise this summons ~Cthulhu~ _Tim_.

wcfsetup/install/files/lib/system/condition/type/user/UserRegistrationDaysConditionType.class.php
```diff
@@ -3,56 +3,32 @@
...
     }
 
     #[\Override]
-    public function getIdentifier(): string
+    public function getFormField(string $id): PrefixConditionFormFieldContainer
     {
-        return 'registrationDays';
-    }
+        $container = parent::getFormField($id);
+        $filed = $container->getField();
```
dtdesign: Typo: `$field`.

wcfsetup/install/files/lib/system/user/group/assignment/command/UserGroupAssignmentMigrateCondition.class.php
```diff
@@ -0,0 +1,42 @@
...
+
+    public function __invoke(): void
+    {
+        if (!$this->assignment->needMigration) {
+            return;
+        }
+
+        $migratedData = ConditionHandler::getInstance()->migrateConditionData(
+            new UserConditionProvider(),
+            JSON::decode($this->assignment->conditions)
```
dtdesign: This can yield an exception which will brick the process.

The deserialization should be a separate step and if it fails, this condition must be discarded because its broken anyway.

dtdesign (changes requested):

wcfsetup/install/files/lib/system/condition/type/user/AbstractUserBooleanConditionType.class.php
Cyperghost: `null` should of course not allow migration, as is the case everywhere else.

wcfsetup/install/files/lib/system/condition/provider/ConditionMigration.class.php
Cyperghost: No, it is possible that only 99 out of 100 different conditions were migrated. This should not be considered a full migration. This can occur if, for example, a condition type was added by a third party plugin and does not (yet) exist.
The previous user group assignment must therefore be deactivated, otherwise users could be assigned under different circumstances than before.

However, if we only allow/enable the migration of data if all data has been transferred correctly, it _may happen_ that the admin has to re-enter all data for a user group assignment because a plugin is no longer supported.

wcfsetup/install/files/lib/system/condition/type/IMigrateConditionType.class.php
Cyperghost: It is not easy to determine whether all data has been migrated correctly and completely without using a reference.
For example, [`UserGroupCondition`](https://github.com/WoltLab/WCF/blob/6.1/wcfsetup/install/files/lib/system/condition/UserGroupCondition.class.php) now consists of two classes:  `UserNotInGroupConditionType` and `UserInGroupConditionType`.
Therefore, I cannot tell whether, for example, after migrating `UserInGroupConditionType`, further steps are necessary for the object type `com.woltlab.wcf.userGroup` or not.

If I were to remove the data `$conditionData[‘com.woltlab.wcf.userGroup’]` after the first call to `migrateConditionData()`, only `UserInGroupConditionType::migrateConditionData()` would be executed, but not `UserNotInGroupConditionType::migrateConditionData()`.

Therefore, the class must remove all entries it can process so that `ConditionHandler::migrateConditionData()` can recognize whether everything has been transferred or not.

dtdesign: Understood. In that case please expand the existing documentation to explain this which also serves as a guideline for developers to properly use this API.

wcfsetup/install/lang/de.xml
```diff
@@ -870,6 +870,7 @@ Sie erreichen das Fehlerprotokoll unter: {link controller='ExceptionLogView' isE
 		<item name="wcf.acp.group.assignment.userGroup"><![CDATA[Benutzergruppe]]></item>
 		<item name="wcf.acp.group.assignment.isDisabled"><![CDATA[Automatische Zuordnung deaktivieren]]></item>
 		<item name="wcf.acp.group.assignment.title.error.tooLong"><![CDATA[Der eingegebene Titel ist zu lang. Die maximale Länge beträgt 255 Zeichen.]]></item>
+		<item name="wcf.acp.group.assignment.legacyNotice"><![CDATA[Führen Sie für die Automatische Benutzergruppen-Zuordnung die <a href="{link controller='RebuildData'}{/link}">Anzeigen aktualisieren</a> durch.]]></item>
```
dtdesign:
```suggestion
		<item name="wcf.acp.group.assignment.legacyNotice"><![CDATA[Die automatischen Benutzergruppen-Zuordnungen wurden noch nicht aktualisiert. Bitte führen Sie dies unter <a href="{link controller='RebuildData'}{/link}">Anzeigen aktualisieren</a> durch.]]></item>
```

wcfsetup/install/lang/en.xml
```diff
@@ -846,6 +846,7 @@ You can access the error log at: {link controller='ExceptionLogView' isEmail=tru
 		<item name="wcf.acp.group.assignment.userGroup"><![CDATA[User Group]]></item>
 		<item name="wcf.acp.group.assignment.isDisabled"><![CDATA[Disable automatic assignment]]></item>
 		<item name="wcf.acp.group.assignment.title.error.tooLong"><![CDATA[The entered title is too long. The maximum length is 255 characters.]]></item>
+		<item name="wcf.acp.group.assignment.legacyNotice"><![CDATA[For Automatic User Group Assignments, perform <a href="{link controller='RebuildData'}{/link}">Rebuild Data</a>.]]></item>
```
dtdesign:
```suggestion
		<item name="wcf.acp.group.assignment.legacyNotice"><![CDATA[The automated user group assignments need to be recalculated. Please perform this task by visiting  <a href="{link controller='RebuildData'}{/link}">Rebuild Data</a>.]]></item>
```

dtdesign (changes requested):
Please also add an ACP status box of type warning to notify about outstanding migrations.

wcfsetup/install/files/lib/system/acp/dashboard/box/StatusMessageAcpDashboardBox.class.php
```diff
@@ -271,4 +273,42 @@ private function getPHPExtensionMessage(): array
...
+                new StatusMessage(
+                    StatusMessageType::Warning,
+                    WCF::getLanguage()->getDynamicVariable('wcf.acp.dashboard.box.migrationNeeded', [
+                        'titles' => $event->needsMigration(),
+                    ])
+                ),
+            ];
+        }
+
+        return [];
```
dtdesign: Please invert this check, the happy path should always be at the bottom.

dtdesign (changes requested):

PR #6348, opened 2025-06-13
Unique key for `wcf1_background_job.identifier`

### LINKED ISSUE #6303: Set a unique key on `wcf1_background_job.identifier`
This is a follow-up issue for #6299 which rewrote the handling of the background queue. The current design is a `SELECT` followed by an `INSERT` which works because while there is a key on `identifier`, it is not required to be unique.

We can simplify this logic and also prevent situations where a unique job is inserted in parallel (we run into a TOCTOU problem here!) by changing the existing key into a unique key. Since non-unique jobs have their `identifier` set to `null`, these will be ignored by the index.

This requires a separate update script to remove any duplicate unique jobs before changing the key type. From then on we simplify the handler to just issue an `INSERT IGNORE` to gracefully handle conflicts without requiring a write lock on the table.

Cyperghost:
Closes #6303

### DIFF
--- a/wcfsetup/install/files/lib/system/background/BackgroundQueueHandler.class.php
+++ b/wcfsetup/install/files/lib/system/background/BackgroundQueueHandler.class.php
@@ -111,13 +111,14 @@ final class BackgroundQueueHandler extends SingletonFactory
             }
         }
 
-        $sql = "INSERT INTO wcf1_background_job
-                            (job, time, identifier)
-                VALUES      (?, ?, ?)";
-        $statement = WCF::getDB()->prepare($sql);
-        foreach ($jobs as $job) {
-            $identifier = null;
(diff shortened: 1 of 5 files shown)


wcfsetup/install/files/lib/system/background/BackgroundQueueHandler.class.php
```diff
@@ -81,32 +81,21 @@ public function enqueueAt(AbstractBackgroundJob|array $jobs, int $time): void
         $committed = false;
         try {
             WCF::getDB()->beginTransaction();
```
dtdesign: Why did you keep the transaction plus the `try … catch` around? This was only required for the write lock to work.

dtdesign (changes requested):

dtdesign (approved):

PR #6353, opened 2025-06-16
Refactor language templates

Cyperghost:
- Remove unnecessary `@`
- Use `unsafe:` instead of `@`
- Add missing `{unsafe:…|encodeJS}`

### DIFF
--- a/com.woltlab.wcf/templates/shared_languageChooser.tpl
+++ b/com.woltlab.wcf/templates/shared_languageChooser.tpl
@@ -4,9 +4,9 @@
 {if $languages|count}
 	<dl{if $errorField|isset && $errorField == $__languageChooserPrefix|concat:'languageID'} class="formError"{/if}>
 		<dt>{lang}{$label}{/lang}</dt>
-		<dd id="{@$__languageChooserPrefix}languageIDContainer">
+		<dd id="{$__languageChooserPrefix}languageIDContainer">
 			<noscript>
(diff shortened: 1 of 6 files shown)


com.woltlab.wcf/templates/shared_multipleLanguageInputJavascript.tpl
```diff
@@ -5,16 +5,17 @@
 				'wcf.global.button.disabledI18n': '{jslang}wcf.global.button.disabledI18n{/jslang}'
 			});
 
-			var availableLanguages = { {implode from=$availableLanguages key=languageID item=languageName}{@$languageID}: '{$languageName}'{/implode} };
-			var values = { {implode from=$i18nValues[$elementIdentifier] key=languageID item=value}'{@$languageID}': '{$value}'{/implode} };
+			var availableLanguages = { {implode from=$availableLanguages key=languageID item=languageName}{$languageID}: '{unsafe:$languageName|encodeJS}'{/implode} };
+			{* `$value` already encoded in `I18nHandler::assignVariables()` *}
```
dtdesign: And that is a bad assumption, because `WoltLabSuite/Core/Language/Input` (and `Text`) unescape the HTML. While it is incorrect to escape the value here, not escaping it makes it worse. Arguably, this piece of code has aged like fine milk.

Please return to the previous behavior which is known to be working.

dtdesign (changes requested):

Issue #6360, opened 2025-06-21
Upgrade `@fancyapps/ui` to v6


dtdesign:
This version has been a significant rewrite, please follow the migration docs found at https://fancyapps.com/resources/upgrading/. Please test all changes extensively.

PR #6362, opened 2025-06-22
Upgrade `@fancyapps/ui` to v6

Cyperghost:
Closes https://github.com/WoltLab/WCF/issues/6360


ts/WoltLabSuite/Core/Component/Image/Viewer.ts
```diff
@@ -1,45 +1,43 @@
...
+      autoplay: false,
     },
   };
 }
 
 export async function getLocalization(): Promise<Record<string, string>> {
   let locale = document.documentElement.lang;
 
-  if (!LOCALES.includes(locale)) {
+  if (!Object.prototype.hasOwnProperty.call(LOCALES, locale)) {
```
dtdesign: Tip: `Object.hasOwn()`

ts/WoltLabSuite/Core/Component/Image/Viewer.ts
```diff
@@ -1,45 +1,43 @@
-import { Fancybox } from "@fancyapps/ui";
-import { userSlideType } from "@fancyapps/ui/types/Carousel/types";
-import { OptionsType } from "@fancyapps/ui/types/Fancybox/options";
+import { Fancybox, CarouselSlide, FancyboxInstance } from "@fancyapps/ui";
 import { getPageOverlayContainer } from "WoltLabSuite/Core/Helper/PageOverlay";
 
-const LOCALES = ["cs", "de", "en", "es", "fr", "it", "lv", "pl", "sk"];
+const LOCALES = { de: "de_DE", en: "en_EN" };
```
dtdesign: Why do we even bother with this instead of providing the phrases ourselves? I mean, those are not a whole lot phrases and at least the German translations are a bit sketchy at times anyway.

dtdesign (changes requested):

dtdesign (changes requested):
I don’t like the name of the phrases at all because they align with nothing. Either use camelCase like we always do or preserve the uppercase style of their identifiers. Lowercasing it but keeping the underscores is highly inconsistent.

Considering that we have traditionally been mapping phrases using lookup tables (like it is the case right now), you should stick to camelCase for the phrase identifiers.

PR #6365, opened 2025-06-24
Unify grid / list view filters

BurntimeX:


wcfsetup/install/files/lib/system/gridView/filter/DateFilter.class.php
```diff
@@ -0,0 +1,135 @@
...
+    {
+        $values = explode(';', $value);
+        if (\count($values) !== 2) {
+            return '';
+        }
+
+        $locale = WCF::getLanguage()->getLocale();
+        $fromString = $toString = '';
+        if ($values[0] !== '') {
+            $fromDateTime = \DateTime::createFromFormat(
```
dtdesign: Please prefer using `\DateTimeImmutable` at all times.

wcfsetup/install/files/lib/system/listView/filter/DateFilter.class.php
```diff
@@ -0,0 +1,135 @@
...
+    {
+        $values = explode(';', $value);
+        if (\count($values) !== 2) {
+            return '';
+        }
+
+        $locale = WCF::getLanguage()->getLocale();
+        $fromString = $toString = '';
+        if ($values[0] !== '') {
+            $fromDateTime = \DateTime::createFromFormat(
```
dtdesign: Please always use `\DateTimeImmutable`.

wcfsetup/install/files/lib/system/listView/filter/TimeFilter.class.php
```diff
@@ -0,0 +1,136 @@
...
+    {
+        $values = explode(';', $value);
+        if (\count($values) !== 2) {
+            return '';
+        }
+
+        $locale = WCF::getLanguage()->getLocale();
+        $fromString = $toString = '';
+        if ($values[0] !== '') {
+            $fromDateTime = \DateTime::createFromFormat(
```
dtdesign: Please always use `\DateTimeImmutable` instead.

dtdesign (changes requested):

PR #6369, opened 2025-06-26
Allow the configuration of the effect that is to be applied after an interaction has been executed

BurntimeX:


ts/WoltLabSuite/Core/Component/Interaction/LegacyDboAction.ts
```diff
@@ -31,21 +34,25 @@ async function handleDboAction(
     .payload(confirmationResult.reason ? { reason: confirmationResult.reason } : {})
     .dispatch();
 
-  if (confirmationType == ConfirmationType.Delete) {
+  if (interactionEffect === InteractionEffect.ReloadItem) {
```
dtdesign, line 37: I would like to see this being a `switch` instead in order to handle unknown effects by raising an exception.

ts/WoltLabSuite/Core/Component/Interaction/Rpc.ts
```diff
@@ -42,25 +43,25 @@ async function handleRpcInteraction(
     }
   }
 
-  if (confirmationType === ConfirmationType.Delete) {
+  if (interactionEffect === InteractionEffect.ReloadItem) {
```
dtdesign, line 46: This should be a `switch` statement too.

dtdesign (changes requested):

PR #6370, opened 2025-06-27
ListView user filter

Cyperghost:


BurntimeX:
The `UserFilter` got already added with this PR:
https://github.com/WoltLab/WCF/pull/6365

PR #6382, opened 2025-07-15
Overhaul notice conditions

Cyperghost:
TODO

- [x] Migration of old data
- [x] Info that rebuild data for notice is required
  - [x] In the ACP Dashboard box
  - [x] In the list of notices


wcfsetup/install/files/lib/data/notice/NoticeAction.class.php
```diff
@@ -134,4 +147,22 @@ public function update()
...
+    #[\Override]
+    public function getLanguageCategory(): string
+    {
+        return 'wcf.notice';
+    }
+
+    #[\Override]
+    public function getPackageID(): int
+    {
+        return PACKAGE_ID;
```
dtdesign: Do not do this! `PACKAGE_ID` maps to the currently loaded app which means that calling this action from anywhere else but the notice form in the admin panel has the risk of binding the phrases to the active app. Uninstalling that app would then remove the phrases too.

Notices are tied to the core, thus you should set this to the explicit value of `1`.

wcfsetup/install/files/lib/system/condition/type/IGlobalConditionType.class.php
```diff
@@ -0,0 +1,20 @@
...
+/**
+ * @author Olaf Braun
+ * @copyright 2001-2025 WoltLab GmbH
+ * @license GNU Lesser General Public License <http://opensource.org/licenses/lgpl-license.php>
+ * @since 6.3
+ *
+ * @template TFilter
+ * @extends  IConditionType<TFilter>
+ */
+interface IGlobalConditionType extends IConditionType
```
dtdesign: I think `IContextualConditionType` is a more fitting name for this interface.

wcfsetup/install/files/lib/system/condition/type/request/NotDayOfWeekRequestConditionType.class.php
```diff
@@ -0,0 +1,94 @@
...
+
+            return [];
+        }
+
+        $conditions = [];
+
+        // We must remove all selected week of days to convert the previous condition from an “or” to an “and” condition.
+        $daysOfWeeks = \array_diff_key(DateUtil::getWeekDays(), $daysOfWeeks);
+
+        foreach ($daysOfWeeks as $dayOfWeek => $_) {
```
dtdesign:
```suggestion
        foreach (\array_keys($daysOfWeeks) as $dayOfWeek) {
```

dtdesign (changes requested):

PR #6384, opened 2025-07-16
Generic styling for a discussion list

BurntimeX:


wcfsetup/install/files/style/ui/discussionList.scss
```diff
@@ -0,0 +1,303 @@
...
+	display: flex;
+	gap: 10px;
+
+	color: var(--wcfContentDimmedText);
+
+	@include wcfFontSmall;
+
+	.icon {
+		color: inherit;
+	}
```
dtdesign:
```suggestion
```

We do not use the `icon` class anymore. Besides that the icons will always inherit the text color since 6.0.

wcfsetup/install/files/style/ui/discussionList.scss
```diff
@@ -0,0 +1,303 @@
...
+	align-items: center;
+	display: flex;
+	gap: 10px;
+	color: var(--wcfContentDimmedText);
+
+	@include wcfFontSmall;
+
+	.icon {
+		color: inherit;
+	}
```
dtdesign:
```suggestion
```

dtdesign (changes requested):

PR #6386, opened 2025-07-17
Improve error handling of filter parameters in grid/list views

BurntimeX:


wcfsetup/install/files/lib/system/gridView/filter/InvalidFilterValue.class.php
```diff
@@ -0,0 +1,13 @@
+<?php
+
+namespace wcf\system\gridView\filter;
```
dtdesign: This should be placed one level further down inside an `exception/` directory.

wcfsetup/install/files/lib/system/listView/filter/InvalidFilterValue.class.php
```diff
@@ -0,0 +1,13 @@
+<?php
+
+namespace wcf\system\listView\filter;
```
dtdesign: This should be placed one level further down inside an `exception/` directory.

dtdesign (changes requested):

PR #6391, opened 2025-07-22
Overhaul trophy conditions

Cyperghost:
- Store trophy image via `wcf\data\file\File`
- Migrate to the new condition system
- Migrate the Add/Edit Form to a FormBuilderForm


wcfsetup/install/files/lib/data/trophy/TrophyCache.class.php
```diff
@@ -122,29 +85,41 @@ public function getEnabledTrophiesByCategoryID($categoryID)
...
+    {
+        $fileIDs = [];
+        foreach ($trophies as $trophy) {
+            if ($trophy->imageFileID) {
+                $fileIDs[] = $trophy->imageFileID;
+            }
+        }
+
+        FileRuntimeCache::getInstance()->cacheObjectIDs($fileIDs);
     }
```
dtdesign: Why is this here, are you trying to reinvent runtime caches?

wcfsetup/install/files/lib/data/user/trophy/UserTrophyList.class.php
```diff
@@ -71,4 +72,18 @@ public static function getUserTrophies(array $userIDs, $includeDisabled = false)
...
+    {
+        parent::readObjects();
+
+        $fileIDs = [];
+        foreach ($this->getObjects() as $trophy) {
+            if ($trophy->getTrophy()->imageFileID !== null) {
+                $fileIDs[] = $trophy->getTrophy()->imageFileID;
+            }
+        }
+        FileRuntimeCache::getInstance()->cacheObjectIDs(\array_unique($fileIDs));
```
dtdesign: This is inconsistent, you already have a weird `cacheFileIDs()` method in your `TrophyCache`, why do you not use it here?

dtdesign (changes requested):
Overall I don’t like how the runtime cache is being used. Littering the code with `cache*IDs()` is not a solution but a hacky workaround.

Generally speaking, we usually know which trophies are being requested, so why not group them in a collection and fix this instead of making this an afterthought?

PR #6395, opened 2025-07-24
Migrate members list page to list view

BurntimeX:


com.woltlab.wcf/templates/membersList.tpl
```diff
@@ -1,172 +1,25 @@
...
-	{if $pageNo < $pages}
-		<link rel="next" href="{link controller='MembersList' pageNo=$pageNo+1}{unsafe:$canonicalURLParameters}{/link}">
+	{if $listView->getPageNo() < $listView->countPages()}
+		<link rel="next" href="{link controller='MembersList'}pageNo={$listView->getPageNo() + 1}{/link}">
 	{/if}
-	{if $pageNo > 1}
-		<link rel="prev" href="{link controller='MembersList'}{if $pageNo > 2}pageNo={$pageNo-1}&{/if}{unsafe:$canonicalURLParameters}{/link}">
+	{if $listView->getPageNo() > 1}
+		<link rel="prev" href="{link controller='MembersList'}{if $listView->getPageNo() > 2}pageNo={$listView->getPageNo() - 1}{/if}{/link}">
 	{/if}
```
dtdesign: This is a repeated pattern and it is an annoying piece of copy & paste. In addition, the `pageNo` parameter is hidden from the request route.

We could solve this by introducing two new helper methods to `AbstractListView()`:

- `public function getNextPageNo(): ?int`
- `public function getPreviousPageNo(): ?int`

They essentially do the same but can now be added as an actual parameter. Also the `{if}` checks can be turned into a simple equality check again `null`.

com.woltlab.wcf/templates/shared_interactionButton.tpl
```diff
@@ -1,9 +1,16 @@
-<div class="dropdown">
-	<button type="button" class="button small dropdownToggle" aria-label="{lang}wcf.global.button.more{/lang}">
-		{icon name='ellipsis'}
+<div class="dropdown {$configuration->cssClassName}">
+	<button
+		type="button"
+		class="dropdownToggle {$configuration->buttonCssClassName}{if !$configuration->label} jsTooltip{/if}"
+		{if !$configuration->label}title="{lang}{$configuration->tooltip}{/lang}"{/if}
+	>
+		{icon name=$configuration->icon size=$configuration->iconSize}
```
dtdesign: This causes an unintentional limitations because this only supports non-solid icons and excludes the ability to use brand icons. Please see my remarks on the configuration class.

wcfsetup/install/files/lib/system/interaction/InteractionContextMenuComponentConfiguration.class.php
```diff
@@ -0,0 +1,29 @@
...
+ */
+final class InteractionContextMenuComponentConfiguration
+{
+    public function __construct(
+        public readonly string $cssClassName = '',
+        public readonly string $buttonCssClassName = '',
+        public readonly string $dropdownMenuCssClassName = '',
+        public readonly string $label = '',
+        public readonly string $tooltip = 'wcf.global.button.moreOptions',
+        public readonly string $icon = 'ellipsis',
```
dtdesign: This should expect `IFontAwesomeIcon` instead, enabling both solid and brand icons.

Add a `getIconHtml()` method that just returns `$this->icon->toHtml($this->iconSize)`.

You can support a default icon by making the constructor parameter nullable and implicitly setting it to `FontAwesomeIcon::fromValue('ellipsis')`.

com.woltlab.wcf/templates/shared_standaloneInteractionButton.tpl
```diff
@@ -1,16 +1,16 @@
...
 		type="button"
-		class="button dropdownToggle {$buttonCssClassName}"
-		{if !$label}aria-label="{lang}wcf.global.button.more{/lang}"{/if}
+		class="dropdownToggle {$configuration->buttonCssClassName}{if !$configuration->label} jsTooltip{/if}"
+		{if !$configuration->label}title="{lang}{$configuration->tooltip}{/lang}"{/if}
 	>
-		{icon name=$icon}
-		{if $label}
-			<span>{$label}</span>
+		{icon name=$configuration->icon size=$configuration->iconSize}
```
dtdesign: This will need to be updated too.

com.woltlab.wcf/templates/userCard.tpl
```diff
@@ -7,15 +7,41 @@
...
-			{user object=$user type='avatar64' ariaHidden='true' tabindex='-1'}
+			{unsafe:$user->getAvatar()->getImageTag(64)}
 
 			{if $user->isOnline()}<span class="userCard__onlineIndicator jsTooltip" title="{lang username=$user->username}wcf.user.online.title{/lang}"></span>{/if}
 		</div>
+
+		<div class="userCard__header__interactions">
+			{if $__wcf->user->userID && $user->userID != $__wcf->user->userID}
+				{if !$__wcf->getUserProfileHandler()->isIgnoredByUser($user->userID)}
+					{if $__wcf->getUserProfileHandler()->isFollowing($user->userID)}
```
dtdesign, line 18: I am aware that this is most likely copied from somewhere else but maybe introduce two helper methods instead?

com.woltlab.wcf/templates/userProfileHeader.tpl
```diff
@@ -102,46 +102,36 @@
...
-								{else}
-									<button type="button" {unsafe:$interactionOption->attributes}>{$interactionOption->title}</button>
-								{/if}
-							</li>
-						{/foreach}
-					</ul>
-				</div>
+			{if $__wcf->user->userID && $user->userID != $__wcf->user->userID}
+				{if !$__wcf->getUserProfileHandler()->isIgnoredByUser($user->userID)}
+					{if $__wcf->getUserProfileHandler()->isFollowing($user->userID)}
```
dtdesign, line 108: My remark on `userCard.tpl` applies here too.

wcfsetup/install/files/lib/system/listView/user/TeamListView.class.php
```diff
@@ -0,0 +1,44 @@
...
+ * List view for the team list.
+ *
+ * @author      Marcel Werk
+ * @copyright   2001-2025 WoltLab GmbH
+ * @license     GNU Lesser General Public License <http://opensource.org/licenses/lgpl-license.php>
+ * @since       6.2
+ */
+class TeamListView extends MembersListView
+{
+    public function __construct(private readonly int $groupID)
```
dtdesign, line 17: I think it would be better to expect a `UserGroup` object instead. This is already available in `TeamPage` and serves as a validation that the ID is most likely correct.

dtdesign (changes requested):

PR #6419, opened 2025-08-07
Implement a command to copy attachments

Cyperghost:


wcfsetup/install/files/lib/command/attachment/CopyAttachments.class.php
```diff
@@ -0,0 +1,122 @@
...
+ * @license GNU Lesser General Public License <http://opensource.org/licenses/lgpl-license.php>
+ * @since 6.3
+ */
+final class CopyAttachments
+{
+    public function __construct(
+        public readonly string $sourceObjectType,
+        public readonly int $sourceObjectID,
+        public readonly string $targetObjectType,
+        public readonly int $targetObjectID,
```
dtdesign:
```suggestion
        private readonly string $sourceObjectType,
        private readonly int $sourceObjectID,
        private readonly string $targetObjectType,
        private readonly int $targetObjectID,
```

There is no reason to make these properties public. Always set the lowest possible visibility, because this can be reversed later if needed without breaking compatibility.

At the same time this keeps the API surface as low as possible without impacting the functionality.

dtdesign (changes requested):
Apart from this you can merge it.

PR #6439, opened 2025-08-22
Unify bulk interactions with normal interactions

BurntimeX:


wcfsetup/install/files/lib/action/AbstractModerationAction.class.php
```diff
@@ -40,16 +40,16 @@ public function handle(ServerRequestInterface $request): ResponseInterface
             <<<'EOT'
                 array {
                     id?: positive-int,
-                    objectIDs?: positive-int[]
+                    ids?: positive-int[]
                 }
```
dtdesign, line 44: This should possibly be a union type instead: `array { id: positive-int } | array { ids: positive-int[] }`. This avoids dealing with this in line 48-50 and makes the intent clearer.

dtdesign (changes requested):

PR #6471, opened 2025-09-23
Migrate grid view filters to list view filters

### LINKED ISSUE #6459: Migrate grid view filters to list view filters
Filtering in grid views is currently based on the defined columns. Although this is very practical, it reaches its limits when complex filtering is required that cannot be mapped using a single column or that requires information that are not part of the defined columns.

BurntimeX:
Closes #6459


wcfsetup/install/files/lib/system/view/filter/BooleanFilter.class.php
```diff
@@ -16,6 +16,15 @@
  */
 class BooleanFilter extends AbstractFilter
 {
+    public function __construct(
+        string $id,
+        string $languageItem,
+        string $databaseColumn = '',
+        private readonly bool $reverseValue = false
```
dtdesign, line 23: This property is unused.

Apart from this extra parameter the signature matches the parent constructor which means that you can remove it.

dtdesign (changes requested):
Please make sure to run PHPStan before merging.

PR #6508, opened 2025-11-04
Add missing support for the `staticDictionary` to the `PasswordFormField`

BurntimeX:


wcfsetup/install/files/lib/system/form/builder/field/PasswordFormField.class.php
```diff
@@ -63,6 +63,12 @@ class PasswordFormField extends AbstractFormField implements
      */
     protected array $relatedFieldsId = [];
 
+    /**
+     * @var string[]
```
dtdesign, line 67: This should be a `list<string>` instead because that is what the JS implementation expects. This affects the methods below too.

com.woltlab.wcf/templates/shared_passwordFormField.tpl
```diff
@@ -27,6 +27,11 @@
 					document.getElementById('{unsafe:$fieldId|encodeJS}'),
 				{/foreach}
 				],
+				staticDictionary: [
+				{foreach from=$field->getStaticDictionary() item=staticDictionary}
+					'{unsafe:$staticDictionary|encodeJS}',
+				{/foreach}
+				],
```
dtdesign, line 34:
```suggestion
				staticDictionary:  {unsafe:$field->getStaticDictionary()|json},
```

This is just a list of strings after all.

dtdesign (changes requested):

PR #6515, opened 2025-11-11
Show absolute path to WSC installation in path selection overlay on `DevtoolsProjectListPage`

Fighter456:
During the test of Suite 6.2 I've discovered, that the path selection overlay does not provide the absolute path to the Suite Core installation on the `DevtoolsProjectListPage`.

Because the path is shown on the `DevtoolsProjectAddForm` it makes sense to unify this behavior.

Note: The usage of `\WCF_DIR` directly in the template is disallowed. Therefore the page-class needs to assign this information through the template engine.

### DIFF
--- a/wcfsetup/install/files/acp/templates/devtoolsProjectList.tpl
+++ b/wcfsetup/install/files/acp/templates/devtoolsProjectList.tpl
@@ -95,7 +95,7 @@
 	<dl>
 		<dt>{lang}wcf.acp.devtools.project.quickSetup.path{/lang}</dt>
 		<dd>
-			<input type="text" name="projectQuickSetupPath" id="projectQuickSetupPath" class="long" />
+			<input type="text" name="projectQuickSetupPath" id="projectQuickSetupPath" class="long" placeholder="{unsafe:$wcfDir}" />
 			<small>{lang}wcf.acp.devtools.project.quickSetup.path.description{/lang}</small>
 		</dd>
 	</dl>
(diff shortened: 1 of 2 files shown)


wcfsetup/install/files/acp/templates/devtoolsProjectList.tpl
```diff
@@ -95,7 +95,7 @@
 	<dl>
 		<dt>{lang}wcf.acp.devtools.project.quickSetup.path{/lang}</dt>
 		<dd>
-			<input type="text" name="projectQuickSetupPath" id="projectQuickSetupPath" class="long" />
+			<input type="text" name="projectQuickSetupPath" id="projectQuickSetupPath" class="long" placeholder="{unsafe:$wcfDir}" />
```
dtdesign, line 98: Why would you use `unsafe:` here?

dtdesign:
I’m not sure what the point of adding the placeholder is, because a placeholder is merely a hint and not available for text completion?

Fighter456:
> I’m not sure what the point of adding the placeholder is, because a placeholder is merely a hint and not available for text completion?

As stated in the PR description, the `DevtoolsProjectAddForm` shows the path as placeholder, too. In the past I've copied the value from the placeholder (using the browser developer tools).

What about show the path in a `kdb` element within a field description in both places to make it easier to copy?

wcfsetup/install/files/acp/templates/devtoolsProjectList.tpl
Fighter456: IIRC I've copied it from another template. Not necessary at this point?

dtdesign:
Thanks. I’ve opted to make the changes myself since there I made some small changes at didn’t want to bother you with them, the result is effectively the same.

wcfsetup/install/files/acp/templates/devtoolsProjectList.tpl
dtdesign: As a general guideline: Never use `unsafe:` unless you actually _need_ it. The performance gains are negligible but you risk malformed HTML or XSS. This also makes it much easier to test for security issues because in HTML any variable that is escaped before outputting is safe by definition.

PR #6524, opened 2025-11-20
Sync the package version when synchronizing an entire project

dtdesign:


ts/WoltLabSuite/Core/Api/DevtoolsProjects/SyncVersion.ts
```diff
@@ -0,0 +1,19 @@
+/**
+ * Synchronizes the version of an installed package with the versio number in
```
BurntimeX:
```suggestion
 * Synchronizes the version of an installed package with the version number in
```

BurntimeX (changes requested):

PR #6583, opened 2026-02-14
Use the real installation number instead of hardcoded

Fighter456:
Currently the validation of MySQL respective the foreign keys are always falsy [in 6.2](https://github.com/WoltLab/WCF/commit/32bd58238b0e9a296c953491b34c8153d41c95d5) if the installation number of the instance is not `1`.

The parameters of the condition builder are not replaced by `WCF::getDB()->prepare()` during statement preparation.


dtdesign:
Good catch, thank you!

PR #6586, opened 2026-02-15
Localize the position and type column of `BoxGridView`

Fighter456:
In contrast to version 6.1, where the output in the table was localized using language variables, currently only the raw value from the database table is displayed.

This change localizes the output again.

Before:
image

After:
image

Previous (6.1):
image


BurntimeX (approved):

BurntimeX:
Thanks!

PR #6607, opened 2026-02-27
Use the PHP-DDL for the WCFSetup

dtdesign:
This PR does a few things:

1. Migrate the table creation from the legacy `install.sql` to the PHP-DDL to keep it in sync with all other packages.
2. Deprecate the `SQLParser` and `sql`-PIP in favor of the PHP-DDL.
3. Optimize the performance of `DatabaseTableChangeProcessor` and add an option to efficiently and conveniently call it programmatically.
4. Use the `install.sql` to keep the basic `INSERT` statements around but execute them directly without parsing them.


BurntimeX (approved):
LGTM

PR #6608, opened 2026-02-27
Add auto-discover for RPC API controllers

### LINKED ISSUE #6606: Consider auto-discover for RPC API controllers
Registering API controllers is a tedious task and could be made obsolete by an auto-discover feature.

BurntimeX:
Closes #6606


dtdesign (changes requested):
Please also mark the `ControllerCollecting` event as deprecated, indicating that this is no longer required.

PR #6611, opened 2026-02-28
Update / unify property documentation of database objects

BurntimeX:


wcfsetup/install/files/lib/data/trophy/Trophy.class.php
```diff
@@ -23,20 +23,20 @@
...
- * @property-read   string $iconName           the icon name
- * @property-read   string $iconColor          the icon color
- * @property-read   string $badgeColor         the icon badge color
- * @property-read   int $isDisabled         `1` if the trophy is disabled
- * @property-read   int $awardAutomatically     `1` if the trophy is awarded automatically
- * @property-read   int $revokeAutomatically        `1` if the trophy should be automatically revoked once the conditions are no longer met.
- * @property-read   int $trophyUseHtml              `1` if the trophy use a html description
- * @property-read   int $showOrder              position of the trophy in relation to the other trophies at the same location
+ * @property-read   int     $trophyID               unique id for the trophy
+ * @property-read   ?string $title
```
dtdesign: The title is mandatory.

wcfsetup/install/files/lib/data/trophy/Trophy.class.php
```diff
@@ -23,20 +23,20 @@
...
+ * @property-read   ?string $description
+ * @property-read   int     $categoryID
+ * @property-read   ?int    $type
```
dtdesign: The `type` is actually an enum in disguise with the only valid values being `1` (image) and `2` (badge). This column must not be null.

dtdesign (changes requested):

dtdesign (approved):

PR #6649, opened 2026-03-31
`ReactionHandler` performance optimizations

BurntimeX:


wcfsetup/install/files/lib/system/reaction/ReactionHandler.class.php
```diff
@@ -248,6 +261,12 @@ public function loadLikeObjects(ObjectType $objectType, array $objectIDs): int
             $i++;
         }
 
+        foreach ($objectIDs as $objectID) {
+            if (!isset($this->likeObjectCache[$objectType->objectTypeID][$objectID])) {
+                $this->likeObjectCache[$objectType->objectTypeID][$objectID] = null;
+            }
```
dtdesign: This is equivalent to this:

```php
$this->likeObjectCache[$objectType->objectTypeID][$objectID] ??= null;
```

dtdesign (approved):

PR #6654, opened 2026-04-11
Add event for collecting user avatar form fields during the form dialog

Cyperghost:


BurntimeX:
What is the purpose of this event? The corresponding dialog does not have a “Submit” button. Input fields added by this event cannot be saved.

Cyperghost:
> What is the purpose of this event? The corresponding dialog does not have a “Submit” button. Input fields added by this event cannot be saved.

I need a simple way to add extra fields there. It doesn’t matter, and I’m well aware, that the form can’t be submitted or processed.
At the moment, I have no way of adding anything there other than using JavaScript events. This makes it possible in the dialog.

PR #6689, opened 2026-06-15
Add `DatabaseObjectBuilder` with `TagBuilder` implementation

BurntimeX:
Introduces an abstract builder for creating, updating and deleting database objects with a fluent setter API, batched transactional deletes and an `INSERT IGNORE`-style helper. `TagBuilder` is the first concrete implementation.

### DIFF
--- a/wcfsetup/install/files/lib/data/DatabaseObjectBuilder.class.php
+++ b/wcfsetup/install/files/lib/data/DatabaseObjectBuilder.class.php
@@ -302,4 +302,36 @@ abstract class DatabaseObjectBuilder
 
         return $this;
     }
+
+    /**
+     * Updates counters for the given object.
+     *
+     * @param TDatabaseObject $object
+     * @param array<string, int|float> $counters
+     */
+    final public static function updateCounters(DatabaseObject $object, array $counters): void
+    {
+        if ($counters === []) {
+            throw new \InvalidArgumentException("The list of counters to update must not be empty.");
+        }
+
+        \assert($object instanceof (static::getBaseClass()));
+
+        $updateSQL = '';
+        $statementParameters = [];
+        foreach ($counters as $key => $value) {
+            if ($updateSQL !== '') {
+                $updateSQL .= ', ';
+            }
+            $updateSQL .= $key . ' = ' . $key . ' + ?';
+            $statementParameters[] = $value;
+        }
+        $statementParameters[] = $object->getObjectID();
+
+        $sql = "UPDATE  " . static::getBaseClass()::getDatabaseTableName() . "
+                SET     " . $updateSQL . "
+                WHERE   " . static::getBaseClass()::getDatabaseTableIndexName() . " = ?";
+        $statement = WCF::getDB()->prepare($sql);
+        $statement->execute($statementParameters);
+    }
 }
--- a/wcfsetup/install/files/lib/data/DatabaseObjectBuilder.class.php
+++ b/wcfsetup/install/files/lib/data/DatabaseObjectBuilder.class.php
@@ -23,17 +23,17 @@ abstract class DatabaseObjectBuilder
     /**
      * @var array<string, string|int|float|null>
      */
-    protected array $properties = [];
+    public protected(set) array $properties = [];
 
     /**
      * @var array<string, string|int|float|null>
      */
-    protected array $customProperties = [];
+    public protected(set) array $customProperties = [];
 
     /**
      * @var array<string, int|float>
      */
-    protected array $incrementProperties = [];
+    public protected(set) array $incrementProperties = [];
 
     /**
      * Use forCreate() or forUpdate() to obtain a builder instance.
--- a/wcfsetup/install/files/lib/data/DatabaseObjectBuilder.class.php
+++ b/wcfsetup/install/files/lib/data/DatabaseObjectBuilder.class.php
@@ -86,10 +86,10 @@ abstract class DatabaseObjectBuilder
         $statement = WCF::getDB()->prepare($sql);
         $statement->execute($statementParameters);
 
-        if (static::getBaseClass()::getDatabaseTableIndexIsIdentity()) {
-            $id = WCF::getDB()->getInsertID(static::getBaseClass()::getDatabaseTableName(), static::getBaseClass()::getDatabaseTableIndexName());
-        } elseif (isset($this->properties[static::getBaseClass()::getDatabaseTableIndexName()])) {
+        if (isset($this->properties[static::getBaseClass()::getDatabaseTableIndexName()])) {
             $id = $this->properties[static::getBaseClass()::getDatabaseTableIndexName()];
+        } elseif (static::getBaseClass()::getDatabaseTableIndexIsIdentity()) {
+            $id = WCF::getDB()->getInsertID(static::getBaseClass()::getDatabaseTableName(), static::getBaseClass()::getDatabaseTableIndexName());
         } else {
             throw new \BadMethodCallException("Missing value for '" . static::getBaseClass()::getDatabaseTableIndexName() . "'");
         }
@@ -165,7 +165,7 @@ abstract class DatabaseObjectBuilder
      * Deletes the rows identified by the given primary keys in batches inside
      * a single transaction.
      *
-     * @param (string|int)[] $objectIDs
+     * @param (int|string)[] $objectIDs
      */
     final public static function deleteAll(array $objectIDs = []): void
     {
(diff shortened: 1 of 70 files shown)


wcfsetup/install/files/lib/data/tag/TagBuilder.class.php
```diff
@@ -0,0 +1,46 @@
...
+ * @author      Marcel Werk
+ * @copyright   2001-2026 WoltLab GmbH
+ * @license     GNU Lesser General Public License <http://opensource.org/licenses/lgpl-license.php>
+ * @since       6.3
+ *
+ * @extends DatabaseObjectBuilder<Tag>
+ */
+final class TagBuilder extends DatabaseObjectBuilder
+{
+    public function setTagID(int $tagID): static
```
dtdesign: This must not be available inside the `forUpdate()` flow. Maybe introduce a generic method `DatabaseObjectBuilder::setID()` that does that. After all, this is only ever required for (a) sessions and (b) data imports.

wcfsetup/install/files/lib/data/DatabaseObjectBuilder.class.php
```diff
@@ -0,0 +1,250 @@
...
+     * @param ?TDatabaseObject $object
+     */
+    private function __construct(protected readonly ?DatabaseObject $object = null) {}
+
+    /**
+     * Persists the pending changes and returns the resulting database object.
+     *
+     * @return TDatabaseObject
+     */
+    public function save(): DatabaseObject
```
dtdesign: This and all other methods declared by `DatabaseObjectBuilder` should be declared as `final` to allow us to make adjustments to this crucial component.

wcfsetup/install/files/lib/data/DatabaseObjectBuilder.class.php
```diff
@@ -0,0 +1,250 @@
...
+    }
+
+    /**
+     * Inserts a new row and returns the primary key of the created object.
+     */
+    private function create(): int|string
+    {
+        $keys = $values = '';
+        $statementParameters = [];
+        foreach (array_merge($this->properties, $this->customProperties) as $key => $value) {
```
dtdesign: Should be `\array_merge()`, not `array_merge()`.

wcfsetup/install/files/lib/data/DatabaseObjectBuilder.class.php
```diff
@@ -0,0 +1,250 @@
...
+
+        if (static::getBaseClass()::getDatabaseTableIndexIsIdentity()) {
+            $id = WCF::getDB()->getInsertID(static::getBaseClass()::getDatabaseTableName(), static::getBaseClass()::getDatabaseTableIndexName());
+        } elseif (isset($this->properties[static::getBaseClass()::getDatabaseTableIndexName()])) {
+            $id = $this->properties[static::getBaseClass()::getDatabaseTableIndexName()];
+        } else {
+            throw new \BadMethodCallException("Missing value for '" . static::getBaseClass()::getDatabaseTableIndexName() . "'");
+        }
+
+        return $id;
```
dtdesign: This can be untangled into two conditional returns and an unconditional exception. The `return $id` just obfuscates the flow.

dtdesign (changes requested):

wcfsetup/install/files/lib/data/DatabaseObjectBuilder.class.php
BurntimeX: Obsolete bc the `afterCreate` hook was added.

wcfsetup/install/files/lib/data/tag/TagBuilder.class.php
```diff
@@ -0,0 +1,115 @@
...
+
+    /**
+     * @param list<string> $synonyms
+     */
+    public function setSynonyms(array $synonyms): static
+    {
+        $this->synonyms = $synonyms;
+
+        return $this;
+    }
```
dtdesign, line 59: These two are mutually exclusive. `setSynonymFor()` should set `$this->synonyms = [];` and `setSynomyms()` should set `$this->properties['synonymFor'] = null;`.

wcfsetup/install/files/lib/system/form/builder/field/wysiwyg/WysiwygFormField.class.php
```diff
@@ -393,4 +393,16 @@ public function getValue()
...
         return $upcastProcessor->getHtml();
     }
+
+    /**
+     * @since 6.3
+     */
+    public function getHtmlInputProcessor(): HtmlInputProcessor
+    {
+        if ($this->htmlInputProcessor === null) {
+            throw new \BadMethodCallException("The html input processor is not available before validate() has been called.");
```
dtdesign:
```suggestion
            throw new \BadMethodCallException("The HTML input processor is not available before validate() has been called.");
```

wcfsetup/install/files/lib/system/form/builder/field/IFormField.class.php
```diff
@@ -94,6 +95,80 @@ public function getValidators();
...
 
+    /**
+     * Sets a callback that transfers this field's save value into a
+     * `DatabaseObjectBuilder` instance and returns this field.
+     *
+     * The callback is invoked by `DatabaseObjectBuilderFormDocument` when the
+     * builder is populated from the form's fields, for example:
+     *
+     *     $field->saveValueCallback(
+     *         static fn(DatabaseObjectBuilder $builder, IFormField $formField) => $builder->setName($formField->getSaveValue())
```
dtdesign: This still uses the shorthand function which returns the builder, violating the `void` return type, turn this into a regular `static function` instead.

wcfsetup/install/files/lib/system/form/builder/DatabaseObjectBuilderFormDocument.class.php
```diff
@@ -0,0 +1,67 @@
...
+/**
+ * Represents a form document whose field values are written into a
+ * `DatabaseObjectBuilder` instead of being collected into a parameters array
+ * for a database object action.
+ *
+ * Fields contribute their value by registering a callback via
+ * `IFormField::saveValueCallback()`:
+ *
+ *     $field->saveValueCallback(
+ *         static fn(DatabaseObjectBuilder $builder, IFormField $formField) => $builder->setName($formField->getSaveValue())
```
dtdesign: Should be a `static function` instead.

dtdesign (changes requested):

dtdesign:
## Pattern for Builders

We need to formalize the pattern at some point, for now I think it’s reasonable to go with:

* Methods should be named `set*(): static` and use camel case, for example, `setTags()` or `setIsDeleted()`.
* Whenever possible, methods should be designed to make invalid states unrepresentable. Do not offer a separate `setUserID()` and `setUsername()` method, instead offer `setUser(User $user)` and `setGuest(string $username)` which internally sets both properties (`userID` and `username`).
* Parameters must be as strictly typed as reasonably (!) possible. Instead of asking for `?int $languageID`, require a `?Language $language` to be provided instead.
* Any extra state must be tracked in private properties that are both typed and not initialized. A definition like `private ?Foo $foo;` allows us to distinguish between ”do not consider” and `null`. Use this for all types in order to have a common pattern.
* Do not perform any validation beyond checking for the minimum required properties to _formally_ create an object. Use `afterValidateCreate()` to run checks for extra state only that must be present.
* Any methods that runs in `afterCreate()` or `afterUpdate()` should be named `save*()` in order to distinguish them from the `set*()` methods.

## Things not yet decided

### Support for Implicit Default Values

Not every value can be mapped through MySQL’s default values, for example, we cannot set default values for `*TEXT` columns nor are we able to utilize `UNIX_TIMESTAMP()` for timestamps since we require them to be consistent within a request. A common use case is to implicitly set the time column to `\TIME_NOW`.

However, this can be extended to other kind of columns, such as auto-filling the user. DBOActions usually set it to whatever is present in `WCF::getUser()` which is dangerous when executed in a programmatic fashion. It is also debatable wether or not this actually saves a lot of work, typically these builders are part of a form and the effort to just manually invoke `->setTime(\TIME_NOW)` is pretty much negligible.

We could add this as an indirect feature to the `Create*` commands but this has the issue that the contents of the builder is not visible on the outside.

### Asymmetric Visibility on Builder Properties (PHP 8.4)

PHP 8.4 has the asymmetric visibility that would allow us to define the properties as `private(set)`. This would also make it possible to make the changes performed in  `forUpdate()` transparent to event listeners by providing the builder. Listeners can then inspect the changes and decide if they want to take an action.

The major drawback of this is that Ubuntu 22.04 LTS ships with PHP 8.3 (Debian Trixie offers PHP 8.4) natively, which would mean that this would be a bit of a steep increase in system requirements. That said, PHP 8.3 is already out of support, with security support ending Jan 1, 2028. It is already a pretty old version of PHP.

See https://www.php.net/releases/8.4/en.php#asymmetric_visibility to learn more.

dtdesign:
Both the `Create*` and `Update*` command expect a builder instance but they do not verify if this is actually for an update. They will simply request the object and later on interact with it, expecting it to be not `null`, eventually causing an error when accessing properties on null.

Maybe add a new helper method that fails loud when accessing the object on a create instance? Or just turn the `getObject()` method into something that throws an exception? The intention is to avoid having the same boiler-plate code in each `Update*` command just to check it and throw.

dtdesign:
## Methods for Counter-Style Columns

Some columns aren’t really made for being updated directly, at least not using a regular update. Counters are prone to race conditions when two concurrent requests want to modify it by providing an absolute number. The `DatabaseObjectBuilder::updateCounters()` method exists for this purpose as it allows making relative changes in order to achieve atomic updates.

Not offering a method for absolute adjustments (for example, `setComments()`) has some merit here but it would mean that the callee has to know the column name in order to provide it for `updateCounters()`. This doesn’t matter for imports (which usually don’t provide these numbers at all) because when the item has a counter of zero, any provided number becomes an absolute adjustment.

### Alternative Approach Using Helper Methods

We could also go a different route and make `updateCounters()` protected and instead only offer regular helper methods. Instead of `set*()` these would be called `increment*()` like `incrementComments()`, taking advantage of the fact that the implementation always use addition and only the sign of the value changes the outcome. Technically it is applying a delta which by definition can be a positive or negative value.

This has the benefit of once more using a name pattern that is not already taken in the base implementation. A new property `$incrementProperties` can be used to keep track of these, with the columns being merged in both the `create()` and `update()` call with little adjustments.

Bonus points for getting rid of a public method `updateCounters()` and having a fully type-safe API that does not accept arbitrary arrays.

PR #6693, opened 2026-06-23
Fix Sass mixed declaration deprecations

SoftCreatR:
Fix Sass mixed-decls deprecation warnings in WCF core and ACP styles by avoiding declarations after nested media-producing mixins.

- Add `wcfFontHeadlineBold` for bold headline text without chaining `wcfFontHeadline` and `wcfFontBold`
- Reorder declarations around `wcfFontDefault`, `wcfFontSmall`, `screen-lg`, and related nested mixins
- Wrap the intentional post-input overrides in `itemListInput.scss` with `& {}`

Even after fixing, there are Sass warnings remaining. These are from WSF, etc. where these fixes need to be applied as well.

See also https://www.woltlab.com/community/thread/318717-style-compilation-schl%C3%A4gt-unter-windows-fehlt

### DIFF
--- a/wcfsetup/install/files/acp/style/layout.scss
+++ b/wcfsetup/install/files/acp/style/layout.scss
@@ -674,8 +674,7 @@ html[data-color-scheme="dark"] {
 }
 
 .packageSearchName {
-	@include wcfFontHeadline;
-	@include wcfFontBold;
+	@include wcfFontHeadlineBold;
 }
 
 .packageSearchVersion {
--- a/wcfsetup/install/files/style/bootstrap/mixin/font.scss
+++ b/wcfsetup/install/files/style/bootstrap/mixin/font.scss
@@ -37,6 +37,19 @@
 	}
 }
 
+@mixin wcfFontHeadlineBold {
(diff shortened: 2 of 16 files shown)


dtdesign:
I’ve looked into this and came to the conclusion that this does not fit into 6.2. The mixed declarations thing isn’t actionable at this point because it is the implicit result of combining `@imports`. The stopgap solution of introducing a combined import is merely a bandaid for the symptom but does not address the inherent issue with these kind of selectors.

Looking into both our apps and third party plugins, these kind of combinations appear more often than I anticipated. The most reasonable way is to silence the deprecation for the time now and then later add a custom logger that properly propagates these issues to our exception handler, making it more visible. This would effectively break installations running 6.2 in a patch release which is against our policy.

We may address this in 6.3 although I’m not entirely confident about that given that this is more of a systemic weakness of SCSS rather than an isolated bug. Maybe we can ignore this issue until 7.0 and just drop SCSS, which would solve a boatload of issues at once … These days SCSS serves no real purpose anymore.

Issue #5925, opened 2024-05-26
Poll - status bar, dark theme version


eveneme:
The status bar of the poll in the dark theme is not very distinct (in the light version, the bar is blue). I suggest changing the status bar color to blue in the dark version as well. The result is shown below.

Before:
![Zrzut ekranu 2024-05-27 002707](https://github.com/WoltLab/WCF/assets/85762221/79b94709-e08d-4791-924b-9c5203f5289f)

After:
![Zrzut ekranu 2024-05-27 002753](https://github.com/WoltLab/WCF/assets/85762221/6283371c-aa4b-485b-91c5-eb2c62745d84)

PR #6697, opened 2026-06-29
Improve poll meter bar contrast in dark mode

jensfr1:
## What

In dark mode the poll result bar (`.pollMeterValue`) was filled with a
faint translucent white (`rgb(255 255 255 / 34%)`), which is hard to
distinguish from the underlying track. This makes it difficult to read
the relative share of votes at a glance.

This changes the dark mode fill to the accent color (`--wcfContentLink`),
so the filled portion is clearly visible and consistent with the
appearance of the bar in light mode.

## Why

Reported in #5925 with before/after screenshots.

## Notes

Single-line style change in `wcfsetup/install/files/style/ui/poll.scss`.
Using the `--wcfContentLink` variable keeps the color in sync with the
active style's accent instead of hardcoding a value.

### DIFF
--- a/wcfsetup/install/files/style/ui/poll.scss
+++ b/wcfsetup/install/files/style/ui/poll.scss
@@ -97,5 +97,5 @@
 }
 
 html[data-color-scheme="dark"] .pollMeterValue {
-	background-color: rgb(255 255 255 / 34%);
+	background-color: var(--wcfContentLink);
 }


dtdesign:
This is actually a worse contrast ratio plus you are using a variable that you have no control over in terms of it matching the underlying bar. The problem with contrast is, that the perception is misleading because it needs to account for different visual impairments.

It’s easy to think of it as a binary “blind – not blind” where in reality this is not only a gradient but _also_ a mix-and-match of different impairments, for example the inability to perceive color. To better illustrate the problem, here is a screenshot of the referenced issue but converted to grayscale:

PR #6701, opened 2026-07-29
Fix one-time consent for external media

SoftCreatR:
The Fancybox consent plugin only handles permanently enabling external media. Clicking “Display content once” still invokes the generic user-consent handler, which expects a target or payload and can attempt to insert content relative to an already detached consent element.

This results in an empty Fancybox slide and the following error:

> The reference element has no parent, but the insert position was set to 'before'.

This PR handles one-time consent directly in the Fancybox plugin:

- Restores the original media type for the current slide.
- Refreshes the current slide so the external media is loaded.
- Prevents the generic payload-based consent handler from processing the Fancybox button.
- Leaves permanent consent behavior unchanged.


dtdesign (approved):

dtdesign:
Thanks!

